PR #6, opened 2017-09-06
Fix location of node binary on Windows

alexeagle:


defs.bzl
```diff
@@ -16,5 +16,5 @@
 
 Users should not load files under "/internal"
 """
-load("//internal:node.bzl", "nodejs_binary")
 load("//internal:node_install.bzl", "node_repositories")
+load("//internal:node.bzl", "nodejs_binary")
```
mprobst: is this change intentional?

internal/node_install.bzl
```diff
@@ -48,6 +41,8 @@ alias(name = "npm", actual = "bin/npm")
         stripPrefix = "node-v6.10.2-win-x64",
         sha256 = "d778ed84685c6604192cfcf40192004e27fb11c9e65c3ce4b283d90703b4192c"
     )
+    node = "node.exe"
```
mprobst: why is this `bin/node` above, but just `node.exe` here? Add a comment?

internal/node_install.bzl
```diff
@@ -86,14 +90,18 @@ package(default_visibility = ["//visibility:public"])
 exports_files(['yarn.sh'])
 alias(name = "yarn", actual = ":yarn.sh")
 """)
+  if ctx.os.name.lower().find("windows") != -1:
+    node = Label("@nodejs//:node.exe")
```
mprobst: why are you not using the alias defined above, but special case node.exe for windows?

defs.bzl
alexeagle: just changing the order to be more intuitive, but it has no effect on the PR, I'll remove for clarity

internal/node_install.bzl
alexeagle: adding comment

internal/node_install.bzl
alexeagle: the alias isn't resolved in skylark, if you `ctx.path(Label(alias))` you get the location of the alias not the target.
```
Not a file: /usr/local/google/home/alexeagle/.cache/bazel/_bazel_alexeagle/c410d6070ef2d359d26ac4aeb886e1ea/external/nodejs/node
```
I guess the alias resolution happens in the loading phase?

mprobst: Maybe something to task? Also, why do you have the alias then in the first place?

PR #11, opened 2017-09-10
Remove .vscode

jonjonsonjr:


booninite:
Sorely needed fix

alexeagle:
Why do you want to remove the vscode settings?
Seems useful to me that it adds the ignore for `bazel-*`, otherwise the editor tries to look in the outputs

jonjonsonjr:
It's odd to have editor-specific config at the root of a project (except for personal projects). It feels a bit unhygienic 👃 

Feel free to close if you disagree -- this is just a pet peeve of mine.

alexeagle:
Okay, I agree.
I folded this into https://github.com/bazelbuild/rules_nodejs/pull/27 which also stops creating the `bazel-*` symlinks.
Thanks!

PR #12, opened 2017-09-11
windows-compatible runfiles

alexeagle:


alexeagle:
Tested manually on windows:
```
$ bazel build examples/rollup
$ bazel-out/msvc_x64-fastbuild/bin/examples/rollup/rollup --help
```

/cc @dslomov

internal/node_launcher.sh
```diff
@@ -92,4 +92,8 @@ done
 # Note: for debugging it is useful to see what files are actually present, eg:
 # find . -name thingImLookingFor
 
-exec "${RUNFILES}/TEMPLATED_node" "${NODE_OPTIONS[@]}" "${RUNFILES}/TEMPLATED_script_path" "${ARGS[@]}"
+# Lookup the real paths from the runfiles manifest.
+NODE_PATH=$(grep TEMPLATED_node ${RUNFILES} | awk '{print $2}')
```
ocombe: this doesn't seem very robust & optimized, is there an alternative?

alexeagle: It certainly isn't very portable.

/cc @mprobst for ideas

internal/node_loader.js
```diff
@@ -56,8 +58,20 @@ function resolveToModuleRoot(path) {
   return null;
 }
 
-function runfilesDir() {
-  return process.env.RUNFILES || process.env.TEST_SRCDIR;
+function loadRunfiles(manifestPath) {
```
mprobst: this is loading the manifest, not the runfiles, right?

internal/node_loader.js
```diff
@@ -56,8 +58,20 @@ function resolveToModuleRoot(path) {
...
+  return new Promise((resolve, reject) => {
+    const map = Object.create(null);
+    const input = fs.createReadStream(manifestPath, {encoding: 'utf-8'});
+    const manifest = readline.createInterface({input});
+
+    manifest.on('line', line => {
+      if (!line) return;
+      const [key, value] = line.split(" ");
```
mprobst: can you give better names than key/value here?

internal/node_loader.js
```diff
@@ -56,8 +58,20 @@ function resolveToModuleRoot(path) {
...
 
-function runfilesDir() {
-  return process.env.RUNFILES || process.env.TEST_SRCDIR;
+function loadRunfiles(manifestPath) {
+  return new Promise((resolve, reject) => {
+    const map = Object.create(null);
+    const input = fs.createReadStream(manifestPath, {encoding: 'utf-8'});
+    const manifest = readline.createInterface({input});
+
+    manifest.on('line', line => {
```
mprobst: what happens when reading fails here?

internal/node_loader.js
```diff
@@ -94,13 +110,17 @@ module.constructor._resolveFilename =
   throw error;
 }
 
+let runfilesManifest;
```
mprobst: docs?

internal/node_loader.js
```diff
@@ -94,13 +110,17 @@ module.constructor._resolveFilename =
...
   // NB: entry_point below is replaced during the build process.
   var mainScript = process.argv[1] = 'TEMPLATED_entry_point';
   try {
-    module.constructor._load(mainScript, this, /*isMain=*/true);
+    loadRunfiles(process.env.RUNFILES).then(mf => {
+      runfilesManifest = mf;
+      module.constructor._load(mainScript, this, /*isMain=*/true);
+    });
   } catch (e) {
     console.error('failed to load main ', e.stack || e);
```
mprobst: now that the code above is async, won't this exit the catch block before anything happens?

internal/node_launcher.sh
```diff
@@ -46,16 +46,16 @@ esac
 
 if [[ -n "$TEST_SRCDIR" ]]; then
   # Case 4, bazel has identified runfiles for us.
```
mprobst: the docs talk about finding our runfiles tree, but this is finding our manifest. Update?

internal/node_launcher.sh
alexeagle: Updated to use only bash, and test for equality of the templated bit, so it should not need escaping

internal/node_launcher.sh
alexeagle: done

internal/node_loader.js
mprobst: Maybe this helps?

https://stackoverflow.com/questions/21771220/error-handling-with-node-js-streams

internal/node_launcher.sh
```diff
@@ -92,4 +92,20 @@ done
...
-exec "${RUNFILES}/TEMPLATED_node" "${NODE_OPTIONS[@]}" "${RUNFILES}/TEMPLATED_script_path" "${ARGS[@]}"
+# Lookup the real paths from the runfiles manifest with no dependency on posix
+while read line; do
+  declare -a PARTS=($line)
+  if [ "${PARTS[0]}" == "TEMPLATED_node" ]; then
+    readonly node="${PARTS[1]}"
+  elif [ "${PARTS[0]}" == "TEMPLATED_script_path" ]; then
+    readonly script="${PARTS[1]}"
+  fi
+done < ${RUNFILES}
```
mprobst: quotes around runfiles?

mprobst:
You haven't resolved a bunch of comments from me, and also haven't waited for an LGTM. Was that intentional?

internal/node_loader.js
alexeagle: I don't think that applies to the `manifest` object here, the return of `readline.createInterface` is not a stream. But I think I could listen for `error` events on the underlying stream?
It really needs unit tests, I don't know how to verify the error handling. Maybe it's time for `jasmine_node_test`

PR #16, opened 2017-09-12
address code review comments

alexeagle:
/cc @mprobst

### DIFF
--- a/internal/node_loader.js
+++ b/internal/node_loader.js
@@ -24,7 +24,6 @@
 'use strict';
 var path = require('path');
 var fs = require('fs');
-var readline = require('readline');
 
 /**
  * The module roots as pairs of a RegExp to match the require path, and a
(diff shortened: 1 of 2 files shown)


internal/node_loader.js
```diff
@@ -68,14 +68,13 @@ function loadRunfilesManifest(manifestPath) {
...
+      input.on('error', reject);
       const manifest = readline.createInterface({input});
       manifest.on('line', line => {
         if (!line) return;
         const [runfilesPath, realPath] = line.split(" ");
         runfilesManifest[runfilesPath] = realPath;
-      }).on('close', () => {
-        resolve(runfilesManifest);
-      });
+      }).on('close', resolve);
```
mprobst: Maybe document that this is a void promise? loadRunfilesManifest sounds as if it should resolve with the manifest.

alexeagle: Actually I don't like that it's a void promise. Changed to return the manifest, which starts out undefined, and is set before bootstrapping.

PR #19, opened 2017-09-12
Add a check for minimum bazel version.

alexeagle:
Downstream repos (eg. typescript, Angular) may have their own minimal version constraint,
so expose a function they can use.

### DIFF
--- a/internal/check_bazel_version.bzl
+++ b/internal/check_bazel_version.bzl
@@ -1,0 +1,50 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# From https://github.com/tensorflow/tensorflow/blob/5541ef4fbba56cf8930198373162dd3119e6ee70/tensorflow/workspace.bzl#L44
+
+# Parse the bazel version string from `native.bazel_version`.
+# The format is "<major>.<minor>.<patch>-<date> <commit>"
+# Returns a 3-tuple of numbers: (<major>, <minor>, <patch>)
+def _parse_bazel_version(bazel_version):
+  # Remove commit from version.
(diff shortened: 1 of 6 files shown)


internal/check_bazel_version.bzl
```diff
@@ -0,0 +1,47 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# From https://github.com/tensorflow/tensorflow/blob/5541ef4fbba56cf8930198373162dd3119e6ee70/tensorflow/workspace.bzl#L44
+
+# Parse the bazel version string from `native.bazel_version`.
+def _parse_bazel_version(bazel_version):
+  # Remove commit from version.
```
mprobst: it'd make this more readable if you could give an example of the pattern you're parsing and of what you're returning.

internal/check_bazel_version.bzl
```diff
@@ -0,0 +1,47 @@
...
+    fail("\nCurrent Bazel version is lower than 0.2.1, expected at least %s\n" %
+         bazel_version)
+  elif not native.bazel_version:
+    print("\nCurrent Bazel is not a release version, cannot check for " +
+          "compatibility.")
+    print("Make sure that you are running at least Bazel %s.\n" % bazel_version)
+  else:
+    current_bazel_version = _parse_bazel_version(native.bazel_version)
+    minimum_bazel_version = _parse_bazel_version(bazel_version)
+    if minimum_bazel_version > current_bazel_version:
```
mprobst: `('1', '12', '3') < ('1', '2', '1')` is a problem, isn't it?

alexeagle: sounds like it needs unit tests! let's do some proper engineering :)

internal/check_bazel_version.bzl
alexeagle: good point

internal/check_bazel_version.bzl
```diff
@@ -0,0 +1,50 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# From https://github.com/tensorflow/tensorflow/blob/5541ef4fbba56cf8930198373162dd3119e6ee70/tensorflow/workspace.bzl#L44
+
+# Parse the bazel version string from `native.bazel_version`.
+# The format is "<major>.<minor>.<patch>-<date> <commit>"
+# Returns (<major>, <minor>, <patch>)
```
mprobst: ... as numbers?

PR #24, opened 2017-09-20
add bazel CI

alexeagle:


alexeagle:
Going to merge this now even though later PRs will be red, because it's a high priority to fix it, and other contributors don't seem to trigger the bazel CI even if they include this commit.

PR #27, opened 2017-09-21
add jasmine_node_test

### LINKED ISSUE #26: Provide jasmine_node_test rule
API should be like

```
jasmine_node_test(
    name = "test",
    srcs = ["foo.js", ":rule_that_produces_js"],
    data = [":some_nodejs_binary_rule"],
)
```

alexeagle:
fixes #26


.circleci/config.yml
```diff
@@ -3,18 +3,23 @@ jobs:
...
       - checkout
 
       - restore_cache:
           key: angular_bazel_example-{{ .Branch }}-{{ checksum "examples/rollup/yarn.lock" }}
       - run: bazel info release
+
       - run: bazel run @yarn//:yarn
-      - run: bazel build ...
-      - run: bazel test ...
+      - run: bazel build -- ... -examples/program/...
```
LucasSloan: Add a comment why the `-example/program/...`

examples/program/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
...
+        "node_modules/**/*.js",
+        "node_modules/**/*.d.ts",
+        "node_modules/**/*.json"
+    ]),
+    visibility = ["//visibility:public"],
+)
+
+jasmine_node_test(
+    name = "program",
+    srcs = glob(["*.js"]),
```
LucasSloan: As a developer looking at this example, I'd like an explanation of how jasmine_node_test handles sources vs. tests, which this glob doesn't really cover.

examples/program/WORKSPACE
```diff
@@ -0,0 +1,21 @@
+workspace(name = "program_example")
```
LucasSloan: Better name than program?  testable?

internal/devmode_js_sources.bzl
```diff
@@ -0,0 +1,42 @@
...
+  for d in ctx.rule.attr.deps:
+    if DevmodeJs in d:
+      sources += d[DevmodeJs].sources
+  sources += ctx.rule.attr.files
+  return [DevmodeJs(sources = sources)]
+
+def _devmode_js_sources_impl(ctx):
+  files = depset()
+  files += ctx.files.deps
+  # FIXME: needs TDD
```
LucasSloan: FIXME

examples/program/WORKSPACE
alexeagle: I'm planning to extend this example to cover a couple other missing things related to running code that you author under nodejs_binary (to reproduce problems currently seen only in repos that consume this) so I think this name will be okay.

.circleci/config.yml
alexeagle: :+1:

examples/program/BUILD.bazel
alexeagle: Good point, added a separate "library" rule as a filegroup, added that as a dep, so the srcs here are just *.spec.ts

alexeagle:
PTAL

It occurs to me that we should never add another aspect for gathering up the manifest file, just re-use the same aspect that nodejs_binary has for locating sources. (In a subsequent PR)

Issue #25, opened 2017-09-21
Manifest file lookup happens outside of node module resolution


alexeagle:
So if node is looking for 
`pkgA`
we fail to find that in the manifest because the entry is
`pkgA.js`.
Then we don't give any starting path for node to look in. Somehow we need to do the manifest lookup lower down, when node is stat'ing files on disk.

/cc @dslomov

alexeagle:
@hansl tell me more about how you think we could fix this...

hansl:
Our code is basically the following. Note that outside of some magic to add global checks and custom resolution dir, this is pretty much the same logic as Node itself.

```typescript
export function resolve(x: string, options: ResolveOptions): string {
  // We have more logic for options here.
  const basedir = options.basedir || path.dirname(_findPathOfCaller());
  if (isRelative(x)) {
    let res = path.resolve(basedir, x);
    if (x == '..' || x.slice(-1) == '/') {
      res += '/';
    }

    const maybe = _loadAsFileSync(res) || _loadAsDirectorySync(res);
    if (maybe) {
      return maybe;
    }
  } else {
    const maybe = _loadNodeModulesSync(x, basedir);
    if (maybe) {
      return maybe;
    }
  }
  throw new ModuleNotFoundException(x, basedir);
}

function _loadAsFileSync(res: string) {
  if (isFile(res)) {
    return res;
  }
  return require.extensions.map(ex => res + ex).find(x => isFile(x)) || null;
}

function _loadAsDirectorySync(res: string) {
  const pkgfile = path.join(res, 'package.json');
  if (isFile(pkgfile)) {
    try {
      const body = fs.readFileSync(pkgfile, 'UTF-8');
      const pkg = JSON.parse(body);
        if (pkg['main']) {
          if (pkg['main'] === '.' || pkg['main'] === './') {
            pkg['main'] = 'index';
          }

          let maybe = _loadAsFileSync(path.resolve(res, pkg['main']));
          if (maybe) {
            return maybe;
          }
          maybe = _loadAsDirectorySync(path.resolve(res, pkg['main']));
          if (maybe) {
            return maybe;
          }
        }
      }
    } catch (e) {}
  }
  return _loadAsFileSync(path.join(res, 'index'));
}

function _loadNodeModulesSync(x: string, start: string): string | null {
  const dirs = _nodeModulesPaths(start, options);
  for (const dir of dirs) {
    const m = _loadAsFileSync(path.join(dir, x));
    if (m) {
[...]

PR #28, opened 2017-09-22
Workaround #25

alexeagle:
Look for foo.js in the runfiles manifest when asked for foo

### DIFF
--- a/internal/node_loader.js
+++ b/internal/node_loader.js
@@ -85,7 +85,9 @@ function resolveRunfiles(...pathSegments) {
   if (runfilesManifest) {
     // Join on forward slash, because even on Windows the runfiles_manifest
     // file is written with forward slash.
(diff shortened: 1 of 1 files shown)


alexeagle:
it's a small workaround so I'm going to merge without review this time to unblock angular. Leaving the issue open to find a better solution long-term

ocombe:
Ah sorry, got caught up on finishing another PR.
My only question was: ` return runfilesManifest[runfilesEntry] || runfilesManifest[runfilesEntry + '.js'];` --> do we need to worry about other extensions than .js ?

alexeagle:
The right thing to do is to hook underneath node's resolution, so that any
time node looks on disk for a file, we intercept that and use the manifest
file as an indirection.
But I haven't figured out how to do that yet. Probably in general we might
need to look things like runfilesEntry + '/index.js'.
Leaving the issue open...

PR #30, opened 2017-09-28
Remove empty path segments before looking up runfile in manifest

appsforartists:
`resolveRunfiles([ 'karma', '', 'node_modules', 'karma/bin/karma' ])` was looking for `'karma//node_modules/karma/bin/karma'`, which is not a valid path.

See also alexeagle/rules_karma#3.  This doesn't fix that, but it makes progress.

### DIFF
--- a/internal/node_loader.js
+++ b/internal/node_loader.js
@@ -82,15 +82,21 @@ function loadRunfilesManifest(manifestPath) {
 const runfilesManifest = loadRunfilesManifest(`${process.env.RUNFILES}_manifest`);
 
 function resolveRunfiles(...pathSegments) {
+  // Remove any empty strings from pathSegments
+  pathSegments = pathSegments.filter(segment => segment);
+
+  const defaultPath = path.join(process.env.RUNFILES, ...pathSegments);
+
   if (runfilesManifest) {
-    // Join on forward slash, because even on Windows the runfiles_manifest
-    // file is written with forward slash.
+    // Join on forward slash, because even on Windows the runfiles_manifest file
+    // is written with forward slash.
     const runfilesEntry = pathSegments.join('/');
+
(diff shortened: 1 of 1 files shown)


appsforartists:
This is the simplest fix, so it's the one I PRed, but I wonder if there's a more comprehensive solution.

The current code assumes that is the `runfilesManifest` exists, the path will be in there.  Otherwise, it looks at a default path.  Should it be rewritten to fallback to the default path if the lookup fails?  Something like:

```javascript
function resolveRunfiles(...pathSegments) {
  // Remove any empty strings from pathSegments
  pathSegments = pathSegments.filter(segment => segment);

  const defaultPath = path.join(process.env.RUNFILES, ...pathSegments);

  if (runfilesManifest) {
    // Join on forward slash, because even on Windows the runfiles_manifest file
    // is written with forward slash.
    const runfilesEntry = pathSegments.join('/');

    // Add .js as a workaround for https://github.com/bazelbuild/rules_nodejs/issues/25
    return runfilesManifest[runfilesEntry] || runfilesManifest[runfilesEntry + '.js'] || defaultPath;
  }

  return defaultPath;
}
```

appsforartists:
Whoa - I tried running the more comprehensive fix, and it looks like Karma works now!  Amending the PR.

alexeagle:
Filtering empty path segments seems fine, but that seems like a bug further up, who creates the empty path?

My typical approach is likehttps://github.com/bazelbuild/rules_nodejs/blob/master/internal/jasmine_node_test.bzl#L13
to avoid writing such paths.

For the issue with the runfiles manifest - see the commit Lucas just landed - this swaps the logic so we only look at the runfiles_manifest if the runfiles directory is absent, which means only on Windows.

We'll still have to fix these bugs, but now they affect windows usage only.
The proper fix for https://github.com/bazelbuild/rules_nodejs/issues/25 makes me very scared.

PR #31, opened 2017-09-29
Only use the runfiles manifest on windows.

LucasSloan:
If we're on linux or macos, symlinks work, so we don't need to use the runfiles manifest.  Therefore, if the runfiles directory exists, we can ignore the runfiles manifest.

### DIFF
--- a/internal/node_loader.js
+++ b/internal/node_loader.js
@@ -67,7 +67,9 @@ function resolveToModuleRoot(path) {
  * See https://github.com/bazelbuild/bazel/issues/3726
  */
 function loadRunfilesManifest(manifestPath) {
-  if (!fs.existsSync(manifestPath)) {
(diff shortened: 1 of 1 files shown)


bazel-io:

internal/node_loader.js
```diff
@@ -67,6 +67,11 @@ function resolveToModuleRoot(path) {
  * See https://github.com/bazelbuild/bazel/issues/3726
  */
 function loadRunfilesManifest(manifestPath) {
+  // If the runfiles directory exists, we're not running on Windows, so
+  // we don't need the runfiles manifest
+  if (fs.existsSync(process.env.RUNFILES)) {
+    return;
+  }
   if (!fs.existsSync(manifestPath)) {
```
alexeagle: either remove this second conditional, or make it throw an error if the manifestPath doesn't exist.

PR #37, opened 2017-10-11
Add directory of current BUILD file to module resolve locations

### LINKED ISSUE #32: Support for entry points relative to current BUILD file directory
Currently, users provide the full path to a local js file to use it as an entry point, including the workspace name. It's a bit of extra typing, but it's workable.

This gets a bit more cumbersome for nodejs_binary using entry points that aren't at fixed locations e.g. from other rules in macros. (FWIW this isn't an issue for rules e.g. `ctx.workspace_name + "/" + ctx.file.foo.path`, but it's unrelated because rules can't call other rules.)

Below is an example: we have an intermediary rule which outputs a js file, and we want the nodejs_binary to use it as an entry point. But the macro doesn't know the path to the js file, so we have to get the user to specify the path.

``` python
def foo_rule(
        name,
        # "workspace_name/path/to/BUILD/directory/"
        current_path,
        visibility = None):

    # Intermediary rule outputs "{name}.js".
    intermediary_name = "%s.intermediary" % name
    intermediary_rule(
        name = intermediary_name,
    )

    # Nodejs rule combines current path with known output name
    nodejs_binary(
        name = name,
        data = [
            "@//:node_modules",
            intermediary_name,
        ],
        entry_point = "%s%s.js" % (current_path, intermediary_name),
        visibility = visibility,
    )
```

# Proposal

Add the current path to the node module resolver's list of resolve paths.

The above entry point can be changed to a relative path:

```python
    nodejs_binary(
        # ...
        entry_point = "%s.intermediary.js" % name,
```

## Caveats

Files supplied in `data` would clash with node_modules. Convoluted example:

``` python
    nodejs_binary(
        # ...
        data = [
            "@//:node_modules",
            "some_node_module/foo.js",
        ],
        entry_point = "example.js",
    )
```

```javascript
const foo = require('some_node_module/foo.js'); // Which one?
```

# Alternatives considered

## Change entry_point to be [`attr.label`](https://docs.bazel.build/versions/master/skylark/lib/attr.html#label) instead of attr.string

Ruled out because this would break references to node_modules.

## Split entry_point into two attributes, one attr.label for local references, one attr.string for node_modules references

Not as straightforward for users.

## Split nodejs_binary into two rules, one for local entry points, one for node_modules entry points

Not as straightforward for users.

rzhw:
Closes #32.

Also adds an example that uses a relative entry point path. Fails CI because multiple node repositories isn't possible -- resolve by using a single root package.json? (Or split into separate PR)


bazel-io:

alexeagle:
I think we want #42 instead

PR #36, opened 2017-10-06
Fix Windows CI

alexeagle:

PR #38, opened 2017-10-13
Support testing targets in different workspaces.

tbosch:
@alexeagle How do test this?

### DIFF
--- a/internal/jasmine_node_test.bzl
+++ b/internal/jasmine_node_test.bzl
@@ -8,8 +8,10 @@ def jasmine_node_test(name, srcs, data = [], args = [], deps = [], **kwargs):
       testonly = 1,
   )
 
+  # Note: REPOSITORY_NAME is something like "@angular"
   args = ["/".join([p
-      for p in [".", PACKAGE_NAME, "%s_devmode_srcs.MF" % name]
+      for p in [".", "external" if len(REPOSITORY_NAME)>1 else "", REPOSITORY_NAME[len("@"):],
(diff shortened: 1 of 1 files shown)


bazel-io:

alexeagle:
I think the subtraction pattern here is why we didn't have test coverage for this:
https://github.com/bazelbuild/rules_nodejs/blob/master/.circleci/config.yml#L18
this was to work around issues with nested WORKSPACEs but I think we just need to correctly understand how they work.
(note that the Windows PR #36 is planning to drop the nested WORKSPACE as we also ran into issues there, hopefully learning how they are meant to work means we can remove that part of the PR)

internal/jasmine_node_test.bzl
```diff
@@ -9,7 +9,8 @@ def jasmine_node_test(name, srcs, data = [], args = [], deps = [], **kwargs):
   )
 
   args = ["/".join([p
-      for p in [".", PACKAGE_NAME, "%s_devmode_srcs.MF" % name]
+      for p in [".", "external" if REPOSITORY_NAME else "", REPOSITORY_NAME[1:],
```
alexeagle: better to show what character is being stripped, either add a comment, or maybe
`REPOSITORY_NAME[len("."):]` to show which prefix we expect to be removed
or both
(isn't this a bug though, don't you want to split on `/` and take the second part?)

tbosch: Will change to use `len("@")`.
`REPOSITORY_NAME` is something like `@angular`.

tbosch:
@alexeagle PTAL

PR #41, opened 2017-10-23
Fix += for lists in jasmine_node_test

dslomov:
'+=' for lists now mutates the lhs list instead of creating a new one and assigning to lhs (https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#mutating).

This is a fix for https://ci.bazel.io/blue/organizations/jenkins/Global%2Frules_nodejs/detail/rules_nodejs/56/pipeline/


dslomov:
Friendly pin @alexeagle

alexeagle:
Thanks!

PR #42, opened 2017-10-26
Add support for $(location) in nodejs_binary#entry_point

alexeagle:


alexeagle:
@vladmos I would appreciate an opinion from the Bazel team: I'm using `$(location :label)` expansion in a different way than intended I think - I want these to reference output locations in the RUNFILES.
According to https://docs.bazel.build/versions/master/be/make-variables.html#location it should give paths relative to the runfiles of the _bin or _test target, but I find I have to do munging on the path.

alexeagle:
@rzhw any comments? I'm still a bit unsure if users should be exposed to the `$(location)` syntax as I've done here.
Another possibility would be to allow either string or label in the `entry_point` attr, and if the value starts with `@`, `:`, or `//` then treat it as requiring location expansion (and also maybe add it to `data` automatically)

examples/program/BUILD.bazel
```diff
@@ -13,21 +27,51 @@ filegroup(
...
+    node_modules = "//:node_modules",
+)
+
+# An end-to-end test for the program.
+# It's written in Python just because this is convenient and cross-platform.
+# See https://docs.bazel.build/versions/master/be/python.html#py_test
+py_test(
+    name = "bin_test",
+    srcs = ["bin_test.py"],
+    deps = ["@build_bazel_rules_nodejs//internal:runfiles"],
```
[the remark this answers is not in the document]
rzhw: Whoops, didn't realise this example was a separate workspace.

rzhw:
bazelbuild/bazel#2745 looks relevant, particularly [this proposal](https://github.com/bazelbuild/bazel/issues/2475#issuecomment-339296177). `$(rootpath)` would let `expand_path_into_runfiles` be deprecated. Though it still leaves the concern of exposing users to that syntax in the first place.

alexeagle:
Awesome job tracking down that feature request. It looks like Ulf is close to introducing the `$(rootpath)` expansion.
Since that change would affect how users write their `entry_point`, I think I should hold this PR for a couple weeks to see if that lands and is released.

rzhw:
Looks like `$(rootpath)` has [landed in 0.8.0](https://github.com/bazelbuild/bazel/releases/tag/0.8.0).

Tests [here](https://github.com/bazelbuild/bazel/blob/master/src/test/java/com/google/devtools/build/lib/analysis/LocationExpanderIntegrationTest.java#L79), doesn't appear to test for e.g. `$(rootpath @other_wksp//:thing)`

rzhw:
What's currently blocking this from landing?

I've tried rebasing once but looks like I'm hitting some conflicts again :cry:

alexeagle:
Nothing blocking this other than prioritization. I haven't had time, would appreciate some help!

(Also I'm switching to Windows to force myself to make all the Bazel + Angular builds work on Windows so it's going to be slow going for a while....)

rzhw:
Gotcha, what's needed after resolving the merge conflicts if any? Verifying it works on Windows, design input from someone else from Bazel?

alexeagle:
Sorry @rzhw just never prioritized this one. Dusting it off now...

nimerritt:
@alexeagle Any reason this can't be merged? I could help out if it's a time thing. It is confusing for others in my team who don't know the details of the rules_nodejs implementation to understand what this field should be. Using a build target instead hides the implementation details better.

gregmagolan:
@nimerrit This has been collecting some dust. Its I've been looking at related code recently with https://github.com/bazelbuild/rules_nodejs/pull/240. I'll discuss this PR with @alexeagle when he's back from vacation next week and to see what else needs to be done so that it can get in. It would be a good one to land for users like yourself.

internal/common/expand_into_runfiles.bzl
```diff
@@ -1,55 +0,0 @@
...
-    path: the path to expand
-
-  Returns:
-    The expanded path or the original path
-  """
-  if path.find('$(location') < 0:
-    return path
-  return expand_path_into_runfiles(ctx, path)
-
-def expand_path_into_runfiles(ctx, path):
```
gregmagolan: expand_path_into_runfiles is used in rules_typescript so likely need to keep it around as deprecated

internal/node/node_loader.js
```diff
@@ -221,7 +221,12 @@ module.constructor._resolveFilename =
   var failedResolutions = [];
   var resolveLocations = [
     request,
-    resolveRunfiles(request),
+    // See https://github.com/bazelbuild/rules_nodejs/issues/84 for discussion of paths
+    // The request should have either external/other_wksp/path/to/thing
+    // or just path/to/thing with the implied local workspace
+    // This matches Bazel semantics for cwd of toolchain processes
+    // and helpers like ctx.expand_location("$(rootpath foo)")
+    resolveRunfiles('TEMPLATED_user_workspace_name', request),
```
gregmagolan: This seems make sense. It should also continue to work with https://github.com/bazelbuild/rules_nodejs/pull/240.

nimerritt:
@alexeagle @gregmagolan Any updates on this? I'm trying to calculate the entry_point in a macro and can't figure out how to do it currently. The discussion in https://github.com/bazelbuild/rules_nodejs/issues/32 would have me believe that I can use `native.respitory_name` but that actually returns `@` and not the required `rules_nodejs`. I had it hard-coded for the specific repository to work around this, but now I'm consuming my rules from another workspace, so this is now more important. Any suggestions? Do you need help with this PR?

Globegitter:
Given a discussion here: https://github.com/bazelbuild/bazel/issues/6001 it seems that entry_point should not be a string and not use the `$(location x)` syntax but rather be a label. I think it also makes the most sense to me and it should be able to make this work with node_modules also.

PR #43, opened 2017-10-30
Add missing test for arguments to node programs

alexeagle:
This would have prevented the need to revert a commit:
https://github.com/bazelbuild/rules_nodejs/commit/404c4b0c096cd438d1f190a2a0d8da59649b52f3


alexeagle:
This commit is included in #42 let's just review there

PR #48, opened 2017-11-03
Revert "Add documentation clarifying which node_modules are used"

alexeagle:
This reverts commit 6b9979b5c1b7f213acdfcf29fa258f76be1ec055.

The idea here is misguided, it depends on how the package manager
lays out the node_modules.
Just because @bazel/typescript depends on typescript doesn't mean
we can find node_modules/@bazel/typescript/node_modules/typescript/...
(relative to node_modules/@bazel/typescript/WORKSPACE)
because the user might also depend on typescript at the same
version, and there will only be
node_modules/typescript
(not under the @build_bazel_rules_typescript workspace)

### DIFF
--- a/internal/node.bzl
+++ b/internal/node.bzl
@@ -53,12 +53,37 @@ def _write_loader_script(ctx):
           "TEMPLATED_module_roots": "\n  " + ",\n  ".join(module_mappings),
           "TEMPLATED_entry_point": ctx.attr.entry_point,
           "TEMPLATED_label_package": ctx.attr.node_modules.label.package,
-          # If the label being built is in another workspace, look for runfiles
-          # produced by that workspace
-          "TEMPLATED_workspace_name": (
+          # There are two workspaces in general:
+          # A) The user's workspace is the one where the bazel command is run
+          # B) The label's workspace contains the target being built/run
+          #
+          # If A has an npm dependency on B, then we'll look in the node_modules
+          # to find B's dependency D. It could be in two different places
+          # depending on hoisting [1]:
+          # A/node_modules/B/node_modules/D
+          # A/node_modules/D
+          # That means we must resolve runfiles relative to A.
+          #
+          # However if A has a bazel dependency on B, then B is not under A's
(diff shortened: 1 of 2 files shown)


internal/node.bzl
```diff
@@ -53,12 +53,35 @@ def _write_loader_script(ctx):
...
-          # If the label being built is in another workspace, look for runfiles
-          # produced by that workspace
-          "TEMPLATED_workspace_name": (
+          # There are two workspaces in general:
+          # A) The user's workspace is the one where the bazel command is run
+          # B) The label's workspace contains the target being built/run
+          #
+          # If A has an npm dependency on B, then we'll look in the node_modules
+          # to find B's dependency D. It could be in two different places
+          # depending on hoisting:
```
rkirov: I tried to find a good reference for 'module hoisting' but all I got this sentence in https://docs.npmjs.com/cli/install

'If <folder> sits inside the root of your project, its dependencies may be hoisted to the toplevel node_modules as they would for other types of dependencies.'

Not quite the same use case. Are you sure 'hoisting' is the right term?

internal/node.bzl
```diff
@@ -152,15 +175,6 @@ _NODEJS_EXECUTABLE_ATTRS = {
         allow_files = True,
         single_file = True),
     "node_modules": attr.label(
-        # We expect most users declare a binary/test and run it within the same
```
rkirov: Why remove this comment? Maybe you don't stand by its recommendation "should override", but I think it is still worth documenting what the attr node_modules is, why it is tricky to pick a "correct" value in presence of npm hoisting.

internal/node.bzl
alexeagle: Yeah, seems like hoisting is the term, might be yarn only. Will link to https://yarnpkg.com/lang/en/docs/workspaces/#toc-limitations-caveats

internal/node.bzl
alexeagle: Updated the comment rather than remove it.

PR #50, opened 2017-11-10
devmode_js_sources: make public

alexeagle:
Also fix it: default output is .MF and the other .js files are collected in runfiles

This should have some local testing within this repo before merging...

### DIFF
--- a/internal/node.bzl
+++ b/internal/node.bzl
@@ -99,7 +99,7 @@ _NODEJS_EXECUTABLE_ATTRS = {
         cfg = "data",
         aspects=[_sources_aspect, module_mappings_runtime_aspect]),
     "_node": attr.label(
-        default = "@nodejs//:node",
+        default = Label("@nodejs//:node"),
         allow_files = True,
         single_file = True),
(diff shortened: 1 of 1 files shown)


alexeagle:
CI failure is because previously the manifest had lines like `build_bazel_rules_nodejs/external/program_example/index.spec.js` but now it's `build_bazel_rules_nodejs/../program_example/index.spec.js`
the correct path should be
`build_bazel_rules_nodejs/external/program_example/index.spec.js` in order to look it up in the manifest file.

alexeagle:
bazel 0.8 is currently at RC2 and adds
> runfiles, sh: Shell scripts may now depend on //src/tools/runfiles:runfiles_sh_lib and source runfiles.sh. The script defines the rlocation function which returns runfile paths on every platform.
> In addition to $(location), Bazel now also supports $(rootpath) to obtain the root-relative path (i.e., for runfiles locations), and $(execpath) to obtain the exec path (i.e., for build-time locations)

So I think we should just wait until we can use that.

PR #52, opened 2017-11-14
Change node module load order

alexeagle:
Try looking in the directory containing the label first, before the one in the working directory.
Needed for tsickle tests to use a different TypeScript version in one test, and probably generally useful.


alexeagle:
This commit broke loading in another case (a deep require() chain under karma).
Also I can no longer reproduce the need for it in tsickle, either locally or on CI.
I think I am missing some understanding of how this loading works, in the meantime I'll revert this.

PR #53, opened 2017-11-14
Set Jasmine timeout to the same as Bazel

alexeagle:
Also remove a stray console.log

### DIFF
--- a/internal/jasmine_runner.js
+++ b/internal/jasmine_runner.js
@@ -27,8 +27,27 @@ function main(args) {
   }
 
   const jrunner = new JasmineRunner();
+
+  // Unless the user overrides the timeout for a test, base the Jasmine timeout
+  // on the Bazel one.
+  //
+  // Jasmine's timeout is per-test, not for the whole process. Also Jasmine
+  // takes additional time to start. So setting Jasmine's timeout the same as
+  // Bazel's should guarantee that users will see Bazel's timeout status, not a
+  // failed status due to Jasmine internally throwing
+  // "Async callback was not invoked within timeout specified by
+  // jasmine.DEFAULT_TIMEOUT_INTERVAL"
+  // This allows users to have a simpler mental model between languages: test
+  // timeouts look similar, and you can increase the `timeout` attribute on any
+  // `*_test` target to give it more time.
+  //
+  // This environment variable is set by Bazel, see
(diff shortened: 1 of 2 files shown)


internal/jasmine_runner.js
```diff
@@ -27,8 +27,27 @@ function main(args) {
   }
 
   const jrunner = new JasmineRunner();
+
+  // Unless the user overrides the timeout for a test, base the Jasmine timeout
+  // on the Bazel one.
+  //
+  // Jasmine's timeout is per-test, not for the whole process. Also Jasmine
+  // takes additional time to start. So setting Jasmine's timeout the same as
```
mprobst: *to* the same ...

internal/jasmine_runner.js
```diff
@@ -27,8 +27,27 @@ function main(args) {
...
+  // Bazel's should guarantee that users will see Bazel's timeout status, not a
```
mprobst: *as* Bazel's

internal/jasmine_runner.js
```diff
@@ -27,8 +27,27 @@ function main(args) {
...
+  // failed status due to Jasmine internally throwing
+  // "Async callback was not invoked within timeout specified by
+  // jasmine.DEFAULT_TIMEOUT_INTERVAL"
+  // This allows users to have a simpler mental model between languages: test
```
mprobst: So when you hit the bazel timeout, you'll just get a terminated process. When you hit Jasmine's timeout, you'll get a stack trace and a message from Jasmine what failed, and potentially the genesis of the Promise that failed to resolve.

I think it'd be a better developer UX to reduce this so that users will see Jasmine's timeout at least most of the time. E.g. set to timeout - 5s or so so that usually you hit Jasmine's timeout first.

alexeagle: Hmm, if you want to see Jasmine's message, then when the timeout is too short, you'll add a jasmine timeout. Users wouldn't know about the `timeout` attribute.
This makes me thing it's already WAI - you really do want two different timeout mechanisms.

mprobst: Yeah, come to think of it, that's actually maybe fine. We just happen to have two different timeouts, one for the suite, one for individual tests, and that's WAI.

Might still be worth documenting that?

PR #54, opened 2017-11-16
Upgrade to node 8, npm 5, yarn 1.3

alexeagle:
/cc @jelbourn


alexeagle:
Note, breaking change so this should be tagged with 0.3.0

See #4

PR #55, opened 2017-11-16
Handle version that looks like "x.x.xrcx" in check_bazel_version.bzl

gregmagolan:
0.8.0rc2 is labeled like this


bazel-io:

alexeagle:
Have time to add the missing unit tests?

PR #56, opened 2017-11-18
If release is an RC, pretend it's the final release

achew22:


gregmagolan:
Looks like you solved the same issue I did in #55 in parallel.

achew22:
Huh, I should look at the existing PRs before I do stuff like this in the future. I'll leave it to @alexeagle on which one to merge and which to revert. They are subtly different but I doubt it matters.

TBH this seems like a thing that should be upstreamed into `@bazel_tools` so individual projects don't have to write it any more.

alexeagle:
this thing is actually unit tested (unit testing a .bzl file? does such a thing really exist?)
so I'll merge whichever PR adds the test :)

achew22:
@gregmagolan and @alexeagle just for your reference Skylib has a [unittest](https://github.com/bazelbuild/bazel-skylib/blob/master/lib/unittest.bzl) library. I don't think I'll be able to get to this today, sorry.

alexeagle:
I see, the unittest thing is pushed, I wonder if it is released yet.

alexeagle:
going with #55 instead, thanks!

PR #61, opened 2017-11-30
Add an express example.

mattmoor:
This also adds a trivial non-http example (which I can delete if you prefer).

### DIFF
--- a/defs.bzl
+++ b/defs.bzl
@@ -21,4 +21,4 @@ load("//internal:node.bzl",
      nodejs_binary = "nodejs_binary_macro",
      nodejs_test = "nodejs_test_macro")
 load("//internal:jasmine_node_test.bzl", "jasmine_node_test")
-load("//internal:node_install.bzl", "node_repositories")
+load("//internal:node_install.bzl", "node_repositories", "npm_install")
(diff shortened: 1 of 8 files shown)


alexeagle:
The contents look fine, but what's the objective? There is already an example of writing a program in js that links against a third-party library. If we add an express example, does that mean we ought to add an example for every third-party library? Or does this illustrate something the rollup example does not?

mattmoor:
Writing web apps seems pretty core to Node.js.

If you don't think it's useful here, I'm happy to discard.  It'll probably end up being the basis for a `node_image` K8s "http" example / test [here](https://github.com/bazelbuild/rules_k8s/tree/master/examples/hellohttp).

... that is once I use it to figure out how to properly build `node_image` :)

laat:
npm_install could be made more reproducible with a parameter specifying a `package-lock.json` file

mattmoor:
@laat Sorry, I updated this branch without realizing we'd left this open.  See the discussion [here](https://github.com/bazelbuild/rules_nodejs/issues/63), but it is very much my intention that folks could (and probably would) use lock files with that rule.

Issue #63, opened 2017-12-01
repository rule for node_modules


mattmoor:
Currently these rules support accessing `node_modules` from some place instantiated within the source tree.  This is consistent with Node.js user expectations and makes sense to offer users as an option.

However, requiring `node_modules` within the source tree is inconsistent with the way just about all other Bazel tools leverage `WORKSPACE`, which enables a user to effectively clone the repo and `bazel {build,test} ...` with no additional setup.

Essentially, we offer a Node.js idiomatic option, but not a Bazel idiomatic option.

I would like to propose one small addition to this repository, which is to offer a mechanism by which users can choose to instantiate the `:node_modules` `filegroup` during `WORKSPACE` evaluation.

The idea would be that users could still opt to use today's approach, where they instantiate `node_modules` outside of `WORKSPACE`, then expose and reference it via:
```python
=== BUILD ===
# Expose
filegroup(name = "node_modules", srcs = glob(["node_modules/**/*"]))

# Reference
nodejs_binary(
    name = "bar",
    node_modules = "//:node_modules",
    ...
)
```

Alternately users could write:
```python
=== WORKSPACE ===
# Expose (or yarn_install)
npm_install(
    name = "foo",
    packages = ":package.json",  # Or the lock-file
)

=== BUILD ===
# Reference
nodejs_binary(
    name = "bar",
    node_modules = "@foo//:node_modules",
    ...
)
```

I have a working prototype of this [here](https://github.com/mattmoor/rules_nodejs-1/blob/express-example/examples/express/BUILD.bazel#L5) that I've validated with my `express` "hello world" app.  I believe beyond the new rule definitions I only needed one small change to `internal/node.bzl`.

If you'd be receptive to adding this as an option, I'd be receptive to trying to clean it up and get it upstreamed.  WDYT?

@jonjohnsonjr @davidstanke @pubref @samertm FYI

alexeagle:
This landed a while ago

PR #67, opened 2017-12-02
Working npm_install rule with example and test.

mattmoor:
Fixes: https://github.com/bazelbuild/rules_nodejs/issues/63


mattmoor:
I think I need to prefix commands with `node`...

internal/packages.bzl
```diff
@@ -0,0 +1,52 @@
...
+# limitations under the License.
+
+"""Rules to install NodeJS dependencies during WORKSPACE evaluation."""
+
+def _npm_install_impl(repository_ctx):
+  """Core implementation of npm_install."""
+
+  repository_ctx.file("BUILD", """
+package(default_visibility = ["//visibility:public"])
+filegroup(name = "node_modules", srcs = glob(["node_modules/**/*"]))
```
alexeagle: there's a problem: many npm packages ship with files that are not valid labels. sometimes the user currently needs to give specific glob patterns under node_modules which are safe, eg. https://github.com/angular/angular/blob/master/BUILD.bazel#L7

If that's not possible with bazel-managed-deps, could you at least document that in the tradeoffs section in the readme?

internal/packages.bzl
```diff
@@ -0,0 +1,52 @@
...
+  result = repository_ctx.execute(
+    [repository_ctx.path(repository_ctx.attr._npm),
+     "install", repository_ctx.path("")])
+
+  if result.return_code:
+    fail("npm_install failed: %s (%s)" % (result.stdout, result.stderr))
+
+npm_install = repository_rule(
+    attrs = {
+        "packages": attr.label(
```
alexeagle: a plural name looks like it should be a `label_list`
in other places I call this attribute "package_json" so it's more clear what value it expects

examples/packages/WORKSPACE
```diff
@@ -0,0 +1,40 @@
...
+    name = "npm_json",
+    packages = "//:package.json",
+)
+
+npm_install(
+    name = "npm_lock",
+    packages = "//:package-lock.json",
+)
+
+# load("@build_bazel_rules_nodejs//:defs.bzl", "yarn_install")
```
alexeagle: remove commented block, let's file an issue for yarn support in bazel-managed-deps and see if anyone actually wants it

examples/packages/BUILD.bazel
```diff
@@ -0,0 +1,31 @@
+load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary")
+
+VARIANTS = ["npm_json", "npm_lock"]
```
alexeagle: this doesn't look much like an example for users to follow. I suspect there should be some e2e testing folder under `/internal` where this kind of exhaustive checking lives.

The better example would just be to convert `examples/program` to use bazel-managed-deps

internal/packages.bzl
```diff
@@ -0,0 +1,52 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
alexeagle: my style convention has been to name the file after the rule it exports, so this should be
npm_install.bzl

alexeagle:
Hmm, here's a complication:

Many bazel rules in the nodejs ecosystem need some runtime dependency on a nodejs library. For example, rules_typescript exposes a bazel worker so it needs protobuf at runtime. In addition, this bazel-to-npm dependency edge needs to carry version restrictions, for example rules_typescript works with typescript@'>=2.4 <2.6'

The way I do this is by installing npm package `@bazel/typescript`, then tell the user to add to `WORKSPACE`
```
local_repository(name = "build_bazel_rules_typescript", path="node_modules/@bazel/typescript")
```

Now if they use bazel-managed dependencies, how will they point bazel to the repository rule that creates the `@build_bazel_rules_typescript` workspace? Same issue for most rules.

alexeagle:
Merged most of this as 2b07bb6853ac2def1a493f251674f3e54d31e475 with a follow-up commit I added for the comments above.

There was still a test failing, which I left a TODO for, if you have time to re-base this on master and turn on the `@packages_example//:test_escape_npm_lock` that would be nice

alexeagle:
Was merged in 2b07bb6853ac2def1a493f251674f3e54d31e475

PR #69, opened 2017-12-04
Npm install

alexeagle:


alexeagle:
Yeah, one of the tests tries to `npm install` with the lockfile in place of `package.json` and this results in no `node_modules` being installed at all. Hopefully @mattmoor can figure that out in a follow-up change.
Now that we are building on top of this, I'll merge this as-is.

gregmagolan:
> Yeah, one of the tests tries to npm install with the lockfile in place of package.json and this results in no node_modules being installed at all. Hopefully @mattmoor can figure that out in a follow-up change.
Now that we are building on top of this, I'll merge this as-is.

Sounds good

mattmoor:
Sorry I was MIA today, it turns out airplane wifi doesn't scale to techies descending on Kubecon from Cloud Central (Seattle).

thanks for so proactively getting this in :)

PR #70, opened 2017-12-05
[WiP] Re-enable all of the tests for the npm_install example.

mattmoor:


mattmoor:
Hmm, this is a bit odd, it works fine for me locally.  @alexeagle if you have any tips for reproducing this kind of thing locally, I'm open to suggestions.

One thing that pops out about this: Can I just copy `package-lock.json` to `package.json` and expect things to install properly?  When I make the rules name the symlink appropriately (e.g. `package-lock.json`) I actually see the file getting mutated (it empties it out, presumably because no `package.json`?), which is an unexpected side effect.  Strangely things still seem to pass for me...

alexeagle:
What version of npm do you have? The failing test runs under https://hub.docker.com/r/angular/ngcontainer/

```
$ docker run -it --rm angular/ngcontainer:0.0.2
root@29c8750fbe20:/# npm -version  
3.10.10
```

alexeagle:
updating it to `angular/ngcontainer:0.0.6` is probably the fix

mattmoor:
I'll give that a shot.  I thought it was using the version of NPM you were fetching... :(

alexeagle:
Oh the reason it works locally is that you are not on Windows. I'll try to fix this up and merge it.

alexeagle:
This packages_example is also missing coverage on windows, I don't think you can `cd` to the `examples/packages` directory though based on https://github.com/bazelbuild/continuous-integration/blob/d7081d8d34e2f8527b03d1c48f18b6c392d9dca3/docs/owner.md

gregmagolan:
This was resolved in #124. Can be closed.

PR #71, opened 2017-12-08
npm_install: correct path

filipesilva:
The npm path is not within `bin`.

I ran into this problem while trying to address https://github.com/bazelbuild/rules_typescript/issues/56. Starting from master, these were the commands I ran in order to see the error:

```
kamik@T460p MINGW64 /d/sandbox/rules_typescript (master)
$ bazel clean --expunge
INFO: Starting clean (this may take a while). Consider using --expunge_async if the clean takes more than several minutes.

kamik@T460p MINGW64 /d/sandbox/rules_typescript (master)
$ bazel run @yarn//:yarn
...............
INFO: Analysed target @yarn//:yarn (7 packages loaded).
INFO: Found 1 target...
Target @yarn//:yarn up-to-date:
  C:/users/kamik/appdata/local/temp/_bazel_kamik/y9xg3tkg/execroot/build_bazel_rules_typescript/bazel-out/x64_windows-fastbuild/bin/external/yarn/yarn
  C:/users/kamik/appdata/local/temp/_bazel_kamik/y9xg3tkg/execroot/build_bazel_rules_typescript/bazel-out/x64_windows-fastbuild/bin/external/yarn/yarn.exe
INFO: Elapsed time: 33.328s, Critical Path: 0.03s
INFO: Build completed successfully, 5 total actions

INFO: Running command line: C:/users/kamik/appdata/local/temp/_bazel_kamik/y9xg3tkg/execroot/build_bazel_rules_typescript/bazel-out/x64_windows-fastbuild/bin/external/yarn/yarn.exe
yarn install v1.3.2
[1/4] Resolving packages...
success Already up-to-date.
Done in 0.27s.

kamik@T460p MINGW64 /d/sandbox/rules_typescript (master)
$ bazel build internal/tsc_wrapped
ERROR: D:/sandbox/rules_typescript/internal/tsc_wrapped/BUILD.bazel:27:1: no such package '@build_bazel_rules_typescript_deps//': Traceback (most recent call last):
        File "C:/users/kamik/appdata/local/temp/_bazel_kamik/y9xg3tkg/external/build_bazel_rules_nodejs/internal/npm_install.bzl", line 36
                repository_ctx.execute([repository_ctx.path(repository_...("")])
        File "C:/users/kamik/appdata/local/temp/_bazel_kamik/y9xg3tkg/external/build_bazel_rules_nodejs/internal/npm_install.bzl", line 37, in repository_ctx.execute
                repository_ctx.path(repository_ctx.attr._npm)
[...]

PR #74, opened 2017-12-08
Npm path

alexeagle:
should work instead of
https://github.com/bazelbuild/rules_nodejs/pull/71


alexeagle:
I wish I had time to try `select/config` but merging this to unblock Filipe.

PR #78, opened 2017-12-21
Add a utility to re-root files from nested trees (eg. typescript)

alexeagle:
It will be tested in rules_typescript shortly. We'll modify the rollup example to consume it as well, in a followup PR.

### DIFF
--- a/internal/collect_es6_sources.bzl
+++ b/internal/collect_es6_sources.bzl
@@ -1,0 +1,54 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


internal/collect_es6_sources.bzl
```diff
@@ -0,0 +1,57 @@
...
+      non_rerooted_files = depset(transitive = [non_rerooted_files, dep.typescript.transitive_es6_sources])
+    else:
+      fail(
+          ("%s is neither a TypeScript nor a Closure JS library producing rule." % dep.label) +
+          "\nDependencies must be ts_library, ts_declaration, or closure_js_library.")
+
+  rerooted_files = []
+  for file in non_rerooted_files.to_list():
+    rerooted_file = ctx.actions.declare_file(
+      "%s.closure/%s" % (
```
mrmeku: es6

internal/collect_es6_sources.bzl
```diff
@@ -0,0 +1,57 @@
...
+        ctx.label.name,
+        file.short_path.replace(".closure.js", ".js")))
```
mrmeku: Leave a comment to update this to only target the end of the short path (there's potential for a stupidly named file being sad)

alexeagle:
merged to master manually

PR #80, opened 2017-12-22
add node_modules replace

thelgevold:
Tweaking the rerooting to handle rxjs/node_modules

This is basically the original code from @mrmeku

@alexeagle


bazel-io:

alexeagle:
@mrmeku do we still need this change?

mrmeku:
I need to copy my latest version of this code from my closure_ts_binary PR
from rules typescript. It should be more or less the same, but that should
be the reference for what we merge since that code has been tested with
both an angular and angularjs app

alexeagle:
Okay, please open a new PR with the thing that we should merge.

mrmeku:
Donezo: https://github.com/bazelbuild/rules_nodejs/pull/83

PR #81, opened 2017-12-22
Added skylint and fixed lint errors

mrmeku:


package.json
```diff
@@ -0,0 +1,15 @@
+{
```
alexeagle: I think it's confusing to add a `package.json` even though we don't publish this as an npm package, nor do we install any packages from npm. Could we avoid adding it right now, and just have `skylint.sh` in place of `npm run skylint`?

Also, if there's no CI enforcement included in this PR, then we'll probably introduce more lint violations. Have you thought about adding to the circleci config?

alexeagle:
Sorry this got stale - @gregmagolan did this now and did the TODOs for skydocs as well.

Note, I spent some time trying to make skydoc work so we can render some markdown/HTML API docs from the sources. It would be nice to move example code inline so we keep it up to date. Might be a good side project.

PR #83, opened 2018-01-04
Tweak collect_es6_sources to be compatible with the closure compiler

mrmeku:


internal/collect_es6_sources.bzl
```diff
@@ -19,37 +19,33 @@ The tree will be flattened, such that all the es6 files are under a single tree.
...
-  #TODO(mrmeku): we should include the files and closure_js_library contents too
-  return depset(direct = rerooted_files)
+      for es6_source in dep.typescript.transitive_es6_sources:
+        rerooted_es6_source = ctx.actions.declare_file("/".join([f for f in [
+          ctx.label.name + ".es6",
+          "node_modules",
+          "" if es6_source.owner.workspace_root else ctx.workspace_name,
+          es6_source.short_path.replace("../", "").replace(".closure.js", ".js")
+        ] if f]))
+        ctx.actions.expand_template(
```
alexeagle: restore the missing TODO

internal/collect_es6_sources.bzl
```diff
@@ -19,37 +19,33 @@ The tree will be flattened, such that all the es6 files are under a single tree.
...
-    rerooted_files += [rerooted_file]
-
-  #TODO(mrmeku): we should include the files and closure_js_library contents too
-  return depset(direct = rerooted_files)
+      for es6_source in dep.typescript.transitive_es6_sources:
+        rerooted_es6_source = ctx.actions.declare_file("/".join([f for f in [
+          ctx.label.name + ".es6",
+          "node_modules",
+          "" if es6_source.owner.workspace_root else ctx.workspace_name,
+          es6_source.short_path.replace("../", "").replace(".closure.js", ".js")
```
alexeagle: the `replace("../")` looks suspicious. what's an example before/after path that we need here. Is this due to the `my_repo/external/your_repo` paths?

PR #88, opened 2018-01-18
Fix CI cache

gregmagolan:
This fixes the CI cache. CI still works on master but caching is broken:

```
Error computing cache key: template: cacheKey:1:25: executing "cacheKey" at <checksum "yarn.lock">: error calling checksum: open /home/circleci/ng/yarn.lock: no such file or directory
```


bazel-io:

alexeagle:
thanks!

PR #89, opened 2018-01-20
fix es6 rerooted paths to include external workspaces

alexeagle:
First step to fix https://github.com/alexeagle/angular-bazel-example/issues/62


gregmagolan:
Merged with #92

PR #92, opened 2018-01-23
Handle module mappings in rollup_bundle rule

gregmagolan:
Includes @alexeagle commit from #89

### DIFF
--- a/internal/rollup/rollup.config.js
+++ b/internal/rollup/rollup.config.js
@@ -4,37 +4,47 @@ const rollup = require('rollup');
 const nodeResolve = require('rollup-plugin-node-resolve');
 const commonjs = require('rollup-plugin-commonjs');
 const path = require('path');
-const fs = require('fs');
 
 const binDirPath = "TMPL_bin_dir_path";
 const workspaceName = "TMPL_workspace_name";
 const buildFileDirname = "TMPL_build_file_dirname";
 const labelName = "TMPL_label_name";
+const moduleMappings = TMPL_module_mappings;
 
 class NormalizePaths {
   resolveId(importee, importer) {
     // process.cwd() is the execroot and ends up looking something like /.../2c2a834fcea131eff2d962ffe20e1c87/bazel-sandbox/872535243457386053/execroot/<workspace_name>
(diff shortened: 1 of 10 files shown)


bazel-io:

internal/rollup/rollup.config.js
```diff
@@ -9,25 +9,43 @@ const binDirPath = "TMPL_bin_dir_path";
...
+      resolved = path.join(process.cwd(), binDirPath, buildFileDirname, `${labelName}.es6`, importee);
+    } else {
+      // possible workspace import or external import if importee matches a module mapping
+      for (const k in moduleMappings) {
+        if (importee == k || importee.startsWith(`${k}/`)) {
+          var v = moduleMappings[k];
+          var userWorkspace = workspaceName;
+          // replace the root module name on a mappings match
+          if (v.startsWith('external/')) {
+            // for external workspaces, drop the 'external/' from the module mapping and clear the user workspace anme
```
pshields: sp: name

internal/rollup/rollup_bundle.bzl
```diff
@@ -30,6 +51,7 @@ def _rollup_bundle(ctx):
           "TMPL_workspace_name": ctx.workspace_name,
           "TMPL_build_file_dirname": buildFileDirname,
           "TMPL_label_name": ctx.label.name,
+          "TMPL_module_mappings": str(mappings),
```
alexeagle: instead of just stringifying, could you use the same serialization as https://github.com/bazelbuild/rules_nodejs/blob/master/internal/node.bzl#L46

gregmagolan: do you think stringifying could fail in some cases?

gregmagolan:
test added for module_name using js_library... unfortunately, I had to put the js_library target dependency in the rollup_bundle `srcs` instead of `deps` since putting it in `deps` breaks the jasmine_node_test rule. jasmine_node_test uses nodejs_test which has a source aspect that passes the es6 sources from the js_library to jasmine which can't load them.

gregmagolan:
rollup e2e test is fixed up. `:bundle` belongs in the `data` attribute of the jasmine_test rule instead of `:deps`.

PR #95, opened 2018-01-23
Cleanups for rollup_bundle

alexeagle:


gregmagolan:
Windows is still red

alexeagle:
Windows is red on master it turns out. Firing up the surface tablet now to debug it ...

PR #100, opened 2018-01-31
Allow sources to be in multiple root dirs

alexeagle:
For es6, we currently have this clumsy thing:
https://github.com/bazelbuild/rules_nodejs/blob/master/internal/collect_es6_sources.bzl#L52
that flattens the re-rooted directories from all the libraries in the transitive closure.

The more performant thing is just to update the module resolution to understand a `rootDirs` option, just like TypeScript does.

It turns out Angular's optimizer doesn't work with ES6, so I need to create another flavor, es5_esm. This refactoring allows Angular to re-use the rollup rule, but pass in that output instead, and leave it in the original roots.

### DIFF
--- a/internal/e2e/rollup/package.json
+++ b/internal/e2e/rollup/package.json
@@ -1,5 +1,7 @@
 {
   "devDependencies": {
-    "jasmine": "~2.8.0"
+    "jasmine": "~2.8.0",
+    "rollup": "0.52.3",
+    "rollup-plugin-node-resolve": "3.0.0"
   }
 }
--- a/internal/rollup/uglify.config.json
+++ b/internal/rollup/uglify.config.json
@@ -1,7 +1,6 @@
 {
   "compress": {
     "pure_getters": true,
-    "pure_funcs": ["Object.defineProperty"],
     "passes": 3,
     "global_defs": {
       "ngDevMode": false
--- a/internal/collect_es6_sources.bzl
+++ b/internal/collect_es6_sources.bzl
@@ -37,17 +37,16 @@ def collect_es6_sources(ctx):
 
   rerooted_files = []
   for file in non_rerooted_files.to_list():
-    if file.short_path.startswith(".."):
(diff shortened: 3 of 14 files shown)


internal/e2e/rollup/package.json
```diff
@@ -1,5 +1,8 @@
 {
   "devDependencies": {
-    "jasmine": "~2.8.0"
+    "jasmine": "~2.8.0",
+    "rollup": "0.52.3",
```
gregmagolan: Are these needed? commonjs could go for sure as its not used. ```  npm_install(
      name = "build_bazel_rules_nodejs_rollup_deps",
      package_json = "@build_bazel_rules_nodejs//internal/rollup:package.json",
  )``` should handle rollup?

internal/rollup/uglify.config.json
```diff
@@ -1,7 +1,6 @@
 {
   "compress": {
     "pure_getters": true,
-    "pure_funcs": ["Object.defineProperty"],
```
gregmagolan: 👍

internal/e2e/rollup/package.json
alexeagle: Ah, but that one doesn't have jasmine.
I thought "oh, no problem, I could union them, I'll make a filegroup with a 
`srcs=['@bbrj_rollup_deps//:node_modules', '@bbrj//internal/e2e/rollup:node_modules']`
but this doesn't work because the resulting filegroup isn't flattened

gregmagolan: Ohh. Ok. That makes sense. commonjs could go still.

internal/rollup/uglify.config.json
alexeagle: yeah this broke the code for new Angular renderer. I'll have to add a way for Angular (and others) to set their own uglify settings...

internal/e2e/rollup/package.json
alexeagle: oh, right commonjs was still here, fixed

PR #103, opened 2018-02-01
set uglify option sequences:false in debug mode

alexeagle:
Requested by @mhevery - I hope he knows what he's doing...


alexeagle:
Would like to add some tests before merging this

alexeagle:
okay now with tests, @gregmagolan would you like to look?

PR #104, opened 2018-02-02
WIP: Prototype of rollup_bundle with code splitting

gregmagolan:


bazel-io:

internal/rollup/rollup_bundle.bzl
```diff
@@ -117,14 +138,14 @@ def run_uglify(ctx, input, output, debug = False):
...
-  run_rollup(ctx, rollup_config, ctx.outputs.build_es6)
-  run_tsc(ctx, ctx.outputs.build_es6, ctx.outputs.build_es5)
-  run_uglify(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min)
-  run_uglify(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min_debug, debug = True)
-  return DefaultInfo(files=depset([ctx.outputs.build_es5_min]))
+  output_dir_es6 = run_rollup(ctx, rollup_config)
+  output_dir_es5 = run_tsc(ctx, output_dir_es6)
+  #run_uglify(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min)
+  #run_uglify(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min_debug, debug = True)
+  return DefaultInfo(runfiles=ctx.runfiles([output_dir_es6, output_dir_es5]), files=depset([]))
```
gregmagolan: runfiles probably not the right place for these

internal/rollup/rollup_bundle.bzl
```diff
@@ -48,47 +53,63 @@ def write_rollup_config(ctx, plugins=[]):
...
+          "TMPL_include_path": input_dir_rerooted + "/*",
+          "TMPL_out_dir": output_dir_rerooted,
+      })
+
   args = ctx.actions.args()
-  args.add(["--target", "es5"])
-  args.add("--allowJS")
-  args.add(input.path)
-  args.add(["--outFile", output.path])
+  args.add(["--project", config.path])
```
gregmagolan: tsconfig needed for tsc as its the only way to tell it to compile all files in a given folder. in this case the output folder of the rollup rule

internal/rollup/rollup.config.js
```diff
@@ -53,12 +52,14 @@ class NormalizePaths {
...
-  output: {format: 'iife'},
+  input: [TMPL_inputs],
+  output: {dir: path.join(process.cwd(), binDirPath, buildFileDirname, 'bundles.es6'), format: 'cjs'},
+  experimentalCodeSplitting: true,
+  experimentalDynamicImport: true,
   plugins: [
       TMPL_additional_plugins
   ].concat([
       new NormalizePaths(),
-      commonjs(),
```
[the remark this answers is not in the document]
alexeagle: Yeah I also removed it today as it was giving rxjs when I didn't want it

internal/rollup/rollup.config.js
```diff
@@ -53,12 +52,14 @@ class NormalizePaths {
 }
 
 export default {
-  output: {format: 'iife'},
+  input: [TMPL_inputs],
+  output: {dir: path.join(process.cwd(), binDirPath, buildFileDirname, 'bundles.es6'), format: 'cjs'},
```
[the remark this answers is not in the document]
alexeagle: I think we need a formatter in this repo

internal/rollup/rollup.config.js
```diff
@@ -53,12 +52,14 @@ class NormalizePaths {
...
+  experimentalCodeSplitting: true,
```
alexeagle: not that we're really rolling it out yet - but I think we should consider exposing an experimental flag as a rule attribute when we enable these things - we could have users depend on us to provide a higher guarantee than the thing we depend on

internal/rollup/rollup_bundle.bzl
```diff
@@ -117,14 +138,14 @@ def run_uglify(ctx, input, output, debug = False):
...
-  return DefaultInfo(files=depset([ctx.outputs.build_es5_min]))
+  output_dir_es6 = run_rollup(ctx, rollup_config)
+  output_dir_es5 = run_tsc(ctx, output_dir_es6)
+  #run_uglify(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min)
+  #run_uglify(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min_debug, debug = True)
+  return DefaultInfo(runfiles=ctx.runfiles([output_dir_es6, output_dir_es5]), files=depset([]))
 
 ROLLUP_ATTRS = {
-    "entry_point": attr.string(mandatory = True),
+    "entry_points": attr.string_list(mandatory = True),
```
alexeagle: in the future we have to think about backward compat for these changes, and avoid breaking.
Maybe not yet? maybe soon? hard to say how many users we have

internal/rollup/rollup_bundle.bzl
```diff
@@ -150,15 +208,11 @@ ROLLUP_ATTRS = {
         single_file = True),
 }
 
-ROLLUP_OUTPUTS = {
```
alexeagle: We can't land it if we remove these. Can we still output the files for the "main" bundle or something?

gregmagolan: We could make it so that the rule could output both a regular bundle or a code split one but without a devserver that can support code splitting is there any value in a landing the production bundler at this point?

alexeagle: I think we need to land this somewhere with CI so that it doesn't rust, and people can try it and build on it.
I think there are some users who would be satisfied with lazy-loading only for production, if they have simple enough code splitting and are okay with having to debug issues in production mode.
I dunno it seems like  a tough call.

alexeagle:
Not planning to land this here, yet. Will live on Greg's fork.

PR #105, opened 2018-02-03
Use bazel managed node when running npm

### LINKED ISSUE #77: npm_install uses "/usr/bin/env node" instead of the node managed by bazel
fyi - this is evident on my build host where there's no node available in the environment. The build fails with:

`no such package '@npm_json//': npm_install failed: (/usr/bin/env: 'node': No such file or directory`

I note that `external/nodejs/bin/npm` starts with `#!/usr/bin/env node`, which, if I understand bazel philosophy correctly, seems like a problem.

gregmagolan:
Fixes #77

/cc @alexeagle

### DIFF
--- a/internal/node_install.bzl
+++ b/internal/node_install.bzl
@@ -22,7 +22,7 @@ node binary to other rules.
 def get_executables(repository_ctx):
   if repository_ctx.os.name.lower().find("windows") != -1:
     # The windows distribution of nodejs has the binaries in different paths
-    return {"node": "node.exe", "npm": "npm.cmd"}
+    return {"node": "node.exe", "npm": "node_modules/npm/bin/npm-cli.js"}
   else:
     return {"node": "bin/node", "npm": "bin/npm"}
 
--- a/internal/node_install.bzl
+++ b/internal/node_install.bzl
@@ -86,6 +86,8 @@ _node_repo = repository_rule(_node_impl, attrs = {})
 #     attrs = { "package_json": attr.label() },
 # )
 
+load(":node_labels.bzl", "get_node_label")
+
 def _yarn_impl(ctx):
   # Yarn is a package manager that downloads dependencies. Yarn is an improvement over the `npm` tool in
   # speed and correctness. We download a specific version of Yarn to ensure a hermetic build.
@@ -96,10 +98,7 @@ sh_binary(
   srcs = ["yarn.sh"],
 )
 """)
-  if ctx.os.name.lower().find("windows") != -1:
-    node = Label("@nodejs//:node.exe")
-  else:
-    node = Label("@nodejs//:bin/node")
+  node = get_node_label(ctx)
   ctx.file("yarn.sh", "#!/bin/bash" + "".join(["""
 ROOT="{}"
 NODE="{}"
--- a/internal/npm_install.bzl
+++ b/internal/npm_install.bzl
@@ -14,6 +14,8 @@
 
 """Rules to install NodeJS dependencies during WORKSPACE evaluation."""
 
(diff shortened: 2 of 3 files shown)


bazel-io:

internal/npm_install.bzl
```diff
@@ -32,16 +34,14 @@ filegroup(
...
-    npm = Label("@nodejs//:bin/npm")
+  node = get_node_label(repository_ctx)
+  npm = get_npm_label(repository_ctx)
 
+  # This runs node, not npm directly, as the latter will
+  # use #!/usr/bin/node (see https://github.com/bazelbuild/rules_nodejs/issues/77)
   # To see the output, pass: quiet=False
   result = repository_ctx.execute(
-    [repository_ctx.path(npm), "install", repository_ctx.path("")])
+    [repository_ctx.path(node), repository_ctx.path(npm), "install", repository_ctx.path("")])
```
alexeagle: I think when I tried to fix this before, I had trouble getting it to work on Windows. And the CI isn't triggering it seems. Do you have access to a Windows environment? If you need, I'm sure I can expense a machine for you.

gregmagolan:
> As part of fixing this issue, let's make sure there is no node/npm installed on our CI environment (or the CI environment of some downstream example repo), so that we can ensure nothing breaks this again.
(Maybe we should have a -no-node variant of the ngcontainer image?)

Agree with that

internal/npm_install.bzl
gregmagolan: Yup. I have a slow windows VM. I'll try it out.

gregmagolan: Oh nevermind. The examples/packages test is broken on windows on master as well. require can't find `jsesc`. I'll commit a fix for the npm path on windows and then figure out why the test is broken.

internal/node_install.bzl
```diff
@@ -22,7 +22,7 @@ node binary to other rules.
 def get_executables(repository_ctx):
   if repository_ctx.os.name.lower().find("windows") != -1:
     # The windows distribution of nodejs has the binaries in different paths
-    return {"node": "node.exe", "npm": "npm.cmd"}
+    return {"node": "node.exe", "npm": "node_modules/npm/bin/npm-cli.js"}
   else:
     return {"node": "bin/node", "npm": "bin/npm"}
```
alexeagle: this looks strange - if windows points to a .js file for npm, shouldn't we do the same on other platforms?

internal/node_install.bzl
```diff
@@ -22,7 +22,7 @@ node binary to other rules.
 def get_executables(repository_ctx):
   if repository_ctx.os.name.lower().find("windows") != -1:
     # The windows distribution of nodejs has the binaries in different paths
-    return {"node": "node.exe", "npm": "npm.cmd"}
+    return {"node": "node.exe", "npm": "node_modules/npm/bin/npm-cli.js"}
```
alexeagle: should this re-use node_label.bzl?

ppodolsky:
Hi!
There is another problem that using `bazel run @nodejs//:npm install` also invokes system node binary.

internal/node_install.bzl
gregmagolan: oddly on osx `bin/npm` is the javascript equivalent of `node_modules/npm/bin/npm-cli.js` but it doesn't have an `.js` extension

alexeagle:
Discussed in meeting - instead of publishing separate container, we should just have a circle workflow that moves node out of the way

alexeagle:
@ppodolsky that's a good point - and also if you have a multi-package repo, there should be a command to install all the package.json files in a loop. Filed https://github.com/bazelbuild/rules_nodejs/issues/108 for this.

PR #106, opened 2018-02-03
Adds yarn_install rule

gregmagolan:
Also adds package_lock_json attribute to npm_install and fixes examples/packages tests.

/cc @alexeagle 
/cc @mattmoor


bazel-io:

internal/node_install.bzl
```diff
@@ -101,14 +101,14 @@ sh_binary(
   else:
     node = Label("@nodejs//:bin/node")
   ctx.file("yarn.sh", "#!/bin/bash" + "".join(["""
-ROOT="{}"
 NODE="{}"
 SCRIPT="{}"
-(cd "$ROOT"; "$NODE" "$SCRIPT" "$@")
+ROOT="{}"
+"$NODE" "$SCRIPT" --cwd "$ROOT" "$@"
```
gregmagolan: Seems a little cleaner IMO

defs.bzl
```diff
@@ -23,4 +23,5 @@ load("//internal:node.bzl",
 load("//internal:jasmine_node_test.bzl", "jasmine_node_test")
 load("//internal:node_install.bzl", "node_repositories")
 load("//internal:npm_install.bzl", "npm_install")
```
alexeagle: I'd like to drop `npm_install`. I don't think we need both, it will be more work to maintain both, and it's broken.

A polite way to roll that out would be to keep `npm_install` in this file but make it an alias of the new yarn impl

Or do we have to keep it for users who have a package-lock.json for their regular workflow and want Bazel to manage a copy of all the dependencies?

defs.bzl
```diff
@@ -23,4 +23,5 @@ load("//internal:node.bzl",
...
+load("//internal:yarn_install.bzl", "yarn_install")
```
alexeagle: do users need to know or care which package manager we use?
1. maybe we should just keep the name 'npm' even though we don't run the npm tool. After all, we are installing NPM dependencies from the NPM registry.
2. other ideas for a name?

internal/node_install.bzl
```diff
@@ -136,7 +136,8 @@ def node_repositories(package_json):
 
   _yarn_repo(name = "yarn", package_json = package_json)
 
-  npm_install(
+  yarn_install(
       name = "build_bazel_rules_nodejs_rollup_deps",
       package_json = "@build_bazel_rules_nodejs//internal/rollup:package.json",
+      yarn_lock = "@build_bazel_rules_nodejs//internal/rollup:yarn.lock",
```
alexeagle: can we make it work with only a lock file? or does yarn require we keep a pair of files and pass them both?

obviously for our purposes we don't care about `package.json`
maybe users would be confused though, even if it were possible.

defs.bzl
alexeagle: Answering my own question - since the user has to give us a lock file, they do have to know which is used... I guess that's okay then.

alexeagle:
For some reason it looks like this didn't get a CI result on Windows.
We should look into Appveyor rather than depending on the Bazel team's CI I think...

defs.bzl
gregmagolan: Yeah. The rules are slightly different. yarn_install takes a yarn_lock attribute and npm_install takes a package_lock_json attribute. For users wanting to use an npm lock file they'll want to keep using npm_install.

internal/node_install.bzl
gregmagolan: Both files are needed for yarn.

defs.bzl
alexeagle: I think we should at least prevent anyone new from depending on `npm_install` if we're not going to make it deterministic yet. I don't want users to fall into an unhappy path by accident just because they started with a `package-lock.json` file

.circleci/config.yml
```diff
@@ -24,8 +24,10 @@ jobs:
 
       # We should also be able to test targets in a different workspace
       - run: bazel test --config=ci @program_example//...
-      # TODO(mattmoor): figure out why other tests in this example fail
       - run: cd examples/packages && bazel test //:test_escape_npm_json
+      - run: cd examples/packages && bazel test //:test_escape_npm_lock
```
alexeagle: this isn't a good model - I wasn't able to get the Bazel CI to do this so these are missing windows coverage.
I'm starting to think that nested workspaces are more trouble than they are worth, given that we need some multi-repo CI setup anyway.

gregmagolan:
Versions fetched by npm_install and yarn_install for lock and no cases are now tested by examples/packages

internal/npm_install.bzl
```diff
@@ -53,6 +60,10 @@ npm_install = repository_rule(
             mandatory = True,
             single_file = True,
         ),
+        "package_lock_json": attr.label(
+            allow_files = True,
```
alexeagle: for future reference - there is a shortcut `allow_single_file` that does both things

PR #110, opened 2018-02-06
Add a release stamping feature to rollup_bundle

alexeagle:


gregmagolan:
Note: the minified bundle ends up being a bit a larger with multiple license banners. It looks like this ```**var __extends=this&&this.__extends||...
/**
 * @license A dummy license banner that goes at the top of the file.
 * This is version v1.2.3
 */
/**
 * @license A dummy license banner that goes at the top of the file.
 * This is version v1.2.3
 */
!function(){"use strict";var e=Object.setPrototypeOf...
/**
     * @license Angular v5.2.3
     * (c) 2010-2018 Google, Inc. https://angular.io/
     * License: MIT
     */
/**
     * @license
     * Copyright Google Inc. All Rights Reserved.
     *
     * Use of this source code is governed by an MIT-style license that can be
     * found in the LICENSE file at https://angular.io/license
     */
var Q=function(){function e(e){this._desc=e,...
/**
     * @license
     * Copyright Google Inc. All Rights Reserved.
     *
     * Use of this source code is governed by an MIT-style license that can be
     * found in the LICENSE file at https://angular.io/license
     */function X(e,t,n,r){var o=ee(t);...
**```

alexeagle:
Which minified bundle is that? The duplication looks like a smell that we're doing something wrong somwehere

gregmagolan:
> Which minified bundle is that? The duplication looks like a smell that we're doing something wrong somwehere

`bazel-out/darwin-fastbuild/bin/src/bundle.min.js`

alexeagle:
Oh, you mean in the angular-bazel-example project, but not reproducible
here? Interesting

gregmagolan:
> Oh, you mean in the angular-bazel-example project, but not reproducible here? Interesting

Right. Forgot to mention that.

PR #111, opened 2018-02-07
Fix rollup_bundle module resolution for user workspace files

gregmagolan:
/cc @alexeagle

### DIFF
--- a/internal/rollup/rollup.config.js
+++ b/internal/rollup/rollup.config.js
@@ -11,6 +11,7 @@ const fs = require('fs');
 const DEBUG = false;
 
 const moduleMappings = TMPL_module_mappings;
+const workspaceName = "TMPL_workspace_name";
 const rootDirs = TMPL_rootDirs;
 const banner_file = TMPL_banner_file;
 const stamp_data = TMPL_stamp_data;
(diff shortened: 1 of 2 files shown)


bazel-io:

internal/rollup/rollup.config.js
```diff
@@ -62,6 +63,11 @@ function resolveBazel(importee, importer, baseDir = process.cwd(), resolve = req
 
   if (!resolved) {
     // workspace import
+    const userWorkspace = `${workspaceName}/`;
```
alexeagle: doesn't look Windows-friendly, would be better to use path.relative

gregmagolan: hmmm. maybe I need more coffee but how would check if the path starts with a `workspaceName` segment with path.relative?

alexeagle: I use this
```
> let path = "mywksp/path/to/file"
undefined
> require('path').relative("mywksp", path)
'path/to/file'
> require('path').relative("otherwksp", path)
'../mywksp/path/to/file'
```

so it would be `if (!path.relative(TMPL_workspaceName, resolved).startsWith("..")`

alexeagle: and it also yields the path you want when it doesn't start with '..'

gregmagolan: Nice. Much better.

gregmagolan:
roll e2e test didn't catch this as it is not in its own nested WORKSPACE

Issue #98, opened 2018-01-30
npm_install is non-deterministic


alexeagle:
Thanks to @buchgr for debugging this.

The npm_install rule runs `npm install` and the resulting package.json file is not stable:
```
buchgr@buchgr0:~/code$ diff package1.json package2.json 
5c5
<   "_integrity": "sha1-2oEVQQzCps14cUVRrJKkLd4mdnY=",
---
>   "_integrity": "sha1-R2/vTKo8o0qUVsI0pBPoRMV9gcc=",
```

Since the value is different, it's not sufficient to just sort the keys.

As a result, all actions that rely on this toolchain get cache misses since the input doesn't have a stable hash.

pcj:
A bit surprised you're using npm here.  

I struggled with non-determinism issues with npm in `rules_node` for a while and ultimately ditched npm altogether for yarn.  The problems included npm adding fields that are themselves non-deterministic like you're seeing above (requiring stripping certain "blacklisted" fields), and even weird "flaky" problems where the package *contents* were occasionally different (i.e., sometimes I'd get a tarball back  where the hash was unexpected, only to find an extra README file in there).

alexeagle:
Changing to Yarn is totally an option. It's only npm because the
contributor who added the feature did it that way, and I wasn't aware of
the determinism issues.
Thanks Paul!

gregmagolan:
https://github.com/bazelbuild/rules_nodejs/pull/106 adds yarn_install which can accept a yarn.lock file

alexeagle:
@gregmagolan I think this is fixed?

gregmagolan:
> @gregmagolan I think this is fixed?

Yes. npm_install should now be deterministic. I'll verify and add something to CI to check for both npm_install and yarn_install.

PR #112, opened 2018-02-08
Makes package.json files generated by npm_install deterministic

gregmagolan:
/cc @alexeagle


gregmagolan:
Fixes #98

gregmagolan:
Makes use of https://www.npmjs.com/package/removeNPMAbsolutePaths which removes all fields starting with "_" from the generated package.json files

internal/npm_install/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+licenses(["notice"])  # Apache 2.0
+
+package(default_visibility = ["//internal:__subpackages__"])
```
alexeagle: why does this matter if there are no rules in the package?

gregmagolan:
For reference: https://github.com/npm/npm/issues/10393

gregmagolan:
Optional dependencies can still make npm non-deterministic but this has been fixed in npm 5.6.0. See https://github.com/npm/npm/issues/18135

internal/npm_install/BUILD.bazel
gregmagolan: you mean the package()? yeah. That can go.

alexeagle: whole file can just be
`# Marker that this directory is a bazel package`

gregmagolan: You still need the BAZEL.build file in the folder if there is a .bzl file there right? But it can be an empty file. Is the convention to just leave it empty?

gregmagolan:
> Just clean up the new BULID file

Good to go

alexeagle:
Do we need to upgrade npm to 5.6.0 to fix it all the way?

gregmagolan:
> Do we need to upgrade npm to 5.6.0 to fix it all the way?

The issue that 5.6.0 fixes is with optional dependencies. Apparently, there were cases where optional dependencies would result in non-deterministic node_modules. Reference issue is here: https://github.com/npm/npm/issues/18135

gregmagolan:
There is still an open issue so it may not be fixed yet in npm 5.6.0. https://github.com/npm/npm/issues/17722

gregmagolan:
And related https://github.com/npm/npm/issues/18712

gregmagolan:
And this is related to npm determinism as well for reference: https://github.com/npm/npm/issues/18286

PR #114, opened 2018-02-09
Use local yarn cache

gregmagolan:
/cc @alexeagle


bazel-io:

gregmagolan:
For reference: https://github.com/yarnpkg/yarn/issues/1275

PR #115, opened 2018-02-09
Turn on node option --preserve-symlinks

### LINKED ISSUE #109: make --preserve-symlinks the default
Such things as webpack use `__dirname` extensively so it can become a problem. The solution is to use `--preserve-symlinks` flag in `nodejs_binary` by default but I have no idea if it breaks something else.

alexeagle:
Unless the nodejs_binary has `templated_args=["--node_options=--nopreserve-symlinks"]` to disable it

Fixes #109


gregmagolan:
LGTM

PR #118, opened 2018-02-10
Run npm across all package.json files in node_repositories

### LINKED ISSUE #108: Need a way for user to call npm install over all package.json files
Referenced in https://github.com/bazelbuild/rules_nodejs/pull/105#issuecomment-363161414

gregmagolan:
Fixes #108 

Also, running `bazel run @nodejs//:npm install` will no longer invoke the system node. However, both `bazel run @nodejs//:npm install` and `bazel run @yarn//:yarn install` will now run on all package.json files in node_repositories. @alexeagle, this may not be the expected or desired behavior for manual invocation.

/cc @alexeagle


bazel-io:

internal/node_install.bzl
```diff
@@ -19,6 +19,8 @@ We then create a repository @nodejs which provides the
 node binary to other rules.
 """
 
+load(":node_labels.bzl", "get_node_label")
```
alexeagle: can you eliminate the uses of `get_executables` at the same time?

internal/node_install.bzl
```diff
@@ -58,26 +60,41 @@ def _node_impl(repository_ctx):
...
 package(default_visibility = ["//visibility:public"])
-exports_files([
-  "{0}",
-  "{1}",
-])
+exports_files(["{0}"])
 alias(name = "node", actual = "{0}")
-alias(name = "npm", actual = "{1}")
-""".format(executables["node"], executables["npm"]))
+sh_binary(
```
alexeagle: we have to avoid bash dependencies - but since we're underneath nodejs binary here I think that means authoring this again in Windows Batch
don't need to fix in this PR

internal/node_install.bzl
gregmagolan: 👍 will do

PR #119, opened 2018-02-11
Windows manifest resolution

### LINKED ISSUE #25: Manifest file lookup happens outside of node module resolution

gregmagolan:
Proper file and directory resolution for npm package from MANIFEST file for windows.

Fixes #25 

/cc @alexeagle


bazel-io:

gregmagolan:
Uses resolution logic from https://github.com/bazelbuild/rules_nodejs/issues/25#issuecomment-333263309

gregmagolan:
> I end up having to debug the resolution a LOT. Could you add a const DEBUG at the top, and leave some debug-guarded console.error's that help us track down issues in the future?

> In the same PR can you resolve the TODO(#25)'s like in the rollup.config.js where I loaded the plugins with a silly long path?

Done

gregmagolan:
> Other than that, what's the testing story for this? Feels like we could use a simple unit test for the resolution logic (like the one I added for rollup resolution)

Yes. Something to test all the different cases would be good. There are a few. Its even possible to have nested package.json files in an npm module and recursively resolve the "main" by directory. Will come up with something tomorrow.

PR #130, opened 2018-02-22
Exclude "node_modules/**/.*/**" from yarn_install and npm_install

gregmagolan:


bazel-io:

PR #131, opened 2018-02-23
fix rollup module resolution

alexeagle:
- the tests were not running because of an fit, and some were broken at head :(
- I changed re-rooting to look like execroot, so it uses external/wksp
- handle the case of a relative import across roots

### DIFF
--- a/internal/rollup/rollup.config.js
+++ b/internal/rollup/rollup.config.js
@@ -21,6 +21,8 @@ Rollup: running with
   moduleMappings: ${JSON.stringify(moduleMappings)}
 `);
 
+// This resolver mimics the TypeScript `rootDirs` feature, which lets us pretend
+// that multiple roots are logically merged into a single tree.
 function resolveBazel(importee, importer, baseDir = process.cwd(), resolve = require.resolve) {
   function resolveInRootDirs(importee) {
     for (var i = 0; i < rootDirs.length; i++) {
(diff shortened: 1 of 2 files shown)


internal/rollup/rollup.config.spec.js
```diff
@@ -13,49 +13,66 @@ TMPL_stamp_data = '';
...
                join(
-                   baseDir, 'bazel-bin', 'path', 'to', 'b.esm5', 'some_wksp', 'path', 'to', 'b',
-                   'sub')))
-        .toEqual(`${baseDir}/bazel-bin/path/to/b.esm5/some_wksp/path/to/a/index.js`);
+                   baseDir, 'bazel-bin', 'path', 'to', 'b.esm5', 'external', 'some_wksp', 'path',
+                   'to', 'b', 'sub')))
+        .toEqual(`${baseDir}/bazel-bin/path/to/b.esm5/external/some_wksp/path/to/a/index.js`);
+  });
+
+  fit('should resolve relative imports from other root', () => {
```
gregmagolan: Another fit? Intentional?

internal/rollup/rollup.config.js
```diff
@@ -46,8 +48,20 @@ function resolveBazel(importee, importer, baseDir = process.cwd(), resolve = req
...
+      for (var i = 0; i < rootDirs.length; i++) {
+        var root = rootDirs[i];
+        const relative = path.relative(path.join(baseDir, root), importerRootRelative);
+        if (!relative.startsWith('.')) {
+          importerRootRelative = relative;
+        }
+      }
+      resolved = path.join(importerRootRelative, importee);
+    } else {
+      throw new Error('cannot resolve relative paths without an importee');
```
gregmagolan: "without an importer"?

PR #133, opened 2018-02-26
Add Windows to .bazelci/presubmit.yml.

philwo:
Our Buildkite CI has working Windows agents now, so do you want to give this a try?

Do you need anything installed on the Windows machines (e.g. nodejs)?

(FYI @buchgr)


alexeagle:
No, I don't want nodejs installed since Bazel should manage it as an external repository. Having it on the machine would actually mask problems where we rely on system node by accident.

PR #139, opened 2018-03-05
Include node directory in PATH for all node sub-commands

### LINKED ISSUE #7: bazel run @yarn//:yarn uses system node instead of nodejs target
This means that rules_nodejs requires node to be installed on the system before it is able to do anything.

```
$ bazel run @yarn//:yarn
INFO: Analysed target @yarn//:yarn (6 packages loaded).
INFO: Found 1 target...
Target @yarn//:yarn.sh up-to-date (nothing to build)
INFO: Elapsed time: 7.735s, Critical Path: 0.01s

INFO: Running command line: [redacted]/a3cee343d71b0f2101c79d88
606a48aa/external/yarn/yarn.sh
yarn install v0.22.0
warning example@0.0.1: License should be a valid SPDX license expression
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
warning "@angular/bazel@5.0.0-beta.5-4ec5e28" has incorrect peer dependency "@angular/compiler-cli@5.0.0-be
ta.5-4ec5e28".
[4/4] Building fresh packages...
$ ngc -p angular.tsconfig.json
/usr/bin/env: node: No such file or directory
error Command failed with exit code 127.
```

It may be be possible to invoke ngc's index.js from the included nodejs and not have this problem but by invoking it directly it uses the /usr/bin/env node shebang.

dflemstr:
Create a shell script that wraps real `node` and adds the directory containing `node` to `PATH`, which makes it so that sub-processes invoke the right `node` (for example during `postinst` hooks)

Fixes #7. Maybe other issues too?


alexeagle:
This would address https://github.com/bazelbuild/rules_nodejs/issues/140
Looks like it might not work on Windows? And I'm not sure why the delta is so big, is this needed or is it just refactoring?
Will let Greg review

alexeagle:
Thanks for the PR - we didn't hear back from you so Greg fixed it in 585a8d601ad2bf61aa47e69accb677df7e795c2c

By the way, for the future, try to minimize the PR so it contains only changes that are required to fix the issue. Otherwise it's much harder to review.

dflemstr:
@alexeagle Apologies for not replying earlier; I'm getting swamped by GitHub notifications :) I thought the changes I made were minimal, but Greg's fix also works so I'm happy with that change!

Issue #138, opened 2018-03-04
rollup_bundle fails to locate node_modules if /node_modules is not located in the workspace root


erickj:
repro:

1. create a new node_repositories and declare package.json in a subdirectory of the workspace, yarn_install with the same package.json:

```
node_repositories(package_json = [
    "//javascript:package.json",
])

yarn_install(
    name = "yarn_install",
    package_json = "//javascript:package.json",
    yarn_lock = "//:yarn.lock"
)
```
2. add some npm's to package json, and yarn install
3. create a js entry point that requires an npm, and a rollup_bundle target:
```
rollup_bundle(
    name = "bundle",
    srcs = [
        "foo.js",
    ],
    entry_point = "javascript/foo.js",
    node_modules = "//@yarn_install:node_modules",
)
```
At this point node_modules will be located in /javascript/node_modules.

4. `bazel build javascript:bundle`

Rollup will fail to resolve the npm imports. And the js bundle will be created without the imports, with log errors 'Missing global varaible names' and 'Unresolved dependencies' like:

```
javascript/foo.js → bazel-out/k8-fastbuild/bin/javascript/bundle.es6.js...
Rollup: try to resolve at /home/erick/.cache/bazel/_bazel_erick/4fbe1c2851c54abfd26b77789ffadbff/bazel-sandbox/4506426477410011211/execroot/io_vos/bazel-out/k8-fastbuild/bin/javascript/bundle.es6/javascript/foo.js
Rollup: try to resolve at /home/erick/.cache/bazel/_bazel_erick/4fbe1c2851c54abfd26b77789ffadbff/bazel-sandbox/4506426477410011211/execroot/io_vos/bazel-out/k8-fastbuild/bin/javascript/bundle.es6/home/erick/.cache/bazel/_bazel_erick/4fbe1c2851c54abfd26b77789ffadbff/execroot/io_vos/bazel-out/k8-fastbuild/bin/javascript/bundle.es6/javascript/protobufjs_es6_bundle.js
Rollup: try to resolve at /home/erick/.cache/bazel/_bazel_erick/4fbe1c2851c54abfd26b77789ffadbff/bazel-sandbox/4506426477410011211/execroot/io_vos/bazel-out/k8-fastbuild/bin/javascript/bundle.es6/protobufjs_es6_bundle.js
[...]

PR #141, opened 2018-03-06
Fixes bazelbuild/rules_nodejs#138, extracts the directories named

erickj:
'node_modules' from the rollup_bundle/node_modules attr and passes the
directory list to the rollup config template. This list is used in the
nodeResolve config per recommendations here:

https://github.com/rollup/rollup-plugin-node-resolve/issues/75#issuecomment-290728570

TESTED=bazel test //...


erickj:
@googlebot I signed it

erickj:
@googlebot Try again!

alexeagle:
Looks like this was resolved in https://github.com/bazelbuild/rules_nodejs/pull/173 instead?

PR #142, opened 2018-03-07
Add replacements feature to npm_package

alexeagle:
This is needed to produce the `@angular/bazel` package.
Also add missing docs to the README

### DIFF
--- a/internal/npm_package/packager.js
+++ b/internal/npm_package/packager.js
@@ -24,8 +24,12 @@ function mkdirp(p) {
   }
 }
 
-function write(p, content) {
+function write(p, content, replacements) {
   mkdirp(path.dirname(p));
+  replacements.forEach(r => {
+    const [regexp, newvalue] = r;
+    content = content.replace(regexp, newvalue);
+  });
   fs.writeFileSync(p, content);
 }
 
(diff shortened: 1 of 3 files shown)


internal/npm_package/packager.js
```diff
@@ -24,8 +24,12 @@ function mkdirp(p) {
   }
 }
 
-function write(p, content) {
+function write(p, content, replacements) {
   mkdirp(path.dirname(p));
+  for (let i = 0; i < replacements.length; i++) {
```
gregmagolan: Could do a ```replacements.forEach(r => {
   const [regexp, newvalue] = r;
   content = content.replace(regexp, newvalue);
})``` if you wanted to avoid the for loop

gregmagolan: or
```
for (let r of replacements) {
   const [regexp, newvalue] = r;
   content = content.replace(regexp, newvalue);
}
```
works too

internal/npm_package/packager.js
```diff
@@ -34,13 +38,22 @@ function write(p, content) {
...
 function main(args) {
-  const [outDir, baseDir, srcsArg, binDir, genDir, depsArg, packPath, publishPath] = args;
+  const [outDir, baseDir, srcsArg, binDir, genDir, depsArg, replacementsArg, packPath, publishPath] = args;
 
+  const replacements = [
+    // Strip content between BEGIN-INTERNAL / END-INTERNAL comments
+    [/(#|\/\/)\s+BEGIN-INTERNAL[\w\W]+END-INTERNAL/g, ''],
+  ];
+  const rawReplacements = JSON.parse(replacementsArg);
+  for (let key of Object.keys(rawReplacements)) {
```
gregmagolan:
```
for (const key in rawReplacements)
```
is a little cleaner and avoids Object.keys

alexeagle: We discourage that pattern in the TypeScript style guide since you can pick up prototype properties by mistake

internal/npm_package/packager.js
alexeagle: sure

PR #144, opened 2018-03-07
Make node path known to env for npm & yarn when executing...

### LINKED ISSUE #140: postinstall scripts use system node
Easy to reproduce on a machine without node installed at all.

We should set `--scripts-prepend-node-path` when calling `npm install` so that the node on the path is the one invoking the `npm` command, this makes `"postinstall": "node my_script.js"` work when the user does `bazel run @nodejs//:npm install`

However the equivalent feature for `bazel run @yarn//:yarn` seems to be broken, so we need to find some workaround? File a bug with yarn? Bazel manages its own yarn installation so patching yarn is totally an option.

Finally, what do we do about cases where you run an npm bin, like `postinstall: "ngc -p angular.tsconfig.json"`? I think maybe this just works? but needs testing

gregmagolan:
...so it is available globally and used instead of system node

/cc @alexeagle 

Fixes #140


gregmagolan:
This fixes yarn on osx/linux and npm on all platforms. Fix for yarn on Windows is in progress.

gregmagolan:
Updated to PR to work with yarn on Windows

PR #147, opened 2018-03-09
Test npm_install and yarn_install determinism

gregmagolan:
CI test for issue for #98

### DIFF
--- a/examples/packages/WORKSPACE
+++ b/examples/packages/WORKSPACE
@@ -24,6 +24,12 @@ npm_install(
 )
 
 npm_install(
+    name = "npm_install_duplicate_for_determinism_testing",
+    package_json = "//:package.json",
+    package_lock_json = "//:package-lock.json",
+)
+
+npm_install(
     name = "npm_no_lockfile",
     package_json = "//:package.json",
 )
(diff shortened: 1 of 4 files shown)


examples/packages/WORKSPACE
```diff
@@ -23,6 +23,12 @@ npm_install(
     package_lock_json = "//:package-lock.json",
 )
 
+npm_install(
```
alexeagle: add comments explaining the purpose of these duplicates

examples/packages/WORKSPACE
```diff
@@ -23,6 +23,12 @@ npm_install(
...
+    name = "npm_install_dup",
```
alexeagle: also consider a longer name like `npm_install_duplicate_for_determinism_testing` - no one needs to type this out so it's better to add documentation in the name

examples/packages/npm_install_dup.spec.js
```diff
@@ -0,0 +1 @@
+require('./test_version').test('1.2.0');
```
alexeagle: instead of checking in the extra file, can you create it with a `genrule` that copies from the other during the build?

examples/packages/WORKSPACE
gregmagolan: 👍

examples/packages/npm_install_dup.spec.js
gregmagolan: just removed the two extra version test. not needed.

examples/packages/WORKSPACE
gregmagolan: still think a comment is needed or is the run name as `npm_install_duplicate_for_determinism_testing` sufficient?

alexeagle: good enough w/o comment

PR #154, opened 2018-03-14
Add ability for npm_package to include other npm_package's

alexeagle:
this allows ng_package rule to be a macro composing some special packaging with a regular npm_package, reducing duplication


gregmagolan:
WindowsCI failures expected?

alexeagle:
Windows CI failure was interesting actually - the artifact we need to resolve from the runfiles is the package, which is a TreeArtifact. But `require.resolve` can't be used to resolve the path of a directory on disk. So I had to resolve a sibling to make it work against the runfiles MANIFEST.

/cc @dslomov runfiles MANIFEST strikes again...

PR #158, opened 2018-03-15
Add export support for rollup bundles

mhevery:
Add export support for rollup bundles which export symbols

If the bundle, which rollup creates, has an export symbol, than the whole bundle needs to be reachable under some symbol. This feature adds name to the exported bundle, so that the exported symbols from the bundle can be reached.


internal/rollup/rollup.config.spec.js
```diff
@@ -76,4 +76,9 @@ describe('rollup config', () => {
     expect(doResolve('external/some_wksp/path/to/a'))
         .toEqual(`${baseDir}/bazel-bin/path/to/b.esm5/external/some_wksp/path/to/a/index.js`);
   })
+
+  it('should contain name for iife', () => {
+    expect(rollupConfig.output.format).toEqual('iife');
+    expect(rollupConfig.output.name).toEqual('TMPL_label_name');
```
alexeagle: this is a change detector test - it doesn't assert the thing you care about which is that something actually interpolates the name into this location, and that rollup works.
I think a change detector is worse than no test - it has zero assertive power and yet requires maintenance anytime we change the settings.
I think the test belongs in the e2e, can you repro your actual failure there?

mhevery: Fixed and improved. PTAL

PR #162, opened 2018-03-23
Use ctx.version_file in place of a genrule with stamp=True

### LINKED ISSUE #156: Documentation for stamping is incorrect
In the readme, the explanation of how to do stamping is wrong. Basically, if you are writing a genrule that has `stamp = True`, that genrule will only work as expected if it has at least one file in its `srcs`. See https://github.com/bazelbuild/bazel/issues/4842 for a full explanation.

alexeagle:
Fixes #156

See discussion in https://github.com/bazelbuild/bazel/issues/4842
and this fix is based on https://github.com/bazelbuild/bazel/issues/1054#issuecomment-219151351

### DIFF
--- a/internal/npm_package/npm_package.bzl
+++ b/internal/npm_package/npm_package.bzl
@@ -39,11 +39,15 @@ def create_package(ctx, devmode_sources, nested_packages):
   args.add([p.path for p in nested_packages], join_with=",")
   args.add(ctx.attr.replacements)
   args.add([ctx.outputs.pack.path, ctx.outputs.publish.path])
-  args.add(ctx.file.stamp_data.path if ctx.file.stamp_data else '')
+  args.add(ctx.version_file.path if ctx.version_file else '')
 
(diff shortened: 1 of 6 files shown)


internal/npm_package/npm_package.bzl
```diff
@@ -39,11 +39,11 @@ def create_package(ctx, devmode_sources, nested_packages):
   args.add([p.path for p in nested_packages], join_with=",")
   args.add(ctx.attr.replacements)
   args.add([ctx.outputs.pack.path, ctx.outputs.publish.path])
-  args.add(ctx.file.stamp_data.path if ctx.file.stamp_data else '')
+  args.add(ctx.version_file.path if ctx.version_file else '')
 
   inputs = ctx.files.srcs + devmode_sources + nested_packages + [ctx.file._run_npm_template]
-  if ctx.file.stamp_data:
-    inputs.append(ctx.file.stamp_data)
+  if ctx.version_file:
```
mprobst: It might be worth adding a comment here, too.

PR #167, opened 2018-03-28
Support multiple NodeJS versions.

chrisleck:
Chipping away at https://github.com/bazelbuild/rules_nodejs/issues/4

Leave v8.9.1 as the default, add support for v8.9.4 as proof of concept.

Questions:
- [ ] How should this be tested?

Future work:
- [ ] Consider allowing users to pass in the version_info parameters directly as parameters to this repository rule.
- [ ] Consider generating package URLs from functions using string substitution instead of writing them all out.

### DIFF
--- a/internal/node/node_repositories.bzl
+++ b/internal/node/node_repositories.bzl
@@ -23,35 +23,80 @@ load(":node_labels.bzl", "get_node_label")
 load("//internal/common:check_bazel_version.bzl", "check_bazel_version")
 load("//internal/npm_install:npm_install.bzl", "yarn_install")
 
+_node_versions = {
+  "v8.9.1": {
+    "mac": {
+      "repos": [
+        "https://mirror.bazel.build/nodejs.org/dist/v8.9.1/node-v8.9.1-darwin-x64.tar.gz",
+        "https://nodejs.org/dist/v8.9.1/node-v8.9.1-darwin-x64.tar.gz",
+      ],
+      "stripPrefix": "node-v8.9.1-darwin-x64",
+      "sha256": "05c992a6621d28d564b92bf3051a5dc0adf83839237c0d4653a8cdb8a1c73b94"
+    },
+    "windows": {
+      "repos": [
+        "https://mirror.bazel.build/nodejs.org/dist/v8.9.1/node-v8.9.1-win-x64.zip",
+        "http://nodejs.org/dist/v8.9.1/node-v8.9.1-win-x64.zip",
+      ],
+      "stripPrefix": "node-v8.9.1-win-x64",
+      "sha256": "db89c6e041da359561fbe7da075bb4f9881a0f7d3e98c203e83732cfb283fa4a"
+    },
+    "linux": {
+      "repos": [
+        "https://mirror.bazel.build/nodejs.org/dist/v8.9.1/node-v8.9.1-linux-x64.tar.xz",
+        "http://nodejs.org/dist/v8.9.1/node-v8.9.1-linux-x64.tar.xz",
+      ],
+      "stripPrefix": "node-v8.9.1-linux-x64",
+      "sha256": "8be82805f7c1ab3e64d4569fb9a90ded2de78dd27cadbb91bad1bf975dae1e2d"
+    }
+  },
+  "v8.9.4": {
+    "mac": {
+      "repos": [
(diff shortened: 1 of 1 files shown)


chrisleck:
Not sure if I need to do a notification; sorry for the spam if I don't.
PTAL @alexeagle

internal/node/node_repositories.bzl
```diff
@@ -23,35 +23,80 @@ load(":node_labels.bzl", "get_node_label")
...
+        "http://nodejs.org/dist/v8.9.1/node-v8.9.1-linux-x64.tar.xz",
+      ],
+      "stripPrefix": "node-v8.9.1-linux-x64",
+      "sha256": "8be82805f7c1ab3e64d4569fb9a90ded2de78dd27cadbb91bad1bf975dae1e2d"
+    }
+  },
+  "v8.9.4": {
+    "mac": {
+      "repos": [
+        "https://mirror.bazel.build/nodejs.org/dist/v8.9.4/node-v8.9.4-darwin-x64.tar.gz",
```
alexeagle: I need to actually create these entries on the bazel.build mirror, last time I tried my permissions did not work...

internal/node/node_repositories.bzl
```diff
@@ -23,35 +23,80 @@ load(":node_labels.bzl", "get_node_label")
...
+    "linux": {
+      "repos": [
+        "https://mirror.bazel.build/nodejs.org/dist/v8.9.1/node-v8.9.1-linux-x64.tar.xz",
+        "http://nodejs.org/dist/v8.9.1/node-v8.9.1-linux-x64.tar.xz",
+      ],
+      "stripPrefix": "node-v8.9.1-linux-x64",
+      "sha256": "8be82805f7c1ab3e64d4569fb9a90ded2de78dd27cadbb91bad1bf975dae1e2d"
+    }
+  },
+  "v8.9.4": {
```
alexeagle: How likely that someone needs 8.9.1 and cannot upgrade to 8.9.4? I understand the intent here is to allow choosing versions, but I thought this was more for major versions. this doesn't seem like a great example, maybe we can just change 8.9.1 to 8.9.4 without introducing configurability?

internal/node/node_repositories.bzl
```diff
@@ -198,11 +246,12 @@ def node_repositories(package_json):
...
     package_json: a list of labels, which indicate the package.json files that need to be installed.
+    node_version: the version of NodeJS to install.
   """
   # Windows users need sh_binary wrapped as an .exe
   check_bazel_version("0.5.4")
 
-  _node_repo(name = "nodejs", package_json = package_json)
+  _node_repo(name = "nodejs", package_json = package_json, node_version = node_version)
 
   _yarn_repo(name = "yarn", package_json = package_json)
```
alexeagle: do we need to give a way to control the yarn version too?

chrisleck:
Regarding the choice of `8.9.4`, that's what my team is currently standardized on :)

internal/node/node_repositories.bzl
chrisleck: It looks like we fall back to the public URL in these cases, so probably fine?

internal/node/node_repositories.bzl
chrisleck: Yes, we could (and it would solve my immediate needs), but doing this paves the way for multiple version support which is what we're really after here.

That said, nodejs has made some breaking changes in minor version updates in the past, so it's probably worth considering supporting even minor versions in this way.

internal/node/node_repositories.bzl
chrisleck: Probably - but we can tackle that in a subsequent PR :)

chrisleck:
@alexeagle - do you think it's worth keeping a hardcoded list of supported versions like this, or should we just open things up and let users specify the d/l URL, filenames, and hashes as parameters?  The latter has the nice property of allowing users to move between nodejs versions at their own pace.

alexeagle:
I had imagined solving this by giving users control of the download (maybe
even hand us a label pointing to the archive file for each platform, with
their own download_and_extract calls in their WORKSPACE. That does keep our
code shorter, and we'd only have to provide one default. It also lets you
vendor the nodejs into your repo and avoid downloading it, if you choose
(that's what we do internally at google)

Also we should include documentation in the PR if possible?

alexeagle:
Replaced by #183

PR #173, opened 2018-04-03
Update rollup_bundle to handle node_modules that are not in root WORKSPACE folder

### LINKED ISSUE #138: rollup_bundle fails to locate node_modules if /node_modules is not located in the workspace root

gregmagolan:
Fixes #138


internal/test/hello/index.js
```diff
@@ -0,0 +1 @@
+export default 'Hello';
```
alexeagle: why are there changes in both `internal/e2e` and `internal/test`? Seems like this file and the `internal/test/hello/package.json` don't do anything?

internal/rollup/rollup_bundle.bzl
```diff
@@ -65,6 +65,13 @@ def write_rollup_config(ctx, plugins=[], root_dir=None, filename="_%s.rollup.con
   if not root_dir:
     root_dir = "/".join([ctx.bin_dir.path, build_file_dirname, ctx.label.name + ".es6"])
 
+  node_modules_path = ["."]
+  if ctx.attr.node_modules.label.workspace_root:
```
alexeagle: nit: we have this logic in a bunch of other places, usually expressed as a const expression
```
"/".join[f for f in [workspace_root, package.split(), "thing"] if f]
```
would be nice to be consistent

WORKSPACE
```diff
@@ -35,6 +35,7 @@ node_repositories(package_json = [
     "//:package.json",
     "//examples/rollup:package.json",
     "@program_example//:package.json",
+    "//internal/e2e/rollup:package.json",
```
alexeagle: opt: rather than actually do the install, you could just have the `node_modules` directory checked in

internal/test/hello/index.js
gregmagolan: whoops. didn't mean to commit those. initially had the hello module in test before moving it to e2e/rollup

internal/rollup/rollup_bundle.bzl
gregmagolan: Thanks. Just thinking about this and although this logic works it assumes that the `node_modules` filegroup is defined in the same package as the `node_modules` folder is. Do you think this is reasonable restriction?

gregmagolan: Something like this from erickj's fix is more robust:

```
 node_paths = []
 for f in ctx.files.node_modules:
   # This is a bit of a hack, assuming the node_modules label is
   # actually referencing a directory called "node_modules".
   idx = f.path.find("node_modules")
   if idx >= 0:
      # Resolve paths relative to execroot.
      node_paths.append("./%snode_modules" % f.path[0:idx])
uniq_node_paths = {el: None for el in node_paths}.keys()
```

Or a similar approach is taken in this PR which just got put up:

https://github.com/bazelbuild/rules_nodejs/pull/176/files

Thoughts?

gregmagolan: Looks like this assumption is already made in rules_typescript as well
```
  config["bazelOptions"]["nodeModulesPrefix"] = "/".join([p for p in [
    ctx.attr.node_modules.label.workspace_root,
    ctx.attr.node_modules.label.package,
    "node_modules"
  ] if p])
```

PR #174, opened 2018-04-03
Adds data attribute to npm_install and yarn_install

### LINKED ISSUE #169: Can't run some postinstall scripts with npm_install or yarn_install
If a postinstall script requires additional files (i.e. https://github.com/alexeagle/angular-bazel-example/blob/master/package.json#L30), then it fails with `Error: ENOENT: no such file or directory`.

I'm guessing this is because the path to the file is no longer correct, but that's just a guess.

Is there a way to get that to work?

gregmagolan:
Fixes #169

### DIFF
--- a/internal/npm_install/npm_install.bzl
+++ b/internal/npm_install/npm_install.bzl
@@ -54,6 +54,14 @@ filegroup(
           repository_ctx.attr.package_lock_json,
           repository_ctx.path("package-lock.json"))
 
+  # Add data dependencies to the repository
+  for f in repository_ctx.attr.data:
+    to = []
+    if f.package:
+      to += [f.package]
+    to += [f.name]
+    repository_ctx.symlink(f, repository_ctx.path("/".join(to)))
+
   node = get_node_label(repository_ctx)
   npm = get_npm_label(repository_ctx)
 
@@ -93,6 +101,7 @@ npm_install = repository_rule(
             allow_files = True,
             single_file = True,
         ),
+        "data": attr.label_list(),
     },
     implementation = _npm_install_impl,
 )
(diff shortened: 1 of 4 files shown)


internal/npm_install/npm_install.bzl
```diff
@@ -54,6 +54,14 @@ filegroup(
           repository_ctx.attr.package_lock_json,
           repository_ctx.path("package-lock.json"))
 
+  # Add data dependencies to the repository
+  for f in repository_ctx.attr.data:
+    to = []
+    if f.package:
+      to += [f.package]
+    to += [f.name]
+    repository_ctx.symlink(f, repository_ctx.path("/".join(to)))
```
alexeagle: does it work on Windows? seems like this relies on file symlinks?

internal/npm_install/npm_install.bzl
```diff
@@ -130,6 +139,14 @@ filegroup(
           repository_ctx.attr.yarn_lock,
           repository_ctx.path("yarn.lock"))
 
+  # Add data dependencies to the repository
```
alexeagle: nit: seems like the code duplication between npm and yarn is getting to the point where it's worth sharing

internal/npm_install/npm_install.bzl
gregmagolan: It does. Not sure what bazel does but maybe it just copies the file instead of symlinking it. Same function is used to get 'package.json' into the repository.

gregmagolan:
```
  repository_ctx.symlink(
      repository_ctx.attr.package_json,
      repository_ctx.path("package.json"))
```

internal/npm_install/npm_install.bzl
gregmagolan: good call. factored out the common code.

internal/npm_install/npm_install.bzl
alexeagle: okay, interesting

PR #176, opened 2018-04-04
Pass node_modules location to rollup bundler.

SirVer:
The assumption is that all `node_modules` have a common root path, which should be safe since most javascript tools seem to operate with this assumption anyways.

Without this patch, external node_modules, i.e. tracked by bazel are not found by rollup. This change does not change anything if node_modules is in your root path - which seems to be the common case for users of this repository.


gregmagolan:
@SirVer Thanks! I put up PR #173 for the same issue just a day before you that just got merged.

SirVer:
@gregmagolan All the better! And your fix is much shorter than mine too. :)

PR #178, opened 2018-04-09
Add source-map-explorer output to rollup_bundle

alexeagle:
/cc @mhevery


gregmagolan:
LGTM

alexeagle:
merged as 1931156

PR #179, opened 2018-04-09
Allow users to control output.name in rollup_bundle

alexeagle:


alexeagle:
Needed for P0 Angular regression: https://github.com/angular/angular/issues/23238

PR #183, opened 2018-04-16
Support multiple versions of NodeJS and Yarn

chrisleck:
Taking #167 further per discussion with @alexeagle.

* Provides new repository rules for installing remote or locally vendored copies of NodeJS and Yarn
* Updates the existing `node_repositories` rule to enable passing in a version string for NodeJS and Yarn, for ease of installing a specific, known version.
* Lots of documentation updates I found useful when reading through the code in this area.


chrisleck:
FWIW, I based this on a read-through of the Go rules.

README.md
```diff
@@ -34,6 +34,96 @@ load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories")
...
+    name = "build_bazel_rules_nodejs",
+    remote = "https://github.com/bazelbuild/rules_nodejs.git",
+    tag = "0.3.1", # check for the latest tag when you install
+)
+
+load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories", "node_download_runtime", "yarn_download")
+
+node_download_runtime(
+    name = "nodejs",
+    version = "8.10.0",
```
alexeagle: not sure what we can do about this, but we've seen a request to have the version of npm be free to modify as well given a fixed version of node.

internal/node/node_repositories.bzl
```diff
@@ -198,13 +379,40 @@ def node_repositories(package_json):
...
+      yarn_download(
+        name = "yarn",
+        version = yarn_version,
+        filename = filename,
+        strip_prefix = strip_prefix,
+        sha256 = sha256,
+        package_json = package_json,
+      )
+    else:
+      fail("Unknown Yarn version {}".format(yarn_version))
```
alexeagle: how would the user know which versions are supported?

internal/node/node_repositories.bzl
```diff
@@ -198,13 +379,40 @@ def node_repositories(package_json):
...
+    node_version: optional; the specific version of NodeJS to install.
+    yarn_version: optional; the specific version of Yarn to install.
   """
   # Windows users need sh_binary wrapped as an .exe
   check_bazel_version("0.5.4")
 
-  _node_repo(name = "nodejs", package_json = package_json)
+  # If the caller already executed node_download_runtime or node_local_runtime, and correctly used the "nodejs"
+  # repo name, then we can skip this step.
+  if "nodejs" not in native.existing_rules():
```
alexeagle: ooh I never saw this existing_rules() before...

internal/node/node_repositories.bzl
```diff
@@ -156,27 +266,98 @@ export PATH="$NODE_PATH":$PATH
...
+  - Normal Usage:
+    Specify no explicit versions. This will download and use the latest NodeJS & Yarn that were available when the
+    version of rules_nodejs you're using was released.
+  - Forced version(s):
+    You can select the version of NodeJS and/or Yarn to download & use by specifying it when you call node_repositories,
+    but you must use a value that matches a known version.
+  - Using a custom version:
+    You can call node_download_runtime or node_local_runtime to setup a particular NodeJS version or 
+    yarn_download or yarn_local to setup a particular Yarn version before calling node_repositories.
+    The "local" versions of these rules enable a repo to use vendored copies of tools rather than requiring a download.
```
alexeagle: probably easier to understand if it's documented as a fourth way

README.md
chrisleck: They're packaged together, so I imagine this will be difficult.  Deferring that for now.

internal/node/node_repositories.bzl
chrisleck: We could try to keep this list in sync with something in the top level readme?  I'll add something there and we can see how it works out.

internal/node/node_repositories.bzl
chrisleck: Done.

chrisleck:
Updated to address feedback.  PTAL @alexeagle

alexeagle:
Awesome thank you!!

chrisleck:
Would it be possible to get a new release tag which includes this?

PR #188, opened 2018-04-30
Only include .js files when collecting devmode outputs of transitive …

alexeagle:
…deps


internal/common/sources_aspect.bzl
```diff
@@ -34,7 +34,8 @@ def _sources_aspect_impl(target, ctx):
   if hasattr(target, "typescript"):
     result = depset(transitive=[result, target.typescript.es5_sources])
   elif hasattr(target, "files"):
-    result = depset(transitive=[result, target.files])
+    result = depset([f for f in target.files if f.path.endswith(".js")],
```
gregmagolan: Intentionally outside of transitive?

alexeagle: yes, because this is a list now that I've iterated and filtered it, not a depset

PR #190, opened 2018-04-30
Collect amd names from js_library targets

alexeagle:
Not sure we love this design yet, but this is an internal-only API right now so it seems harmless to add a bit more complexity here? WDYT?


internal/js_library/js_library.bzl
```diff
@@ -17,8 +17,40 @@
...
+  """
+
+  amd_names_shim_content = """// GENERATED by ts_web_test.bzl
+// Shim these global symbols which were defined by a bootstrap script
+// so that they can be loaded with named require statements.
+"""
+  for t in targets:
+    if AmdNames in t:
+      for n in t[AmdNames].names.items():
+        amd_names_shim_content += "define(\"%s\", () => %s);\n" % n
```
gregmagolan: Make this `define(\"%s\", function() { return %s; });` so its es5?

alexeagle: yeah good idea to make it as portable as possible

PR #191, opened 2018-05-02
Ensure "files" from data attribute of nodejs_binary rule make it into runfiles

gregmagolan:
This fixes `//internal/npm_package/test:test` when preserve-symlinks is turned on

### DIFF
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -85,6 +85,8 @@ def _nodejs_binary_impl(ctx):
     for d in ctx.attr.data:
       if hasattr(d, "node_sources"):
         sources += d.node_sources.to_list()
+      if hasattr(d, "files"):
+        sources += d.files.to_list()
 
     _write_loader_script(ctx)
 


gregmagolan:
@alexeagle Should we include an optional way to turn _on_ preserve-symlinks for use in repos we control?

internal/node/node.bzl
```diff
@@ -85,6 +85,8 @@ def _nodejs_binary_impl(ctx):
     for d in ctx.attr.data:
       if hasattr(d, "node_sources"):
         sources += d.node_sources.to_list()
+      if hasattr(d, "files"):
```
gregmagolan: Note: this does have to be an "if" and not an "elif" to fix `//internal/npm_package/test:test` when preserve-symlinks is turned on

alexeagle:
Turning on per-repo would be cool, that gives users some time to get past the breaking change without having to stay on an old version.
Can we do it in the node_repositories rule in the WORKSPACE - I haven't found a good way to do that.

internal/node/node.bzl
gregmagolan: This makes me think that other code that has
```
  for d in ctx.attr.deps:
    if hasattr(d, "node_sources"):
      files = depset(transitive=[files, d.node_sources])
    elif hasattr(d, "files"):
      files = depset(transitive=[files, d.files])
```
like in `devmode_js_sources.bzl` and also in rules_typescript `ts_devserver.bzl` and `ts_webtest.bzl` in incorrect

gregmagolan:
> Would be nice to include some form of regression test at the same time - that's probably why you're thinking about enabling the flag one repo at a time

Exactly. I'll think about how to turn it on from WORKSPACE

PR #192, opened 2018-05-03
Add WORKSPACE preserve_symlinks option for node

gregmagolan:


internal/node/node_repositories.bzl
```diff
@@ -383,6 +400,7 @@ def node_repositories(package_json, node_version=DEFAULT_NODE_VERSION, yarn_vers
     package_json: a list of labels, which indicate the package.json files that need to be installed.
     node_version: optional; the specific version of NodeJS to install.
     yarn_version: optional; the specific version of Yarn to install.
+    preserve_symlinks: Turn on --node_options=--preserve-symlinks for nodejs_binary and nodejs_test rules
```
alexeagle: needs more explanation about how we are switching to this mode as default, failures you might see, how to resolve those, what to do if one of your deps is non-compliant

internal/node/node_repositories.bzl
```diff
@@ -337,10 +353,11 @@ yarn_local = repository_rule(
   attrs = {
     "path": attr.string(),
     "package_json": attr.label_list(),
+    "preserve_symlinks": attr.bool(default = False),
```
alexeagle: TODO to switch default

internal/node/node_repositories.bzl
```diff
@@ -285,7 +296,9 @@ def _yarn_download_impl(repository_ctx):
     urls to use.
 
     package_json: a list of labels, which indicate the package.json files that the npm label will point to.
-  
+
+    preserve_symlinks: Turn on --node_options=--preserve-symlinks for nodejs_binary and nodejs_test rules
```
alexeagle: maybe share the doc using a constant?

alexeagle: and why is there so much duplication here?

gregmagolan: There are the 4 variants now. npm_download_runtime, npm_local_runtime, yarn_download, yarn_local, which are all very similar.

internal/node/node_repositories.bzl
gregmagolan: 👍

internal/node/node_repositories.bzl
gregmagolan: 👍

internal/node/node_repositories.bzl
gregmagolan: Can't think of how to reduce the duplicity. Linter will probably complain if the doc is not inline

PR #195, opened 2018-05-07
Include declarations when bundling typescript sources

nimerritt:
When using the npm_package rule and referencing ts_library rules only the direct dependencies declaration files are included. This causes the TypeScript compiler to complain when trying to resolve the import statements.

### DIFF
--- a/examples/program/BUILD.bazel
+++ b/examples/program/BUILD.bazel
@@ -15,7 +15,10 @@ filegroup(
 
 mock_typescript_lib(
     name = "mock_ts_lib",
-    srcs = ["decrement.js"],
+    srcs = [
+        "decrement.d.ts",
+        "decrement.js",
+    ],
 )
 
 # This is like a "js_library", but there are no actions to run on JS files so a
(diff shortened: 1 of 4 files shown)


nimerritt:
I'm not sure if I should update the `mock_typescript_lib` to include the declarations attribute or whether it should only conditionally include the declarations if the attribute is defined. I'm guessing this could benefit from a minimal example of the problem and a test to show that this fixes it.

alexeagle:
Thanks for digging in! I'll fix it in #198 

A note on the difference in our PRs - you put the .d.ts files in the node_sources, but I don't think this is what we want, it means that every program we execute retains the typings. I think the node_sources should just be the files which are required at runtime. We want a different side-channel to propagate typings to downstream rules; for internal bazel deps we just have ts_library dependencies, and as you point out when shipping to npm we need them included in the package

nimerritt:
I totally agree with the `node_sources` I just didn't know where to put them. Thanks for being so responsive and fixing this. I'll probably write another issue soon as it's not possible to package ts_library targets that are siblings to the npm_package target. I'll modify my example and see if I can reproduce it again while we're on the topic :)

PR #193, opened 2018-05-04
Refactor node_repositories

gregmagolan:
Simplifies code & fixes #181.

Ability to specify version of node & yarn as well as link to local packages is now baked into node_repositories() function and no repository rules are exposed from node_repositories.bzl.

Custom repositories & URLs for download can be configured via node_repositories() parameters.

Yarn is now downloaded to @nodejs workspace. `@nodejs//:yarn` is now the default and recommended target. `@yarn//:yarn` is still available for backward compat.

nodejs repository folder `repository_bin` is put on the PATH when running node. This path includes entry points for node, npm & yarn. These are `repository_bin/node`, `repository_bin/npm`, `repository_bin/yarn` for osx/linux && `repository_bin/node.cmd`, `repository_bin/npm.cmd`, `repository_bin/yarn.cmd` for windows. These entry points handle adding repository_bin to the PATH.

PR #202, opened 2018-05-12
Propogate the exit code from node on osx/linux

gregmagolan:
Fixes node exit code breakage introduced in #193


gregmagolan:
Added xjasmine_node_test which is identical to jasmine_node_test except it expects the test to fail. This tests that node can produce a non-zero exit code.

gregmagolan:
Suppressed node exit code also hid a regression in npm_install and yarn_install. Included fixes for those as well here

defs.bzl
```diff
@@ -18,20 +18,25 @@ Users should not load files under "/internal"
...
 load("//internal/rollup:rollup_bundle.bzl", _rollup_bundle = "rollup_bundle")
 load("//internal/npm_package:npm_package.bzl", _npm_package = "npm_package")
 
 check_bazel_version = _check_bazel_version
+get_host = _get_host
 nodejs_binary = _nodejs_binary
 nodejs_test = _nodejs_test
 node_repositories = _node_repositories
 jasmine_node_test = _jasmine_node_test
+xjasmine_node_test = _xjasmine_node_test
```
alexeagle: why expose these in the public API? I don't think this file should be touched in this PR

internal/common/get_host.bzl
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Get the host for a repository rule
+
+Returns a string describing the host for a repository rule
```
alexeagle: shouldn't this pyDoc be inside the function body?

internal/e2e/node/BUILD.bazel
```diff
@@ -1,7 +1,16 @@
-load("@build_bazel_rules_nodejs//:defs.bzl", "jasmine_node_test")
+load("@build_bazel_rules_nodejs//:defs.bzl", "jasmine_node_test", "xjasmine_node_test")
```
alexeagle: load it from //internal

internal/jasmine_node_test/jasmine_node_test.bzl
```diff
@@ -44,11 +32,71 @@ def jasmine_node_test(name, srcs = [], data = [], deps = [], **kwargs):
...
+    data: Runtime dependencies which will be loaded while the test executes
+    deps: Other targets which produce JavaScript, such as ts_library
+    **kwargs: remaining arguments are passed to the test rule
+  """
+  _jasmine_node_test(name = name, srcs = srcs, data = data, deps = deps, nodejs_rule = nodejs_test, **kwargs)
+
+def _xjasmine_node_test(
+  name,
+  size = None,
+  timeout = None,
```
alexeagle: make this minimal, don't need all these attributes

gregmagolan: These are here so that they can be passed to native.sh_test while `**kwargs` is passed to `_jasmine_node_test`.

internal/e2e/node/BUILD.bazel
gregmagolan: 👍

defs.bzl
gregmagolan: Good point. Its not a useful rule for users 👍

internal/common/get_host.bzl
```diff
@@ -0,0 +1,35 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Helper function for repository rules
+"""
+
+def get_host(repository_ctx):
```
alexeagle: why the word "host"? seems a bit confusing, this isn't related to the hostname or the compiler host.
how about "os_name"?

internal/jasmine_node_test/jasmine_node_test.bzl
alexeagle: the new rule can be short and undocumented since it's not in the public API
then you don't need this refactoring of the current rule, and don't need to pass through all the attrs

internal/jasmine_node_test/xtest.sh
```diff
@@ -0,0 +1,19 @@
...
+"${TEST_BINARY}" "$@"
+RESULT="$?"
+set -e
+
+if (( ${RESULT} == 0 )); then
+  echo "Expected test to fail, but it succeeded" >&2
+  # This exit code is handled specially by Bazel:
+  # https://github.com/bazelbuild/bazel/blob/486206012a664ecb20bdb196a681efc9a9825049/src/main/java/com/google/devtools/build/lib/util/ExitCode.java#L44
+  BAZEL_EXIT_TESTS_FAILED = 3;
+  exit ${BAZEL_EXIT_TESTS_FAILED}
```
alexeagle: this all seems like too much added code

how about an attribute to jasmine_node_test which is the expected exit code, which defaults to zero?

internal/common/get_host.bzl
gregmagolan: 👍

internal/jasmine_node_test/xtest.sh
gregmagolan: 👍 Good idea

internal/node/node_launcher.sh
```diff
@@ -129,4 +129,22 @@ for ARG in "${ALL_ARGS[@]}"; do
...
+RESULT="$?"
+set -e
+
+readonly EXPECTED_EXIT_CODE="TEMPLATED_expected_exit_code"
+if [ "${EXPECTED_EXIT_CODE}" -ne "0" ]; then
+  if (( ${RESULT} != ${EXPECTED_EXIT_CODE} )); then
+    echo "Expected exit code to be ${EXPECTED_EXIT_CODE}, but got ${RESULT}" >&2
+    # This exit code is handled specially by Bazel:
+    # https://github.com/bazelbuild/bazel/blob/486206012a664ecb20bdb196a681efc9a9825049/src/main/java/com/google/devtools/build/lib/util/ExitCode.java#L44
+    readonly BAZEL_EXIT_TESTS_FAILED = 3;
```
alexeagle: I'm a little surprised that this talks about test handling - what if I just `bazel run :my_binary` and look for the exit code 128 as listed in the documentation for my_binary?
I think we either need to:
1) check here for whether we are in a test, like whether $TEST_SRCDIR is defined
2) just exit with the same code as the program - it's up to the program we are running to be aware of Bazel exit codes if it's a test.

I think I prefer #2

internal/node/node_launcher.sh
```diff
@@ -129,4 +129,22 @@ for ARG in "${ALL_ARGS[@]}"; do
...
+    exit ${BAZEL_EXIT_TESTS_FAILED}
+  else
+    exit 0
```
alexeagle: can just fall through in this case, don't need an else clause right?

internal/node/node_launcher.sh
alexeagle: oh I see, if actual exit code is zero then using BAZEL_EXIT_TESTS_FAILED seems reasonable.

internal/node/node_launcher.sh
gregmagolan: yes, since exit code (`$RESULT`) != 0, we need to `exit 0` since the correct non-zero exit code found

gregmagolan: right. I should use the program exit code if it is non-zero and not expected. updated this.

PR #210, opened 2018-05-17
Export targets required for rules_docker

nimerritt:
This also renames `repository_bin` to `bin` as suggested by @gregmagolan.
See https://github.com/bazelbuild/rules_docker/pull/403 for motivation.


internal/node/node.bzl
```diff
@@ -177,12 +177,12 @@ _NODEJS_EXECUTABLE_ATTRS = {
         # dependency on a package like @bazel/typescript.
         # See discussion: https://github.com/bazelbuild/rules_typescript/issues/13
         default = Label("@//:node_modules")),
-    "_node": attr.label(
+    "node": attr.label(
         default = Label("@nodejs//:node"),
```
gregmagolan: Can you add a doc attribute since this is now a public attribute? Something like `doc = """The node entry point target.""",`

nimerritt: Yeah sure but not until tomorrow since I'm in Europe :) I'll also remove the un-needed targets.

PR #212, opened 2018-05-21
Mark _bin test target in test macro as manual

mmikulicic:
The nodejs_test macro uses a sh_test wrapper around an internal `%s_bin` target.
However this target is not tagged as "manual" and thus will get matched by
target wildcard, e.g.: `bazel test //...` would run the same test twice
(and probably fail on windows, since that's the documented rationale for having the sh_test
wrapper in the first place).


mmikulicic:
friendly ping

alexeagle:
Thanks!

Issue #211, opened 2018-05-18
babel not being able to load plugins


Globegitter:
Not sure if this is an issue with babel or bazel or a combination, but I am just trying to write a `js_library` that uses babel to downlevel es6 to amd. I am running into issues with babel trying to load its plugins. I started out just using the cli and passing it a .babelrc - but now also tried using the node api for easier debugging.

So quite simple, for testing now I have:

```
nodejs_binary(
    name = "babel",
    entry_point = "__main__/common/bazel/rules/js/babel.js",
    data = [
        "babel.js",
        "@js_library_deps//:node_modules",
    ],
    node_modules = "@js_library_deps//:node_modules",
    visibility = ["//visibility:public"],
)
```

```
// babel.js
const babel = require("@babel/core");
// This require does work.
// const transform_plugin = require("@babel/plugin-transform-modules-amd");

console.log(babel.transformFileSync("./babel.js", {
    plugins: ["@babel/plugin-transform-modules-amd"]
}).code);
```

Then, when executing this via bazel run I get:

```failed to load main  Error: Cannot find module '@babel/plugin-transform-modules-amd' from '/home/markus/.cache/bazel/_bazel_markus/c83dd5dc66625baf386a59b750f50a10/execroot/__main__/bazel-out/k8-fastbuild/bin/common/bazel/rules/js/babel.runfiles/__main__'
    at Function.module.exports [as sync] (/home/markus/.cache/bazel/_bazel_markus/c83dd5dc66625baf386a59b750f50a10/execroot/__main__/bazel-out/k8-fastbuild/bin/common/bazel/rules/js/babel.runfiles/js_library_deps/node_modules/resolve/lib/sync.js:42:15)
    at resolveStandardizedName (/home/markus/.cache/bazel/_bazel_markus/c83dd5dc66625baf386a59b750f50a10/execroot/__main__/bazel-out/k8-fastbuild/bin/common/bazel/rules/js/babel.runfiles/js_library_deps/node_modules/@babel/core/lib/config/files/plugins.js:100:31)
    at resolvePlugin (/home/markus/.cache/bazel/_bazel_markus/c83dd5dc66625baf386a59b750f50a10/execroot/__main__/bazel-out/k8-fastbuild/bin/common/bazel/rules/js/babel.runfiles/js_library_deps/node_modules/@babel/core/lib/config/files/plugins.js:53:10)
[...]

Issue #200, opened 2018-05-09
rollup_bundle cache support


Globegitter:
Just started experimenting with the rollup_bundle and in general it works quite well. I was just  wondering, is any sort of cache being utilized inbetween builds? As in if just one file is changes, does it r
ollup reanalyse everything again?

From my understanding it would currently not be utilised but not quite sure.

What it says in the rollup docs about it:
```
cache
Object A previously-generated bundle. Use it to speed up subsequent builds — Rollup will only reanalyse the modules that have changed.
```

alexeagle:
Hmm, that's interesting. Typically a bazel action runs in isolation, just given some inputs it produces some outputs, and if any input changes, the whole action is re-run.
However with Bazel workers it is possible to keep a warm instance of some tool, and that instance could hold some state internally. There's no guarantee you get the same worker every time, there are a pool of them to handle parallel builds.
That adds quite a bit of complexity though. We try to squeeze performance in development mode, but bundlers like rollup should only be used in production mode, which isn't in the critical path of your typical development round-trip. So it's better to keep these simple.
I have a draft post about how the code-splitting ought to work, which would allow better parallelization in rollup, but that's a long way out.

I'll close this since it's working as intended that we don't do caching internally in rollup. Thanks for asking!

Globegitter:
@alexeagle Thanks for the answer, makes sense. I think it is reasonably common to also bundle the js in development mode, but you are suggesting not bundling and just using a `nodejs_binary` that can then be served or straight using `http_server` (would be nice to have referenced on the website btw) and the browser can just resolve the es6 modules itself now? Which should have a fairly low development round trip time i suppose. 

Also looking forward to see the post on code-splitting.

alexeagle:
Under devmode there is a different highly scalable bundling strategy. If
you use `ts_web_test` or `ts_devserver` this happens automatically. It's
part of the TypeScript layering because this devmode speed comes from
assuming that all JS source files have named UMD/AMD format, so they can be
concatenated together without any module format transpilation and without
losing the module names.
See
https://github.com/alexeagle/angular-bazel-example/wiki/Running-a-devserver-under-Bazel
Is your use case just authoring in JavaScript, or is it TypeScript?

Globegitter:
Ahh that sounds very interesting I'll have a read. Our case is just Javascript / es2015+, should that be possible here also?

Globegitter:
[...]

Issue #149, opened 2018-03-10
promote the js_library rule to the public API


javatlacati:
I looked through the definitions file and couldn't find any `nodejs_library` definition, does that means that I can only have exactly one global `BUILD` and have a monolithic building process?

alexeagle:
what would you expect such a `nodejs_library` rule to do?

If you want to compile libraries in isolation, use a compile-to-JS rule like `ts_library`
If you want to publish a package to npm, use `npm_package`

an-kumar:
I think the point would be to have multiple nodejs packages that reference one another in the same repository. e.g:

/path/to/first/package
/path/to/second/package (dependent on first package)
/path/to/some/binary (dependent on second package; transitively dependent on first package)

alexeagle:
@an-kumar yes, this is supported already

For TypeScript, you have
```
/path/to/first/package
  BUILD.bazel
    ts_library(name = "first", module_name = "first", srcs =["first.ts"])
/path/to/second/package
  BUILD.bazel
    ts_library(name = "second" module_name = "second", srcs = ["second.ts"], deps = ["//path/to/first/package:first"])
  second.ts
    import * as f from 'first';
/path/to/some/binary
  BUILD.bazel
    ts_library(name = "program", deps = ["//path/to/second/package:second"], srcs = ["program.ts"])
    nodejs_binary(name = "bin", deps = [":program"], ...)
  program.ts
    import * as s from 'second';
```

If you don't use TypeScript, this is still supported, but not yet documented because we aren't sure we understand the use case (why use Bazel if you're not running any compilers?)

https://github.com/bazelbuild/rules_nodejs/blob/befbd8961c18de98e254bea94db08ede78889f86/internal/js_library/js_library.bzl
lets you assign the `module_name` just like above TS example

Globegitter:
@alexeagle Our use case would be to have shared code between different rollup bundles. Both would then just define that common js_library as a dep. Or would you there just directly depend on the files?

alexeagle:
you can just use a plain `filegroup` rule if you want to be able to reference the same group of `.js` inputs in several different bundling rules.

`babel_library` sounds like a great addition, want to try a PR for that?

nimerritt:
@Globegitter we would also benefit from a babel rule as we are migrating a large service to TypeScript and it's nice to use Bazel to mix between the two as we move forward, especially given that the TypeScript rules don't take .js files.

an-kumar:
[...]

Issue #137, opened 2018-03-04
rollup_bundle should use write_amd_names_shim


erickj:
Hi all,

Thanks for this package, it's coming along nicely. What are the plans for supporting commonjs style dependencies in rollup bundles? In rollup_bundle.bzl I see that there is currently an unused `TMPL_additional_plugins`, however I'm unsure of the planned format of this, given that in the rollup.config.js is simply expecting the contents to either be previously required, or a full require+call expression (e.g. `require('myplugin')()`)?

https://github.com/bazelbuild/rules_nodejs/blob/master/internal/rollup/rollup.config.js#L108

In any case, the rollup-plugin-commonjs is common enough that it probably warrants inclusion in the base set of plugins by default (bazelResolve, nodeResolve, and commonjsResolve).

I've hacked in default support for commonjsResolve into my own fork, however it was done w/o any thought to integrating it with the `rollup_bundle` rule give it's only for my use case.

https://github.com/erickj/rules_nodejs/commit/d4c83225f06200c94726b363fd6c4e1f33702b45

What's the future direction here w/ either actually implementing TMPL_additional_plugins or first party support for the cjs plugin?

erickj:
"for example it tries to load rxjs commonjs distro which doesn't tree-shake well" 

:/ i've noticed

"I assume you'd like to bundle some off-the-shelf JS library with your app"

exactly, specifically [protobufjs](https://github.com/dcodeIO/protobuf.js/blob/master/minimal.js) at the moment. there are several other libs as well. Unfortunately cjs is just unavoidable ATM, practically speaking.

erickj:
One update re:

"The commonjs plugin causes problems, for example it tries to load rxjs commonjs distro which doesn't tree-shake well."

IIUC that's actually the fault of the rxjs package.json file previously only declaring `pkg.main` to point to the root index.js file, which is a cjs module. [Now that rxjs is properly using `pkg.module` to point to the _esm5 targets](https://github.com/ReactiveX/rxjs/issues/3227) plus per subpackage package.json files this should be fixed and a non-issue for including rollup-commonjs.

It appears then that the real fix is for package maintainers to migrate towards explicit es6 module support. In the meanwhile, providing cjs support by default (or at least optionally) seems reasonable.

Globegitter:
I think it would be great in general if it was possible to have an easy way to provide further rollup plugins. Just testing this out and we are also in the need of the json plugin, the babel plugin, etc.

[...]

PR #217, opened 2018-05-28
babel_library for babel support and compatibility with ts_devserver

Globegitter:
This is a first proof of concept that picks up on the ideas talked about in https://github.com/bazelbuild/rules_nodejs/issues/149, https://github.com/bazelbuild/rules_nodejs/issues/137 and https://github.com/bazelbuild/rules_nodejs/issues/200 to have full js (using es6 modules) support with `ts_devserver` as well with the possibility of some incremental transpilation before passing to a `rollup_bundle` or equivalent.

I ended up using babel for the js_library rules as it seems to be the easiest way to downlevel es6 modules to amd, this also opens up the nice possibility to reuse this for further transpilation in the future. This does not support node_module dependencies yet and probably has some other ways how it can be improved (e.g. there is the issue with https://github.com/bazelbuild/rules_nodejs/issues/211) but I thought I would push this out before I go on holiday to have something to get the conversation started.


alexeagle:
Hey @Globegitter would you like some help with this? I think Babel rules are an important missing area, and would help to get a chunk of the community interested in Bazel (I talked to James Kyle a bit this week)

Globegitter:
@alexeagle Yeah would be very open to help. I am back from holiday on Friday. Then I can look more at #211. A specific open question I had was how to handle node modules. To me it seems the more bazel way would be to specify each direct dependency directly on the js_library but it seems the direction for the js rules seems to just specify the whole node modules. And I was wondering if it even makes sense to specify that for each js_library target, or if that is just expected to be handled fully by the top level rules, such as ts_devserver and rollup_bundle?

alexeagle:
Okay let's try to schedule some time on Friday to work together on it?

Globegitter:
I just sent you an email.

nimerritt:
@Globegitter @alexeagle I'm looking forward to this! We want to incrementally convert our legacy Babel built projects to TypeScript with Bazel but without a js_library rule it's pretty awkward!

alexeagle:
@Globegitter how is this going? LMK when you're ready for another review pass.

Globegitter:
@alexeagle yeah sorry for the long delay on this, it has not been on my immediate priority list for a while as we found other solutions but it is starting to become more relevant again. I am very close for it to be ready for another review, next week I should have time for the last touches. The benefit of it taking that time is I have gotten a much better sense of how things work with bazel internally :)

Globegitter:
@alexeagle This is now in a revieable state - the build is failing because rules_typescript is expecting bazel 0.15 rather than 0.14 - how can we update? Also there are a few points to discuss still:
 - js_library does not take node_modules now as it does not really make sense with babel. Babel does not load any require/import statements, it just compiles the files one passes in through the cli, so we would have to either statically analyse the code ourselves or try and compile all node_modules. And even when nothing is compiled it already makes things quite a bit slower, in my tests reloads went from ~1.4s to ~2.3s or so.
 - I am not fully happy with the outputs - right now we expose two things, the sourcefiles themselves, which are passed to rollup_bundle and an es5 compiled umd module for the dev_server which works, but I have to admin I expected rollup_bundle to not do es5 compilation itself but rather each js_library doing it to get caching there but I guess that would mess with the sourcemaps? Also I it is a quite common use-case to compile the code to es5 but leave es6 imports in-place, which is currently not supported. But this is probably all fine for a first release and we can iterate on it.
 - The last thing is the output naming/locaion, right now the outputs are put into a `es5` directory under the package directory, which I am not sure is the best approach. I kind of tried to mimic a typical babel behaviour as much as bazel allows, where usually people just compile the whole src directory and put all the output in a `dist` dir or whatever. This obviously does not work with bazel that way. So happy to do it another way if that should work better. I will soon try and use this library with jest rules, so will get some practical experience on how that will work there.
- Also added some fixes to rollup_bundle I ran into.
- Not sure why I have some random style changes, I just ran `yarn format` and it changed it.

an-kumar:
what's the status on this PR?

Globegitter:
@an-kumar it is in a usable state and just needs review. There was a few minor changes I want to revert again, but that is not directly related to the `js_library`. If you could try out this branch and see how it works for you that would be very valuable feedback.

an-kumar:
i'll try it out. is js_library meant to work with nodejs_binary as well?

Globegitter:
@an-kumar ha good question - it should but it is something I have actually not tried out. It is a very good point though, I will add an example for it.

an-kumar:
I’m not sure it will, the module resolution may not work since the babelified code is in a different directory (“es5”). The sources aspect wouldn’t find it I don’t think

Globegitter:
The build is getting the error
```
Current Bazel version is 0.14.0, expected at least 0.15.0
```

Globegitter:
@an-kumar I removed the `es5` output dir again, because I think it would make integration with other tools in general more difficult and I think it is easier to have a requirement either one depends on the `js_library` output or the source file, as most use-cases should be fine with one or the other. I also added an example for it, give it a try and let me know what you think.

an-kumar:
You should remove the leading underscore from `_babelrc_tmpl`. I have a custom babelrc for example -- there's no need to disallow others from overriding your babelrc as far as i can tell.

an-kumar:
few issues i found playing with this:

1- it seems that babelification doesnt work with files that are not top-level (e.g, it didn't work if i had nested files under directories)
2- not all files are being created by the babelification process. I had more than 2 `srcs` and got output was not created. When I went and looked at the file that was created, it seems to include the babelification results from both srcs. Not sure what's going on there, the babel binary seems fine to me.

internal/js_library/v2/babel.js
```diff
@@ -0,0 +1,45 @@
...
+  // Note: We do not want to use bazel resolve mechanisms, so use path resolve to get the absolute
+  // path and load that.
+  babelConfig = require(path.resolve(program['configFile']));
+}
+
+for (let i = 0; i < program.args.length; i += 1) {
+  const input = program.args[i];
+  const output = path.join(outDir, path.basename(input));
+
+  return babel.transformFile(input, babelConfig, function(err, result) {
```
an-kumar: removing `return` here fixes the issue where only one file is created. Of course, `return` here is simply mistaken.

internal/js_library/v2/babel.js
```diff
@@ -0,0 +1,45 @@
...
+let babelConfig = {};
+if (program['configFile']) {
+  // Note: We do not want to use bazel resolve mechanisms, so use path resolve to get the absolute
+  // path and load that.
+  babelConfig = require(path.resolve(program['configFile']));
+}
+
+for (let i = 0; i < program.args.length; i += 1) {
+  const input = program.args[i];
+  const output = path.join(outDir, path.basename(input));
```
an-kumar: this is the source of the issue around nested files. We should pass into this binary directly the output paths for each input, and they should be based on the relative path to the package for each file.

an-kumar:
I have some code that fixes the above issues -- what's the best way of proposing some changes?

Globegitter:
@an-kumar Ah great, yeah any code improvements/fixes are appreciated - you can just make a PR against this branch and I will take a look.

an-kumar:
I’m actually not sure mechanically how to do that. Do I open a or against your fork? Or against this repo?

Globegitter:
@an-kumar It is against this fork, so against https://github.com/ecosia/rules_nodejs/tree/js-library-poc, but if you want you can also push it to a branch in your fork and we can take it from there.

an-kumar:
https://github.com/ecosia/rules_nodejs/pull/2 check it out

Globegitter:
@an-kumar I merged your PR and applied some changes here https://github.com/bazelbuild/rules_nodejs/pull/217/commits/8a74cd6364085914b65ee4268d3e7753f1292405 for it to be working with CI as well as the path stuff to be working with generated files and when executing external workspaces. Can you confirm the CLA here in the PR?

Globegitter:
@alexeagle code is now ready for another review.

package.bzl
```diff
@@ -104,11 +104,31 @@ def rules_nodejs_dev_dependencies():
         sha256 = "d8b663c41039dfd84f3ad26d04f9df3122af090f73816b3ffb8c0df660e1fc74",
     )
 
+    http_archive(
```
alexeagle: we don't want rules_nodejs to depend on rules_typescript as it makes it easier to get new layering violations

theres' a mock_ts_library in this repo for testing

internal/rollup/rollup_bundle.bzl
```diff
@@ -425,6 +430,7 @@ def _rollup_bundle(ctx):
     # Generate the bundles
     rollup_config = write_rollup_config(ctx)
     run_rollup(ctx, collect_es6_sources(ctx), rollup_config, ctx.outputs.build_es6)
+    # TODO - fix source maps - run babel???
```
alexeagle: we have https://github.com/bazelbuild/rules_nodejs/issues/175 already, don't need to add TODO

internal/rollup/rollup.config.spec.js
```diff
@@ -72,7 +72,7 @@ describe('rollup config', () => {
     expect(doResolve(`other${sep}thing`))
         .toEqual(`${baseDir}/bazel-bin/path/to/a.esm5/external/other_wksp/path/to/other_lib/thing`);
     expect(doResolve(`@bar${sep}baz${sep}foo`))
-      .toEqual(`${baseDir}/bazel-bin/path/to/a.esm5/path/to/bar/baz_lib/foo`);        
+        .toEqual(`${baseDir}/bazel-bin/path/to/a.esm5/path/to/bar/baz_lib/foo`);
```
alexeagle: why formatting changes? could you generally go through the PR and remove non-load-bearing deltas so it's easier to review?

examples/devserver/src/BUILD.bazel
```diff
@@ -0,0 +1,51 @@
...
+    srcs = ["module.js"],
+)
+
+js_library(
+    name = "test",
+    srcs = ["test.js"],
+    # Note: This rc file will only be found by babel when run in the devserver_example workspace
+    # but not in the top-level one, due to how bazel currently handels pathing
+    data = [
+        "//:.browserslistrc",
```
alexeagle: this looks like a design flaw to me

if my js_library is depended on by two different root targets that provide binaries for two different platforms, how will those targets select the right outputs? The user would have to have parallel js_library trees right?

internal/js_library/v2/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
...
+nodejs_binary(
+    name = "babel",
+    data = [
+        "@build_bazel_rules_nodejs//internal/js_library/v2:babel.js",
+        "@build_bazel_rules_nodejs_js_library_deps//@babel/core",
+        "@build_bazel_rules_nodejs_js_library_deps//@babel/preset-env",
+        "@build_bazel_rules_nodejs_js_library_deps//commander",
+        "@build_bazel_rules_nodejs_js_library_deps//mkdirp",
+    ],
+    entry_point = "build_bazel_rules_nodejs/internal/js_library/v2/babel.js",
```
alexeagle: why is there a `v2` ? I'd expect this to replace the current `js_library`

internal/js_library/v2/js_library.bzl
```diff
@@ -0,0 +1,172 @@
+
+"""js_library allows defining a set of javascript sources to be used with ts_devserver"""
```
alexeagle: It's really about writing ESnext Javascript while running in runtimes of today

The fact that it's made to interop with `ts_devserver` is only interesting if you also author typescript code

internal/js_library/v2/package.json
```diff
@@ -0,0 +1,9 @@
+{
+    "description": "runtime dependencies for js_library",
+    "devDependencies": {
+        "@babel/core": "^7.1.0",
+        "@babel/preset-env": "^7.1.0",
+        "commander": "^2.18.0",
+        "mkdirp": "^0.5.1"
```
alexeagle: can you avoid this extra dependency? the function is trivial to write out and we don't like dependencies (this one has licensing issues IIRC)

internal/js_library/v2/js_library.bzl
```diff
@@ -0,0 +1,172 @@
...
+        "srcs": attr.label_list(
+            doc = """JavaScript source files from the workspace.
+            These can use ES2015 syntax and ES Modules (import/export)""",
+            allow_files = [".js"]
+        ),
+        "deps": attr.label_list(
+            doc = """Other rules that produce JavaScript outputs, such as `ts_library`.""",
+        ),
+        "data": attr.label_list(
+            doc = """Other files useful for babel such as .browserslistrc""",
```
alexeagle: we need some design discussion about whether this rule should leak the fact that babel is used.

Doing so means that users could write things that only work with a given version of babel, or that don't correctly interoperate with other bazel rules.

Generally the philosophy is that the built-in rules provide a much simpler configuration surface and mental model than using the tools directly, and that customization happens through rule composition, not by changing settings to a particular tool. If power users want to have more control, they can write `.bzl` code that extends from these for their purposes (like `ng_rollup_bundle` extends `rollup_bundle`)

internal/js_library/v2/babel.js
```diff
@@ -0,0 +1,45 @@
+// A simple wrapper around babel-core due to https://github.com/babel/babel/issues/8193
```
alexeagle: note, now that we have "idiomatic install", the existence of this helper code suggests there should be a `@bazel/babel` package containing this JS, and its package.json should have the dependencies currently listed in `internal/js_library/v2/package.json`, and we don't need the extra `build_bazel_rules_nodejs_js_library_deps`.

(Doing it the way you've done is an option too, but we are moving away from that, can discuss tradeoffs more)

internal/js_library/v2/js_library.bzl
Globegitter: I think that is a tricky one here, for non-js rules I would totally agree with you, but I think that is what js developers expect. It is one of the first things I hear internally for our rules: how can I pass in my custom eslintrc, how can I pass in my custom nuxt/jest config, etc. So these are power users within the js world but some of them are totally unfamiliar with the internals of bazel, so asking them to learn yet another thing, rather than just passing in a config they are familiar with makes the selling point for bazel much harder.

And to be honest I have been in the same position where I wanted to customize the rollup_bundle and while I theoretically understand the benefits of not exposing custom configs in practice I find it much preferable to have to ensure my config is still working after upgrading as this also means I get to reap any potential new features and don't have to keep my internal rules in sync. And if for some reason `rollup_bundle` would become deprecated I could still go ahead and create an internal rule for it.

It seems we at ecosia are not the only ones who would want this type of customisation in the js world, given there has been some positive response on https://github.com/bazelbuild/rules_nodejs/pull/234 and there are a bunch of issues around customising rollup, there is also https://github.com/bazelbuild/rules_nodejs/pull/365 and it was actually @an-kumar who pushed to allow a custom babel config.

Maybe the naming is part of the issue here? Should we call this `babel_library`?

Also to take this a bit further it seems to me that `rules_typescript` are leaking the fact that they use `tsc` to compile by mentioning and exposing `tsconfig.json`, right? If we where to swap out `tsc` with `babel@7` I would assume all usages around `tsconfig.json` would start breaking.

But does it maybe make sense to have a discussion about that off of github? As it does seem to be a bigger one.

internal/js_library/v2/babel.js
Globegitter: Cool, I did see some happenings around this but was not fully aware - happy to do so.

internal/js_library/v2/js_library.bzl
alexeagle: I think we should have a living document about the philosophy of the Bazel toolchain, and what things should be the same as a pure JS toolchain and which things are different.

Yes, we have some rules which expose bits of the native config. Protractor is another one (https://angular.github.io/bazel-builds/src/protractor/protractor_web_test.html#protractor_web_test ). In these, we must take some care because parts of the config can be controlled by the user but other parts are related to inputs, outputs, dep graph, and downstream rules and therefore should be managed by bazel. The `--output` flag for Typescript is the obvious one - we output the correct ES syntax level for whatever downstream rule requested the output. Users who want control over that option don't understand how the build graph works. (Not that they should have to understand it, but we don't give them that "foot gun" either).

I agree that if this exposes Babel internals, naming the rule with `babel` in the name is a good convention to indicate that.

Yeah we still need to work through the rollup_bundle customization as well. I think this is a tricky balance of allowing nearly all use cases without making the rules brittle such that they no longer compose easily.

internal/rollup/rollup.config.spec.js
Globegitter: Looking at this file though the formatting fix seems correct:
https://github.com/bazelbuild/rules_nodejs/blob/master/internal/rollup/rollup.config.spec.js#L75
This one line has indentation out of order and trailing whitespaces which does seem incorrect to me.

Either way I will remove it.

examples/devserver/src/BUILD.bazel
Globegitter: Not sure I understand that comment. Yeah so the added `.browserslistrc` can change how babel transpiles the output js file - but you mean that instead of the `js_library` deciding on that the root target should do that? While I do like the idea I am not sure how that would possible.

internal/js_library/v2/package.json
Globegitter: Removed this dependency and noticed that node 10.12 (the new LTS) provides this built-in, so just using that functionality now insread (and submitted a PR to upgrade the version: #392)

internal/js_library/v2/js_library.bzl
Globegitter: Started for now with renaming to `babel_library`. Still happy to have some further discussions.

internal/js_library/v2/babel.js
Globegitter: Done - Then I suppose I just add a `npm_package` rule and what is the best way to make it deploy?

Globegitter: Also one thing that does feel a bit odd now (if I have done everything correctly of course) that now we always have to specify a custom babel if the deps are not installed under the `@npm` workspace. But I guess in most cases that should be fine and people should be able to just install it under the default.

Globegitter:
@alexeagle Managed to update most things you commented on. Re-review would be appreciated.

alexeagle:
I think we should schedule some time to VC with @an-kumar to discuss through design of this and other PRs. It's too slow to do roundtrips in code review for things that could be in a design doc.
We should get @gregmagolan too.
Maybe Monday or Friday of next week?

an-kumar:
I'm up for it. Babel support is a huge part of the javascript ecosystem and if we want to get Bazel actual adoption in js we really need to do it right. Happy to help if I can.

Globegitter:
Yep that sounds good, I could do both days but Friday would work a bit better. How do we best schedule this?

pward123:
fyi, I was getting the following error while trying to create a babel_library.
```
INFO: From Compiling Javascript (Babel) //my-es6-lib:my-es6-lib:
{ Error: EEXIST: file already exists, mkdir 'bazel-out/k8-fastbuild/bin/my-es6-lib/src'
  errno: -17,
  code: 'EEXIST',
  syscall: 'mkdir',
  path: 'bazel-out/k8-fastbuild/bin/my-es6-lib/src' }
```

I resolved it by lifting the mkdirp code from npm_package/packager.js

pward123:
Should the output folder retain transpiled files that I've removed from the source package?

Globegitter:
@pward123 which version of node are you using? You need to be using at least 10.12, though I would recommend 10.13 as it is the latest LTS.

Globegitter:
And not sure I understand your second question - you mean if the transpiled files will stay in `bazel-bin`. As far as I am aware yes they would stay in bazel-bin as it does not do any kind of automatic cleanup etc. If you want to get rid of it (and all of your local cache in fact) then I think you would have to run `bazel clean`

pward123:
Ah, I've been sticking with 8.11 since I have a few legacy apps that can't move to 10 yet.

Globegitter:
If necessary I can also make the code backwards-compatible, it just seemed too nice to now have it available as native node api.

Globegitter:
But how is it going, are you getting these rules to work for you?

Globegitter:
Ahh, you are talking particularly about ibazel, I am not sure, it might just do the right thing if you are using `glob`. Definitely if you are using a `nodejs_binary` that depends on the `babel_library` and run that binary with ibazel. Then I am pretty sure that file would be removed for the binary, as it would just recreate the runfiles directory.

But yeah copying into you repo manually sounds like a good solution for now. Happy to hear any further feedback in how this can be improved.

pward123:
fyi, I've finished my first pass at moving the babel_library code into my own repo. One thing you might be interested in is the minor cleanup and removal of sync operations in [babel.js](https://github.com/pward123/bazel-toy/blob/play-with-babel/packages/internal/babel_library/babel.js)

alexeagle:
Sorry I missed the thread about scheduling a VC.
Could we try again for end of next week? Email me alexeagle@google.com and I'll include @mgechev also

examples/devserver/src/BUILD.bazel
```diff
@@ -0,0 +1,53 @@
...
+    srcs = ["module.js"],
+    babel = "@devserver_example_yarn_install//@bazel/babel/bin:babel",
+)
+
+babel_library(
+    name = "test",
+    srcs = ["test.js"],
+    babel = "@devserver_example_yarn_install//@bazel/babel/bin:babel",
+    # Note: This rc file will only be found by babel when run in the devserver_example workspace
+    # but not in the top-level one, due to how bazel currently handels pathing
```
wtgtybhertgeghgtwtg: I think it should be "handles".

vmax:
@Globegitter thank you for the PR! there are some things I'd kindly ask you to address:
1. `babel_library` now cannot be used as `deps` in `npm_package` due to this error:
```
ERROR: /Users/vmax/work/GraknLabs/grakn/client-nodejs/BUILD:52:1: in npm_package rule //client-nodejs:client-nodejs: 
	File "/Users/vmax/work/GraknLabs/grakn/client-nodejs/BUILD", line 52
		npm_package(name = 'client-nodejs')
	File "/private/var/tmp/_bazel_vmax/a61cd25b9823cba82c925ec919cf6ad0/external/build_bazel_rules_nodejs/internal/npm_package/npm_package.bzl", line 73, in _npm_package
		transitive.append(d.typescript.transitive_declarat...)
	File "/private/var/tmp/_bazel_vmax/a61cd25b9823cba82c925ec919cf6ad0/external/build_bazel_rules_nodejs/internal/npm_package/npm_package.bzl", line 73, in transitive.append
		d.typescript.transitive_declarations
'struct' object has no attribute 'transitive_declarations'
Available attributes: es5_sources, es6_sources, transitive_es5_sources, transitive_es6_sources
ERROR: Analysis of target '//client-nodejs:client-nodejs' failed; build aborted: Analysis of target '//client-nodejs:client-nodejs' failed; build aborted
INFO: Elapsed time: 0,448s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (46 packages loaded)
```

Possible fix I've figured out for it is to include `transitive_declarations = js_providers.transitive_es5_sources` in struct returned from rule implementation. Is it a correct way and if it is, could you add the support to the upstream [`ecosia/rules_nodejs@js-library-poc`](https://github.com/ecosia/rules_nodejs/tree/js-library-poc/)?

2. Pretty common use case for `babel` is using `babel-preset-minify` for minification. Could you include `"babel-preset-minify": "^0.5.0"` requirement in `internal/babel_library/package.json`, and if no, how could I add it?

3. Although `babel`'s configuration and rule's `attr` to pass it is called `.babelrc`, I find it confusing that with `babel` starting with version `7` JSON configs don't seem to be supported and instead JS files are expected (you are using one as well named `babel.rc.js`). Do you know what's the matter here? Is it the limitation of `babel@7` (and should I address `babel` team instead) or it's the limitation of currently-implemented rule?

Again, thank you so much for submitting this PR 🙂

Globegitter:
@vmax the `npm_package` error is a limitation indeed will look at fixing that. The eventual goal of this would be that you can install it as a package `@bazel/babel` and then you provide the installed babel binary to `babel_library` (if you do not use the default `@npm` name) and you should also be able to add any further babel dependencies as needed.

The json `.babelrc` should work in theory but the problem is that babel uses its own method to resolve its plugins so it can not find them when you jsut pass them in as a string, that is why we use the js file and `require` them all at the top.

mattjs:
Hey guys this looks exciting! Any plans to move forward with it? It would be a great tool to have before piping stuff to rollup_bundle.

alexeagle:
I'll pick this up now - I agree that babel is a good step before rollup (maybe we'd also want to use it after rollup to downlevel for older browsers)

alexeagle:
Hey @Globegitter do you have some patience to work with me to get this going again?

I think we ought to start over and pull in some of your ideas and tests, because things have changed a lot since you worked on this:
- don't need to use custom module resolver
- things in their own packages/ directory
- more of a "have the tool behave the same as outside of bazel" mentality

I started a new branch at https://github.com/alexeagle/rules_nodejs/tree/babel - could we pair up to get that to parity with what you need? I haven't used Babel much.

@pward123 are you still using rules_nodejs? I see that you also did some hacking on this rule. Want to help get it upstreamed?

Thanks!

alexeagle:
We now run babel in the examples, so in theory you don't need a `babel_library` rule anymore.
I think we had more ideas here, but it's the oldest PR and is probably too stale to imagine we'll clean it up to merge.

PR #223, opened 2018-06-08
Nodejs macro private visibility & testonly option

gregmagolan:
Some updates to rollup needed for authoring protractor rule

Set `nodejs_binary` in `nodejs_binary_macro` to private visibility and add optional `testonly` attribute so that exported `nodejs_binary` can be used as a test dependency.

Add `patterns` attribute to `node_modules_filegroup`.

### DIFF
--- a/defs.bzl
+++ b/defs.bzl
@@ -38,7 +38,7 @@ rollup_bundle = _rollup_bundle
 npm_package = _npm_package
 # ANY RULES ADDED HERE SHOULD BE DOCUMENTED, run yarn skydoc to verify
 
-def node_modules_filegroup(packages, **kwargs):
+def node_modules_filegroup(packages, patterns=[], **kwargs):
   native.filegroup(
     srcs = native.glob(["/".join([
         "node_modules",
@@ -49,6 +49,6 @@ def node_modules_filegroup(packages, **kwargs):
         "*.js",
         "*.json",
         "*.d.ts",
-    ]]),
+    ]] + patterns),
     **kwargs
   )
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -249,7 +249,7 @@ The runtime will pause before executing the program, allowing you to connect a
 remote debugger.
 """
 
-def nodejs_binary_macro(name, args=[], visibility=None, tags=[], **kwargs):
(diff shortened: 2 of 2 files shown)


gregmagolan:
@mrmeku Seems like rollup 0.59.3 contained a bug that broke the angular build (a few npm_package targets ran forever). Does 0.60.1 work for your purposes?

internal/test/package.json
```diff
@@ -1,19 +1,19 @@
...
     "node_resolve_main": "file:./node_resolve_main",
     "node_resolve_main_2": "file:./node_resolve_main_2",
     "node_resolve_nested_main": "file:./node_resolve_nested_main",
-    "rollup": "0.52.3",
-    "rollup-plugin-node-resolve": "3.0.0",
-    "rollup-plugin-sourcemaps": "^0.4.2",
-    "unidiff": "^0.0.4"
+    "rollup": "0.60.1",
+    "rollup-plugin-node-resolve": "3.3.0",
+    "rollup-plugin-sourcemaps": "0.4.2",
```
alexeagle: since we have lock files, any particular reason to remove usage of ranges?

defs.bzl
```diff
@@ -38,7 +38,7 @@ rollup_bundle = _rollup_bundle
 npm_package = _npm_package
 # ANY RULES ADDED HERE SHOULD BE DOCUMENTED, run yarn skydoc to verify
 
-def node_modules_filegroup(packages, **kwargs):
+def node_modules_filegroup(packages, patterns=[], **kwargs):
```
alexeagle: where would this be used? like for adding `history-server`? I think we should move away from this rule for when `preserve-symlinks` is on by default

internal/test/package.json
gregmagolan: Best practice was the thought. Took that out and made another PR to update rollup.

defs.bzl
gregmagolan: Agreed on moving away from it. I needed this for the protractor rule:
```
node_modules_filegroup(
    name = "node_modules",
    packages = [...]
    patterns = [
        "node_modules/protractor/**",
    ],
)
```

gregmagolan: I think its useful to get in anyway even if angular doesn't use it in the long run

alexeagle: okay makes sense

PR #229, opened 2018-06-13
Downgrade to bazel 0.13.0

gregmagolan:
Bazel 0.14.0 & 0.14.1 show random out of memory crashes on CircleCI. Downgrading to bazel 0.13.0 and removing the 0.14.0 requirement recently introduced because of `@bazel_tools//tools/bash/runfiles`.


alexeagle:
We have a downstream fix so let's leave this release live

PR #230, opened 2018-06-14
Add Overview Text to Generated NodeJS Docs

spomorski:


spomorski:
Ah, I should have put it in the .bzl files like you said. Let me look where
it will fit.

alexeagle:
Ping! @spomorski are you still planning to work on this?

Issue #177, opened 2018-04-05
`jasmine_node_test` should auto restart the JasmineRunner on file change when in debug


Toxicable:
Currently to debug `jasmine_node_test` you must manually restart the process since running `bazel test :target --config=debug` will only do a single run on the jasmine runner.
It should re-run jasmine and hold the process open when in `config=debug`

for ref: https://github.com/bazelbuild/rules_nodejs/blob/master/internal/jasmine_node_test/jasmine_runner.js#L27

alexeagle:
I'm leaning to close this as WAI - I don't think we necessarily need to support the intersection of watch mode and debug mode at the same time. I could see how it's useful, but it also adds complexity and will be hard for us to add tests that we don't break it.

Toxicable:
If neither this is done nor a Jest target is made, then i don't see how were meant to run Nodejs tests with bazel.
Have you tried using this target yourself? It is very tedious and time consuming.
If you want to attach a debugger then you have to manually kill the process and restart the test after each run.
If you want to use ibazel to restart it then you're reduced to using console.log for debugging.

alexeagle:
@filipesilva is looking into this as part of improving round-trip time for running jasmine tests under ibazel

joeljeske:
This would be very helpful! 

I agree with @Toxicable that maintaining the same NodeJS process is critical for properly debugging node tests.

joeljeske:
I ended up implementing this behavior by overriding the jasmine runner. In iBazel it uses worker threads to run the tests. In normal mode, it launches the jasmine runner the standard way. 

I was concerned with an approach that relies on clearing the require cache and reloading the modules. That assumes that the global scope is left clear during prior runs and no other side effects have been performed. This seemed like a better way to encapsulate the tests runs. And any performance hit would occur only in iBazel mode.

PR #233, opened 2018-06-15
feat(jasmine_node_test): add ibazel support

Toxicable:


internal/jasmine_node_test/jasmine_runner.js
```diff
@@ -11,45 +12,94 @@ const UTF8 = {
...
+  // Remove the manifest, some tested code may process the argv.
+  process.argv.splice(2, 1)[0];
+  
+  const watch = true;
+  const jrunner = createRunner(watch);
+  addFiles(manifestLocation, jrunner);
 
-  jrunner.execute();
-  return 0;
+  // TODO: find out how to determine if we're running under ibazel
```
alexeagle: use the environment variable
https://github.com/bazelbuild/bazel-watcher/pull/60/files#diff-79db8648ad38b76a407bbc7744e0a7aeR74

internal/jasmine_node_test/jasmine_node_test.bzl
```diff
@@ -41,10 +41,12 @@ def jasmine_node_test(name, srcs = [], data = [], deps = [], **kwargs):
   all_data += [Label("//internal/jasmine_node_test:jasmine_runner.js")]
   all_data += [":%s_devmode_srcs.MF" % name]
   entry_point = "build_bazel_rules_nodejs/internal/jasmine_node_test/jasmine_runner.js"
+  tags = ["ibazel_notify_changes"]
 
   nodejs_test(
       name = name,
       data = all_data,
+      tags = tags,
```
alexeagle: this prevents the user adding other tags on the target - should amend this macro to have argument `tags = []` and add the ibazel one to the given ones

internal/jasmine_node_test/jasmine_runner.js
```diff
@@ -11,45 +12,113 @@ const UTF8 = {
...
+      console.error('ibazel failed to stop jasmine after 5s; probably a bug');
+      process.exit(1);
+    }, 5000);
+  });
+}
+
+function deleteFromCache(fileName) {
+  // jasmine uses `require(fileName)` to read the files into the program
+  // so we have to delete them from the cache to ensure they read again on subsequent runs
+  // when in ibazel mode
```
alexeagle: be consistent whether it's called "watch mode" or "ibazel mode"

internal/jasmine_node_test/jasmine_runner.js
```diff
@@ -11,45 +12,113 @@ const UTF8 = {
...
+  delete require.cache[require.resolve(fileName)];
+}
+
+function addFiles(manifestLocation, jrunner){
+  deleteFromCache(manifestLocation)
+
+  const manifest = require.resolve(manifestLocation);
+  const file = fs.readFileSync(manifest, UTF8);
```
alexeagle: seems like an extra read? what do you do with this variable?

internal/jasmine_node_test/jasmine_runner.js
```diff
@@ -11,45 +12,113 @@ const UTF8 = {
...
+  fs.readFileSync(manifest, UTF8).split('\n').filter(l => l.length > 0).forEach(f => {
+    // ensure it's actually read from disk again
+    deleteFromCache(f);
+    try {
+      // try requiring it incase there are syntax errors that kills the program later on
+      require(f);
+      jrunner.addSpecFile(f);
+    } catch (error) {
+      // dont add the files to jasmine since it'll run them and it'll kill the process
+      console.error('Error reading file: ', require.resolve(f))
```
alexeagle: message should indicate that we are ignoring the file instead of running it as a spec

internal/jasmine_node_test/jasmine_runner.js
```diff
@@ -11,45 +12,113 @@ const UTF8 = {
...
+  const file = fs.readFileSync(manifest, UTF8);
+  fs.readFileSync(manifest, UTF8).split('\n').filter(l => l.length > 0).forEach(f => {
+    // ensure it's actually read from disk again
+    deleteFromCache(f);
+    try {
+      // try requiring it incase there are syntax errors that kills the program later on
+      require(f);
+      jrunner.addSpecFile(f);
+    } catch (error) {
+      // dont add the files to jasmine since it'll run them and it'll kill the process
```
alexeagle: is this feature related to the watch mode? what happens with the code at HEAD when we run into some file that has syntax errors?
if it's orthogonal to the watch mode, let's move to different PR

internal/jasmine_node_test/jasmine_runner.js
```diff
@@ -11,45 +12,113 @@ const UTF8 = {
...
+  const manifestLocation = args[0];
+  // Remove the manifest, some tested code may process the argv.
+  process.argv.splice(2, 1)[0];
+
+  // only watch when under ibazel
+  const watch = Boolean(process.env[IBAZEL_NOTIFY_CHANGES]);
 
-  jrunner.execute();
-  return 0;
+  const jrunner = createRunner(watch);
```
alexeagle: instead of passing a boolean here, how about passing an onComplete callback? that way anything to do with watch mode is localized here.
As you have it, there is stuff about exit codes in the onComplete above which is unused in single-run mode

Toxicable: I've moved all of the watch logic into `iBazelMode`

internal/jasmine_node_test/jasmine_runner.js
Toxicable: done

internal/jasmine_node_test/jasmine_runner.js
Toxicable: whops, was debugging with that, fixed

internal/jasmine_node_test/jasmine_runner.js
Toxicable: changed all to iBazelMode

internal/jasmine_node_test/jasmine_node_test.bzl
Toxicable: fixed

internal/jasmine_node_test/jasmine_runner.js
Toxicable: It's not directly related to watch mode, but you wouldn't see this specific error in any other scenario.
On HEAD the process only ever runes once, so it dosen't matter if jasmine tries using require since it fails the test anyway.

Globegitter:
Is there anything still blocking this from being merged? Apart from the merge conflict?

Toxicable:
@Globegitter I ran into a few issues with this not actually working correctly and had very little time recently to do any open source stuff.

alexeagle:
@Toxicable need someone from our team to pick this up?

Toxicable:
@alexeagle yeah that would be good on this one

Globegitter:
@alexeagle I think this might be running into an issue with bazel/ibazel itself. See https://github.com/bazelbuild/bazel-watcher/issues/184 and https://github.com/bazelbuild/bazel/pull/8322

Would be good to get https://github.com/bazelbuild/bazel/pull/8322 pushed forward.

Toxicable:
this probably needs to be rewritten, will redo if there's interest in the linked ticket

joeljeske:
I am interested in this capability. I think it could greatly enhance the the DX while writing tests.

@Toxicable, sorry, what linked ticked are you referring to? (Both tickets mentioned by @Globegitter appear resolved at the time of this ticket closing?)

Toxicable:
@joeljeske this would be the one https://github.com/bazelbuild/rules_nodejs/issues/177

PR #234, opened 2018-06-18
Allow for custom rollup config

### LINKED ISSUE #216: How to include node builtins in rollup bundle
I'm trying to move our application from a mix of Gulp, Webpack to Bazel, and I'm getting an error that `events` is not defined, and I'm not sure how to fix it.

Below a simplified version, and [here](https://github.com/kalbasit/bazel-reproducible/tree/master/events-not-defined) a fully reproducible workspace.

lib.ts

```TypeScript
import { EventEmitter } from "events";

const focusBlurEventEmitter = new EventEmitter();
focusBlurEventEmitter.emit('focus');
```

BUILD.bazel
```python
load("@build_bazel_rules_typescript//:defs.bzl", "ts_library")
load("@build_bazel_rules_nodejs//:defs.bzl", "rollup_bundle")

ts_library(
    name = "lib",
    srcs = ["lib.ts"],
    node_modules = "@root_node_modules//:node_modules",
    tsconfig = "//:tsconfig.json",
    visibility = ["//visibility:private"],
)

rollup_bundle(
    name = "bundle",
    entry_point = "lib",
    node_modules = "@root_node_modules//:node_modules",
    deps = [":lib"],
)
```

bundle.js (the output)

```Javascript
(function (events) {
    'use strict';
    var focusBlurEventEmitter = new events.EventEmitter();
    focusBlurEventEmitter.emit('focus');
}(events));
//# sourceMappingURL=bundle.es6.js.map
```

The above obviously does not work in a browser due to events missing. I tried setting the globals in the `rollup_bundle` to `{"events":"events"}` and `{"events":"events.EventEmitter"}` with no luck.

Isn't Rollup supposed to `browserify` the build?

Globegitter:
Even though the default rollup setup is pretty good and should work for most use cases, I think there are also quite a few use-cases that would want to utilize further rollup plugins (us included) and having to write a custom rule for each of these separate use-cases seems a bit cumbersome. So I came up with a way of providing a custom config without having to write your own rule.

Here one would have to provide their own node_modules, which is a superset of the existing node_modules (have not found a satisfying way to have the nodejs_binary depend on two sets of node_modules). Then one would have to copy the existing `rollup.config.js` and ensure to keep the `TMPL_` variables in place. It is not quite ideal but imo stil makes it easier than creating a custom rule.

This would allow it to quite easily fix #216 and commonjs support mentioned in #137

Also internally we are experimenting with a few different plugins and this makes it easier for us than having to write/modify a custom rule.


Globegitter:
@kalbasit nice, happy to hear that

kalbasit:
@alexeagle This PR makes it easier to use Rollup, is it something you would consider, or is there some other way being worked on or considered?

kalbasit:
@Globegitter can you please rebase your PR over master? It looks like one file is conflicting.

alexeagle:
New rollup_bundle rule lets you pass rollup.config.js. Should launch with next release

PR #235, opened 2018-06-18
Use rollup-plugin-typescript2 for downleveling es2015+ rather than separate tsc.

### LINKED ISSUE #175: `rollup_bundle` loses sourcemap context due to running tsc for downleveling
Here is my understand of how `rollup_bundle` works:
1. copy files into `bundles.es6` directory
2. Rollup the files in `bundles.es6` directory into `bundle.es6.js` file
3. Downlevel the file from ES6 -> ES5 into `bundle.js` using `tsc`
4. Run uglify to generate `bundle.min.js`

The pipeline above works fine for the source code, but it brakes source map. There are 2 issues:
1. Many of the actions do not declare `*.js.map` as output of the action. This can be easily fixed as shown [here](https://github.com/mhevery/rules_nodejs/commit/bb7009ade9172abb67f03d743fc747d36a764ed8)
2. The source map pipeline is broken. The reason for this is that when `tsc` is used to downlevel the code from ES2015->ES5 `tsc` assumes that `bundle.es6.js` is the source file. This is incorrect assumption. Instead `tsc` should transform the source maps in `bundel.es6.js` to the next step. I am not even sure if `tsc` supports such mode. In any case the issue is that at this point the source map have been destroyed and so the subsequent source maps are of no value.

Globegitter:
Rollup has at least two plugins to downlevel es2015+: https://github.com/rollup/rollup-plugin-babel and https://github.com/rollup/rollup-plugin-buble, imo it makes sense to use one of these tools for downleveling rather than tsc. This should make builds a bit faster (due to not having to run two binaries), the rule a bit simpler (it took me a while to realize the role of tsc in this context) and should also fix #175

I decided to go with [buble](https://buble.surge.sh/guide/) rather than babel, as it is simpler if all we want is es2015+ downleveling. There are certain things it will not downlevel, such as Iterators and generators as they don't downlevel well. But if #234 also gets merged it would me simple for anyone to substitute buble with the babel plugin if needed.


rogerhub:
Does this change also drop typescript support? Why are we using a package that just sounds like a 
misspelling of "babel"?

Globegitter:
As far as I am aware at this point any typescript has already been converted to js and the compiler was Purley used for downleveling. Well buble does state it was inspired by Babel and is just meant to be a no config alternative to Babel if all you need is downleveling. There are bigger projects that use it, e.g. Vue.js is using it rollup itself is making use of it as well. But yeah also happy to swap it out with Babel if people feel more comfortable with it.

Globegitter:
Just found https://github.com/ezolenko/rollup-plugin-typescript2 which could also be an option to still use typescript to downlevel but hopefully not break sourcemaps.

Toxicable:
>certain things it will not downlevel, such as Iterators and generators as they don't downlevel well

As far as I understand this will be a breaking change then.
More than that, it'll just straight up break applications.

Typescript has Iterator and generator downleveling, which remain when the code is initially compiled from TS -> es2015 (ng_module or ts_libraray).
This change must be able to downlevel those into es5, for reference TS does this by generating helpers.

Globegitter:
Good to know, then I will look into babel/typescript as the alternatives - this change would then also make it easier for other people to use any other rollup plugin for downleveling if they want to.

Globegitter:
This is using https://github.com/ezolenko/rollup-plugin-typescript2, i.e. typescript now for downleveling but it being done through a rollup-plugin should still preserver sourcemaps.

internal/rollup/rollup_bundle.bzl
```diff
@@ -313,7 +294,6 @@ ROLLUP_ATTRS = {
 }
 
 ROLLUP_OUTPUTS = {
-    "build_es6": "%{name}.es6.js",
```
alexeagle: I have no idea how to tell if anyone relies on this output...

internal/e2e/rollup/bundle-umd_golden.js_
```diff
@@ -9,22 +9,24 @@
...
+  var name = 'Alice';
 
+  console.log(hello + ", " + name + " in " + fum);
   // Test for sequences = false
-  class A {
-    a() {
-      return document.a;
-    }
-  }
+  var A = /** @class */ (function () {
```
alexeagle: I need to study this a bit to understand why the syntax is changing here. It seems like the previous output was just wrong, and should never have had a class keyword?

Globegitter: Yes that is a good point, I'll see if I can have a look into that myself the next few days.

alexeagle:
Maybe I should pick this one up? I think we need a test for sourcemaps since this should make such a test go green.

alexeagle:
I think this has gone stale.

I'm about to start working on a `babel_library` rule which I hope will serve our purposes of downleveling for older non-evergreen browsers (the ones without esmodule support following Angular CLI "differential loading" logic)

Also I think we want to avoid running rollup twice with different output plugins enabled - we should take rollup output, send it to evergreen browsers, then take that output and downlevel it for older browsers.

PR #239, opened 2018-06-22
Prevent accident manifest resolve from require('manifest')

gregmagolan:


internal/node/node_loader.js
```diff
@@ -231,8 +231,15 @@ module.constructor._resolveFilename =
     resolveLocations.push(resolveRunfiles(
         'TEMPLATED_label_workspace_name', 'TEMPLATED_label_package', 'node_modules', request));
   }
+  var manifestLocation = resolveRunfiles('manifest').toLowerCase();
```
alexeagle: Nit: I'd normalize the case in the place where you make the comparison

gregmagolan: 👍

PR #241, opened 2018-06-26
Fix bazel run @yarn//:yarn

gregmagolan:
Fixes `@yarn//:yarn` and adds some console output when running `@nodejs//:yarn`, `@nodejs//:npm` & `@yarn//:yarn`


gregmagolan:
Oh. Nevermind. It is @yarn//:yarn that is broken. Will fix that :)

Issue #155, opened 2018-03-14
'.bin' directory not available?


mmorearty:
When you `npm install` a module, if that module's `package.json` includes a `"bin":` attribute, then a `node_modules/.bin/<filename>` file will be created. But although rules_nodejs does seem to create a `.bin/` directory, that directory is not available in all of the places where I think it should be.

To reproduce:

* Download and unzip [example.zip](https://github.com/bazelbuild/rules_nodejs/files/1809665/example.zip)
* `bazel build :main`; then look at `bazel-bin/out.txt`, which contains a list of all of its input files. This include everything from `node_modules/typescript/` (since the "typescript" package is specified in package.json); but it does _not_ include `node_modules/.bin/tsc`. This is surprising, because if you just ran `npm install typescript` without involving Bazel, you would get a `node_modules/.bin/tsc`.
* `bazel build :tsc1`; then `bazel-bin/tsc1`. It works. The `tsc1` target runs `node_modules/typescript/bin/tsc`.
* But the equivalent with `.bin` does not work. `bazel build :tsc2`; then `bazel-bin/tsc2`. You get an error. `tsc2` depends on `node_modules/.bin/tsc`.

alexeagle:
Could you try `bazel query 'deps(@name_of_the_installed_npm_install_repo//:node_modules)'`
to see if the file is being filtered by Bazel's globbing?

mmorearty:
The output of `bazel query 'deps(@foo//:node_modules)` does not include anything from `.bin`. But it DOES include some other filenames that starts with `.`. Here is the output for the above example.zip — as you can see, at the bottom there are three files whose names start with `.`:

```
@foo//:node_modules
@foo//:node_modules/typescript/yarn.lock
@foo//:node_modules/typescript/package.json
@foo//:node_modules/typescript/lib/zh-TW/diagnosticMessages.generated.json
@foo//:node_modules/typescript/lib/zh-CN/diagnosticMessages.generated.json
@foo//:node_modules/typescript/lib/watchGuard.js
@foo//:node_modules/typescript/lib/typingsInstaller.js
@foo//:node_modules/typescript/lib/typescriptServices.js
@foo//:node_modules/typescript/lib/typescriptServices.d.ts
@foo//:node_modules/typescript/lib/typescript.js
@foo//:node_modules/typescript/lib/typescript.d.ts
@foo//:node_modules/typescript/lib/tsserverlibrary.js
[...]

PR #242, opened 2018-06-28
Add node_modules/.bin/* to yarn_install and npm_install

lukaszrek:
According to this issue: https://github.com/bazelbuild/rules_nodejs/issues/155 .bin directory is not available in node_modules target. 

This behaviour was introduced with https://github.com/bazelbuild/rules_nodejs/pull/130


alexeagle:
Thanks!

ryanrasti:
Is there a regression here? I'm running into the following issue:

```
yarn_install(
    name = "npm",
    exports_directories_only=False,
    package_json = "//frontend:package.json",
    yarn_lock = "//frontend:yarn.lock",
)
```

Specifically, I have a `gen_rule` that depends on `@npm//:node_modules`. But when I run it, the `.bin` directory is not exported, even though I can see the `.bin` directory after running `bazel build @npm//:node_modules` directly.

PR #246, opened 2018-07-19
Forward termination signals to node process.

Globegitter:
When using nodejs_image of rules_docker there is a problem that `SIGTERM` and `SIGINT` do not get forward to the node process by bash. So when interactively running the container it is not possible to Ctrl+C out and further it would not be possible to implement any custom `SIGTERM` behaviour if needed.

As far as my bash knowledge goes this should be backwards-compatible.

### DIFF
--- a/internal/node/node_launcher.sh
+++ b/internal/node/node_launcher.sh
@@ -135,11 +135,15 @@ _term() {
   kill -TERM "$child" 2>/dev/null
 }
 
+_int() {
+  kill -INT "$child" 2>/dev/null
+}
+
 set +e
 "${node}" "${NODE_OPTIONS[@]}" "${script}" "${ARGS[@]}" &
 child=$!
 trap _term SIGTERM
-trap _term SIGINT
+trap _int SIGINT
 wait "$child"
 RESULT="$?"
 set -e
--- a/internal/node/node_launcher.sh
+++ b/internal/node/node_launcher.sh
@@ -129,8 +129,19 @@ for ARG in "${ALL_ARGS[@]}"; do
   esac
 done
 
+# Note: Bash does not forward any termination signals to any child process when running in docker
+# so need to manually trap and forward the signals
+_term() {
+  kill -TERM "$child" 2>/dev/null
+}
+
+trap _term SIGTERM
+trap _term SIGINT
+
 set +e
-"${node}" "${NODE_OPTIONS[@]}" "${script}" "${ARGS[@]}"
+"${node}" "${NODE_OPTIONS[@]}" "${script}" "${ARGS[@]}" &
+child=$!
+wait "$child"
 RESULT="$?"
 set -e
 
--- a/internal/node/node_launcher.sh
+++ b/internal/node/node_launcher.sh
@@ -129,18 +129,17 @@ for ARG in "${ALL_ARGS[@]}"; do
   esac
 done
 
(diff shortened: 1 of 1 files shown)


internal/node/node_launcher.sh
```diff
@@ -129,8 +129,19 @@ for ARG in "${ALL_ARGS[@]}"; do
...
 done
 
+# Note: Bash does not forward any termination signals to any child process when running in docker
+# so need to manually trap and forward the signals
+_term() {
+  kill -TERM "$child" 2>/dev/null
+}
+
+trap _term SIGTERM
+trap _term SIGINT
```
rogerhub: Why not do this after bringing up the subprocess? (Or declare the `child` variable first.)

internal/node/node_launcher.sh
```diff
@@ -129,8 +129,19 @@ for ARG in "${ALL_ARGS[@]}"; do
   esac
 done
 
+# Note: Bash does not forward any termination signals to any child process when running in docker
+# so need to manually trap and forward the signals
+_term() {
+  kill -TERM "$child" 2>/dev/null
```
rogerhub: Does this mean that if docker *isn't* being used, the process ends up receiving two signals? (once as part of the foreground process group, and once from this code)

Another thought: If the subprocess spawns even more subprocesses, does this work correctly?

internal/node/node_launcher.sh
```diff
@@ -129,8 +129,19 @@ for ARG in "${ALL_ARGS[@]}"; do
   esac
 done
 
+# Note: Bash does not forward any termination signals to any child process when running in docker
```
rogerhub: nit: can you wrap this at the same length as the other comments? (looks like 80 columns)

internal/node/node_launcher.sh
Globegitter: Just tested it:

```
process.on('SIGINT', function (code) {
  console.log('SIGINT received...');
  setTimeout(() => process.exit(0), 5000);
});
```

and the process just prints the log once.

I also tested spawning up a child process and the behaviour is the same before and after the change, the child process does not get the termination signal forwarded.

internal/node/node_launcher.sh
```diff
@@ -129,8 +129,18 @@ for ARG in "${ALL_ARGS[@]}"; do
...
+_term() {
+  kill -TERM "$child" 2>/dev/null
+}
+
 set +e
-"${node}" "${NODE_OPTIONS[@]}" "${script}" "${ARGS[@]}"
+"${node}" "${NODE_OPTIONS[@]}" "${script}" "${ARGS[@]}" &
+child=$!
+trap _term SIGTERM
+trap _term SIGINT
```
alexeagle: For correctness, could you preserve the same signal? this sends a sigterm to node when the user sent a sigint, and it's possible those have different behavior under some circumstance

Globegitter:
In docker the signal is sent to the process with PID 1 and shell/bash does not automatically forward termination signals. There are various ways around that, usually if you just execute one command in a bash script at the end you can do `exec command`, then that would become PID 1 and receive signals fine, but somehow that did not work. Otherwise one could use something like https://github.com/krallin/tini in the docker, but that seems a bit heavy-handed, so this solution seemed like the most simple one to me.

PR #247, opened 2018-07-21
Do not hardcode bash to /bin/bash, use /usr/bin/env instead

kalbasit:
I'm currently using NixOS as my daily driver, and NixOS does not ship with `/bin/bash` intentionally. This PR changes the shebang to `/usr/bin/env bash`.


alexeagle:
LGTM, but it's broken on Windows:
```
ERROR: C:/build/buildkite-worker-windows-java8-zdk8-1/bazel/rules-nodejs-nodejs/internal/e2e/rollup/BUILD.bazel:15:1: Couldn't build file internal/e2e/rollup/bundle.umd.js: SkylarkAction internal/e2e/rollup/bundle.umd.js failed (Exit 127)
--
  | /usr/bin/env: 'bash': No such file or directory
```

Any ideas how to make it work on both? If you don't have access to a Windows machine I can pair with you on it.

kalbasit:
@alexeagle I don't have a Windows machine at the moment. But does window even provide `/bin/bash`? is it through Cygwin? I'm thinking I could create a variable, say `shebang` set to `/bin/bash` if `is_windows` is True?

kalbasit:
@alexeagle can you kick off the windows build now? I think I fixed the issue in my last two commits. (I'll squash once you give the LGTM).

kalbasit:
most likely :(. How do you suggest I fix this?

alexeagle:
On Windows, Bazel gets bash from http://www.msys2.org/

If I drop into an msys2 terminal, I'm able to run `/bin/bash` and `/usr/bin/env bash` so I'm not sure why these don't work under the bazel environment. I think there is some other layer involved.

@dslomov can you redirect to one of the windows experts pls?

dslomov:
@laszlocsomor @meteorcloudy any thoughts?

laszlocsomor:
> More weirdly, in MSYS you can then explicitly `unset PATH`. Now it's really not set, you can no longer run `echo`, just `/usr/bin/echo`, and `/usr/bin/echo $PATH` prints an empty string.

Correction: `echo` is a Bash built-in, so that will *always* work. But `env` is not a built-in, so it's a good way to check if you have PATH defined.

kalbasit:
In that case, I'll probably change the node_launcher.sh to be a script written by `repository_ctx.file()` and that file sources the actual `node_launcher`. That was shebang is controlled by runtime. Does this work for you @alexeagle ?

alexeagle:
I'm not sure what you mean by 
> sources the actual `node_launcher`

Ideally the fix here is that we shouldn't have a bash dependency to run node, it makes the Windows install story bad. We need a node_launcher.cmd for windows.

@gregmagolan any thoughts?

evahowe:
@kalbasit @gregmagolan @alexeagle - What is the state of this?  Is it done?

kalbasit:
> @kalbasit @gregmagolan @alexeagle - What is the state of this?  Is it done?

I haven't gotten a chance to finish it myself, perhaps someone with Windows machine/experience can take over?

alexeagle:
I think we should do this by cutting the bash dependency altogether, but it's a hard project and not many contributors here on Windows

alexeagle:
Do you need help to get this rebased and green?

kalbasit:
> Do you need help to get this rebased and green?

Yea I wouldn't be able to make this work for Windows as I don't own/don't develop on it.

PR #249, opened 2018-07-23
Use relative paths in generated node/yarn scripts

huangw5:
Currently, we don't include the node/yarn in runfiles of the generated scripts. As a result, we cannot run builds/tests remotely. This change will searches the actual node/yarn binaries via relative paths and also includes the node binaries in runfiles.

Verified rules_nodejs still works locally on both Linux and Windows. It can now build remotely as well.

BREAKING CHANGE:
* This PR stops supporting vendored node as it is not compatible with remote build execution #250
* This PR uses bazelbuild/bazel_skylib and the downstream has to add that repo to the WORKSPACE.


huangw5:
It is now ready to review. I changed it to always use relative paths.

huangw5:
Note that this PR uses bazelbuild/bazel_skylib libraries. The downstream should add that repo to their WORKSPACE.

internal/node/node_repositories.bzl
```diff
@@ -176,44 +178,57 @@ exports_files([
...
+process.env['PATH'] = `${process.argv[2]}:${process.env['PATH']}`;
+const proc = spawn(process.argv0, process.argv.slice(3), {stdio: [process.stdin, process.stdout, process.stderr]});
+proc.on('close', (code) => { process.exit(code); });
+""")
 
     # Sets PATH and runs bin/node.js passing all arguments
-    repository_ctx.file("bin/node", content="""
-export PATH="{root}":$PATH
-"{node}" "{script}" "$@"
+    repository_ctx.file("bin/node", content="""#!/bin/bash
```
kalbasit: can you please change the shell to `/usr/bin/env bash`? Please see #247 for context.

kalbasit:
is this PR supposed to detect node on the host, if present? I can't seem to get it to do that so it's not really fixing #254

internal/node/node_repositories.bzl
```diff
@@ -176,44 +178,57 @@ exports_files([
...
 alias(name = "yarn", actual = "{yarn}")
+filegroup(
+  name = "node_runfiles",
+  srcs = glob(
+    [
+      "bin/node.js",
+      "{node_dir}/**",
+      "{yarn_dir}/**",
+    ],
+    exclude = [
```
alexeagle: I assume you checked if there was anything more we could exclude, to try to bring down the count/size of the action input files

internal/node/node_repositories.bzl
alexeagle: That PR is hung up on Windows compatibility issues, I think we should rebase later.
Also that one has no tests so we won't know if we broke it :(

huangw5:
I found this PR failed on buildkite. But the "Details" page returned an error to me: https://buildkite.com/bazel/rules-nodejs-nodejs/builds/435: Page not found

@alexeagle Do you know how can I figure out what the failure is? Thanks!

gregmagolan:
@huangw5 Here's the buildkite error. Not sure why you can't see it but you probably need to be logged into buildkite.
```
bazel run --curses=yes --color=yes --verbose_failures @nodejs//:yarn
--
  | INFO: Analysed target @nodejs//:yarn (1 packages loaded).
  | INFO: Found 1 target...
  | Target @nodejs//:bin/yarn_node_repositories up-to-date (nothing to build)
  | INFO: Elapsed time: 3.191s, Critical Path: 0.02s
  | INFO: 0 processes.
  | INFO: Build completed successfully, 1 total action
  | INFO: Build completed successfully, 1 total action
  | /private/var/tmp/_bazel_buildkite/38b428035f149a8c4d773a54f778415f/external/nodejs/bin/yarn_node_repositories: line 5: realpath: command not found
  | Traceback (most recent call last):
  | File "bazelci.py", line 1254, in <module>
  | sys.exit(main())
  | File "bazelci.py", line 1241, in main
  | args.build_only, args.test_only, args.monitor_flaky_tests)
  | File "bazelci.py", line 377, in execute_commands
  | execute_bazel_run(bazel_binary, config.get("run_targets", None))
  | File "bazelci.py", line 667, in execute_bazel_run
  | "--color=yes", "--verbose_failures", target])
  | File "bazelci.py", line 844, in execute_command
  | return subprocess.run(args, shell=shell, check=fail_if_nonzero, env=os.environ).returncode
  | File "/usr/local/Cellar/python/3.6.5_1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/subprocess.py", line 418, in run
  | output=stdout, stderr=stderr)
  | subprocess.CalledProcessError: Command '['bazel', 'run', '--curses=yes', '--color=yes', '--verbose_failures', '@nodejs//:yarn']' returned non-zero exit status 127.
  | 🚨 Error: The command exited with status 1
```
`realpath` command not found on osx build. the other builds were successful. looks like realpath is not on osx by default.

gregmagolan:
PR looks great. Just the OSX failure and a BREAKING CHANGE comment on the commit message for the new transitive dep? @alexeagle How are we documenting breaking changes?

huangw5:
@gregmagolan Thanks a lot! I removed realpath as it is not really needed (We always invoke the scripts via Bazel labels which would be resolved to absolute paths). I also updated the first comment with breaking change.

kalbasit:
@huangw5 @alexeagle with this PR, is `nodejs` always expected to be downloaded by the rules? How can I override where the `node` binary is in this case?

huangw5:
@kalbasit Yes, nodejs and yarn are always downloaded by the rules. As stated in #250, we plan to change `node_path` and `yarn_path` to labels. But as for now, I don't think you can override the paths with PR. Can you let us know what are your use cases?

kalbasit:
@huangw5 see #254 for context of my use case, I'm thinking of making a PR that adds `use_host_node` a boolean defaulting to False, when true, will override the `node_exec` to point to it? We can continue the discussion over on #254, if it's out of topic here.

alexeagle:
I moved the BREAKING CHANGE note to the commit message, as that's what users see when they fetch the repo (the PR comments are basically lost after the PR is merged and closed)

Merged as ce8148b

PR #256, opened 2018-07-27
Simplify node_launcher.sh

gregmagolan:


rogerhub:
Can you add some more context to this? I don't really understand what this fixes.

alexeagle:
Moving all node processes into the background changed the stdout in such a way that the streaming protocol between `tsc_wrapped` and Bazel stopped working:

```
ERROR: /Users/greg/google/gregmagolan/rules_typescript/examples/generated_ts/BUILD.bazel:11:1: Compiling TypeScript (devmode) //examples/generated_ts:generated_ts failed: Worker process quit or closed its stdin stream when we tried to send a WorkRequest:

---8<---8<--- Exception details ---8<---8<---
java.io.IOException: Broken pipe
    at java.io.FileOutputStream.writeBytes(Native Method)
    at java.io.FileOutputStream.write(FileOutputStream.java:326)
...com.google.devtools.build.lib.concurrent.AbstractQueueVisitor$WrappedRunnable.run(AbstractQueueVisitor.java:355)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
    at java.lang.Thread.run(Thread.java:748)
---8<---8<--- End of exception details ---8<---8<---
```

rogerhub:
Reverting #246 sgtm. I thought of exec while reviewing #246, but was hesitant to suggest it because it wouldn't have worked with the `EXPECTED_EXIT_CODE` stuff. It's a bit unintuitive that we're now coupling whether we use exec with the value of `EXPECTED_EXIT_CODE`, but I suppose there's not much else we can do when the caller expects us to transform the exit status. In the common case (expect a zero), I think exec is the right thing to do.

internal/node/node_launcher.sh
```diff
@@ -129,38 +129,35 @@ for ARG in "${ALL_ARGS[@]}"; do
...
-      exit ${BAZEL_EXIT_TESTS_FAILED}
-    fi
-  else
-    exit 0
+if (( ${RESULT} != ${EXPECTED_EXIT_CODE} )); then
+  echo "Expected exit code to be ${EXPECTED_EXIT_CODE}, but got ${RESULT}" >&2
+  if [ "${RESULT}" -eq "0" ]; then
+    # This exit code is handled specially by Bazel:
+    # https://github.com/bazelbuild/bazel/blob/486206012a664ecb20bdb196a681efc9a9825049/src/main/java/com/google/devtools/build/lib/util/ExitCode.java#L44
+    readonly BAZEL_EXIT_TESTS_FAILED = 3;
```
rogerhub: preexisting issue, but this semicolon is unnecessary, right?

Globegitter:
Yeah I agree, exec is the nicer solution, somehow when I tested that though it did not work. Let me test again with this branch though. The other option would be to split this script up into one for the binary rule and one for the test rule but don't think there is much to gain tbh.

gregmagolan:
I slept on it and realized this morning that the issue is caused because putting the node process in the background means that it no longer receives stdin. tsc_wrapped in rules_typescript is fed work via stdin so this breaks it.

gregmagolan:
Here's a simpler fix for this (#257) that attaches stdin to the node background process. Testing with rules_typescript and it solves the problem.

alexeagle:
Okay, merged that simpler fix #257 and tagged `0.11.1` so we can continue the release.

I'd still like to merge this one as well but I'll wait for @Globegitter to have a chance to confirm it doesn't break the docker case.

As for the interaction between these two features, I only anticipate using the expected exit code in our own testing, so I don't think it will intersect with this docker edge case.

alexeagle:
oh and @Globegitter your breakage is probably around here:
https://github.com/bazelbuild/rules_nodejs/commit/ce8148b7de3a3d7621e47d26dc30e21fd0849a14#diff-67b0f7c2b38c3bbb6dd40c85eb923174R108

you could maybe re-use our helper to locate binaries:
https://github.com/bazelbuild/rules_nodejs/blob/master/internal/node/node_labels.bzl
but that points to yet another wrapper around node, maybe you need the underlying binary

Globegitter:
Yep, just got around it and submitted a fix: https://github.com/bazelbuild/rules_docker/pull/470. Yeah for some reason this does not fix the termination signal forwarding in docker. I do not fully understand why because I have done that in the past and it should work fine. But also do not have time right now to investigate further. Can probably do later next week or happy to create a reproduction repo so someone else can investigate further.

alexeagle:
Hmm we need to use rules_docker ourselves in something downstream so we have a chance to catch these issues too...

Globegitter:
@wlsf82 I have unfortunately not had the time yet to investigate why this was not working for me with a rules_docker created nodejs_image. If you have some time to investigate the issue then it would unblock this.

Globegitter:
@gregmagolan I would say this can be closed now as my PR included all changes here :)

gregmagolan:
Replaced by #494

PR #262, opened 2018-07-31
Add check_rules_nodejs_version for downstream rules that depend on rules_nodejs

gregmagolan:


package.bzl
```diff
@@ -0,0 +1,45 @@
...
+against a minimum dependent build_bazel_rules_nodejs version.
+"""
+
+load("//internal/common:parse_version.bzl", "parse_version")
+
+VERSION = "0.11.2"
+
+def check_rules_nodejs_version(minimum_version_string):
+  """
+  Verify that a miminum build_bazel_rules_nodejs is loaded a WORKSPACE.
```
alexeagle: spelling: minimum

examples/packages/WORKSPACE
```diff
@@ -26,3 +26,8 @@ node_repositories(package_json = ["//:package.json"])
 load(":setup_workspace.bzl", "packages_example_setup_workspace")
 
 packages_example_setup_workspace()
+
+load("@build_bazel_rules_nodejs//:package.bzl", "check_rules_nodejs_version")
```
alexeagle: users should load it through `//:defs.bzl` so they don't need two load statements in their WORKSPACE

package.bzl
```diff
@@ -0,0 +1,45 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
alexeagle: nit: I think we should use the current year in new files?

package.bzl
```diff
@@ -0,0 +1,45 @@
...
+  load("@build_bazel_rules_nodejs//:package.bzl", "check_rules_nodejs_version")
+  check_rules_nodejs_version("0.11.2")
+  ```
+
+  Args:
+    minimum_version_string: a string indicating the minimum version
+  """
+  current_version = parse_version(VERSION)
+  minimum_version = parse_version(minimum_version_string)
+  if minimum_version > current_version:
```
alexeagle: if you move this assertion into the new library file, you could just modify the existing test to load the library rather than the check_bazel_version.
So far we don't have a use case for parsing the version but not asserting on it.
Does that make sense?

package.bzl
gregmagolan: thanks 🙄  I need a spell check in my vscode.

package.bzl
gregmagolan: yup.. gotcha

package.bzl
gregmagolan: 👍

examples/packages/WORKSPACE
gregmagolan: 👍

package.bzl
```diff
@@ -0,0 +1,43 @@
...
+
+"""Package file which defines build_bazel_rules_nodejs version in skylark
+
+check_rules_nodejs_version can be used in downstream WORKSPACES to check
+against a minimum dependent build_bazel_rules_nodejs version.
+"""
+
+load("//internal/common:check_version.bzl", "check_version")
+
+VERSION = "0.11.2"
```
alexeagle: 1) comment here saying it needs to be synced with package.json
2) edit the end of the README.md so I know how to do the next release
3) where is todo/issue reminding us to fix the duplication?

PR #273, opened 2018-08-08
[WIP] Add data test to node e2e 

filipesilva:


internal/e2e/node/BUILD.bazel
```diff
@@ -16,3 +17,18 @@ jasmine_node_test(
...
+    data = ["data/data.json"],
+    node_modules = "//internal/test:node_modules",
+    deps = ["//internal/e2e/node/lib1"],
+)
+
+genrule(
+    name = "lib",
+    srcs = ["module_resolution.spec.js"],
+    outs = ["module_resolution_copy.spec.js"],
+    cmd = "cat $< > $@",
```
alexeagle: nit: could be `cp $< $@`

internal/e2e/node/data/missing-data.json
```diff
@@ -0,0 +1,3 @@
+{
+  "value": 42
```
alexeagle: put a distinct value in here so there's never a question that a test passed by accident by reading the value `42` from the wrong file

internal/e2e/node/module_resolution.spec.js
```diff
@@ -7,6 +7,8 @@ const node_resolve_main_2 = require('node_resolve_main_2');
 const node_resolve_nested_main = require('node_resolve_nested_main');
 const lib1 = require('lib1');
 const lib1some = require('lib1/src/some');
+const data_resolve_relative = require.resolve('./data/data.json')
```
alexeagle: hmm we don't have a linter here for semicolons I guess

internal/e2e/node/module_resolution.spec.js
```diff
@@ -37,3 +39,19 @@ describe('node npm resolution', () => {
...
+    expect(JSON.parse(dataContent)).toEqual({ "value": 42 });
+  });
+  it('should throw when resolving missing data files', () => {
+    if (process.platform.startsWith('win')) {
+      // On Windows, file location is the original one and not sandboxed.
+      // This means we cannot correctly exclude relative paths that aren't part of the runfiles.
+      return;
+    }
+    // This file exists in the source folder but is not in the data array.
+    expect(() => require.resolve('./data/missing-data.json')).toThrow();
```
alexeagle: this test fails with sandboxing disabled too right?
How about change the desc
`it('should throw when resolving files that are outside the sandbox')`?

internal/node/node_loader.js
```diff
@@ -83,20 +83,24 @@ function resolveToModuleRoot(path) {
...
+  return {runfilesManifest, reverseRunfilesManifest};
 }
-const runfilesManifest =
+const { runfilesManifest, reverseRunfilesManifest } =
     // On Windows, Bazel sets RUNFILES_MANIFEST_ONLY=1.
     // On every platform, Bazel also sets RUNFILES_MANIFEST_FILE, but on Linux
     // and macOS it's faster to use the symlinks in RUNFILES_DIR rather than resolve
-    // through the indirection of the manifest file
+    // through the indirection of the manifest file.
+    // We also need to construct a reverse map to make sure to always resolve files from
```
alexeagle: this doesn't seem to match the implementation below - you don't add a new place where we throw or reject input right? We don't need any guarantee that we only ever resolve declared inputs, that should be bazel's job (like with the docker sandboxing). We only need happy path cases, I believe

filipesilva: Yeah this comment was old, back when I was trying a different approach. Have updated it to the current implementation.

internal/e2e/node/module_resolution.spec.js
filipesilva: Yes, it should fail without sandboxing in general.

internal/e2e/node/module_resolution.spec.js
filipesilva: Fixed

internal/e2e/node/data/missing-data.json
filipesilva: Fixed

internal/e2e/node/BUILD.bazel
filipesilva: Fixed

PR #281, opened 2018-08-14
Resolve absolute runfiles on windows

filipesilva:


internal/node/node_loader.js
```diff
@@ -84,18 +88,48 @@ function resolveToModuleRoot(path) {
...
+  let workspaceRoot;
   for (const line of input.split('\n')) {
     if (!line) continue;
     const [runfilesPath, realPath] = line.split(' ');
     runfilesManifest[runfilesPath] = realPath;
     reverseRunfilesManifest[realPath] = runfilesPath;
+
+    // Determine workspace root to convert absolute paths into runfile paths.
+    // This only works if there is at least one runfile in the workspace root, but that is
+    // also the only case when we need to map back to the runfiles.
```
alexeagle: include a pointer to the bug you filed about how Bazel ought to do this

filipesilva: Done

PR #285, opened 2018-08-15
Remove usage of deprecated Bazel features

jjudd:
This makes rules_nodejs compatible with `--all_incompatible_changes`, best I can tell.

### DIFF
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -132,7 +132,6 @@ _NODEJS_EXECUTABLE_ATTRS = {
     "data": attr.label_list(
         doc = """Runtime dependencies which may be loaded during execution.""",
         allow_files = True,
-        cfg = "data",
         aspects=[sources_aspect, module_mappings_runtime_aspect]),
     "templated_args": attr.string_list(
         doc = """Arguments which are passed to every execution of the program.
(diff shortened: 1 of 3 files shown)


jjudd:
I was updating our internal version of Bazel to 0.16 and these errors showed up with `--all_incompatible_changes` enabled, so I fixed them.

alexeagle:
could you also add the `--all_incompatible_changes` flag to our repo(s) so we locally catch regressions?

pauldraper:
@jjudd the Circle CI builds are using Bazel 0.14. If you update angular/ngcontainer to 0.5.0, those are using Bazel 0.16.1 and it should work.

(BTW, holy frigging crap, angular/ngcontainer is over 2GB.)

alexeagle:
@pauldraper I would love to make that docker container smaller, I do a few things in there to trim it down (like delete downloads and intermediate output directories) but I struggled to even make a docker container that works. Would love some help there :)

pauldraper:
@alexeagle looks like almost all of that comes from circleci/node:10.9.0-browsers. And a lot that is browsers.

alexeagle:
while we are at it, this change ought to include whatever config change or linter so that it doesn't regress

gregmagolan:
@alexeagle @pauldraper The browsers in ngcontainer aren't needed in many cases now (angular-bazel-example for example) since ts_web_test_suite now uses browsers from rules_webtesting. I think angular/angular may still depend on the browsers being in the docker image for some old tests.

alexeagle:
This got stale, and Bazel team has been sending PRs to flip individual `--incompatible_*` flags so I think we'll just let that process unfold

PR #299, opened 2018-08-30
Add node 10.3.0 as an available version for node_repositories

cristiannancu:
Node 10.3.0 is needed to enable usage of `npm ci`, which only uses the `package-lock.json` to install dependencies and does not resolve differences.
Adding this version enables consistency through projects that use Bazel and move to node 10.3.0.

Using `npm ci` should result in much faster dependency installation as outlined below (up to 2x-3x speedup):
* https://docs.npmjs.com/cli/ci
* https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable
* https://medium.com/better-things-digital/npm-ci-a-command-you-should-know-d8d67847884c

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,7 @@ node_repositories(package_json = ["//:package.json"])
 You can choose a specific version of Node.js that's built into these rules.
 Currently these versions are:
 
+* 10.3.0
 * 9.11.1
 * 8.11.1
 * 8.9.1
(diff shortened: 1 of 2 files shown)


cristiannancu:
Hey @alexeagle, could you please take another look at this? 😄

alexeagle:
Thanks!!

PR #300, opened 2018-08-30
Make configuration env vars optional.

### LINKED ISSUE #290: Setting `--define` flags makes nodejs_binary not reproducible
We are using `--define` flags for some parts in the repository unrelated to node, but when these values change on `bazel build ...` commands then it rebuilds every node binary and everything that depends on it making this behaviour less then ideal for our use case because it reduces the cache hit rate and means we get different  resulting docker containers even when nothing has changed in the js stack

The culprit is this line: https://github.com/bazelbuild/rules_nodejs/blob/master/internal/node/node.bzl#L82 as `define` flags are part of the configuartion variables. So if I have `bazel build --define test=test` the resulting `nodejs_binary` file includes `export test="test"`.

What is the reason this has been put in? Could it just be removed, or at least made configurable, as removing it for us seems to have no effect. Happy to make a PR for whatever is best.

Globegitter:
Fixes #290 I don't like the fact that this is adding yet another attribute, but this is starting to bite us a bit in our setup so would like a fix for it.


Globegitter:
@alexeagle any news on that? I would really love to bring our ci timw down. Maybe it also makes sense to turn the default of the attribute around, so it is false per default and we just set it to true for the cases where it was needed. WDYT?

internal/node/node.bzl
```diff
@@ -131,6 +132,12 @@ _NODEJS_EXECUTABLE_ATTRS = {
         Enable this to get stack traces that point to original sources, e.g. if the program was written
         in TypeScript.""",
         default = True),
+    "configuration_env_vars": attr.bool(
+        doc = """Pass configuration environment variables to the resulting binary
+        Enable this to get configuration environment variables (taken from ctx.var), which also
+        includes anything specified via the --define flag, which can lead to different outputs
+        produced by this rule.""",
+        default = True),
```
alexeagle: hmm, I agree that this was a hasty addition, we shouldn't pollute the environment of every program with the entire list.
I think instead of a boolean, it should be a string_list.

PR #303, opened 2018-08-31
Fix default directory served for history-server when target is in root WORKSPACE folder

gregmagolan:


gregmagolan:
http-server can run without any arguments (defaults to cwd) but history-server requires the serving directory as an argument

PR #307, opened 2018-09-06
mocha_node_test rule

an-kumar:
This is a simple PR that adds a mocha_node_test rule that generally follows the example set by jasmine_node_test.

Mocha is a very popular node.js testing framework, so it seems like a reasonable addition.

I was not 100% sure where to put tests, so I included some in the mocha_node_test folder, but let me know if they should go elsewhere.


an-kumar:
I signed the CLA.

Issue #168, opened 2018-03-29
Add mocha_test rule


chrisleck:
Similar to jasmine_node_test, create a mocha wrapper for running simple TS / javascript / node unit tests.

Other similar nodejs rulesets for bazel already support this: https://github.com/pubref/rules_node#mocha_test

pscarey:
Any issues using something like:

```
def mocha_ts_test(name, data = [], **kwargs):
    """Runs mocha on TS files using ts-node."""
    data += [
        "@npm//ts-node",
        "@npm//mocha",
        "@npm//sinon",
        "@npm//chai",
        "@npm//@types/node",
        "//:tsconfig.json"
    ]

    args = [
        "--reporter list",
        "--require ts-node/register/transpile-only",
        "**/*.test.ts"
    ]

    nodejs_test(
        name = name,
        data = data,
        entry_point = "npm/node_modules/mocha/bin/mocha",
        templated_args = args,
        expected_exit_code = 0,
        **kwargs
    )
```

We've implemented that in a defs.bzl alongside the WORKSPACE file, with:

```
load("//:defs.bzl", "mocha_ts_test")

mocha_ts_test(
    name = "assert",
    data = glob([ "src/assert.*", "src/has.ts" ])
)
```

It just works by running any .test.ts file which is included in `data`. The typical usage would be 1 test file, and any required source files + other dependencies.

Let me know if this is the right way to go, and I'll look at tidying up the above & a pull request etc.

I don't know if you want to end up with a test rule for every test runner and/or bundling program. Surely anything provided through NPM with a CLI could fit into the above format?

Another alternative would be a short example of how to do something like the above to help people new to Bazel.

(As a side note r.e. other test runners - I had a lot of headaches trying to make Jest work, including postinstall patches to fix it's symlink handling behaviour as it would just hang with no errors due to the Bazel symlinks. Was kinda buggy and slow, not really sure why.).

pscarey:
Cool, I'll do some more work on it if there's nothing fundamentally wrong with the approach.

1. Easy - the 'npm' is just a reference to the workspace used by `yarn_install` and `npm_install`
2. OK
3. I'll look into configuration options and the best ways to offer them.

alexeagle:
After doing some cleanup work on #383 I found that `mocha_node_test` was indistinguishable from `jasmine_node_test` from a user point of view.

[...]

PR #308, opened 2018-09-06
mocha_node_test rule

kitofans:
(re-opening PR on my github account that has a signed CLA)

This is a simple PR that adds a mocha_node_test rule that generally follows the example set by jasmine_node_test.

Mocha is a very popular node.js testing framework, so it seems like a reasonable addition.

I was not 100% sure where to put tests, so I included some in the mocha_node_test folder, but let me know if they should go elsewhere. There are more tests to add but I figured I'd get some feedback on if this is something you guys want before adding them.


Toxicable:
tests should be under the examples directory.
Take a look around there, you should see similar ones for testing withe jasmine

internal/e2e/node/BUILD.bazel
```diff
@@ -1,4 +1,4 @@
-load("@build_bazel_rules_nodejs//:defs.bzl", "jasmine_node_test")
+load("@build_bazel_rules_nodejs//:defs.bzl", "jasmine_node_test", "mocha_node_test")
```
jupp0r: You need to also reexport this in //:defs.bzl

an-kumar: done

internal/mocha_node_test/mocha_node_test.bzl
```diff
@@ -0,0 +1,50 @@
...
+"""
+load("//internal/node:node.bzl", "nodejs_test")
+
+def mocha_node_test(
+  name,
+  test_entrypoints = [], # these should NOT contain the preceding workspace name.
+  data = [],
+  expected_exit_code = 0,
+  **kwargs):
+
```
gregmagolan: Needs a docstring here for this function. See the jasmine_node_test.bzl for an example. This will fix the lint failure: `./internal/mocha_node_test/mocha_node_test.bzl:30:1-37:2: function 'mocha_node_test' has no docstring`

internal/mocha_node_test/mocha_runner.js
```diff
@@ -0,0 +1,47 @@
+const Mocha = require('mocha');
+const fs = require('fs');
```
gregmagolan: not: `fs` and `path` not used

internal/mocha_node_test/BUILD
```diff
@@ -0,0 +1,47 @@
...
+    name = "exampleLib",
+    srcs = glob(["exampleLib/*.js"]),
+    module_name = "example-lib",
+    module_root = "exampleLib/index.js",
+)
+
+mocha_node_test(
+    name = "test",
+    data = [
+        ":testOne.js",
```
gregmagolan: nit: usings snake_case is more consistent with the rest of the repo

internal/mocha_node_test/BUILD
```diff
@@ -0,0 +1,47 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
gregmagolan: Better to nest tests under `/internal/mocha_node_test/test/...` like `/internal/npm_install/test`. Also, if the test can also act as an example of how to use the rule than it can go under `/examples`

internal/mocha_node_test/BUILD
```diff
@@ -0,0 +1,47 @@
...
+
+package(default_visibility = ["//visibility:public"])
+
+load(":mocha_node_test.bzl", "mocha_node_test")
+load("//internal/js_library:js_library.bzl", "js_library")
+
+exports_files([
+    "mocha_runner.js",
+    # Exported to be consumed for generating skydoc.
+    "mocha_node_test.bzl",
```
gregmagolan: Also needs to be added to the `/docs/BUILD` for doc generation

gregmagolan:
@kitofans Thanks for updating the PR. Looks great. Can you rebase onto master one more time? I pushed a fix to master that should fix the flaky CircleCI list job.

alexeagle:
Wait, why are we doing this?

See https://github.com/bazelbuild/rules_nodejs/issues/168#issuecomment-432313574

Globegitter:
For me the reason is that we have a project (not migrates to bazel yet) that uses mocha so presumably that would be more straightforward to migrate with some existing mocha test rules.

evahowe:
@gregmagolan @alexeagle @Globegitter Did this actually happen?

alexeagle:
We don't need a custom rule for this, you can just use the generated `mocha_test` rule that is created for the `mocha` package like the example: https://github.com/bazelbuild/rules_nodejs/blob/1.x/examples/webapp/BUILD.bazel#L25

Issue #250, opened 2018-07-24
vendored node feature is not compatible with remote execution, allows non-portable paths


alexeagle:
Our documentation says

> Finally, you could check Node.js and Yarn into your repository, and not fetch them from the internet. This is what we do internally at Google.
```
load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories")

# NOTE: this rule does NOT install your npm dependencies into your node_modules folder.
# You must still run the package manager to do this.
node_repositories(
  node_path = "path/to/node/base",
  yarn_path = "path/to/yarn/base",
  package_json = ["//:package.json"])
```

it's not clear what `path/to/node/base` means here, but @gregmagolan says it can point to something outside the workspace, like a local path on disk. That's bad because you're checking in code that might not work for another team member.
But also, we want to make everything work with remote execution, which means that all inputs including the node binary must be declared for actions to run. Pointing node somewhere in a script means bazel doesn't know to copy the binary to remote executor.

I think we should change the `node_path` and `yarn_path` attributes of `node_repositories` to be labels, which requires them to be in the bazel workspace. Then our scripts should always reference them by label, using an execroot-relative path so it's portable to a remote executor.

This will be a potentially breaking change. /FYI @chrisleck  since you added the feature. Do you think that requiring these to be labels would break your use case?

Also, we need to figure out how they can be platform-independent - the windows binary is a different location in the node distribution than for other platforms.

PR #309, opened 2018-09-06
Remove unimplemented node_path and yarn_path from node_repositories

gregmagolan:


gregmagolan:
@alexeagle That would be the better option. Issue for that is https://github.com/bazelbuild/rules_nodejs/issues/250.

PR #319, opened 2018-09-14
Don't depend on grep for runfiles.bash init in node_launcher.sh

gregmagolan:


gregmagolan:
Underlying issue was a regression in Bazel. Fixed in Bazel 0.17.1.

PR #322, opened 2018-09-15
Fix yarn_install getting called multiple times since 0.13.

Globegitter:
Since updating some of my projects to rules_nodejs 0.13.1 I have noticed that `yarn_install` gets called 2-3 times in a row on a single `bazel build :target` command (bazel 0.17.1). I don't know why that is the case, maybe it is a bug in bazel, but either way this diff fixes it for me.


Globegitter:
Hmm on my Mac I still see yarn getting executed twice. Really wonder what the issue here is.

PR #323, opened 2018-09-15
Allow yarn to be able to use workspace node_modules for improved performance

Globegitter:
One of the big workflow killers we have when using rules_nodejs is that the bazel managed node_modules are not cached persistently, it affects ci but even more so it affects our developer's workflow, that basically every time some package gets added or removed yarn has to install everything from scratch. And the fact that there is no progress message makes it even less ideal. At first I was working on a caching solution similar to https://github.com/bazelbuild/bazel/pull/5928, but then realised that we could just run yarn within the workspace and then symlink/hardlink/copy the node_modules from there into the external repository. That also has the nice added benefit that this works really well with the `bazel run @nodejs//:bin/yarn -- add package` workflow as that then already installs to the right node_modules, and you also have only one node_modules directory. Right now I am just symlinking the node_modules directory, mostly for performance reasons, but can also be nice for debugging reasons to be able to modify something in the workspace `node_modules` and it is reflected in the external repository.

In its current state this does not work with Windows, but not sure if that is a requirement. It should work on MacOS, but will verify myself as that is also an internal requirement.

Further this PR also includes the changes from #322 and removes the `prod_only` attribute again introduced in #248 as it now is not necessary anymore, because we have the fine-grained modules. Given that was just a small change I included it here, but also happy to factor it out if necessary.


Globegitter:
Tested on my Mac and works like a charm. Going to start rolling this out to our devs tomorrow.

Globegitter:
One issue we now ran into, is that the yarn_cache was living next to the node_modules in the workspace, degrading editor performance. After discussing internally with our devs, we came to the conclusion that we would like to set the `--mutex` flag rather than having individual yarn caches anyway. As with our current setup we would benefit more from keeping the shared cache but locking on that. So added an attribute for that.

internal/npm_install/npm_install.bzl
```diff
@@ -192,11 +212,21 @@ yarn_install = repository_rule(
             mandatory = True,
             single_file = True,
         ),
-        "prod_only": attr.bool(
-            default = False,
-            doc = "Don't install devDependencies",
-        ),
         "data": attr.label_list(),
+        "workspace_node_modules": attr.bool(
+          doc = """When set to true yarn will run in the directory where the
```
alexeagle: this is just "vendoring the dependencies" right?
I wouldn't expect bazel to be involved in running yarn in this case. Up to the user to vendor things into their source directory (I'd put them in `/third_party/npm/node_modules` in a monorepo)
so I'd expect the bazel setup to be only a `filegroup` that points there.

Globegitter: Not quite, the main idea is trying to address a suboptimal workflow with how the current `yarn_install` rules work: 

One is, there is no `repository_cache` support, so the dependencies do not get persisted outside of the bazel external repositories, there are cases where all dependencies will have to be reinstalled even if no dependencies have changed. 

The second issue is, that if dependencies have changed then it will currently discard the whole of `node_modules` as well as `yarn_cache`.

The third issue is when using yarn to add/remove a package via the recommended approach `bazel run @nodejs//:bin/yarn -- add package` this is already adding a `node_modules` to the workspace as this command is running within the workspace and not with the external repository. And then when a node_js binary is started that depends on node_modules managed via bazel then bazel has no idea about the `node_modules` that have been added in the workspace by the previous command, nor does it reuse the same cache, so it just goes and reinstalls all dependencies again from scratch.

This leads to a suboptimal experience in a few cases, it makes caching more tricky in ci, it consumes additional bandwith because there is a lot of unnecessarily fetching of packages and the main reason it leads to a really slow workflow when packages are added/removed as in the current setup all dependencies will get reinstalled twice, once when invoking yarn and once on `bazel run :target`

I am also happy to create an issue on that topic first, or even a small design document, as I do realise it is a bigger change, before going ahead with progressing the PR. I just needed to have some improvements now as this was one of the main critique points about the current setup.

Globegitter: Yes bazel would point yarn at a cached node_modules so to speak, within the repo, but it is not checked in, it is on gitignore and just lives on each developers machine.

alexeagle: maybe https://github.com/bazelbuild/rules_nodejs/commit/3946a16920f6d1383b99995420d2a358b3cd4e8b has already fixed this? instead of yarn using a cache in the project we can just use the standard global yarn cache on the machine

Globegitter: Yeah I was considering this as well and maybe it is the best trade-off (at least until yarn Plug'n'Play is released). The commit you linked does indeed improve the situation but it still does not allow for a nice workflow when one uses `bazel run @nodejs//:bin/yarn -- add package` as the downloaded package will still be copied into two places, but at least it just downloads it once.

Globegitter: Hmm, the symlink approach does seem to cause smaller other issues, getting some i/o error sometimes, where bazel says that the files are a symlink and a resolved file at the same time. Rerunning fixes this. I will have a think if I can come up with a better strategy that resolves any last workflow issues and not having to do any "hacky" symlinks.

gregmagolan:
@Globegitter Does the addition of `use_global_yarn_cache` which defaults to True and the upcoming change to have Bazel use the `node_modules` folder in the user's workspace directory solve all the issues that this PR was addressing?

I'm working finishing yarn_install/npm_install prototype to have Bazel use the `node_modules` folder in the user's workspace this week. If the all goes well then the new Bazel `refresh` feature which this depends on will make it into a upcoming Bazel release and we can support this workflow.

Globegitter:
It does indeed. That is great news.

PR #327, opened 2018-09-18
Filter out files that match test.js and spec.js but are in node_modules

gregmagolan:
Since fine grained deps don't exclude `**/test/**` you can get files such as `node_modules/jasmine/test/test.js` included as specs unintentionally.

This is the same logic that ts_web_test_suite uses: https://github.com/bazelbuild/rules_typescript/blob/master/internal/karma/karma.conf.js#L120


gregmagolan:
It might also be correct to not includes sources from npm fine grained deps in the `devmode_js_sources` aspect similar to what is needs to be done in the `sources_aspect`: https://github.com/bazelbuild/rules_nodejs/pull/329

PR #328, opened 2018-09-18
Support for ts_library npm fine grained deps

gregmagolan:
Generate @wksp//:package__typings filegroups which include only .d.ts fil…es for ts_library support


internal/npm_install/generate_build_file.js
```diff
@@ -340,6 +340,23 @@ filegroup(
     tags = ["NODE_MODULE_MARKER"],
 )
 
+filegroup(
+    name = "${pkg._dir}__dts",
```
alexeagle: we generally call this "typings" instead of "dts"

gregmagolan: much better thanks

Issue #314, opened 2018-09-08
@nodejs//:node crashing in custom rules with 0.11.0+


fwouts:
I've been using the `@nodejs//:node` target to define other Bazel rules, like the following:
```py
def _my_rule_impl(ctx):
  ctx.actions.run(
    executable = ctx.file._internal_nodejs,
    inputs = [
      ctx.file._script,
    ],
    outputs = [ctx.outputs.output],
    arguments = [
      # Run `node my-rule.js`.
      ctx.file._script.path,
      # Output path.
      ctx.outputs.output.path,
    ]
  )

my_rule = rule(
  implementation = _my_rule_impl,
  attrs = {
    "_internal_nodejs": attr.label(
      allow_files = True,
      single_file = True,
      default = Label("@nodejs//:node"),
    ),
    "_script": attr.label(
      allow_files = True,
      single_file = True,
      default = Label("//my_rule:my-rule-impl.js"),
    ),
  },
  outputs = {
    "output": "%{name}_output"
  }
)
```

This rule's implementation would effectively delegate to a NodeJS script like the following:
```js
const fs = require("fs");
const [, , outputPath] = process.argv;
fs.writeFileSync(outputPath, "Hello, World!", "utf8");
```

This worked extremely well, until rules_nodejs@0.11.0 which broke this behaviour. My rule now fails with the following error:
```
ERROR: /Users/work/dev/bazel-bug/example/BUILD.bazel:5:1: SkylarkAction example/example_output failed (Exit 127)
external/nodejs/bin/node: line 5: external/nodejs/bin/nodejs/bin/node: No such file or directory
Target //example:example failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 9.423s, Critical Path: 0.05s
INFO: 0 processes.
FAILED: Build did NOT complete successfully
```

Did something change with the `@nodejs//:node` target? It looks like the path `external/nodejs/bin/nodejs/bin/node` is incorrect, as it should probably be `external/nodejs/bin/node`?

To make it easier for you to debug, I've created the smallest possible example in the following repo: https://github.com/zenclabs/bazel-rules-nodejs-bug-repro

fwouts:
I wonder if https://github.com/bazelbuild/rules_nodejs/commit/1cda064b92e7ee295a8e02aae42c3316166076cd could be the culprit.

gregmagolan:
[...]

PR #342, opened 2018-09-27
Get the node scripts directory more reliably

huangw5:
This is to fix https://github.com/bazelbuild/rules_nodejs/issues/314


internal/node/node_repositories.bzl
```diff
@@ -481,9 +495,8 @@ def node_repositories(
       all required files need to be included in your node_modules filegroup. This option is desirable as it gives
       a stronger guarantee of hermiticity which is required for remote execution.
   """
-  # 0.14.0: @bazel_tools//tools/bash/runfiles is required for nodejs
-  # 0.17.1: allow @ in package names is required for fine grained deps
-  check_bazel_version("0.17.1")
+  # @bazel_tools//tools/bash/runfiles is required for nodejs
```
alexeagle: looks like a bad merge?

huangw5: Ah, right. Thanks! Fixed.

PR #343, opened 2018-09-28
Fix regression in fine grained deps when package does not have dependencies

gregmagolan:


gregmagolan:
Closing in favour of #345 which also adds a test case for the regression

PR #345, opened 2018-09-28
Build file generation should not throw if dependency has only optional dependencies

DevVersion:
Currently if a dependency has only optional dependencies, the `generate_build_file.js` script fails at analysis time because only for NPM, the `optionalDependencies` are generated into `dependencies` as well.

This is not the case for Yarn, and therefore the `dependencies` can be also undefined.

See sample package: https://github.com/jprichardson/node-klaw/blob/1.3.1/package.json

cc. @gregmagolan

### DIFF
--- a/internal/e2e/fine_grained_deps/package-lock.json
+++ b/internal/e2e/fine_grained_deps/package-lock.json
@@ -463,21 +463,25 @@
       "dependencies": {
         "abbrev": {
           "version": "1.1.1",
-          "bundled": true,
+          "resolved": false,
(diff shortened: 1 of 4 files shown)


DevVersion:
Oh damn 😄 you actually found the same a few hours ago.

DevVersion:
Actually we can keep this one because it adds a test case that covers this regression. This is an issue I ran into while converting a CLI project to Bazel.

gregmagolan:
Thanks @DevVersion!

DevVersion:
Oh @gregmagolan, I just meant that yours should be prioritized and this one should be merged afterwards _just_ for the test case. I should have looked for existing PRs before creating a fix, sorry 😞

alexeagle:
breakage on Mac looks real

DevVersion:
@alexeagle Should be fixed now. Let's see how the CI results.

Issue #347, opened 2018-10-01
Generated .pyc files affect cache keys


JaredNeil:
When `@nodejs` runs for the first time, some `.pyc` files are generated.
```
external/nodejs/bin/nodejs/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/__init__.pyc
external/nodejs/bin/nodejs/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/common.pyc
external/nodejs/bin/nodejs/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/generator/__init__.pyc
external/nodejs/bin/nodejs/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/generator/make.pyc
external/nodejs/bin/nodejs/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/input.pyc
external/nodejs/bin/nodejs/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/simple_copy.pyc
external/nodejs/bin/nodejs/lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/xcode_emulation.pyc
```
These generated files are then captured in the [`node_runfiles`](https://github.com/bazelbuild/rules_nodejs/blob/master/internal/node/node_repositories.bzl#L217) glob.
This causes a couple issues:

- The `.pyc` files generated during the first run of node change the cache keys for future runs (https://github.com/bazelbuild/rules_typescript/issues/251).
- Cache keys depend on the absolute path to the workspace because `.pyc` files contain absolute paths to the `.py` file they were compiled from (https://stackoverflow.com/questions/11191680/why-do-python-pyc-files-contain-the-absolute-path-of-their-source-code)

The easiest solution is to just exclude `.pyc` files from the `node_runfiles`, but this means they are generated, but never used. Another solution would be to make sure the `.pyc` files are generated with relative paths using the [`compileall -d` option](https://docs.python.org/3/library/compileall.html#cmdoption-compileall-d). A third solution would be to find a way to not generate them in the first place, but I'm not sure if that's possible.

PR #348, opened 2018-10-01
Exclude .pyc files from node_runfiles

JaredNeil:
Fixes https://github.com/bazelbuild/rules_nodejs/issues/347


alexeagle:
Thanks for tracking that down!!

PR #353, opened 2018-10-02
Update circleci setup

alexeagle:


gregmagolan:
LGMT. Just some lint & CI failures to fix.

alexeagle:
I reduced the `--host_jvm_args` by 1G to get skylint to compile... I think that's similar to what you did in other repos.

Issue #358, opened 2018-10-05
yarn_install takes forever to run


alexeagle:
Trying to do Angular development on my Mac, I observe over 20min spent with Bazel outputting nothing, waiting for the first action to run.
Looking at the Activity Monitor I can see node processes running yarn.

Doing a manual fresh yarn install of the project takes 5m.

```
clean cache (?)

time yarn
✨  Done in 325.88s.
real	5m26.645s

Immediate re-run
✨  Done in 3.21s.

real	0m3.561s
user	0m3.793s
sys	0m0.386s

Under bazel
ERROR: /Users/alexeagle/Projects/angular/BUILD.bazel:18:1: no such package '@angular_deps//': yarn_install failed:  (Timed out) and referenced by '//:node_modules'
ERROR: Analysis of target '//packages/compiler-cli/integrationtest/bazel/ng_module:test_module' failed; build aborted: Analysis failed
INFO: Elapsed time: 1205.702s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (228 packages loaded)
FAILED: Build did NOT complete successfully (228 packages loaded)

real	20m11.610s
user	0m0.162s
sys	0m0.247s

```

alexeagle:
We run yarn with a different `--cache-folder` every time to workaround the lack of thread-safety in yarn's cache.

I found that there's a `--mutex` option that ensures only one yarn process accesses the cache while still re-using previously downloaded artifacts.

gregmagolan:
The global cache is on by default (#359) which significantly speeds up yarn if the cache is populated. Note that using the mutex means that multiple yarn_install will be run serially by default.

PR #359, opened 2018-10-05
Use yarn mutex rather than cache-folder

alexeagle:
This lets us reuse the downloaded artifact cache


alexeagle:
Probably fixes #358

PR #360, opened 2018-10-05
History server fixes

alexeagle:
Needs some automated testing before merge


alexeagle:
history server being deleted in https://github.com/bazelbuild/rules_nodejs/pull/1158

I think we still have a problem with running things on Windows but that's not a rule-shaped problem anymore

PR #363, opened 2018-10-12
Vendored node & yarn support

gregmagolan:
@alexeagle I checked in the linux node distribution. Probably a better way is to download and extract in the yarn script that runs the test since this increases the size of the repo by quite a bit.


gregmagolan:
> so many files! makes the github experience broken and slow
why check in the sources for node? can't you just checki in the compiled binary?

The npm node_modules distro is included in there so you need more than just the compiled binary. But I could probably remove the /lib and /include folders and just have the compiled binary and npm within the included node_modules. If it not a lot of files is that preferable to downloading the archive for the test?

gregmagolan:
> > so many files! makes the github experience broken and slow
> > why check in the sources for node? can't you just checki in the compiled binary?
> 
> The npm node_modules distro is included in there so you need more than just the compiled binary. But I could probably remove the /lib and /include folders and just have the compiled binary and npm within the included node_modules. If it not a lot of files is that preferable to downloading the archive for the test?

I trimmed some of the files out but its still not idea as the checked in npm in node_modules has a nested node_modules which contains quite a few packages that npm depends on.

alexeagle:
filed https://github.com/bazelbuild/rules_nodejs/issues/368 to follow-up on making the example more realistic

PR #364, opened 2018-10-13
feat: support jasmine config in bazel rules

PalanQu:
This PR allow to use custom configuration in jasmine test


PalanQu:
I signed it!

PR #367, opened 2018-10-13
Fix commonjs bundle

### LINKED ISSUE #366: commonjs modules are not resolved in the rollup bundle output
I setup a [demo project](https://github.com/wayou/rules_nodejs_rollup_commonjs_bundle) to reproduce the problem. 

Say we have a `index.js` requires modules from local and npm packages by commonjs syntax:

*index.js*
```javascript
var _ = require("lodash");
var names = require("./lib");

module.exports = _.sample(names);
```

and the local module 

*lib.js*
```javascript
module.exports = ["john", "david", "mike"];
```

then using the `rollup_bundle` label to build:

*BUILD.bazel*
```python
load("@build_bazel_rules_nodejs//:defs.bzl", "rollup_bundle")
rollup_bundle(
    name = "bundle",
    srcs = [
        "index.js",
        "lib.js",
    ],
    entry_point = "rules_nodejs_rollup_commonjs_bundle/src/commonjs/index.js",
    deps = [
        "@npm//lodash",
    ],
)
```

we got the output without the required modules resolved like this:

```javascript
(function () {
    'use strict';
    var _ = require("lodash");
    var names = require("./lib");
    module.exports = _.sample(names);
}());
//# sourceMappingURL=bundle.es6.js.map
```

wayou:
fix #366

### DIFF
--- a/internal/rollup/package.json
+++ b/internal/rollup/package.json
@@ -4,6 +4,7 @@
         "is-builtin-module": "2.0.0",
         "minimist": "1.2.0",
         "rollup": "0.65.0",
+        "rollup-plugin-commonjs": "^9.2.0",
         "rollup-plugin-node-resolve": "3.3.0",
         "rollup-plugin-sourcemaps": "^0.4.2",
         "source-map-explorer": "^1.5.0",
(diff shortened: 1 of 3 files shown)


alexeagle:
thanks @wayou  - did you notice the test failures here? are you debugging them?

ashi009:
This one probably comes with a side effect of how commonjs transforms the code. It creates default exports for commonjs packages.  So any code that imports those packages with `import * as ns from 'pkg'` will fail the rollup build.

PR #380, opened 2018-10-18
Support --nolegacy_external_runfiles flag

gregmagolan:


.circleci/config.yml
```diff
@@ -99,6 +113,29 @@ jobs:
             - "internal/e2e/node_loader_preserve_symlinks/node_modules"
             - "internal/test/node_modules"
 
+  # Run all targets and tests with --nolegacy_external_runfiles flag on
+  build_no_legacy_runfiles:
```
alexeagle: the flag is only present during the rollout window of the breaking change, so we'll have to remove this job later. Do we really need to test under both modes now? What if we only flipped the flag on and leave it that way?

gregmagolan: 👍  I think we can test with just the flag on. Any tests that pass with the flag on should not be broken by adding the legacy external runfiles back in.

gregmagolan: Hmmmm. With --nolegacy_external_runfiles on testing across workspaces is broken in an odd way. It may be within Bazel itself as the line it errors out on looks to be during the Bazel test setup.
```
bazel test @program_example//...
==================== Test output for @program_example//:test:
external/bazel_tools/tools/test/test-setup.sh: line 264: : command not found
```
@alexeagle Are these tests important to keep? We can move them to another CircleCI job with legacy external runfiles for now.

gregmagolan: Interestingly, these cross workspace targets are also failing locally with Bazel 0.18 for different reasons.

PR #392, opened 2018-10-26
Update node LTS, yarn to latest version and change defaults.

Globegitter:
On the 30th  of October Node 10.12 is planned to become the new LTS so I think we should update the default version to that. And for yarn it is imo just worth updating as it contains a lot of fixes and improvements since 1.3.


Globegitter:
It says `Skylint failed to build` so should not be an actual issue with my code as far as I can tell.

Globegitter:
@alexeagle nodejs LTS has now been released: https://nodejs.org/en/blog/release/v10.13.0/ So updated the PR here to reflect that.

alexeagle:
Thanks!

Issue #394, opened 2018-10-26
Rename jasmine_node_test to nodejs_jasmine_test


haikalpribadi:
Hi, as we can see, the naming convention for the rules in rules_nodejs starts with `nodejs_<...>` (such as `nodejs_binary` and `nodejs_test`. However, `jasmine_node_test` breaks this convention. Before rule names starts getting messier, can we rename `jasmine_node_test` to `nodejs_jasmine_test` to maintain consistency?

haikalpribadi:
We can make a PR for this if the @bazelbuild team thinks it makes sense. cc: @vmax

alexeagle:
Actually we discussed last week in the context of https://github.com/bazelbuild/rules_nodejs/issues/168 that this rule shouldn't have `jasmine` in the name. We don't expose which test runner is used, and this rule could be used to run most node unit tests.

So I think the new name should be `nodejs_unit_test`
/cc @vikerman

We're happy to take a PR for that. The tricky bit will be how to roll this out without breaking users. Probably needs to be available under both names during a transition period of several releases.

haikalpribadi:
Well, I would love just see one test rule, such as `nodejs_unit_test`. However, from what I understand is that there are 3 unit test framework that people use for NodeJS. At @graknlabs we use Jest. So we're (@vmax) also making a PR to add `nodejs_jest_test`. So that means we can just have Jasmine to be the default `nodejs_unit_test`, unless it becomes a higher level rule to wrap all 3 test framework (or any others) and have it accept an argument to select the framework to use. What do you think @alexeagle ?

alexeagle:
Actually this one is named "jasmine_node_test" to match the rule we use internally at Google, so I don't think it's worth renaming it. I do agree that we should have `mocha_test` and `jest_test` rules.

PR #395, opened 2018-10-26
Rules for running jest tests

vmax:


Globegitter:
@vmax thanks for getting started on this, that is awesome, we are using jest internally in a quite hacky way so far and I have not been able yet to write a proper js rule, so happy to see something happening there.

Generale comments: I wonder if it makes sense to stick to the existing naming and name this `jest_node_test`, what do you think? Also bazel treats certain exit codes in a special way as also used by the jasmin tests: https://github.com/bazelbuild/rules_nodejs/blob/master/internal/jasmine_node_test/jasmine_runner.js#L11 That is one thing I actually gut stuck with last time, trying to map the exit codes, but maybe you have some ideas.

I am also seeing you are copying the file to get around the jest symlink issue: https://github.com/facebook/jest/issues/5356 - I wonder if there is also a better way of getting around this as I could imagine that impacting performance once you reach a certain number of tests. I know it is possible to run `jest path/to/file.js` but have not tested yet if that fixes the symlink issues. 

Also can you provide examples? This is a simple way of being able to tests the functionality but also to show how to use the rules.

Globegitter:
Ahh I am also just seeing #394 for naming - yeah that also makes some sense.

Globegitter:
I had more of a play with it and no matter how jest is invoked it does not recognize symlinks. I do have started adding symlink support to jest a while ago, here is an incomplete PR: https://github.com/facebook/jest/pull/7364

If anyone wants to help finish it up quicker.

dicarlo2:
@Globegitter AFAICT you can avoid the Jest symlinks issue entirely by using the [`--runTestsByPath`](https://jestjs.io/docs/en/cli#runtestsbypath) option.

In case it helps, here's our entry_point script for running jest tests: https://gist.github.com/dicarlo2/7b920185b28058f7f7379ad39834472b. We use a modified set of rules from this repo, so there may be some differences that I'm not accounting for, but this might give some ideas on how to make jest work. 

At a high level, we just call `jest.runCLI` with the appropriate templated arguments (list of files, whether we're running in ci, etc.). We then have some logic for maintaining snapshots - ww write out to the test output directory all of the added/updated snapshots + a `summary.json` file that contains the paths to those snapshots + the paths of the deleted ones. We have another process running that watches the local bazel out directory and syncs the changes over to the workspace. No idea if this is optimal, but it works for us.

Globegitter:
@dicarlo2 thanks for posting that, that is pretty useful - will take a look.

Globegitter:
@dicarlo2 Do you also have the set of rules that you are using with this runner? Or some more hints? `TEMPLATED_filePaths`, is quite obvious, but what are you setting `TEMPLATED_env` to and when are you setting `TEMPLATED_update`? In any case I will start investigating, maybe that will make some of the things clearer by itseld

dicarlo2:
Here's the rules file: https://gist.github.com/dicarlo2/3f35d9ef183a0cfe856997b7122a105a
(And the test sources aspect: https://gist.github.com/dicarlo2/8254f5b8c3d954515287d5e33f20466f)

Note that we are using a slightly customized variant of `nodejs_binary`, hence the `load` from an internal file. However, I don't think we actually require any of the changes that we made for the jest rule, so in theory dropping in the `nodejs_binary` from this repo should work (with a few modifications to expose the underlying function and arguments).

Also, FWIW this is one of the first rules I wrote for bazel and I haven't updated it since, so it may be doing some janky things.

Globegitter:
@dicarlo2 Thanks for posting this, that was enough for me to also put together a set of jest_rules I have here: https://github.com/ecosia/bazel_rules_nodejs_contrib/tree/jest-rules/internal/nodejs_jest_test

I changed things a bit to make it more conformant with test rules from other languages. Feel free to have a look if you are interested :)

vmax:
We've switched our primary test framework to be Jasmine such that the officially supported rule is used. Therefore, we no longer plan on supporting `nodejs_jest_test`.

PR #398, opened 2018-10-29
jasmine_node_test: pass templated_args through to nodejs_test.

Ubehebe:
Currently, it is not possible to specify `templated_args` on a `jasmine_node_test`. This is because the `jasmine_node_test` implementation unconditionally passes its own `templated_args` to the underlying `nodejs_test` rule. It is an error in the loading phase to have a target with duplicated keywords. 

My use case is to pass `--node_options=--experimental-modules` to the underlying `nodejs_binary`, which allows node to load ES modules [natively](https://nodejs.org/api/esm.html).

General support for ES modules in rules_nodejs is a larger topic that calls for more design and discussion, but I think this commit is independently useful. Thoughts?

### DIFF
--- a/internal/jasmine_node_test/jasmine_node_test.bzl
+++ b/internal/jasmine_node_test/jasmine_node_test.bzl
@@ -60,6 +60,7 @@ def jasmine_node_test(
     all_data += [Label("//internal/jasmine_node_test:jasmine_runner.js")]
     all_data += [":%s_devmode_srcs.MF" % name]
     entry_point = "build_bazel_rules_nodejs/internal/jasmine_node_test/jasmine_runner.js"
+
     # If the target specified templated_args, pass it through.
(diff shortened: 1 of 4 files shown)


Ubehebe:
(The ci/circleci:lint failure looks spurious: `Target @io_bazel//src/tools/skylark/java/com/google/devtools/skylark/skylint:Skylint failed to build`. I notice that the CI page has a banner saying "Partially Degraded Service - components affected: Heroku", which may be related. I ran `yarn skylint` on my local machine and verified the commit was clean.)

alexeagle:
Hey sorry @Ubehebe this was totally missed. I think it's reasonable. I force-pushed to your branch with an added test, and ported the feature to the `@bazel/jasmine` package (we'll delete the old one soon)

PR #400, opened 2018-10-31
Strip entry point prefix correctly.

mtsgrd:
I ended up with a mangled `bundle.js` because of the incorrect use of `lstrip()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,13 @@ npm_install(
 )
 ```
 
+#### yarn_install vs. npm_install
+
+`yarn_install` is the preferred rule for setting up Bazel-managed dependencies for a number of reasons:
+
+* `yarn_install` will use the global yarn cache by default which will improve your build performance (this can be turned off with the `use_global_yarn_cache` attribute)
+* npm has a known peer dependency hoisting issue that can lead to missing peer dependencies in some cases (see https://github.com/bazelbuild/rules_nodejs/issues/416)
+
 #### Fine-grained npm package dependencies
 
 You can then reference individual npm packages in your `BUILD` rules via:


mtsgrd:
It looks like testing this requires a `rollup_bundle` with at least one additional entry point, and doing that generates a new set of files for code-splitting. Does it make sense to do this and add a few more golden files?

```bash
tree -L 2 bazel-bin/internal/e2e/rollup
...
├── additional_entry_bundle.cs
│   ├── foo.js
│   ├── foo.js.map
│   ├── index.js
│   └── index.js.map
├── additional_entry_bundle.cs.es6
│   ├── foo.js
│   ├── foo.js.map
│   ├── index.js
│   └── index.js.map
├── additional_entry_bundle.cs.min
│   ├── foo.js
│   ├── foo.js.map
│   ├── index.js
│   └── index.js.map
├── additional_entry_bundle.cs.min_debug
│   ├── foo.js
│   ├── foo.js.map
│   ├── index.js
│   └── index.js.map
├── additional_entry_bundle.es6
│   └── internal
├── additional_entry_bundle.es6.js
├── additional_entry_bundle.js
├── additional_entry_bundle.min.js
├── additional_entry_bundle.min_debug.js
...
```

alexeagle:
This e2e test has additional entry points

alexeagle:
do you know why that didn't demonstrate the bug? Could you add the missing assertion there?

mtsgrd:
The `lstrip` function strips all occurrences of letters present in the pattern up until the first mismatch. In this case, In this case, we have the variables:

```
main_entry_point_dirname = internal/e2e/rollup_code_splitting
entrypoint = main2.js
```

The first letter of the `entry_point` is not in `main_entry_point_dirname` so we get correct output. If the js file were called `additional_entry.js` then `lstrip` would leave only `y.js`!

.bazelrc
```diff
@@ -6,7 +6,7 @@ test --test_output=errors
 # eventually a surprising failure with auto-discovery of the C++ toolchain in
 # MacOS High Sierra.
 # See https://github.com/bazelbuild/bazel/issues/4603
-build --symlink_prefix=/
+#build --symlink_prefix=/
```
gregmagolan: Looks like this change got in by accident

gregmagolan:
@mtsgrd Not sure how but I messed up pushing the revert of `#build --symlink_prefix=/` up to your fork. This PR got auto-closed. I replaced it with #426 which includes all of your commits.

PR #401, opened 2018-11-01
Retreat to running skylint/buildifier binaries in the docker image

alexeagle:
We have tried and failed several times to find memory settings that allow us to build skylint


alexeagle:
Note, this will re-introduce the previous problem of version skew between the bazel formatter/linter used locally and the one used on CI :(

PR #402, opened 2018-11-01
[WIP] Add internal node profile flag

filipesilva:


internal/node/node_loader.js
```diff
@@ -456,38 +457,77 @@ if (TEMPLATED_install_source_map_support) {
...
+  beforePromise = new Promise((resolve) => {
+    session.post('Profiler.enable', () => {
+      session.post('Profiler.start', () => {
+        
+        // Set the profile writing logic.
+        afterPromise = () => new Promise((resolve, reject) => {
+          session.post('Profiler.stop', (err, { profile }) => {
+            if (!err) {
+              // TODO: new to decide on the real name/path here. Where should it go?
+              const profilePath = path.resolve(process.cwd(), 'PROFILE_NAME_HERE') + '.cpuprofile';
```
filipesilva: Where should these go?

internal/node/node_loader.js
```diff
@@ -456,38 +457,77 @@ if (TEMPLATED_install_source_map_support) {
...
+  beforePromise = new Promise((resolve) => {
+    session.post('Profiler.enable', () => {
+      session.post('Profiler.start', () => {
+        
+        // Set the profile writing logic.
+        afterPromise = () => new Promise((resolve, reject) => {
+          session.post('Profiler.stop', (err, { profile }) => {
+            if (!err) {
+              // TODO: new to decide on the real name/path here. Where should it go?
+              const profilePath = path.resolve(process.cwd(), 'PROFILE_NAME_HERE-' + Math.floor(Math.random() * 99999) + '.cpuprofile');
```
filipesilva: Where should these go, and what should they be named?

alexeagle: same behavior as `mktemp -d` works for me. It doesn't need to be elegant, just need to be able to find them, right? maybe including the program's name like `$0` would help to locate the right one?

filipesilva:
I used to think this was necessary to debug node profiles, but not anymore. 

Instead I think it's better to use a configuration flag like the [`debug` config for Angular](https://github.com/angular/angular/blob/052ef654d2bb279f8a1c797988cc9ac49e43892f/.bazelrc#L19):
```
test:debug --test_arg=--node_options=--inspect-brk --test_output=streamed --test_strategy=exclusive --test_timeout=9999 --nocache_test_results
```

And then running the test with `--config=debug`, and connecting the Chrome debugger.

https://github.com/angular/angular/issues/27425#issuecomment-448600155 as a write-up of gathering and analysing these profiles.

PR #405, opened 2018-11-02
Replace globs in fine grained deps filegroups with explicit file lists

gregmagolan:


internal/npm_install/test/golden/node_modules/jasmine/BUILD.bazel.golden
```diff
@@ -22,47 +22,48 @@ filegroup(
...
-        exclude = [
-          "test/**",
-          "docs/**",
-          "**/* */**",
-          "**/* *",
-        ],
-    ),
+    srcs = [
+        ".bin/jasmine",
+        ".editorconfig",
```
alexeagle: oh yeah this is the test - no .travis.yml

internal/npm_install/generate_build_file.js
```diff
@@ -317,15 +336,55 @@ function flattenDependencies(pkg, dep, pkgsMap) {
...
+              // See https://github.com/bazelbuild/bazel/issues/4327
+              .filter(f => !f.match(/[^a-zA-Z0-9\_\/\.\+\-\=\,\@\~]/))
+              // Don't include bazel BUILD and WORKSPACE files in a filegroup
+              .filter(f => path.basename(f) !== 'BUILD')
+              .filter(f => path.basename(f) !== 'BUILD.bazel')
+              .filter(f => path.basename(f) !== 'WORKSPACE');
+  if (exts.length) {
+    files = files.filter(f => {
+      // include files with no extensions if noExt is true
+      if (noExt && !path.extname(f)) return true;
```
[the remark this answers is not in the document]
gregmagolan: 👍

internal/npm_install/generate_build_file.js
```diff
@@ -317,15 +336,55 @@ function flattenDependencies(pkg, dep, pkgsMap) {
...
+  }
+  return files;
+}
+
 /**
  * Given a pkg, print a skylark `filegroup` target for the package.
  */
 function printPackage(pkg) {
+  const allSources = filterFilesForFilegroup(pkg._files).map(f => `"${f}",`).join('\n        ');
+  const liteSources = filterFilesForFilegroup(pkg._files, ['.js', '.d.ts', '.json', '.proto'])
```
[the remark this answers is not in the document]
gregmagolan: 👍

internal/npm_install/generate_build_file.js
```diff
@@ -317,15 +339,57 @@ function flattenDependencies(pkg, dep, pkgsMap) {
...
+ *             '' empty string denotes to allow files with no extensions, other extensions
+ *             are listed with '.ext' notation such as '.d.ts'.
+ */
+function filterFilesForFilegroup(files, exts = []) {
+  files = files
+              // Bazel target names must be composed entirely of characters drawn
+              // from the set a–z, A–Z, 0–9, and the punctuation symbols _/.+-=,@~.
+              // Files with spaces are also not allowed in Bazel runfiles
+              // See https://github.com/bazelbuild/bazel/issues/4327
+              .filter(f => !f.match(/[^a-zA-Z0-9\_\/\.\+\-\=\,\@\~]/))
```
vmax: @gregmagolan thanks for the patch!

could you please add `$` as valid symbol? not having it breaks [adv@5.5.2](https://github.com/epoberezkin/ajv/blob/v5.5.2/lib/%24data.js) because `$data` module is not symlinked into `bazel`'s runfiles tree

```suggestion
              .filter(f => !f.match(/[^a-zA-Z0-9\_\/\.\+\-\=\,\@\~\$]/))
```

gregmagolan:
* Removed `__lite` targets in favour of a whitelisted file extensions attribute.

* Added exclude_packages attribute that defaults to `['@bazel/bazel']` to prevent double download for bazel over npm workflow

internal/npm_install/generate_build_file.js
gregmagolan: Thanks for pointing that out. I think the regex needs to be permissive. Will figure out exactly which characters are legal tomorrow.

internal/npm_install/generate_build_file.js
```diff
@@ -317,15 +336,55 @@ function flattenDependencies(pkg, dep, pkgsMap) {
...
+
 /**
  * Given a pkg, print a skylark `filegroup` target for the package.
  */
 function printPackage(pkg) {
+  const allSources = filterFilesForFilegroup(pkg._files).map(f => `"${f}",`).join('\n        ');
+  const liteSources = filterFilesForFilegroup(pkg._files, ['.js', '.d.ts', '.json', '.proto'])
+                          .map(f => `"${f}",`)
+                          .join('\n        ');
+  const dtsSources =
```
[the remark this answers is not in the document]
gregmagolan: 👍

internal/npm_install/npm_install.bzl
```diff
@@ -24,14 +24,58 @@ See discussion in the README.
...
+        default = False,
+        doc = "Don't install devDependencies",
+    ),
+    "data": attr.label_list(),
+    "whitelisted_file_extensions": attr.string_list(
+        doc = """List of whitelisted file extensions.
+
+        For example, [".js", ".d.ts", ".proto", ".json", ""].
+
+        If an empty list then all file extentions are included in the package
```
alexeagle: should indicate why you would want to use this option (improve performance by minimizing the number of files copied into remote/sandboxed execution)

internal/npm_install/npm_install.bzl
```diff
@@ -24,14 +24,58 @@ See discussion in the README.
...
+        allow_files = True,
+        mandatory = True,
+        single_file = True,
+    ),
+    "prod_only": attr.bool(
+        default = False,
+        doc = "Don't install devDependencies",
+    ),
+    "data": attr.label_list(),
+    "whitelisted_file_extensions": attr.string_list(
```
alexeagle: did we decide to rename it to `included_files` ?

internal/npm_install/generate_build_file.js
```diff
@@ -108,7 +113,8 @@ function generateRootBuildFile(pkgs) {
 # this filegroup to include those files.
 filegroup(
     name = "node_modules",
-    srcs = glob(["node_modules/.bin/*"]) + [
+    srcs = [
+        ${binFiles}
```
alexeagle: nit: to be consistent, should format this one in this presentation logic too (move the join on newline/spaces)

internal/npm_install/generate_build_file.js
```diff
@@ -317,15 +344,57 @@ function flattenDependencies(pkg, dep, pkgsMap) {
...
+              .filter(f => path.basename(f) !== 'BUILD.bazel')
+              .filter(f => path.basename(f) !== 'WORKSPACE');
+  if (exts.length) {
+    const allowNoExts = exts.includes('');
+    files = files.filter(f => {
+      // include files with no extensions if noExt is true
+      if (allowNoExts && !path.extname(f)) return true;
+      // filter files in exts
+      let found = false;
+      exts.forEach(e => {
```
alexeagle: nit: if you use a for loop instead of forEach, then you could just return instead of having the `found` variable

internal/npm_install/generate_build_file.js
```diff
@@ -317,15 +344,57 @@ function flattenDependencies(pkg, dep, pkgsMap) {
...
+ * @param files array of files to filter
+ * @param exts list of white listed extensions; if empty, no filter is done on extensions;
+ *             '' empty string denotes to allow files with no extensions, other extensions
+ *             are listed with '.ext' notation such as '.d.ts'.
+ */
+function filterFilesForFilegroup(files, exts = []) {
+  files = files
+              // Bazel target names must be composed entirely of characters drawn
+              // from the set a–z, A–Z, 0–9, and the punctuation symbols _/.+-=,@~.$
+              // Files with spaces are also not allowed in Bazel runfiles
```
alexeagle: "and Bazel doesn't support unicode characters in file names"

internal/npm_install/npm_install.bzl
```diff
@@ -24,14 +24,58 @@ See discussion in the README.
...
+        For example, [".js", ".d.ts", ".proto", ".json", ""].
+
+        If an empty list then all file extentions are included in the package
+        filegroups. An empty string is a special string that denotes
+        that files with no extensions such as `README` should be included
+        in the package filegroups""",
+        default = [],
+    ),
+    "exclude_packages": attr.string_list(
+        doc = """List of packages to exclude from install""",
```
alexeagle: a bit more doc like "Use this when you want to install a package during manual `yarn install` but not install it when Bazel manages dependencies"
I suppose if we move `$(bazel info output_base)/external/npm` to `node_modules` then this option should go away. Maybe we should doc that we expect it's temporary?

internal/npm_install/npm_install.bzl
```diff
@@ -147,20 +174,22 @@ npm_install = repository_rule(
...
-      repository_ctx.attr.package_json,
-      repository_ctx.path("package.json"))
+  node = repository_ctx.path(get_node_label(repository_ctx))
+  yarn = get_yarn_label(repository_ctx)
+
   if repository_ctx.attr.yarn_lock:
-      repository_ctx.symlink(
-          repository_ctx.attr.yarn_lock,
-          repository_ctx.path("yarn.lock"))
+    # Copy the file over instead of using a symlink since the lock file
```
alexeagle: It seems like this means that running `bazel` will no longer update the checked-in lockfile. I think it's actually a feature today, that you don't have to manually run `yarn` to update the lockfile. Needs some discussion.
What do you think about moving the excluded_packages bit to a separate PR?

internal/npm_install/npm_install.bzl
gregmagolan: 👍

internal/npm_install/generate_build_file.js
gregmagolan: 👍

internal/npm_install/generate_build_file.js
gregmagolan: 👍

internal/npm_install/npm_install.bzl
gregmagolan: 👍

internal/npm_install/generate_build_file.js
gregmagolan: 👍

internal/npm_install/npm_install.bzl
gregmagolan: i agree

gregmagolan:
@alexeagle FYI Running this commit against angular/angular and running into some symlinks within node_modules issues. Flushing them out.

internal/npm_install/generate_build_file.js
```diff
@@ -155,6 +158,41 @@ function isPackage(p) {
...
+  const dir = path.posix.join(rootDir, subDir);
+  return fs
+      .readdirSync(dir)
+      // Filter out .bin folder since this is a folder created by yarn/npm that contains
+      // only symlinks and it breaks this function on Windows.
+      // For example, `node_modules/foo/.bin/foo` would be a symlink to `node_modules/.bin/foo`
+      .filter(f => !(f === '.bin' && fs.statSync(path.posix.join(dir, f)).isDirectory()))
+      // Also filter out test folders which contain symlinks in some packages
+      // such as `ecstatic/test/public/containsSymlink/more-problematic`
+      .filter(f => !(f === 'test' && fs.statSync(path.posix.join(dir, f)).isDirectory()))
```
alexeagle: I'm sure there's some npm package where files we actually need at runtime are in a directory called `test`
If the problem is symlinks under windows, could we specifically filter out those symlinks somehow instead of this whole directory?

internal/npm_install/generate_build_file.js
```diff
@@ -155,6 +158,41 @@ function isPackage(p) {
...
+      // Filter out directories that are bazel packages so that files do not cross Bazel package
+      // boundaries. npm packages should not generally include BUILD or BUILD.bazel files
+      // but they may as rxjs does temporarily.
+      .reduce((files, file) => {
+        const fullPath = path.posix.join(dir, file);
+        const relPath = path.posix.join(subDir, file);
+        return fs.statSync(fullPath).isDirectory() ?
+            (isPackageDir(fullPath) ? files : files.concat(listFiles(rootDir, relPath))) :
```
alexeagle: rules_node just deletes the BUILD/BUILD.bazel files instead
https://github.com/pubref/rules_node/blob/master/node/internal/clean_node_modules.js#L9
what do you think are the tradeoffs? I can imagine a package is published with a BUILD file in the root because they use bazel and forgot to .npmignore it, and now we filter out all fines in that package?

gregmagolan: Deleting the BUILD files sounds reasonable. I think I prefer that.

internal/npm_install/generate_build_file.js
gregmagolan: It's a very odd issue. It ended up showing itself in OSX as well for the `test` folder but not the `.bin` folder.

gregmagolan: The failure is odd. It fails within `fs.readDirSync()`

PR #406, opened 2018-11-04
Don't assume /bin/bash exists

mboes:
Using `/usr/bin/env bash` instead is more portable (non-Linux
platforms often have `bash` elsewhere). Even some Linux distributions
don't have a `/usr/bash` (e.g. NixOS). Only `/bin/sh` is mandated by
POSIX.


alexeagle:
I recall that we had a similar change in the past, but found that `/usr/bin/env` isn't reliable either.
The right long-term fix is to cut the dependency on bash altogether.
I assume you did some testing though, and this solves some problem for you? I think we can merge it despite those notes.

PR #407, opened 2018-11-04
Set use_default_shell_env in rollup example

mboes:
The `rollup` command needs `grep`, `cut`, `dirname` and other commands
to be available under the hood. But we can't rely on that to be true
without `use_default_shell_env = True`. That's because without this
flag, shell scripts use Bash's default `PATH` (since `/usr/bin/env -`
clears the entire environment). That default `PATH` will be different
on different platforms, and is not guaranteed to contain all the shell
commands this script expects to be able to call. Indeed, on NixOS the
default `PATH` for Bash is `/no-such-path`.

If turning on `use_default_shell_env` makes your nervous, it
shouldn't. According to https://github.com/bazelbuild/bazel/pull/6263,
`--experimental_strict_action_env` will become the default in Bazel
v0.20. This means that the default env will be restricted, standard
and small. (So this shouldn't hurt hermeticity - actually it should
help it.)

### DIFF
--- a/examples/rollup/rollup.bzl
+++ b/examples/rollup/rollup.bzl
@@ -27,6 +27,7 @@ def _rollup(ctx):
       inputs = ctx.files.srcs,
       executable = ctx.executable.rollup,
       outputs = [ctx.outputs.bundle],
+      use_default_shell_env = True,
       arguments = args,
   )
   return [DefaultInfo()]


alexeagle:
ping @mboes 
Is there an easy way for me to try this on nixOS to reproduce the problem?

mboes:
If you have a NixOS lying around, then try this:

```
$ hub clone mboes/rules_nodejs
$ cd rules_nodejs
$ bazel build //examples/rollup/...
INFO: Build options have changed, discarding analysis cache.
INFO: Analysed 6 targets (21 packages loaded).
INFO: Found 6 targets...
ERROR: /tmp/rules_nodejs/examples/rollup/BUILD.bazel:29:1: SkylarkAction examples/rollup/bundle.js failed (Exit 127)
bazel-out/host/bin/examples/rollup/rollup.runfiles/nodejs/bin/node: line 8: dirname: command not found
bazel-out/host/bin/examples/rollup/rollup.runfiles/nodejs/bin/node: line 8: cd: : No such file or directory
bazel-out/host/bin/examples/rollup/rollup.runfiles/nodejs/bin/node: line 9: readlink: command not found
bazel-out/host/bin/examples/rollup/rollup.runfiles/nodejs/bin/node: line 14: dirname: command not found
bazel-out/host/bin/examples/rollup/rollup.runfiles/nodejs/bin/node: line 14: cd: : No such file or directory
bazel-out/host/bin/examples/rollup/rollup.runfiles/nodejs/bin/node: line 17: /nodejs/bin/node: No such file or directory
INFO: Elapsed time: 5.357s, Critical Path: 0.68s
INFO: 0 processes.
```

With this branch:

```
$ git checkout use-default-shell-env
$ git rebase master  # to get #406, which is also necessary
$ bazel build //examples/rollup/... 
INFO: Analysed 6 targets (1 packages loaded).
INFO: Found 6 targets...
ERROR: /tmp/rules_nodejs/examples/rollup/BUILD.bazel:29:1: SkylarkAction examples/rollup/bundle.js failed (Exit 127)
bazel-out/host/bin/examples/rollup/rollup.runfiles/nodejs/bin/node: line 17: /home/mboes/.cache/bazel/_bazel_mboes/6ebb4a176e70acabad01519e5e923b54/external/nodejs/bin/nodejs/bin/node: No such file or directory
INFO: Elapsed time: 0.953s, Critical Path: 0.63s
INFO: 0 processes.
```

Still an error, but it's another error, because the script has at least made progress. It's still failing at line 17 like before, but no longer failing in the lines before that.

mboes:
To have a shell where Bazel and hub are available:

```
$ nix-shell -p bazel git-hub
```

alexeagle:
Any idea how we can run a CI on NixOS? It seems like there is a long tail of bugs to fix among all Bazel rules to run correctly on that distro. I'm not sure it benefits us to fix them one-off with no regression testing to keep it fixed. /cc @buchgr

buchgr:
Can you create a docker image with nixos on it? I believe we can now also run docker images on our Buildkite CI @fweikert?

PR #410, opened 2018-11-07
Add exclude_packages

gregmagolan:
Based on #405 so that should land first


gregmagolan:
From #405 when this change was included re: copying over lock file instead of symlinking it:

@alexeagle: "It seems like this means that running bazel will no longer update the checked-in lockfile. I think it's actually a feature today, that you don't have to manually run yarn to update the lockfile. Needs some discussion."

alexeagle:
Okay based on conversation in Angular's #caretaker slack channel, I think maybe the behavior where bazel updates the lockfile is not desirable. It's simply counterintuitive that performing a build has side-effects of changing files in your workspace. Let's get this in - could you rebase?

alexeagle:
I rebased it.

PR #419, opened 2018-11-14
Set --network-timeout option in yarn_install

gregmagolan:
This should mitigate yarn install timeouts on slow networks. I've seen timeouts while running a Windows VM on OSX and yarn install timeouts have been seen on occasion in CircleCI.

Note: default --network-timeout option in yarn is 30s


alexeagle:
I've felt something is fishy here - yarn install seems to take a lot longer under bazel than running on its own, as if Bazel is doing some network sandboxing or something. Agreed that we should mitigate it this way...

PR #421, opened 2018-11-16
Cleanups for compatibility with future Bazel releases

brandjon:
With these fixes, the following command passes in Bazel 0.19.1:

```
bazel test //... --all_incompatible_changes --incompatible_disallow_filetype=false --incompatible_no_transitive_loads=false --incompatible_disable_deprecated_attr_params=false --incompatible_package_name_is_a_function=false --incompatible_depset_is_not_iterable=false --incompatible_no_support_tools_in_action_inputs=false --incompatible_depset_union=false --incompatible_new_actions_api=false
```


brandjon:
The changes in this PR should be no-ops, so I'd be pretty surprised if the CI failure is related.

alexeagle:
I did a retry on CI, seems like a legit failure.
Looks like the arguments to the npm_package binary are wrong? One of the arguments should be in JSON format but the command run is using a params file that now contains only
```
bazel-out/darwin-fastbuild/bin/internal/npm_package/test/dependent_pkg
internal/npm_package/test
internal/npm_package/test/dependent_file
bazel-out/darwin-fastbuild/bin
bazel-out/darwin-fastbuild/genfiles
bazel-out/darwin-fastbuild/bin/internal/npm_package/test/dependent_pkg.pack
bazel-out/darwin-fastbuild/bin/internal/npm_package/test/dependent_pkg.publish
bazel-out/volatile-status.txt
```

brandjon:
Thanks for the tip! We'll see if this fixes it...

alexeagle:
I found the problem, by default `args.add_joined` will totally drop the argument if there are no elements in the array. But this program parses the whole command line as positional arguments.
Adding `omit_if_empty=False` to these two fixes it.

brandjon:
Thanks for the diagnosis and fix!

PR #422, opened 2018-11-16
Fix fine grained deps symlinks

gregmagolan:


internal/npm_install/generate_build_file.js
```diff
@@ -181,8 +174,19 @@ function listFiles(rootDir, subDir = '') {
...
         const fullPath = path.posix.join(dir, file);
         const relPath = path.posix.join(subDir, file);
-        return fs.statSync(fullPath).isDirectory() ? files.concat(listFiles(rootDir, relPath)) :
-                                                     files.concat(relPath);
+        let stat;
+        try {
+          stat = fs.statSync(fullPath);
+        } catch (e) {
+          if (fs.lstatSync(fullPath).isSymbolicLink()) {
+            // filter out broken symbolic links... these cause fs.statSync(fullPath)
```
DevVersion: What happens if the symlink is valid? I assume it won't even cause an error then right?

DevVersion: Sounds good.

PR #427, opened 2018-11-19
Default to verbose yarn_install and npm_install

gregmagolan:


alexeagle:
I'm not sure you should have to change the rule definition to debug it. Probably doesn't make sense to check in the non-quiet mode for your team and CI to use. What about alternatives like `bazel build --define=VERBOSE_PACKAGE_INSTALL=true` or something?

Globegitter:
just as a heads up with bazel `0.20.0` we do have some basic progress report for external repositories, so we generically see `fetching @workspace 6s` etc. and with 0.21.0 there will be a new api coming: `ctx.report_progress("Fetching npm dependencies"). So we could add a custom progress message. I still think it would be nice to have an option to see the underlying yarn output in addition to that I and quite like @alexeagle's suggestion.

ixdy:
This default has broken our use of `bazel query`, since now we get things like 
```
yarn install v1.12.1
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
```
appearing in stdout.

I tried setting `quiet = True` for the one `yarn_install` call in our workspace, but there's apparently yet another `yarn_install` lurking in some dependency or something which I'm trying to track down.

ixdy:
ah, maybe https://github.com/bazelbuild/rules_typescript/blob/master/internal/ts_repositories.bzl#L51-L70

ixdy:
oh, while that might be problematic, that's not what's causing the issue. I think it might be these:
https://github.com/bazelbuild/rules_nodejs/blob/8684bbeba7dac13584597b7ce7b596f9612d087c/internal/node/node_repositories.bzl#L575-L605

gregmagolan:
Lets change the default to quiet at this point since Bazel is now outputting status when running yarn_install with 0.20.0.

gregmagolan:
@ixdy #493 was merged in which defaults `quiet` to True. This should solve your problem if you pull that commit of rules_nodejs into your WORKSPACE.

ixdy:
@gregmagolan awesome, thanks! I worked around it on our end by just filtering out anything from `bazel query` which didn't start with `//`.

Issue #428, opened 2018-11-20
fine grained deps and catch-all deps broken in latest master


Globegitter:
It seems that commit: https://github.com/bazelbuild/rules_nodejs/commit/b5e0774852d2f30d91fc2442c3316e907c51b92b broke something that now we are getting errors like:

```
(14:20:38) ERROR: /home/markus/.cache/bazel/_bazel_markus/c83dd5dc66625baf386a59b750f50a10/external/yarn/BUILD.bazel:20:1: Target '@yarn//:node_modules/.bin/acorn' contains an error and its package is in error and referenced by '@yarn//:node_modules'

(14:28:11) ERROR: /home/markus/.cache/bazel/_bazel_markus/c83dd5dc66625baf386a59b750f50a10/external/yarn/BUILD.bazel:20:1: no such target '@yarn//:node_modules/.bin/url-loader/node_modules/.bin/mime': target 'node_modules/.bin/url-loader/node_modules/.bin/mime' not declared in package ''; however, a source file of this name exists.  (Perhaps add 'exports_files(["node_modules/.bin/url-loader/node_modules/.bin/mime"])' to /BUILD?) defined by /home/markus/.cache/bazel/_bazel_markus/c83dd5dc66625baf386a59b750f50a10/external/yarn/BUILD.bazel and referenced by '@yarn//:node_modules'
```

no matter if I use the `:node_modules` target or fine grained deps. Switch one commit back, to release 0.16.1 fixes fine grained deps but the catch-all `node_modules` target is still broken with basically the same errors.

Just been playing around a bit and removing the line https://github.com/bazelbuild/rules_nodejs/blob/master/internal/npm_install/generate_build_file.js#L114 fixes the catch-all `node_modules` target.

gregmagolan:
Thanks @Globegitter! Something is up with the symlinks in the `.bin` folder for some npm packages it looks like to me. I'll look for a principled fix today.

Globegitter:
no worries and thanks for looking into this @gregmagolan. If you need any more information or a reproduction case I am happy to provide one.

gregmagolan:
@Globegitter Which OS are you building with? A repro case would be helpful. I haven't been able to reproduce it locally yet.

Globegitter:
This is on Ubuntu 18.04 - let me put something up today.

gregmagolan:
This looks related: https://github.com/bazelbuild/rules_nodejs/issues/438

[...]

PR #430, opened 2018-11-23
Fix fine grained deps .bin issue.

Globegitter:
To be honest I just blindly added this based on https://github.com/bazelbuild/rules_nodejs/commit/b5e0774852d2f30d91fc2442c3316e907c51b92b but seems to fix fine grained deps for us. Fixes part of #428

### DIFF
--- a/internal/npm_install/generate_build_file.js
+++ b/internal/npm_install/generate_build_file.js
@@ -180,6 +180,9 @@ function listFiles(rootDir, subDir = '') {
           }
           throw e;
         }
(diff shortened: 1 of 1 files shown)


Globegitter:
Yeah I am not sure this is the best fix given it breaks other tests. Let's see if @gregmagolan comes up with a better fix otherwise I can try and tackle this again in a few days.

gregmagolan:
Closing in favor of https://github.com/bazelbuild/rules_nodejs/pull/457

PR #432, opened 2018-11-23
Cleanups for compatibility with future Bazel releases

### LINKED ISSUE #431: Fix Bazel incompatible changes
I'd like to be able to build rules_nodejs with `--all_incompatible_changes`:
  `cd rules_nodejs && bazel build //... --all_incompatible_changes`

It may not be possible to fix every issue now due to transitive dependencies, but we should try to fix the issues from this repository.

laurentlb:
- No implicit flattening of depsets
- Separate tools from inputs in an action
- Update some dependencies

Fixes #431

### DIFF
--- a/internal/rollup/rollup_bundle.bzl
+++ b/internal/rollup/rollup_bundle.bzl
@@ -173,21 +173,19 @@ def _run_rollup(ctx, sources, config, output, map_output=None):
     args.add("--globals")
     args.add_joined(["%s:%s" % g for g in ctx.attr.globals.items()], join_with=",")
 
-  # TODO: Avoid flattening the depset
-  inputs = sources.to_list() + [config]
-
-  inputs += _filter_js_inputs(ctx.files.node_modules)
+  direct_inputs = [config]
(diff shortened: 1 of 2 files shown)


laurentlb:
This now works:
`bazel test //... --all_incompatible_changes --incompatible_no_transitive_loads=false --incompatible_no_support_tools_in_action_inputs=false `

`--incompatible_no_transitive_loads=false` is needed because of the Go dependency
`--incompatible_no_support_tools_in_action_inputs=false` is needed because of a bzl file shipped with Bazel

laurentlb:
ping?

laurentlb:
With my latest commits, it's all fixed!

Tested:
`bazel0.20 test //... --all_incompatible_changes`

mgechev:
LGTM, I'll let @alexeagle and @gregmagolan give the final approval because they have more context.

laurentlb:
Thanks!
Yes, we're looking at enabling this on the CI for all rules.

laurentlb:
please merge (I don't have write access)

PR #441, opened 2018-12-01
fix(rollup_bundle): add resolve({..., browser:true})

IgorMinar:
This is useful for bundling zone.js and other npm packages that contains
the browser field in the package.json

### DIFF
--- a/internal/rollup/rollup.config.js
+++ b/internal/rollup/rollup.config.js
@@ -165,8 +165,12 @@ const config = {
   },
   plugins: [TMPL_additional_plugins].concat([
     {resolveId: resolveBazel},
-    nodeResolve(
-        {jsnext: true, module: true, customResolveOptions: {moduleDirectory: nodeModulesRoot}}),
+    nodeResolve({
+      jsnext: true,
+      module: true,
+      browser: true,
+      customResolveOptions: {moduleDirectory: nodeModulesRoot}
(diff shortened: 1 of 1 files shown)


gregmagolan:
Just a note that the `browser` field will take precedence over `module` (and over `es2015`) which may be undesirable in some cases since `browser` code will not be ES6 typically. https://github.com/rollup/rollup-plugin-node-resolve/blob/master/src/index.js#L140

Maybe this should option should be turned on with an attribute?

Also, this will hopefully last soon as well which will support `es2015` needed to bundle rxjs: https://github.com/rollup/rollup-plugin-node-resolve/pull/182 which means we could expose a `mainFields` attribute in rollup_bundle as well.

alexeagle:
This PR ought to include a test of some kind. Also, could you explain why this change? (is there some accompanying change in Angular? what's wrong with how we load zone.js now?)

Per @gregmagolan comment, how can we even know what is the right set of attributes here? Users need `browser` field to resolve if and only if one of their packages uses it, but otherwise it's undesirable?

filipesilva:
https://github.com/rollup/rollup-plugin-node-resolve/pull/182 is now merged into Rollup.

alexeagle:
Not planning to add features to old rollup-bundle rule, will be deleted by 1.0

PR #444, opened 2018-12-03
Web package

alexeagle:


internal/web_package/assembler.js
```diff
@@ -0,0 +1,45 @@
...
+const fs = require('fs');
+const path = require('path');
+
+function main(args) {
+  const params = fs.readFileSync(args[0], {encoding: 'utf-8'}).split('\n').filter(l => !!l);
+
+  const outdir = params.shift();
+  if (!fs.existsSync(outdir)) fs.mkdirSync(outdir);
+
+  for (const f of params) {
```
mgechev: We'll ignore the nested directory structure for now?

internal/web_package/injector.js
```diff
@@ -0,0 +1,79 @@
...
+    const result = findElementByName(f, name);
+    if (result) return result;
+  }
+  return undefined;
+}
+
+function main(args) {
+  const params = fs.readFileSync(args[0], {encoding: 'utf-8'}).split('\n').filter(l => !!l);
+  const output_file = params.shift();
+  const input_file = params.shift();
```
mgechev: Can we use camelCase?

internal/web_package/injector.js
```diff
@@ -0,0 +1,79 @@
...
+    console.error('No <body> tag found in HTML document');
+    return 1;
+  }
+  const head = findElementByName(document, 'head');
+  if (!head) {
+    console.error('No <head> tag found in HTML document');
+    return 1;
+  }
+
+  for (const s of params.filter(s => /\.js$/.test(s))) {
```
mgechev: Do we want to add a hash/timestamp to invalidate caching?

internal/web_package/injector.js
```diff
@@ -0,0 +1,79 @@
...
+    const script = treeAdapter.createElement('script', undefined, [
+      {name: 'type', value: 'text/javascript'},
+      {name: 'src', value: `/${path.basename(s)}`},
+    ]);
+    treeAdapter.appendChild(body, script);
+  }
+
+  for (const s of params.filter(s => /\.css$/.test(s))) {
+    const stylesheet = treeAdapter.createElement('link', undefined, [
```
mgechev: This won't add newlines after script/style tags.

internal/web_package/package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "@bazel/web_package",
+  "private": true,
+  "dependencies": {
+    "parse5": "~5.1.0"
+  }
+}
```
mgechev: Newline

internal/node/node_repositories.bzl
```diff
@@ -576,3 +576,9 @@ def node_repositories(
       package_json = "@build_bazel_rules_nodejs//internal/http-server:package.json",
       yarn_lock = "@build_bazel_rules_nodejs//internal/http-server:yarn.lock",
   )
+
+  yarn_install(
+      name = "build_bazel_rules_nodejs_web_package_deps",
+      package_json = "@build_bazel_rules_nodejs//internal/web_package:package.json",
+      yarn_lock = "@build_bazel_rules_nodejs//internal/web_package:yarn.lock",
+  )
```
mgechev: Newline?

internal/web_package/web_package.bzl
```diff
@@ -0,0 +1,81 @@
...
+    },
+    outputs = {
+        "injected_index": "{}.index.html",
+    }
+)
+"""
+Assembles a web application from source files.
+
+Injects JS and CSS resources into the index.html.
+"""
```
mgechev: Newline

internal/web_package/injector.js
```diff
@@ -0,0 +1,79 @@
...
+
+function main(args) {
+  const params = fs.readFileSync(args[0], {encoding: 'utf-8'}).split('\n').filter(l => !!l);
+  const output_file = params.shift();
+  const input_file = params.shift();
+
+  const document = parse5.parse(fs.readFileSync(input_file, {encoding: 'utf-8'}), {treeAdapter});
+  const body = findElementByName(document, 'body');
+  if (!body) {
+    console.error('No <body> tag found in HTML document');
```
mgechev: Maybe we can consider creating a reusable abstraction for the logger so we can format the output?

internal/npm_install/generate_build_file.js
```diff
@@ -93,7 +93,8 @@ module.exports = {main};
 
 function generateRootBuildFile(pkgs) {
   const srcs = pkgs.filter(pkg => !pkg._isNested);
-  const binFiles = listFiles('node_modules/.bin');
+  const binDir = 'node_modules/.bin';
```
gregmagolan: Can be removed with a rebase

internal/web_package/BUILD.bazel
```diff
@@ -0,0 +1,27 @@
...
+    node_modules = ":node_modules_none",
+    install_source_map_support = False,
+)
+
+nodejs_binary(
+    name = "injector",
+    entry_point = "build_bazel_rules_nodejs/internal/web_package/injector.js",
+    data = [
+      "injector.js",
+      "@build_bazel_rules_nodejs_web_package_deps//parse5"
```
gregmagolan: Side-note: It's much easier to author new rules by using enclosed deps but adds an extra yarn install for every rule downstream that does this which is around 15s per rule since yarn_install are all run in serial with the cache mutex. Should come up with a pattern that we can follow easily so that new rules will end up with transitive deps in the user's node_modules?

internal/web_package/assembler.js
gregmagolan: Maybe use a list files function to get the file list and then just iterator over that. Here's an example: https://github.com/bazelbuild/rules_nodejs/blob/master/internal/npm_install/generate_build_file.js#L165

internal/web_package/injector.js
```diff
@@ -0,0 +1,79 @@
...
+    return 1;
+  }
+  const head = findElementByName(document, 'head');
+  if (!head) {
+    console.error('No <head> tag found in HTML document');
```
gregmagolan: Question: is it better to `throw new Error('No <head> tag found in HTML document')` or do the `console.error()` and `exit 1` like it is done here?

internal/web_package/injector.js
gregmagolan: +1 sounds like a good idea

internal/web_package/BUILD.bazel
alexeagle: yes, we should do that for all the extra yarn_install's in this repo, which means publishing a bunch of new `@bazel/*` packages. Let's do that as a follow-up.

internal/web_package/injector.js
alexeagle: how would we want to format it differently? I can add an error() helper that just forwards to console.error, is that what you mean?

internal/web_package/injector.js
alexeagle: I don't think a stack trace is useful, this is validating the user input, not violating a condition in our code.

internal/web_package/injector.js
alexeagle: yes, agreed, I don't think that's required thugh

internal/web_package/injector.js
alexeagle: Actually, no matter what input I give to parse5 (even invalid HTML) it always manages to find <body> and <head> in the document it parses out.

gregmagolan: Yeah. That's what I thought as well re: stack strack. As long as the console log is not silence by bazel then its better to just console error it out if its not an code exception.

internal/web_package/BUILD.bazel
gregmagolan: 👍

internal/web_package/assembler.js
mgechev: Sounds good

internal/web_package/injector.js
mgechev: Yep, agreed

internal/web_package/injector.js
alexeagle: will add it as a query parameter

internal/web_package/assembler.js
alexeagle: problem with the nested directory structure is that we don't know what is the right base path to strip

we probably need to match ts_devserver with a list of strip_prefix rules
will punt this to the next story... don't have enough time to do that now.

PR #456, opened 2018-12-05
Buildifier

alexeagle:


alexeagle:
probably want to ignore the "ran buildifier" commit when reviewing, it's just automated whitespace changes

Issue #447, opened 2018-12-04
rollup --silent silences build failure errors


ashi009:
In rollup_bundle rule, --silence is passed to make rollup quiet:

https://github.com/bazelbuild/rules_nodejs/blob/c19ce368f9a06cfc3f529540ab009c6c390e6cd7/internal/rollup/rollup_bundle.bzl#L166-L168

However, rollup is a lovely tool, which populates errors with `onwarn` as well. When `--silent` is specified a build will succeed even if there an error:

> [!] Error: 'render' is not exported by '../../../../../execroot/ws/bazel-out/darwin-fastbuild/bin/src/bundle.es6/external/npm/node_modules/react-dom/index.js'
> Error: 'render' is not exported by '../../../../../execroot/ws/bazel-out/darwin-fastbuild/bin/src/bundle.es6/external/npm/node_modules/react-dom/index.js'
>     at Object.onwarn (/private/var/tmp/_bazel_xiaoyi/2258f168242f85dcc9f2c0d3e9d2258a/execroot/ws/bazel-out/darwin-fastbuild/bin/src/_bundle.rollup.conf.js:138:11)
>     at Object.onwarn (/private/var/tmp/_bazel_xiaoyi/2258f168242f85dcc9f2c0d3e9d2258a/sandbox/darwin-sandbox/1/execroot/ws/bazel-out/host/bin/external/build_bazel_rules_nodejs/internal/rollup/rollup.runfiles/build_bazel_rules_nodejs_rollup_deps/node_modules/rollup/bin/rollup:413:50)
>     at Graph.onwarn (/private/var/tmp/_bazel_xiaoyi/2258f168242f85dcc9f2c0d3e9d2258a/sandbox/darwin-sandbox/1/execroot/ws/bazel-out/host/bin/external/build_bazel_rules_nodejs/internal/rollup/rollup.runfiles/build_bazel_rules_nodejs_rollup_deps/node_modules/rollup/dist/rollup.js:22136:50)
>     at Graph.warn (/private/var/tmp/_bazel_xiaoyi/2258f168242f85dcc9f2c0d3e9d2258a/sandbox/darwin-sandbox/1/execroot/ws/bazel-out/host/bin/external/build_bazel_rules_nodejs/internal/rollup/rollup.runfiles/build_bazel_rules_nodejs_rollup_deps/node_modules/rollup/dist/rollup.js:21884:14)
[...]

PR #459, opened 2018-12-06
rollup_bundle: stop --silent rollup to let it report all build errors

ashi009:
This CL removes `--silent` flag when running `rollup` to let it report all the warnings and errors (due to rollup/rollup#2582) during the build. The warning message now contains location info when available.

As a result of this change, some config errors are found and fixed as well:

- set banner with `config.output.banner` instead of `config.banner` as the latter is deprecated in rollup/rollup@ffe4e5b.
- removed `config.experimentalDynamicImport` as rollup/rollup@4d4ec4c made it default and removed it.
- stop exporting `resolveBazel` in `config` to avoid triggering unknown config error, and let the test to find it in plug-in list by name.

Updates #447

### DIFF
--- a/internal/rollup/rollup_bundle.bzl
+++ b/internal/rollup/rollup_bundle.bzl
@@ -169,7 +169,10 @@ def _run_rollup(ctx, sources, config, output, map_output = None):
 
     # We will produce errors as needed. Anything else is spammy: a well-behaved
     # bazel rule prints nothing on success.
-    args.add("--silent")
+    #
+    # TODO(https://github.com/rollup/rollup/issues/2582): Uncomment the following
+    # line once rollup stops reporting build errors with onwarn.
+    # args.add("--silent")
 
     if ctx.attr.globals:
         args.add("--external")
--- a/internal/rollup/rollup.config.js
+++ b/internal/rollup/rollup.config.js
@@ -154,8 +154,6 @@ const inputs = [TMPL_inputs];
 const enableCodeSplitting = inputs.length > 1;
 
 const config = {
-  resolveBazel,
-  banner,
   onwarn: ({loc, frame, message}) => {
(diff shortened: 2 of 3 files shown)


internal/rollup/rollup_bundle.bzl
```diff
@@ -169,7 +169,10 @@ def _run_rollup(ctx, sources, config, output, map_output = None):
 
     # We will produce errors as needed. Anything else is spammy: a well-behaved
     # bazel rule prints nothing on success.
-    args.add("--silent")
+    #
+    # TODO(https://github.com/rollup/rollup/issues/2582): Uncomment the following
+    # line once rollup stops reporting build errors with onwarn.
+    # args.add("--silent")
```
alexeagle: my concern is what does an execution of bazel look like now? Does rollup print a bunch of stuff to the terminal? as said in the original comment, a well-behaved bazel rule prints nothing on success. Can we somehow throw away the stdout from rollup if the compilation is successful?

ashi009: Totally understand. However, in this case I'd prioritize the correctness over the noise made by rollup. OTOH, this is actually inevitable in the nodejs world, even if the rollup is silenced, some poorly written plugin may console.log something from nowhere . So I think we can replace this comment with a more descriptive message pointing to a bug in rules_nodejs saying we should fix the log spam in the future. What do you think?

ashi009:
PTAL @alexeagle

filipesilva:
Now we can use `--silent` together with a custom `onwarn` handler, and still get the errors/warnings: https://github.com/rollup/rollup/pull/2981

alexeagle:
Not planning to add features to old rollup-bundle rule, will be deleted by 1.0

you were right though, `--silent` was causing other problems...

PR #462, opened 2018-12-07
feat(jasmine): add some flexibility to the spec suffix

### LINKED ISSUE #451: Jasmine test files should be flexible.
# 🚀 feature request

Jasmine test file names should be more flexible suffix. I have files that are `*.spec-unit.js` and `*.spec-int.js` that do not work with the bazel build rule because of this filter: https://github.com/bazelbuild/rules_nodejs/blob/master/internal/jasmine_node_test/jasmine_runner.js#L36

My "fix" was to rename my file to end with `*.spec.js`. 

I can come up with a PR with a fix / tests if this makes sense.

cnishina:
- Change the jasmine_runner regex filter to allow for suffix that are
spec-fo0.js, spec_fo0.js, or spec-fo0.js.
- Add more testing examples with the jasmine_node_test rule for JavaScript and TypeScript

closes #451


cnishina:
Will run 'yarn buildifier' later today... whoops.

internal/jasmine_node_test/jasmine_runner.js
```diff
@@ -33,7 +33,7 @@ function main(args) {
       // deps such as "@npm//typescript" if executed may cause the test to
       // fail or have unexpected side-effects. "@npm//typescript" would
       // try to execute tsc, print its help, and process.exit(1)
-      .filter(f => /[^a-zA-Z0-9](spec|test)\.js$/i.test(f))
+      .filter(f => /[^a-zA-Z0-9](spec|test)((-|_|\.)[a-zA-Z0-9]|)\.js$/i.test(f))
```
mgechev: This doesn't seem to match `/[^a-zA-Z0-9](spec|test)((-|_|\.)[a-zA-Z0-9]|)\.js$/i.test('decrement.spec-foo.js')`.

Looks like the `|` in `[a-zA-Z0-9]|` should be `*` or `*?`?

cnishina: It should be `/[^a-zA-Z0-9](spec|test)((-|_|\.)[a-zA-Z0-9]+|)\.js`. The `|)` will allow still allow for `.spec.js` and `.test.js` file endings.

cnishina:
I backed out my examples. I'm not sure how to contribute to the examples directory or if it is better to just have a markdown file.

internal/jasmine_node_test/jasmine_runner.js
mgechev: 👍

alexeagle:
@cnishina so far we avoided having any dependency from rules_nodejs to rules_ts because
1) we might accidentally introduce a layering violation where nodejs rules are specific to TS. to be fair, though, we have such a violation already with the "typescript" provider
2) it's confusing to have a circular dependency between the rulesets, though bazel allows it

alexeagle:
I'd like at least a unit test for this file so we understand what filenames match and don't match the regex before merging this

alexeagle:
Mentioned on the associated issue, how about exposing the pattern as an attribute on the rule? would that be a better way to account for all use cases?

cnishina:
This sounds reasonable. I'll have time once I get to the weekend.

cnishina:
Ping @mgechev for one more review. Let me know if this needs more work, approved, should be closed, etc. Thanks!

mgechev:
LGTM, @gregmagolan is the owner here, he'll be able to provide more accurate feedback.

alexeagle:
Windows CI failure looks maybe legit?

```



FAILED: //examples/define_var:define_var (Summary)
  | D:/b/jbn533dc/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/testlogs/examples/define_var/define_var/test_attempts/attempt_1.log
  | D:/b/jbn533dc/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/testlogs/examples/define_var/define_var/test_attempts/attempt_2.log
  | D:/b/jbn533dc/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/testlogs/examples/define_var/define_var/test.log
  | FAIL: //examples/define_var:define_var (see D:/b/jbn533dc/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/testlogs/examples/define_var/define_var/test.log)
  | INFO: From Testing //examples/define_var:define_var:
  | ==================== Test output for //examples/define_var:define_var:
  | 'test)' is not recognized as an internal or external command,
  | operable program or batch file.
  | ================================================================================
  | ==================== Test output for //examples/define_var:define_var:
  | 'test)' is not recognized as an internal or external command,
  | operable program or batch file.
  | ================================================================================
  | ==================== Test output for //examples/define_var:define_var:
  | 'test)' is not recognized as an internal or external command,
  | operable program or batch file.
  | ================================================================================
  | Target //examples/define_var:define_var up-to-date:
  | D:/b/jbn533dc/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/examples/define_var/define_var_loader.js
  | D:/b/jbn533dc/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/examples/define_var/define_var.sh
 

```

cnishina:
I'll try this on my Windows machine. 🤦‍♂️

gregmagolan:
After discussions with @alexeagle and @IgorMinar we came up with a better approach that solves the underlying problem this PR addresses without introducing a regex `filter` attribute that may be confusing for some users and still has the issue of users having to name their specs to match the regex. Details of the proposed change are here: https://github.com/bazelbuild/rules_nodejs/issues/513. Thanks for pointing out the issue @cnishina. I'll make a PR with the proposed changes shortly.

cnishina:
@gregmagolan Cool. Thanks! I couldn't get the Windows Bazel to work on my home machine.

PR #468, opened 2018-12-13
WIP: fix: Enforce buildifier lint on CI

manekinekko:
## PR Type

- [ ] Bugfix
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [x] CI related changes

- [ ] Yes
- [x] No


internal/node/node.bzl
```diff
@@ -301,10 +301,8 @@ nodejs_binary = rule(
     executable = True,
     outputs = _NODEJS_EXECUTABLE_OUTPUTS,
 )
-"""
-Runs some JavaScript code in NodeJS.
-"""
 
+# Runs some JavaScript code in NodeJS.
```
alexeagle: this will lose the documentation. run `yarn skydoc` to see the corresponding changes in the docs/ folder.

I think we need to enable only a whitelist of buildifier lint rules.

manekinekko: working on a fix

buildifier.sh
```diff
@@ -0,0 +1,118 @@
...
+          unused-variable
+        )
+
+LINT_RULES=$(join , ${LINT_RULES[@]})
+
+echo "Buildifier: (enabled lint rules)"
+echo $LINT_RULES
+echo ""
+
+find . -type f \( -name "*.bzl" -or -name WORKSPACE -or -name BUILD -or -name BUILD.bazel \) ! -path "*/node_modules/*" \
```
alexeagle: This is a nice idea, but it actually belongs in the `@bazel/buildifier` package as a new `bin` entry (I imagined `buildifier-all`) and should be implemented in JS so that it can work on Windows too.
But let's do that as a separate task :)
for now can we just list our lint rules in the package.json script to keep this change small?

manekinekko: We will then have to list all the lint rules (in the package.json) except the rules we want to temporarily exclude (ie, no-effect). Are we okay with that?

manekinekko: fixed: https://github.com/bazelbuild/rules_nodejs/blob/c4b99b9cc66723303d26e4e0d5add8b81e014edd/package.json#L27

PR #477, opened 2018-12-20
Consider all matches in resolveToModuleRoot

filipesilva:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Given `'rxjs/ajax/index.js'` as `path`, and the following resolutions:
```
var MODULE_ROOTS = [
  {module_name: /^rxjs\/ajax\/\b/, module_root: 'rxjs/ajax/'},
  {module_name: /^rxjs\/ajax\b/, module_root: 'rxjs/ajax/index.js'},
]
```

The current algorithm would disregard the first match because it left the string unchanged, even though it was the longest regex.

This caused it to return `rxjs/ajax/index.js/index.js` instead of `rxjs/ajax/index.js`.

Followup to https://github.com/bazelbuild/rules_nodejs/pull/474, related to a failure observed in https://github.com/angular/angular/pull/27764.

## What is the new behavior?
`rxjs/ajax/index.js` is returned for the given example.

- [ ] Yes
- [x] No

## Other information
I couldn't find any tests for this logic. Are there any I should be adding tests to @alexeagle @gregmagolan?

### DIFF
--- a/internal/node/node_loader.js
+++ b/internal/node/node_loader.js
@@ -30,9 +30,12 @@ const DEBUG = false;
 /**
  * The module roots as pairs of a RegExp to match the require path, and a
  * module_root to substitute for the require path.
+ * Ordered by regex length, longest to smallest. 
  * @type {!Array<{module_name: RegExp, module_root: string}>}
  */
-var MODULE_ROOTS = [TEMPLATED_module_roots];
+var MODULE_ROOTS = [
+  TEMPLATED_module_roots
+].sort((a, b) => b.module_name.toString().length - a.module_name.toString().length);
 
 /**
  * Array of bootstrap modules that need to be loaded before the entry point.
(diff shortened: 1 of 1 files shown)


internal/node/node_loader.js
```diff
@@ -59,22 +59,18 @@ function resolveToModuleRoot(path) {
...
-      match = p;
-    }
-  }
-  if (match) {
-    return match;
+  const orderedMatches = MODULE_ROOTS
+    // We want all possible matches.
+    .filter(m => m.module_name.test(path))
+    // Longest regex wins when multiple match.
+    .sort((a, b) => a.module_name.toString().length < b.module_name.toString().length);
```
alexeagle: hmm, can we avoid sorting this list every time we resolve a module?

filipesilva: Yes we can probably pre-sort them at the start.

filipesilva: Done

PR #479, opened 2018-12-20
Fix relative paths in web_package

alexeagle:


internal/web_package/injector.js
```diff
@@ -51,18 +63,33 @@ function main(params, read = fs.readFileSync, write = fs.writeFileSync, timestam
     throw ('No <head> tag found in HTML document');
   }
 
+  /**
+   * Trims the longest prefix from the path
+   */
+  function relative(execPath) {
```
mgechev: Node also exposes `relative` in the `path` module.

internal/web_package/injector.js
```diff
@@ -39,6 +39,18 @@ function findElementByName(d, name) {
 function main(params, read = fs.readFileSync, write = fs.writeFileSync, timestamp = Date.now) {
   const outputFile = params.shift();
   const inputFile = params.shift();
+  const rootDirs = [];
+  while (params.length && params[0] !== '--assets') {
+    let r = params.shift();
+    if (!r.endsWith('/')) {
```
mgechev: Maybe we can use `path.sep` which returns the correct separator depending on the OS.

alexeagle: these paths are constructed by bazel so they should be posix paths always

internal/web_package/injector.js
alexeagle: yeah, and it has weird cross-platform behavior too

PR #482, opened 2018-12-21
Add support for --additional_module_roots flag to nodejs_binary node_loader.js

gregmagolan:
Adds --additional_module_roots flag to nodejs_binary node_loader.js.

prereq for https://github.com/angular/angular/pull/27785.

Needs tests still.

### DIFF
--- a/internal/node/node_loader.js
+++ b/internal/node/node_loader.js
@@ -33,9 +33,29 @@ const DEBUG = false;
  * Ordered by regex length, longest to smallest. 
  * @type {!Array<{module_name: RegExp, module_root: string}>}
  */
-var MODULE_ROOTS = [
-  TEMPLATED_module_roots
-].sort((a, b) => b.module_name.toString().length - a.module_name.toString().length);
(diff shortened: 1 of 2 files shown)


gregmagolan:
> Still waiting to add tests on this one?
> Can you explain why it's needed?

Yes. A test is needed. Not a high priority at the moment.

In short, the problem faced is that `node_loader.js` is only aware of the `module_roots` of the deps that are part of the `nodejs_binary` attributes. If another rule wants to use a nodejs_binary executable (such as `protractor_web_test` in https://github.com/angular/angular/pull/27785) and wants that executable to be aware of the `module_roots` passed to it then currently you must redefine a nodejs_binary target in a macro, add the additional deps, and use that as the executable for the rule.

This PR provides a way to pass additional `module_roots` via a `--additional_module_roots` to the nodejs_binary executable.

I'm not super thrilled with the extra escaping required to pass a stringified JSON as a command line parameter but I couldn't think of a better way to go about it.

PR #483, opened 2018-12-27
Let rollup_bundle() use the CommonJS conversion plugin.

EdSchouten:
React itself, but also some of its dependencies (invariant, warning)
don't use ES6 style imports/exports, causing rollup_bundle() to fail on
them spectacularly. This can be resolved by enabling a plugin that
converts those on the fly.

This allows us to bundle a web application that uses React, React Router, React Router DOM, etc.

**Note:** I don't know what I'm doing. I hardly do any Javascript programming, which is why I've taken little effort so far to polish this up. Is this change correct? Am I overlooking something obvious?

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [X] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

PR #488, opened 2019-01-02
Allow custom rollup config

borgstrom:
Hi 👋 

Lately I've been experimenting with Bazel for a React app and would like to use the `rollup_bundle` rule to produce my final bundle.

To get my bundle working correctly I needed to customize some of the config with things like:

* The replace plugin: https://github.com/rollup/rollup/issues/719#issuecomment-227872509
* The commonjs plugin: #483 

I originally started down the path of trying to customize the existing template and rollup process, but because you need to install npm packages for the plugins I found the existing route very limiting and overly complex for trying to add support for all the different plugins & configurations.

Instead, this PR makes the `rollup` & `rollup_config_tmpl` rule attrs non-private so that users wishing to change the way the rollup process works can simply define their own.

I originally looked at #234, but found it super complicated for my needs.  This was a straight forward change that works well for me locally.

For reference, my files for this setup look like this:

**web/BUILD.bazel**
```python
package(default_visibility = ["//visibility:public"])

load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary")

exports_files([
    "rollup.config.js",
    "tsconfig.json"
])

nodejs_binary(
    name = "rollup",
    entry_point = "rollup/bin/rollup",
    data = [
        "@npm//rollup",
        "@npm//rollup-plugin-commonjs",
        "@npm//rollup-plugin-includepaths",
        "@npm//rollup-plugin-node-resolve",
        "@npm//rollup-plugin-replace",
    ],
)
```

**web/rollup.config.js**
```javascript
const path = require('path');

import includePaths from 'rollup-plugin-includepaths';
import nodeResolve from 'rollup-plugin-node-resolve';
import commonjs from 'rollup-plugin-commonjs';
import replace from 'rollup-plugin-replace';

// The Bazel rollup rule will write these template vars out before running rollup
const inputs = [TMPL_inputs];
const rootDir = 'TMPL_rootDir';
const nodeModulesRoot = 'TMPL_node_modules_root';

const config = {
  onwarn: (warning) => {
    // warnings should be fatal
    throw new Error(warning.message);
  },
  plugins: [
    // include our Bazel root dir
    includePaths({
      paths: [rootDir],
    }),

    // resolve against our node modules
    nodeResolve({
      customResolveOptions: {
        moduleDirectory: nodeModulesRoot,
      }
    }),

    // support commonjs
    commonjs(),

    // replace NODE_ENV with the appropriate env
    replace({
      'process.env.NODE_ENV': JSON.stringify('production'),
    }),
  ],
  output: {
    format: 'TMPL_output_format',
  }
};

if (inputs.length > 1) {
  // code splitting
  config.experimentalCodeSplitting = true;
  config.experimentalDynamicImport = true;
  config.input = inputs;
  if (process.env.ROLLUP_BUNDLE_FIXED_CHUNK_NAMES) {
    config.output.chunkFileNames = '[name].js';
  }
} else {
  // single bundle
  config.input = inputs[0];
  config.output['name'] = 'TMPL_global_name';
}

export default config;
```

**web/admin/BUILD.bazel**
```python
...

rollup_bundle(
    name = "bundle",
    deps = npm_deps + [
        ":admin",
    ],
    entry_point = "web/admin/index",
    rollup = "//web:rollup",
    rollup_config_tmpl = "//web:rollup.config.js",
)

...
```

If you're OK with this change I can sign the CLA and add tests for it.

These issues can be closed in favor of this one if it lands:

* #483 
* #441 
* #235 
* #234 
* #216

### DIFF
--- a/internal/rollup/rollup_bundle.bzl
+++ b/internal/rollup/rollup_bundle.bzl
@@ -111,7 +111,7 @@ def write_rollup_config(ctx, plugins = [], root_dir = None, filename = "_%s.roll
 
     ctx.actions.expand_template(
         output = config,
-        template = ctx.file._rollup_config_tmpl,
+        template = ctx.file.rollup_config_tmpl,
         substitutions = {
             "TMPL_additional_plugins": ",\n".join(plugins),
             "TMPL_banner_file": "\"%s\"" % ctx.file.license_banner.path if ctx.file.license_banner else "undefined",
@@ -196,7 +196,7 @@ def _run_rollup(ctx, sources, config, output, map_output = None):
         outputs += [map_output]
 
     ctx.actions.run(
-        executable = ctx.executable._rollup,
+        executable = ctx.executable.rollup,
         inputs = depset(direct_inputs, transitive = [sources]),
         outputs = outputs,
         arguments = [args],
@@ -594,17 +594,17 @@ ROLLUP_ATTRS = {
         doc = """Other rules that produce JavaScript outputs, such as `ts_library`.""",
         aspects = ROLLUP_DEPS_ASPECTS,
     ),
-    "_no_explore_html": attr.label(
-        default = Label("@build_bazel_rules_nodejs//internal/rollup:no_explore.html"),
-        allow_single_file = True,
-    ),
-    "_rollup": attr.label(
+    "rollup": attr.label(
         executable = True,
         cfg = "host",
         default = Label("@build_bazel_rules_nodejs//internal/rollup:rollup"),
     ),
-    "_rollup_config_tmpl": attr.label(
+    "rollup_config_tmpl": attr.label(
         default = Label("@build_bazel_rules_nodejs//internal/rollup:rollup.config.js"),
+        allow_single_file = True,
+    ),
+    "_no_explore_html": attr.label(
+        default = Label("@build_bazel_rules_nodejs//internal/rollup:no_explore.html"),
         allow_single_file = True,
     ),
     "_source_map_explorer": attr.label(


alexeagle:
Thanks, this is the cleanest proposal for this that I've seen.
My concern is that your copy of the rollup.config.js has some sensitive bits that a user could easily get wrong, like declaring the output location for the bundle which has to match the place Bazel expects the action to write.

Is it possible to pass a config fragment instead? It could `require()` the config that we generate by locating it in the runfiles, then `require()` your other plugins and mutate the `plugins` option of the config?

abhishek-carbon3d:
+1 This would really help us as well. I also need to make commonjs modules work with rollup. 

You can either provide first class support for every feature exposed by rollup config that bazel users ask for (say commonjs=true), leading to a parallel config language. Or provide native coverage for 80% use cases and leave an escape hatch for experts. Downside being you will have to provide and maintain API documentation for the config template. Neither options are very satisfying but the latter is probably more practical imo.

And once you allow running an arbitrary rollup binary, trying to insist on some safe fragment of the config template is not very useful imo. It just gets in the way and will probably never be 100% safe. I say just document the config template API, provide and example and a test.

Let me know if I can help to push this forward.

borgstrom:
Thanks for the comments.  I ended up abandoning these rules in favor of https://github.com/zenclabs/bazel-javascript.  In the end, even with a custom rollup config, I had trouble getting typescript to play nicely with all the features I wanted. Moving to webpack solved all my issues and provided the local development experience I wanted.

I don't have any plans to move forward with this PR.  @abhishek-carbon3d; The changes are pretty simple, if you want to sort out the final implementation with @alexeagle I encourage you to open another PR based on this one.

PR #493, opened 2019-01-10
Default to quiet=True for yarn_install and npm_install now that Bazel outputs progress

gregmagolan:
As of Bazel 0.20.0 progress is reported for repository rules such as yarn_install and npm_install so feedback is on by default for these rules which can take some time to complete.

Quiet defaulting to false was intended as a stop-gap measure while Bazel was working on the above change.

Reference: https://github.com/bazelbuild/rules_nodejs/pull/427#issuecomment-440277731

### DIFF
--- a/internal/npm_install/npm_install.bzl
+++ b/internal/npm_install/npm_install.bzl
@@ -76,7 +76,7 @@ COMMON_ATTRIBUTES = dict(dict(), **{
         doc = "Don't install devDependencies",
     ),
     "quiet": attr.bool(
-        default = False,
+        default = True,
         doc = "If stdout and stderr should be printed to the terminal.",
     ),
 })


Globegitter:
@gregmagolan Bazel 0.21 also got a new API to report custom progress messages: https://docs.bazel.build/versions/0.21.0/skylark/lib/repository_ctx.html#report_progress Maybe that would be a further nice addition, so on can get a better idea of what is happening currently than just seeing the default "fetching @repo".

gregmagolan:
@Globegitter Yes. I was thinking that as well. Created an issue for it https://github.com/bazelbuild/rules_nodejs/issues/498. It's a breaking change since it would require a minimum version of 0.21.0 downstream.

PR #494, opened 2019-01-10
Simplify node_launcher and fix signal propogation again.

Globegitter:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

We had this issue in the past, that where the signal could not be forwarded to the nodejs process, which was fixed in https://github.com/bazelbuild/rules_nodejs/pull/246, not much later it seems that functionality was broken again

Issue Number: N/A

## What is the new behavior?

I took the code from https://github.com/bazelbuild/rules_nodejs/pull/256 and found the remaining pieces of why that code was not enough to fix the issue. Turns out we have a calling chain of bash -> bash -> nodejs (on mac/linux) -> app, so we need to make sure to either propagate the signal all the way through or use exec / remove some of the layers of indirection.

In bash imo it does make the most sense to use `exec` when possible and I decided to remove the nodejs wrapper. I have noticed the commend there, that this node wrapper was put there on purpose, but I can not see how that is any different then setting the `PATH` already in bash and the comment is imo not clear enough as to what the difference is. I also tested the change and at least on our apps it is not breaking anyhting (neither the yarn call, nor running the app, or building the docker image and running that)

- [ ] Yes
- [ ] No
- [x] Not sure if this counts as one

## Other information

Not sure how to best test this.


Globegitter:
I had some more thoughts on testing, I do know `rules_docker` have certain rules they test on each `rules_xxx` docker container, so it could make sense to test this as a `rules_docker` requirement that every docker container should support signal propogation to the app. Otherwise we would have to add this dependency here, adding another circular dependency, or we would have to have some kind of publication system, where a docker container gets built, pushed to some registry outside of this `WORKSPACE` but then we can test the image as part of this test suite here.

Globegitter:
@alexeagle any thoughts on this? Would be nice to get a fix for this in so we do not have to fork and it is something that occasionally bites us in production.

gregmagolan:
Thanks for revisiting this. Looks good at first glance. Noticed your removed the hacky process.env PATH set. I have to remember exactly why that was there and check if its tested against.

gregmagolan:
Note: the underlying reason that process.env PATH hack was in there was that in some cases the process was reverting back to the system `yarn`, `npm` and `node` and not using the Bazel managed entry points.

Globegitter:
Do you have any example to reproduce the issue? I assumed there was a reason for this but I do want to understand why setting the `PATH` here: https://github.com/bazelbuild/rules_nodejs/blob/master/internal/node/node_repositories.bzl#L247 alone is not sufficient. And I would also really prefer not having to have this three layers of indirection before the actual application gets executed.

Globegitter:
@gregmagolan did you have any chance to look at this again?

gregmagolan:
I spent some time on it this morning. I can't reproduce the case where the `external/nodejs/bin` is not making into the `process.env['PATH']`. The comment references Linux as well so I'll will try it on that OS today. If the hack is still needed I'll make a separate PR today to add it to `node_loader.js` as that will be much simpler and will avoid the extra generated script. Either way will get this merged by the end of the day and planning on cutting a release soon as well. Thanks for the fix and cleanup.

PR #495, opened 2019-01-10
Symlink yarn lock file if exclude_packages is empty.

Globegitter:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

In the past rules_nodejs would symlink `yarn.lock` which was nice because I could just update a package in the `package.json` file, then do `bazel run` and my `yarn.lock` would be updated automatically. At some point this behaviour was broken.

Issue Number: N/A

## What is the new behavior?
If `exclude_packages` are empty I see no reason for the copy to take place and it would be nice to have a way to get the previous behaviour back that I kind of got to rely on.

- [ ] Yes
- [x] No

### DIFF
--- a/internal/npm_install/npm_install.bzl
+++ b/internal/npm_install/npm_install.bzl
@@ -166,7 +166,8 @@ cd "{root}" && "{npm}" {npm_args}
         else:
             repository_ctx.symlink(
                 repository_ctx.attr.package_lock_json,
-                repository_ctx.path("package-lock.json"))
+                repository_ctx.path("package-lock.json"),
+            )
 
     _add_package_json(repository_ctx)
     _add_data_dependencies(repository_ctx)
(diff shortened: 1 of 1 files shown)


Globegitter:
It wasn't just me who started to rely on this, some of our team members also. But given there is this proposal to link the workspace node_modules with bazel, would that then also make sure the yarn.lock is up-to-date? Or is there somehow else we can make this controllable? The only way I can think off is via adding a new arg to the rules, or via an env var. It would be nice if repository rules also had access to configuration variables, then we could just make it controllable via a `--define` flag.

Also happy to move this discussion on Slack if you think this is starting to get more involved.

bugzpodder:
When I run yarn with bazel, I run it with the --production flag so it would install less packages (bazel seem to take forever trying to do so).  This will modify yarn.lock and deletes dev dependencies, which is some what annoying.  I guess I can use the exclude_packages option to get around this?

Globegitter:
@bugzpodder does running yarn with --production actually write a different lock file? Would seem a bit strange to me, but I have not tried it.

bugzpodder:
Yeah I agree it is strange.  I'll check out the yarn CLI to see if it can be disabled via a flag.

PR #502, opened 2019-01-16
PARKED: add archive support to npm_install/yarn_install bazelWorkspaces

gregmagolan:
PR PARKED for now. May be revisited in the future.

Based off of #492 which should land first. This feature may or may not be desirable but here for future reference.


internal/npm_install/generate_build_file.js
```diff
@@ -291,44 +299,57 @@ def _maybe(repo_rule, name, **kwargs):
...
+    mkdirp(workspaceSourcePath);
+    bwDetails.pkg._files.forEach(file => {
+      if (/^node_modules[/\\]/.test(file)) {
+        // don't copy over nested node_modules
+        return;
+      }
+      const src = path.posix.join('node_modules', bwDetails.pkg._dir, file);
+      const dest = path.posix.join(workspaceSourcePath, file);
+      mkdirp(path.dirname(dest));
+      console.error(`copying ${src} -> ${dest}`);
```
kyliau: is this meant for debugging?

gregmagolan: Ahh. Yes. Thanks for the catch. This PR is parked for now any but will clean that up.

gregmagolan:
PR PARKED for now. May be revisited in the future.

PR #503, opened 2019-01-17
don't include external/ directory in npm_package

alexeagle:
I can't think of a reason why someone would want this directory in the package. It's only possible when your npm_package rule is declared in the root package anyway. But we should think a bit more about whether there's a case for vendoring your external deps this way, or whether some other files you publish might have references into the external/ directory?

### DIFF
--- a/internal/npm_package/packager.js
+++ b/internal/npm_package/packager.js
@@ -96,7 +96,9 @@ function main(args) {
   }
 
   // deps like bazel-bin/baseDir/my/path is copied to outDir/my/path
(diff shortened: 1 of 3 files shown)


gregmagolan:
LGTM. I can't of a reason why external deps under /external would be desirable in an npm package.

gregmagolan:
Note: this solved the issue of `external/npm/...` deps being added to the root npm_package in https://github.com/bazelbuild/rules_typescript/pull/374

PR #528, opened 2019-02-05
Delete CI configuration which is no longer used on BazelCI

alexeagle:
I think this file became obsolete at some point due to upstream changes on BazelCI?


gregmagolan:
I noticed that as well. It can go.

PR #531, opened 2019-02-05
Replace rollup example with Parcel

alexeagle:
It's confusing that we have a rollup example which doesn't use our rollup_bundle rule.


alexeagle:
It does have a README.md added in the example folder with quite a bit of explanation

PR #539, opened 2019-02-09
Fix jasmine_runner.js so that an already initialized jasmine will be used

gregmagolan:
Replaces #497 and fixes issue found in https://github.com/angular/flex-layout/pull/964.

A jasmine_node_test bootstrap script now only needs the following lines and jasmine_runner will share the initialized jasmine:

```
const jasmineCore: any = require('jasmine-core');
jasmineCore.boot(jasmineCore);
```

`jasmineCore.boot(jasmineCore)` will initialize `global.jasmine`. For reference, the current bootstrap looks like: https://github.com/angular/angular/blob/master/tools/testing/init_node_no_angular_spec.ts

If using zone.js, call `jasmine-patch.js` after `boot()` to patch `global.jasmine`:

```
import 'zone.js/dist/jasmine-patch.js';
```

Now if the bootstrap script modifies the jasmine env with something like this:

```
afterEach(() => testBedInstance.resetTestingModule());
```

jasmine_runner.js will persist that configuration as it will used the initialized jasmine. With the current code, jasmine_runner will not persist the configuration (or the zone patch).

### DIFF
--- a/internal/e2e/jasmine/BUILD.bazel
+++ b/internal/e2e/jasmine/BUILD.bazel
@@ -2,8 +2,8 @@ load("@build_bazel_rules_nodejs//:defs.bzl", "jasmine_node_test")
 
 jasmine_node_test(
     name = "shared_env_test",
+    srcs = ["jasmine_shared_env_test.spec.js"],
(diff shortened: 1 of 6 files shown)


gregmagolan:
> Maybe we should have a local regression test for this? Or somehow include a downstream project in our CI?

Added some tests. Was getting too late on Friday night when I fixed this.

PR #540, opened 2019-02-10
Update release steps

gregmagolan:


DEVELOPING.md
```diff
@@ -5,9 +5,16 @@ We strongly encourage you to review the project's scope described in the `README
 ## Releasing
 
 Start from a clean checkout at master/HEAD.
+
+Note: if you are using a new clone, you'll need to configure `git-clang-format` to be able to commit the release:
```
alexeagle: nit: git-clang-format already prints these messages. instead of duplicating you could just say to run the commands it prints

gregmagolan: the trouble I had was that it printed it during the `npm version` command and it breaks the command half way and interrupts the release process

PR #541, opened 2019-02-11
Don't run vendored_node test unless local system is Linux

gregmagolan:


examples/test_examples.sh
```diff
@@ -4,11 +4,26 @@
...
 
 set -u -e -o pipefail
-cd $(dirname $0)
+
+TESTS_ROOT_DIR=$(cd $(dirname "$0"); pwd)
+NODEJS_ROOT_DIR=$(cd $TESTS_ROOT_DIR/..; pwd)
+
+# Check environment
+unameOut="$(uname -s)"
+case "${unameOut}" in
```
alexeagle: nit: could be simpler, we may never care about the non-linux case

gregmagolan: 👍

PR #542, opened 2019-02-11
Fix //internal/npm_install/test:test for Windows and re-enable Windows tests on bazel CI

gregmagolan:
Also updates the default yarn to 1.12.3 as 1.12.1 was messing up symlinks when testing locally on Windows.

I fixed `//internal/npm_install/test:test` here but disabled two failing web_package tests on Windows (//cc @alexeagle)

@alexeagle Can you mirror yarn v1.12.3 at mirror.bazel.build? `https://github.com/yarnpkg/yarn/releases/download/v1.12.3/yarn-v1.12.3.tar.gz` => `https://mirror.bazel.build/github.com/yarnpkg/yarn/releases/download/v1.12.3/yarn-v1.12.3.tar.gz`


internal/npm_install/test/golden_windows/BUILD.bazel.golden
```diff
@@ -0,0 +1,33 @@
+
+package(default_visibility = ["//visibility:public"])
+filegroup(
+    name = "node_modules",
+    srcs = [
+        "node_modules/.bin/jasmine",
+        "node_modules/.bin/jasmine.cmd",
```
gregmagolan: Need seperate goldens for Windows because of the `.bin/foobar.cmd` files

alexeagle: can we just strip the `.cmd` off those? I think it's better to do the fuzzy assertion than to have twice as many goldens - so that we are able to review golden changes more carefully to catch mistakes

gregmagolan: That would leave you with two `"node_modules/.bin/jasmine"` lines so we'd have to remove the whole line. I think there may also be differences in file order and reading the list of files on Windows so the extra `.cmd` files may not be the only difference.

gregmagolan: I agree that the two sets of goldens are not ideal. It will also be a pain to update since you have to run `.accept` on Linux/OSX and on Windows.

gregmagolan: Not sure what a better approach is tho. There are subtle file system differences between the different platforms so the test should be run in both I think.

gregmagolan: I'll check the full set of differences in the goldens

gregmagolan: Yup. There are also file order difference. For example,
```
4) build file generator should produce a BUILD file for node_modules/ajv/BUILD.bazel
  Message:
    Error: Actual output in node_modules/ajv/BUILD.bazel doesn't match golden file /private/var/tmp/_bazel_greg/37b273501bbecefcf5ce4f3afcd7c47a/sandbox/darwin-sandbox/1/execroot/build_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bin/internal/npm_install/test/test.sh.runfiles/build_bazel_rules_nodejs/internal/npm_install/test/golden_windows/node_modules/ajv/BUILD.bazel.golden.
    
    Diff:
    --- a
    +++ b
    @@ -18,0 +19,2 @@
    +        ":LICENSE",
    +        ":README.md",
    @@ -27,0 +30 @@
    +        ":lib/compile/_rules.js",
    @@ -38 +41,4 @@
    -        ":lib/compile/_rules.js",
    +        ":lib/dot/_limit.jst",
    +        ":lib/dot/_limitItems.jst",
    +        ":lib/dot/_limitLength.jst",
    +        ":lib/dot/_limitProperties.jst",
    @@ -63,4 +69,5 @@
    -        ":lib/dot/_limit.jst",
    -        ":lib/dot/_limitItems.jst",
    -        ":lib/dot/_limitLength.jst",
    -        ":lib/dot/_limitProperties.jst",
    +        ":lib/dotjs/README.md",
    +        ":lib/dotjs/_limit.js",
    +        ":lib/dotjs/_limitItems.js",
    +        ":lib/dotjs/_limitLength.js",
    +        ":lib/dotjs/_limitProperties.js",
    @@ -82 +88,0 @@
    -        ":lib/dotjs/README.md",
    @@ -87,4 +92,0 @@
    -        ":lib/dotjs/_limit.js",
    -        ":lib/dotjs/_limitItems.js",
    -        ":lib/dotjs/_limitLength.js",
    -        ":lib/dotjs/_limitProperties.js",
    @@ -97 +98,0 @@
    -        ":LICENSE",
    @@ -99 +99,0 @@
    -        ":README.md",
    
    Update the golden file:
    
          bazel run //internal/npm_install/test:test.accept
```
I'll add a sort function to fix the order and then its just the .cmd files that are different. I can work around those.

internal/npm_install/test/golden/node_modules/@gregmagolan/test-a/BUILD.bazel.golden
```diff
@@ -12,7 +12,6 @@ filegroup(
 filegroup(
     name = "test-a__files",
     srcs = [
-        ":.bin/test",
```
alexeagle: is this change intentional?

gregmagolan: That comes with the yarn update. It no longer creates those files.

internal/npm_install/test/golden_windows/BUILD.bazel.golden
gregmagolan: Done

Issue #513, opened 2019-01-24
feat(jasmine): change jasmine attributes so that spec files are listed in `srcs` and other runtime deps listed in `data`


gregmagolan:
This would replace the need for a regex to find spec files in jasmine which has led to issues such as https://github.com/bazelbuild/rules_nodejs/pull/462. The same change should be applied to karma test rules in rules_typescript and protractor test rules in angular.

With jasmine, karma and protractor it is necessary to identify spec files from other source files and relying on the filename to do so as we do now is too brittle.

Note: this feature would go against the Bazel convention were `srcs` is only source files and not the outputs of other rules such as `ts_library` but after discussions with @alexeagle and @IgorMinar it seems like it's the best way to go for the test rules under nodejs.

Also note: one downside of this is you can no longer build spec files and other source files in the same `ts_library` upstream of the test rule since the spec files target must only contain spec files when added to `srcs`... however, it's also a nice pattern to keep test targets separate from non-test targets.

IgorMinar:
It's ok to build test helpers or other ts sources in the same ts_library as
specs, as long as evaluating these files has no observable global side
effects (e.g. monkey patching that should be applied conditionally via a
test init script rather than automatically by the jasmine test rule.

gregmagolan:
I think the best approach is to introduce a new attribute to `jasmine_node_test` named `specs` that can accept files or upstream targets the produce js such as `ts_library`. Everything passed to `specs` would be considered a spec regardless of filename. This would preserve the Bazel paradigm that `srcs` should only be files and also not break any current users. User's can update their targets to use `specs` for all specs and `data` for all other runtime deps.

A similar change can be made to the `ts_web_test` and `protractor_test` rules, however, these rules would also need an additional attribute named `non_specs` (or something better) to accept source files or deps that are not specs. `jasmine_node_test` doesn't have this problem since non-spec files can be listed in data. For consistency, `jasmine_node_test` could also have a `non_specs` attribute and users migrating away from `srcs` for all of the above rules can switch to `specs` and `non_specs`.

gregmagolan:
We could also have a brief period of time where specs can still be supplied via `srcs` and warn users that this feature will be removed in the future. This will give users a chance to migrate their targets. Once the feature is removed, `srcs` & `deps` will be for non-spec files and `specs` is for spec files.

gregmagolan:
[...]

PR #543, opened 2019-02-12
Introduce @bazel/jasmine package

alexeagle:
This will eventually replace the built-in jasmine_node_test rule.
It has the advantage that it brings a transitive dep on the jasmine npm package


packages/jasmine/index.bzl
```diff
@@ -0,0 +1,6 @@
+# Public API surface re-exports
```
gregmagolan: @bazel/karma and @bazel/typescript use defs.bzl. Should be consistent. I prefer index.bzl myself since its more node-ish. I think we should change to index.bzl for @bazel/karma & @bazel/typescript. Now's a good time since we've just renamed the workspaces.

gregmagolan: We could have both defs.bzl and index.bzl for those for a few releases before removing the defs..

alexeagle:
Greg suggests dealing with
https://github.com/bazelbuild/rules_nodejs/issues/513
while we're making a breaking change.

WORKSPACE
```diff
@@ -37,6 +37,11 @@ local_repository(
     path = "internal/e2e/packages",
 )
 
+local_repository(
```
gregmagolan: This is not being used yet. `/packages` could get added to the root `.bazelignore` and we don't need this local repository yet.

packages/jasmine/index.bzl
alexeagle: sounds good

Issue #312, opened 2018-09-07
Forked processes are unable to resolve node_modules/


lexor90:
As reported in #304 

> Using cluster.fork() breaks the resolution, as the child process seems to have the original node resolver.

Currently, if a process spawns other processes these don't get the patched `node_loader.js`. It means that every time you fork a process it will start failing because it cannot resolve any `require('xyz')`.

An example use case might be a web server where you `cluster`ize the process sharing a single port (the master waits for connections, workers do the request handling - in node this is so easy that it's very common). This can be worked around starting node in a single process (no master-worker architecture) and then use some orchestrator to have multiple copies of the process (but of course you need an external load balancer, which might not be an option for everyone).

Another valid example might be a process willing to spawn separated processes for heavy data calculations to avoid blocking the main thread and have the master to respond to RPC calls.

alexeagle:
@gregmagolan and I discussed how doing as suggested in #315 would fix this.

We could change our approach to have an action that lays out a node_modules tree with a composition of node_modules from the local repository, remote repositories, and JS outputs of rules. Then at runtime we'd only need to point nodejs at this output directory and could use vanilla nodejs resolution. It would move the complexity from runtime to build-time.

I think we'll do an investigation of this once we're done landing the fine-grained node modules everywhere and confirm the performance improvements.

nimerritt:
@alexeagle I did something very similarly internally at work but the performance was terrible since I needed to construct the node_modules layout for every `ts_module` (similar to ts_library) target in order to have TypeScript resolve everything since I couldn't figure out a good way to re-root a tree. See https://groups.google.com/forum/#!searchin/bazel-discuss/tree$20rooted%7Csort:date/bazel-discuss/JbA19d3vBis/Uv9j4TOwEQAJ In the end, I went back to using the module_mapping functionality provided by the `rules_nodejs`. I could have used the path mapping of the tsconfig.json file to resolve this partly, but wanted to keep the behaviour consistent with the runtime behaviour.

I'd be good to see the custom `require` code go away in any case. I'd be curious to know if you come up with a good solution to this!

lewish:
@nimerritt could you elaborate on your workaround regarding using module_mapping?

lewish:
For anyone else coming across this, I have a workaround.

[...]

PR #553, opened 2019-02-19
feat(@bazel/jasmine): add coverage reporting

Toxicable:
## PR Checklist

## PR Type

jasmine_node_test does not report code coverage

## What is the new behavior?
jasmine_node_test does reports code coverage

- [ ] Yes
- [x] No

Example oputput:
```
:~/Documents/rules_nodejs/packages/jasmine$ bazel run test:coverage
INFO: Invocation ID: c4e9cd85-aad8-4fad-9537-71c8a665861c
INFO: Analysed target //test:coverage (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //test:coverage up-to-date:
  bazel-bin/test/coverage_loader.js
  bazel-bin/test/coverage.sh
INFO: Elapsed time: 0.190s, Critical Path: 0.01s
INFO: 0 processes.
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //test:coverage
-----------------------------------------------------------------------------
Randomized with seed 34724
Started
.


1 spec, 0 failures
Finished in 0.006 seconds
Randomized with seed 34724 (jasmine --random=true --seed=34724)

=============================== Coverage summary ===============================
Statements   : 0% ( 0/48 )
Branches     : 0% ( 0/20 )
Functions    : 0% ( 0/9 )
Lines        : 0% ( 0/47 )
```

### DIFF
--- a/packages/jasmine/src/jasmine_node_test.bzl
+++ b/packages/jasmine/src/jasmine_node_test.bzl
@@ -29,6 +29,7 @@ def jasmine_node_test(
         expected_exit_code = 0,
         tags = [],
         jasmine = Label("@npm//@bazel/jasmine"),
+        coverage = False,
         **kwargs):
     """Runs tests in NodeJS using the Jasmine test runner.
 
@@ -42,6 +43,7 @@ def jasmine_node_test(
       expected_exit_code: The expected exit code for the test. Defaults to 0.
       tags: bazel tags applied to test
       jasmine: a label providing the jasmine dependency
+      coverage: Enables code coverage collection and reporting
       **kwargs: remaining arguments are passed to the test rule
     """
     devmode_js_sources(
@@ -52,15 +54,22 @@ def jasmine_node_test(
     )
 
     all_data = data + srcs + deps + [jasmine]
+
     all_data += [":%s_devmode_srcs.MF" % name]
     all_data += [Label("@bazel_tools//tools/bash/runfiles")]
     entry_point = "@bazel/jasmine/src/jasmine_runner.js"
+
+    templated_args = ["$(location :%s_devmode_srcs.MF)" % name]
+    if coverage:
+        templated_args = templated_args + ["--coverage"]
+    else:
+        templated_args = templated_args + ["--nocoverage"]
 
     nodejs_test(
         name = name,
         data = all_data,
         entry_point = entry_point,
-        templated_args = ["$(location :%s_devmode_srcs.MF)" % name],
+        templated_args = templated_args,
         testonly = 1,
         expected_exit_code = expected_exit_code,
         tags = tags,
--- a/packages/jasmine/src/jasmine_runner.js
+++ b/packages/jasmine/src/jasmine_runner.js
@@ -1,4 +1,5 @@
 const fs = require('fs');
+const path = require('path');
 
 let jasmineCore = null
 let JasmineRunner = require('jasmine/lib/jasmine');
@@ -55,26 +56,51 @@ if (TOTAL_SHARDS) {
 // the maximum (See: https://nodejs.org/api/errors.html#errors_error_stacktracelimit)
 Error.stackTraceLimit = Infinity;
 
+const IS_TEST_FILE = /[^a-zA-Z0-9](spec|test)\.js$/i;
+const IS_NODE_MODULE = /\/node_modules\//
+
 function main(args) {
   if (!args.length) {
     throw new Error('Spec file manifest expected argument missing');
   }
+  // first args is always the path to the manifest
   const manifest = require.resolve(args[0]);
+  // second is always a flag to enable coverage or not
+  const coverageArg = args[1];
+  const enableCoverage = coverageArg === '--coverage';
+
   // Remove the manifest, some tested code may process the argv.
-  process.argv.splice(2, 1)[0];
+  // Also remove the --coverage flag
+  process.argv.splice(2, 2)[0];
+
+  // the relative directory the coverage reporter uses to find anf filter the files
+  const cwd = process.cwd()
 
   const jrunner = new JasmineRunner({jasmineCore: jasmineCore});
-  fs.readFileSync(manifest, UTF8)
-      .split('\n')
-      .filter(l => l.length > 0)
+  const allFiles = fs.readFileSync(manifest, UTF8)
+                       .split('\n')
+                       .filter(l => l.length > 0)
+                       // Filter out files from node_modules
+                       .filter(f => !IS_NODE_MODULE.test(f))
+
+  const sourceFiles = allFiles
+                          // Filter out all .spec and .test files so we only report
+                          // coverage against the source files
+                          .filter(f => !IS_TEST_FILE.test(f))
+                          // the jasmine_runner.js gets in here as a file to run
+                          .filter(f => !f.endsWith('jasmine_runner.js'))
+                          .map(f => require.resolve(f))
+                          // the reporting lib resolves the relative path to our cwd instead of
+                          // using the absolute one so match it here
+                          .map(f => path.relative(cwd, f))
+
+  allFiles
       // Filter here so that only files ending in `spec.js` and `test.js`
       // are added to jasmine as spec files. This is important as other
       // deps such as "@npm//typescript" if executed may cause the test to
       // fail or have unexpected side-effects. "@npm//typescript" would
       // try to execute tsc, print its help, and process.exit(1)
-      .filter(f => /[^a-zA-Z0-9](spec|test)\.js$/i.test(f))
-      // Filter out files from node_modules that match test.js or spec.js
-      .filter(f => !/\/node_modules\//.test(f))
+      .filter(f => IS_TEST_FILE.test(f))
       .forEach(f => jrunner.addSpecFile(f));
 
   var noSpecsFound = true;
@@ -87,10 +113,53 @@ function main(args) {
   // so we need to add it back
   jrunner.configureDefaultReporter({});
 
+
+  let covExecutor;
+  let covDir;
+  if (enableCoverage) {
+    // lazily pull these deps in for only when we want to collect coverage
+    const crypto = require('crypto');
+    const Execute = require('v8-coverage/src/execute');
+
+    // make a tmpdir inside our tmpdir for just this run
+    covDir = path.join(process.env['TEST_TMPDIR'], String(crypto.randomBytes(4).readUInt32LE(0)));
+    covExecutor = new Execute({include: sourceFiles, exclude: []});
+    covExecutor.startProfiler();
+  }
+
   jrunner.onComplete((passed) => {
     let exitCode = passed ? 0 : BAZEL_EXIT_TESTS_FAILED;
     if (noSpecsFound) exitCode = BAZEL_EXIT_NO_TESTS_FOUND;
-    process.exit(exitCode);
+
+    if (enableCoverage) {
+      const Report = require('v8-coverage/src/report');
+      covExecutor.stopProfiler((err, data) => {
+        if (err) {
+          console.error(err);
+          process.exit(1);
+        }
+        const sourceCoverge = covExecutor.filterResult(data.result);
+        // we could do this all in memory if we wanted
+        // just take a look at v8-coverage/src/report.js and reimplement some of those methods
+        // but we're going to have to write a file at some point for bazel coverage
+        // so may as well support it now
+        // the lib expects these paths to exist for some reason
+        fs.mkdirSync(covDir);
+        fs.mkdirSync(path.join(covDir, 'tmp'));
+        // only do a text summary for now
+        // once we know what format bazel coverage wants we can output
+        // lcov or some other format
+        const report = new Report(covDir, ['text-summary']);
+        report.store(sourceCoverge);
+        report.generateReport();
+
+        process.exit(exitCode);
+      });
+    } else {
+      process.exit(exitCode);
+    }
+
+    
   });
 
   if (TOTAL_SHARDS) {
--- a/packages/jasmine/test/BUILD.bazel
+++ b/packages/jasmine/test/BUILD.bazel
@@ -46,3 +46,17 @@ jasmine_node_test(
     tags = ["manual"],
     deps = ["//:jasmine_runner"],
 )
+
+jasmine_node_test(
+    name = "coverage_test",
+    srcs = [
+        "coverage.spec.js",
+        "coverage_source.js",
+    ],
+    coverage = True,
+    jasmine = "@npm//jasmine",
+    deps = [
+        "//:jasmine_runner",
+        "@npm//v8-coverage",
+    ],
+)
--- a/packages/jasmine/test/coverage_source.js
+++ b/packages/jasmine/test/coverage_source.js
@@ -1,0 +1,9 @@
+function isString(input) {
+  if (typeof input === 'string') {
+    return true;
+    } else {
+        return false;
+    }
+}
+
+exports.isString = isString;
--- a/e2e/jasmine/BUILD.bazel
+++ b/e2e/jasmine/BUILD.bazel
@@ -14,3 +14,12 @@ jasmine_node_test(
         "@npm//zone.js",
     ],
 )
+
+jasmine_node_test(
+    name = "coverage_test",
+    srcs = [
+        "coverage.spec.js",
+        "coverage_source.js",
+    ],
+    coverage = True,
+)
--- a/e2e/jasmine/coverage.spec.js
+++ b/e2e/jasmine/coverage.spec.js
@@ -1,0 +1,10 @@
+const {isString} = require('./coverage_source');
+
+describe('coverage function', () => {
+  it('should cover one branch', () => {
+    expect(isString(2)).toBe(false);
+  });
+  it('should cover the other branch', () => {
+    expect(isString('some string')).toBe(true);
+  });
+});
--- a/e2e/jasmine/coverage_source.js
+++ b/e2e/jasmine/coverage_source.js
@@ -1,0 +1,9 @@
+function isString(input) {
+  if (typeof input === 'string') {
+    return true;
+  } else {
+    return false;
+  }
+}
+
+exports.isString = isString;
--- a/packages/jasmine/package.json
+++ b/packages/jasmine/package.json
@@ -17,7 +17,8 @@
     ],
     "main": "index.js",
     "dependencies": {
-        "jasmine": "~3.3.1"
+        "jasmine": "~3.3.1",
+        "v8-coverage": "1.0.8"
     },
     "bazelWorkspaces": {
         "npm_bazel_jasmine": {
--- a/packages/jasmine/test/coverage.spec.js
+++ b/packages/jasmine/test/coverage.spec.js
@@ -1,0 +1,10 @@
+const { isString } = require('./coverage_source');
+
+describe('coverage function', () => {
+  it('should cover one branch', () => {
+    expect(isString(2)).toBe(false);
+  });
+  it('should cover the other branch', () => {
+    expect(isString('some string')).toBe(true);
+  });
+});
--- a/packages/jasmine/yarn.lock
+++ b/packages/jasmine/yarn.lock
@@ -2,6 +2,45 @@
 # yarn lockfile v1
 
 
+ansi-regex@^2.0.0:
+  version "2.1.1"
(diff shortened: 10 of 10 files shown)


Toxicable:
While we do have it working so far I'm running into an issue where istanbul is istrumenting files that aren't apart of the sources.
Since we have a well define list of soruces that should be instrumented I think the best option would be to see if istanbul can implement an includes api rather than trying to use thei excludes api.

Without any exlucdes it'll auto discover these files, producing an incorrect sode coverage report.
```
"/home/f.wiles/Documents/rules_nodejs/third_party/github.com/buffer-from/index.js
/home/f.wiles/Documents/rules_nodejs/third_party/github.com/source-map-support/source-map-support.js
/home/f.wiles/Documents/rules_nodejs/third_party/github.com/source-map/lib/array-set.js
/home/f.wiles/Documents/rules_nodejs/third_party/github.com/source-map/lib/base64-vlq.js
/home/f.wiles/Documents/rules_nodejs/third_party/github.com/source-map/lib/base64.js
/home/f.wiles/Documents/rules_nodejs/third_party/github.com/source-map/lib/binary-search.js
/home/f.wiles/Documents/rules_nodejs/third_party/github.com/source-map/lib/mapping-list.js
/home/f.wiles/Documents/rules_nodejs/third_party/github.com/source-map/lib/quick-sort.js
/home/f.wiles/Documents/rules_nodejs/third_party/github.com/source-map/lib/source-map-consumer.js
/home/f.wiles/Documents/rules_nodejs/third_party/github.com/source-map/lib/source-map-generator.js
/home/f.wiles/Documents/rules_nodejs/third_party/github.com/source-map/lib/source-node.js
/home/f.wiles/Documents/rules_nodejs/third_party/github.com/source-map/lib/util.js
/home/f.wiles/Documents/rules_nodejs/third_party/github.com/source-map/source-map.js
/home/f.wiles/.cache/bazel/_bazel_f.wiles/6b62bb60168067291695fe6dd86d865d/external/nodejs/bin/yarnpkg/bin/yarn.js
/home/f.wiles/.cache/bazel/_bazel_f.wiles/6b62bb60168067291695fe6dd86d865d/external/nodejs/bin/yarnpkg/lib/cli.js
/home/f.wiles/.cache/bazel/_bazel_f.wiles/6b62bb60168067291695fe6dd86d865d/external/nodejs/bin/yarnpkg/lib/v8-compile-cache.js
/home/f.wiles/Documents/rules_nodejs/packages/jasmine/src/jasmine_runner.js"
```

alexeagle:
Seems like Istanbul hacking might not be needed anymore

> In Node.js >=10.10.0 you can now set the environment variable NODE_V8_COVERAGE to a directory, this will result in V8 coverage reports being output in this location.
https://medium.com/the-node-js-collection/rethinking-javascript-test-coverage-5726fb272949

We could try that?

Toxicable:
Oh wow, that looks awesome.
Yeah we'll have a look into that and see how we can intergrate it with the test runner. 

It does mean we'll need to push a requirement on the nodejs version.
Could be done as a warning or maybe a flag to enable coverage collection + an error if the version isn't right.

Toxicable:
@alexeagle  Started inital work on v8-coverage.
the way im planning on doing it is by adding another .js file to the jasmine rule: `jasmine_runner_host.js`, this file will be the new entry point and will fork itself so it can set the `NODE_V8_COVERAGE` flag to a known directory for processing after tests and been executed.

However im running into issues where the sub process (jasmine_runner.js) is unable ro resolve "bazel" nodejs paths ie: `Cannot find module 'npm_bazel_jasmine/test/foo_test.js'` any ideas why this might happen?
Or if there's a better way to set the environment variable to a dir we know and and work on later on

Toxicable:
I think this is ready for an inital review now

From looking at https://github.com/bazelbuild/rules_nodejs/issues/312
I was able to execute the forked script, but it feels pretty hacky and I'm not sure if it'll work in other situations.

```
Executing tests from //test:coverage_test
-----------------------------------------------------------------------------
Randomized with seed 23022
Started
.


1 spec, 0 failures
Finished in 0.014 seconds
Randomized with seed 23022 (jasmine --random=true --seed=23022)

=============================== Coverage summary ===============================
Statements   : 90% ( 9/10 )
Branches     : 60% ( 3/5 )
Functions    : 100% ( 1/1 )
Lines        : 90% ( 9/10 )
```

Toxicable:
So looks like the fork script hack only works when you run the test with `bazel run` if you use bazel test it fails to resolve the path.
Any advice how I can hook up these paths?

alexeagle:
I don't see why you need to fork a process. We have control over the nodejs options passed to the original nodejs process that starts the tests. Can we set the flag there?

Toxicable:
@alexeagle Sure, all we need is to set an environment variable.
I couldn't find how to do that with a macro setup, meaning, at a guess we'd need to translate this into a Rule rather than a Macro?
Also this was the easiest way I knew just to see if it would be possible - which we proved true

Toxicable:
Quick update here:

I tried doing this without using a forked script but it would appear the coverage file is only written once the node process exits. Meaning this must be done from some other process somehow.

Toxicable:
Ignore what I said above, turns out I was all wrong, you can produce coverage without a new process.
I switched to a new lib: `v8-coverage` which internally starts the Nodejs inspector module.
Can be previewed with `bazel run test:coverage_test`, gives the output below.

```
Fabians-MacBook-Air:jasmine fabianwiles$ bazel run test:coverage_test
...
Executing tests from //test:coverage_test
-----------------------------------------------------------------------------
Debugger listening on ws://127.0.0.1:57032/c20d1245-cce8-42e5-b956-437c8b21dddf
For help, see: https://nodejs.org/en/docs/inspector
Randomized with seed 12938
Started
.
1 spec, 0 failures
Finished in 0.008 seconds
Randomized with seed 12938 (jasmine --random=true --seed=12938)

=============================== Coverage summary ===============================
Statements   : 80% ( 8/10 )
Branches     : 75% ( 3/4 )
Functions    : 100% ( 0/0 )
Lines        : 80% ( 8/10 )
```

Globegitter:
@toxicable if you want to be able to set the env variable before the target starts there is a generic solution, you can use the `command` rule from here: https://github.com/atlassian/bazel-tools/blob/master/multirun/README.md

Plus I also opened an issue: https://github.com/bazelbuild/bazel/issues/7364 as I (and some of our devs on the team) expected this to be part of bazel in the same way as one can set args generically.

Globegitter:
Also are there any plans to integrate this with the `bazel coverage` command?

Toxicable:
@Globegitter Ahhh, thanks for the heads up on the `command` rule.
However, for this situation we'd still need to run some JS after the `jasmine_runner` has exited, since it would appear that using  the `V8_COVERAGE` environment variable only writes the coverage once it exits the process. So we would still have to run some JS host that forks into `jasmine_runner`.
I think the current solution should work, aslong as there's no weird interactions with other apis due to using the inspector api.

As for  `bazel coverage` - I'd definitely like to integrate with it and from here it should easy to do so really since we can go to any coverage format, I just don't know how, I havn't been able to find any docs on how to do so.
So for now using `text-summary` will work as a stop gap, I'm not sure if we want to merge it in this state without `bazel coverage` integration though.

alexeagle:
I'll discuss with the Bazel team this week about adding proper `bazel coverage` command. I'll leave it to @Globegitter and @gregmagolan about whether to merge this with a text summary or wait for guidance.

packages/jasmine/src/jasmine_node_test.bzl
```diff
@@ -64,6 +78,7 @@ def jasmine_node_test(
         templated_args = ["$(location :%s_devmode_srcs.MF)" % name],
         testonly = 1,
         expected_exit_code = expected_exit_code,
+        configuration_env_vars = ["NODE_V8_COVERAGE"],
```
Globegitter: Why do we expect this to be passed in by the user?

.bazelrc
```diff
@@ -1,12 +0,0 @@
-# Print test logs for failed tests
```
Globegitter: Why is this removed?

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
```
Globegitter: Maybe we can add this to gitignore

packages/jasmine/src/jasmine_runner.js
```diff
@@ -27,6 +33,9 @@ const BAZEL_EXIT_NO_TESTS_FOUND = 4;
 // the maximum (See: https://nodejs.org/api/errors.html#errors_error_stacktracelimit)
 Error.stackTraceLimit = Infinity;
 
+const IS_TEST_FILE = /[^a-zA-Z0-9](spec|test)\.js$/i;
+const IS_MODULE_MODULE = /\/node_modules\//
```
Globegitter: Did you mean to call it this, or maybe a typo?

Globegitter: The variable name I mean btw.

packages/jasmine/src/jasmine_node_test.bzl
```diff
@@ -51,8 +51,22 @@ def jasmine_node_test(
         tags = tags,
     )
 
-    all_data = data + srcs + deps + [jasmine]
-    all_data += [Label("//:src/jasmine_runner.js")]
+    runner_name = name + "." + "runner"
+
+    nodejs_binary(
```
Globegitter: Any particular reason to wrap this in a `nodejs_binary` compared to how it was before?

packages/jasmine/BUILD.bazel
```diff
@@ -18,6 +18,7 @@ package(default_visibility = ["//visibility:public"])
 
 exports_files([
     "src/jasmine_runner.js",
+    "src/jasmine_runner_host.js",
```
Globegitter: I don't see this file. Leftover?

Toxicable:
@Globegitter Thanks for the review. Will do a tidy up tonight.

Right now it's on no matter what. There is a small perf overhead ~20% with this.
Maybe we should provide an attribute to opt-in incase there's issues we havn't seen with it so far, then later on make a breaking change to switch to opt-out to encourage people to use the coverage report.

Im not sure what you mean by 
>regex for finding the coverage files to be configurable,

This is the part where we filter out the coverage collected from external source files
https://github.com/bazelbuild/rules_nodejs/pull/553/files#diff-b7f7068522dfd058878f39fba971463aR87
which is just a direct file match, I don't see why you'd want it to be configurable though, this will always pick up all user source files that were in the test, is there case where you'd only want a subset of that?

Globegitter:
Yeah I think we should have coverage turned off by default for now and then later we can switch it on. One way I can see it work is by adding a `coverage` parameter to the jasmine_node_test macro and if that is set you can pass a custom arg via `args` to the nodejs_test and then that can be checked in the runner.

Globegitter:
About the file match, yeah the static one you have works for idiomatic bazel usage, but I was thinking of code bases like our own where we have common js code in sloppy file group targets that have a bunch of files grouped together even if an app might just need a subset. So in that case we might want to be able to specify our own regex to filter out specific files. But as I said I think it is good to keep it simple and promote idiomatic use. So I think it makes sense to just leave it as is.

Toxicable:
@Globegitter Ah yes, I see your point, makes sense.

I've implemented the coverage flag and disabled it by default.
Your arguments idea works well here
You can re-enable it with `coverage = True` in your `jasmine_node_test`

One remaining point is that I'm using the dep `"@npm//v8-coverage"` - should we make a `npm_bazel_jasmine_deps` yarn_install to get this dep from instead?
I've noticed there's been a bit of change about where these deps are coming from lately so not sure on the best option.

packages/jasmine/src/jasmine_node_test.bzl
```diff
@@ -51,6 +53,13 @@ def jasmine_node_test(
         tags = tags,
     )
 
+    enableCoverage = False
```
Globegitter: no need for the `enableCoverage` flag, we can just use coverage directly below.

Globegitter:
I am also not familiar with the new process yet, but as far as I understand, once this is merged a new npm package has to be built and then one just adds the latest `@bazel/jasmine` to their package.json and will get the dependencies via normal npm resoultion. Maybe @alexeagle or @gregmagolan can fill in on how the release / dependency process is currently working.

gregmagolan:
Yes @Globegitter you are correct. `/packages/jasmine` gets packaged up by `//packages/jasmine:npm_package` and released as `@bazel/jasmine`. That npm package installs the `npm_bazel_jasmine` workspace via the generated `@npm://install_bazel_workspaces.bzl` and it also pulls in transitive npm deps such as the `jasmine` npm package.

gregmagolan:
For CI & developing & testing locally I made a number shell scripts that build the packages before running the e2e test and examples. You can call `./scripts/build_packages.sh jasmine` to build the jasmine package and its is copied out of the `bazel-out` folder into `/dist/jasmine`. The e2e tests and examples then link to this package in their package.json.

gregmagolan:
So, if `@bazel/jasmine` requires the `v8-coverage` npm package then that just needs to be added as a dependency to `/packages/jasmine/package.json` and it will be pulled into a user's node_modules transitively when the install `@bazel/jasmine`.

Hmmmm. I think I need to spend some time to document this in `DEVELOPING.md` :)

packages/jasmine/src/jasmine_node_test.bzl
Toxicable: Ah, does `1` get converted to a boolean?
I've seen in some places `testonly = 1` and `testonly = True` being treated as equivilent

Toxicable: it would appear not, I think it's reasonable to ask for just `True` or `False` as values here though

Toxicable:
Alrighty, I think we're good to go here now.
I added a bit of a summary to the commit message description of this feature and it's current status.
Also debugging still appears to work even with the internal call the inspector API

packages/jasmine/src/jasmine_node_test.bzl
```diff
@@ -51,7 +55,13 @@ def jasmine_node_test(
         tags = tags,
     )
 
+    # append on any user args to the end
+    args = ["--coverage=%s" % coverage] + args
+
     all_data = data + srcs + deps + [jasmine]
+
+    # code coverage reporting dependancy
+    all_data += ["@npm//v8-coverage"]
```
gregmagolan: You can remove this line. Pulling in `@npm//@bazel/jasmine` (the `jasmine` attribute) is sufficient to also pull in `v8-coverage` since its a transitive dep. Also, we don't want to assume `@npm` is the users node_modules workspace. That's why we have the `jasmine` attribute so a user can set it to `@foobar//@bazel/jasmine` if necessary.

packages/jasmine/test/BUILD.bazel
```diff
@@ -30,3 +30,13 @@ jasmine_node_test(
...
 )
+
+jasmine_node_test(
+    name = "coverage_test",
+    srcs = [
+        "coverage.spec.js",
+        "coverage_source.js",
+    ],
+    coverage = True,
+    jasmine = "@npm//jasmine",
```
gregmagolan: will need to rebase and add `deps = ["//:jasmine_runner"],` to this target

packages/jasmine/src/jasmine_runner.js
```diff
@@ -138,5 +207,5 @@ function getAllSpecs(jasmineEnv) {
 }
 
 if (require.main === module) {
-  process.exitCode = main(process.argv.slice(2));
+    process.exitCode = main(process.argv.slice(2));
```
gregmagolan: nit: extra whitespace

packages/jasmine/src/jasmine_node_test.bzl
```diff
@@ -51,7 +55,13 @@ def jasmine_node_test(
         tags = tags,
     )
 
+    # append on any user args to the end
+    args = ["--coverage=%s" % coverage] + args
```
gregmagolan: You should be able to just add this argument to `templates_args` and avoid the `args` altogether. Also, looking for `--coverage=True` seems pretty brittle as a user could set it to `1`. This would be better:

```
templated_args = ["$(location :%s_devmode_srcs.MF)" % name]
if coverage:
  templated_args = templated_args + ["--coverage"]
else:
  templated_args = templated_args + ["--nocoverage"]
```

and then just look for an argument matching `--coverage` in the `jasmine_runner.js`

packages/jasmine/test/BUILD.bazel
gregmagolan: Can you also add a smoke test to `/e2e/jasmine`? It would test the `v8-coverage` can be required in user land when using the `@bazel/jasmine` package.
A test which verifies the report works would be good but I think we can merge this without one and add that in a follow up PR.

packages/jasmine/src/jasmine_runner.js
```diff
@@ -1,7 +1,13 @@
 const fs = require('fs');
+const path = require('path');
+const crypto = require('crypto');
```
gregmagolan: crypo & Execute & Report requires can be done within the `if (enableCoverage)` blocks so in the no coverage case they are no required at all

packages/jasmine/test/coverage_source.js
```diff
@@ -0,0 +1,9 @@
+function isString(input) {
+    if(typeof input === 'string') {
```
gregmagolan: nit: `if (`

packages/jasmine/src/jasmine_node_test.bzl
Toxicable: I'm running into issues with this one.
If I remove it then im not able to run the tests inside the jasmine package anymore, since nothing is pulling `@npm//v8-coverage` anymore

Globegitter: Yeah I would assume it is just an issue of getting the right package locally now so it pulls in the right deps automatically. I will take a look myself if I spot anything obvious.

gregmagolan: I think you just need to add `@npm//v8-coverage` explicitly to the deps (or data) of the test within `packages/jasmine/test/BUILD.bazel`:
```
jasmine_node_test(
    name = "coverage_test",
    srcs = [
        "coverage.spec.js",
        "coverage_source.js",
    ],
    coverage = True,
    jasmine = "@npm//jasmine",
    deps = [
        "//:jasmine_runner",
        "@npm//v8-coverage,
    ],
)
```

Toxicable: ah! easy as, sorted

packages/jasmine/test/BUILD.bazel
Toxicable: done

Toxicable:
@gregmagolan Thanks for the review, I've addressed all your points, let me know if there's anything else.
With the exception of adding an e2e test for the text-summary output. Since in my next PR ill be turning that off `text-summary` and going to lcov in an attempt to intergrade with bazel coverage

Globegitter:
@gregmagolan @Toxicable  One thing I forgot is documentation, I assume skydoc should pick up the added comment about coverage automatically, but is there anything automatically updating the docs? Looking at https://bazelbuild.github.io/rules_nodejs/jasmine_node_test/jasmine_node_test.html#jasmine_node_test I can not see the new `coverage` attribute commented.

gregmagolan:
> @gregmagolan @Toxicable One thing I forgot is documentation, I assume skydoc should pick up the added comment about coverage automatically, but is there anything automatically updating the docs? Looking at https://bazelbuild.github.io/rules_nodejs/jasmine_node_test/jasmine_node_test.html#jasmine_node_test I can not see the new `coverage` attribute commented.

We typically update the docs with each release so the docstring from jasmine_node_test.bzl will get updated at that point:

```
def jasmine_node_test(
        ...
        **kwargs):

    To debug the test, see debugging notes in `nodejs_test`.

    Args:
      name: name of the resulting label
      srcs: JavaScript source files containing Jasmine specs
      data: Runtime dependencies which will be loaded while the test executes
      deps: Other targets which produce JavaScript, such as ts_library
      expected_exit_code: The expected exit code for the test. Defaults to 0.
      tags: bazel tags applied to test
      jasmine: a label providing the jasmine dependency
      coverage: Enables code coverage collection and reporting
      **kwargs: remaining arguments are passed to the test rule
    """
```

which includes the short description of `coverage`.

@alexeagle Has done more work on docs and some things have changed since moving to separate npm packages such as `@npm//jasmine`. I don't know if we have the docs from each package hosted at `bazelbuild.github.io` yet. The docs there may still be from the legacy `/internal/jasmine_node_test` rule.

PR #559, opened 2019-02-22
Add e2e tests for @bazel/jasmine, @bazel/karma & @bazel/typescript npm packages

gregmagolan:
* some cleanup included
* `@npm_bazel_typescript//:index.bzl` and `@npm_bazel_karma//:index.bzl`  as the public API entry points. I think this is a better default for all the packages in this repo. Its more like index.js. `//:defs.bzl` is still there for backward compat for now.


e2e/jasmine/WORKSPACE
```diff
@@ -0,0 +1,34 @@
...
+workspace(name = "e2e_jasmine")
+
+local_repository(
+    name = "build_bazel_rules_nodejs",
+    path = "../../bazel-bin/local_testing_package",
+)
+
+load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories", "yarn_install")
+
+node_repositories()
```
alexeagle: nit: no longer needed

packages/typescript/BUILD.bazel
```diff
@@ -47,16 +47,19 @@ npm_package(
...
         "version.bzl",
+        "//devserver:npm_package_assets",
         "//internal:npm_package_assets",
         "//ts_auto_deps:npm_package_assets",
     ],
     packages = [
         "@build_bazel_rules_typescript//:npm_bazel_typescript_package",
     ],
     replacements = {
+        "(#|\/\/)\\s+BEGIN-DEV-ONLY[\\w\W]+?(#|\/\/)\\s+END-DEV-ONLY": "",
```
alexeagle: we can just use the built-in `# BEGIN-INTERNAL` fences again now that we aren't in a repo that copybara excludes things from

packages/typescript/WORKSPACE
```diff
@@ -23,6 +15,17 @@ yarn_install(
     yarn_lock = "//:yarn.lock",
 )
 
+# Uncomment for local development
+# local_repository(
+#     name = "build_bazel_rules_typescript",
+#     path = "../../../rules_typescript",
+# )
+http_archive(
```
alexeagle: I wonder if we should just keep git since you only have to update the commit in one place rather than two

packages/typescript/devserver/BUILD.bazel
```diff
@@ -55,3 +47,11 @@ filegroup(
     tags = ["manual"],
     visibility = ["//visibility:public"],
 )
+
+# BEGIN-DEV-ONLY
```
alexeagle: I think we may as well leave these in the distributed file, it means fewer moving parts in deployment of this file, and doesn't cause a transitive dependency to leak

packages/typescript/WORKSPACE
gregmagolan: git_repository flaked out on me in CI when it worked locally so I switched at that point. don't know why but it errored out with not being able to find a commit hash

e2e/jasmine/WORKSPACE
gregmagolan: 👍

packages/typescript/BUILD.bazel
gregmagolan: ahhh.. but I think there is still some `BEGIN-DEV` fencing in the files vendored in from @build_bazel_rules_typescript//:npm_bazel_typescript_package?

packages/typescript/devserver/BUILD.bazel
gregmagolan: 👍

packages/typescript/BUILD.bazel
alexeagle: right...

packages/typescript/index.bzl
```diff
@@ -0,0 +1,35 @@
...
+
+check_rules_typescript_version = _check_rules_typescript_version
+ts_setup_workspace = _ts_setup_workspace
+ts_library = _ts_library
+ts_config = _ts_config
+ts_devserver = _ts_devserver
+
+ts_proto_library = _ts_proto_library
+# DO NOT ADD MORE rules here unless they appear in the generated docsite.
+# Run yarn skydoc to re-generate the docsite.
```
alexeagle: dangling ref to old docgen process, will have a bunch of little things to cleanup later

PR #575, opened 2019-02-22
Refactor /internal/test:node_modules into @npm + fixes

gregmagolan:
Also fixes a few minor bugs that came up during the refactoring:

* `yarn_install/npm_install` `generate_build_file.js` now treats every folder under node_modules as a package regardless of it having a package.json file or not. npm will treat them as such when resolving so we should too.
* jasmine_runner attempts to require jasmine-core from `jasmine/node_modules/jasmine-core` before falling back to `jasmine-core`. this is so that if `jasmine-core`, which is a transitive dep of `jasmine`, is not hoisted it can still be required with only the `@npm//jasmine` dep
* jasmine_node_test (in `@bazel/jasmine`) `jasmine` attribute change to `@npm//@bazel/jasmine` since this is the only guaranteed label. `@npm//jasmine` may not exists if it is not hoisted
* symlinked data files in yarn_install & npm_install cause issues if these are used as npm packages via `file://` on linux. copies are made of files instead


internal/e2e/rollup/BUILD.bazel
```diff
@@ -1,17 +1,6 @@
 load("@build_bazel_rules_nodejs//:defs.bzl", "jasmine_node_test")
 load("//:defs.bzl", "rollup_bundle")
 
-# The local package node_modules filegroup contains just one depenedency to test
-# importing from a non-WORKSPACE-root node_modules folder. This makes it more clear
```
alexeagle: are we losing some coverage of the coarse-grained deps?

internal/jasmine_node_test/jasmine_runner.js
```diff
@@ -8,7 +8,16 @@ try {
...
     // global.jasmine has been initialized which means a bootstrap script
     // has already required `jasmine-core` and called jasmineCore.boot()
-    jasmineCore = global.jasmineCore || require('jasmine-core');
+    jasmineCore = global.jasmineCore;
+    if (!jasmineCore) {
+      try {
+        // Try unhoisted jasmine-core first so that we don't
+        // need an @npm//jasmine-core dep in the unhoisted case.
+        jasmineCore = require('jasmine/node_modules/jasmine-core');
+      } catch (_) {
```
alexeagle: you could check if the error.code is the MODULE_NOT_FOUND one

internal/e2e/rollup/BUILD.bazel
gregmagolan: I don't think so in this case. That comment was from before fine grained deps existed and basically says that the rollup related deps don't need to be in the user's node_modules which is still the case in this test with fine grained dep. I updated it to just the depend on the minimum `@npm//hello` which is a better test.

internal/jasmine_node_test/jasmine_runner.js
gregmagolan: good call thanks

PR #584, opened 2019-02-26
Update rollup_bundle dependencies

alexeagle:
We now tree-shake @PURE-annotated code.

thanks @filipesilva

### DIFF
--- a/internal/e2e/rollup/foo.js
+++ b/internal/e2e/rollup/foo.js
@@ -6,6 +6,18 @@ import {name} from './bar';
 
 console.log(`${hello}, ${name} in ${fum}`);
 
+// Tests for @PURE annotations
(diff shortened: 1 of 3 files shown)


internal/e2e/rollup/foo.js
```diff
@@ -6,6 +6,18 @@ import {name} from './bar';
...
+/*@__PURE__*/
+console.log('side-effect');
+
+class Impure {
+  constructor() {
+    console.log('side-effect')
+  }
+}
+
+/*@__PURE__*/new Impure();
```
gregmagolan: No spec to verify that the bundle is as expected?

alexeagle: The spec is the golden files - they verify that these added lines didn't get retained

PR #587, opened 2019-02-28
Default runfiles to upwards directory as a last fallback

Globegitter:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

When wanting to run the nodejs_binary in a custom `sh_binary`, to have some custom setup or run multiple commands for integration tests for example, one just sees the error message `ERROR: cannot find @bazel_tools//tools/bash/runfiles:runfiles.bash` when executing the binary and it is not really obvious as to why the bash runfiles can not be found.

After digging further it turns out one has to explicitely set the `RUNFILES` env variable to `..` and I wonder if it is not possible to make this use-case work out of the box as this is what we expected.

simple BUILD file example to illustrate the wrapping:
```
nodejs_binary(
    name = "app", 
    entry_point = "__main__/app.js",
    data = ["app.js", "@npm//somedep"],
)

sh_binary(
    name = "app",
    srcs = ["my_script.sh"],
    data = [":app"],
)
```

## What is the new behavior?

Defaulting RUNFILES to `..` if neither `RUNFILES_MANIFEST_FILE` nor `RUNFILES_DIR` can be set in any other way.

I would expect this to not break any existing usages, as it only adds one further case to try for the runfiles dir. And if it is still not there it will still fail.

### DIFF
--- a/internal/node/node_launcher.sh
+++ b/internal/node/node_launcher.sh
@@ -31,7 +31,7 @@ if [[ ! -d "${RUNFILES_DIR:-/dev/null}" && ! -f "${RUNFILES_MANIFEST_FILE:-/dev/
       export RUNFILES_DIR="$0.runfiles"
     fi
 fi
-if [[ -f "${RUNFILES_DIR:-".."}/bazel_tools/tools/bash/runfiles/runfiles.bash" ]]; then
+if [[ -f "${RUNFILES_DIR:-..}/bazel_tools/tools/bash/runfiles/runfiles.bash" ]]; then
   source "${RUNFILES_DIR}/bazel_tools/tools/bash/runfiles/runfiles.bash"
 elif [[ -f "${RUNFILES_MANIFEST_FILE:-/dev/null}" ]]; then
   source "$(grep -m1 "^bazel_tools/tools/bash/runfiles/runfiles.bash " \
--- a/internal/node/node_launcher.sh
+++ b/internal/node/node_launcher.sh
@@ -31,7 +31,7 @@ if [[ ! -d "${RUNFILES_DIR:-/dev/null}" && ! -f "${RUNFILES_MANIFEST_FILE:-/dev/
       export RUNFILES_DIR="$0.runfiles"
     fi
 fi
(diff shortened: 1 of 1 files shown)


Globegitter:
Seems the buildkite tests are very flaky on master also so should be unrelated to my changes.

Globegitter:
@gregmagolan any chance to get this merged in or to get some feedback on this? Would love to be able to remove our internal workaround for this.

Globegitter:
I would also be happy to add an e2e example that utilises `sh_binary` to demonstrate and have a test for this.

gregmagolan:
Something seems fishy here. That bash runfiles init code is copy pasted from `https://github.com/bazelbuild/bazel/blob/master/tools/bash/runfiles/runfiles.bash` so I'm hesitant to modify it and its also duplicated in multiple places in the repo.

We do something similar in `e2e/ts_library/some_module`:

```
nodejs_binary(
    name = "bin",
    data = [
        ":main",
        ":some_module",
    ],
    entry_point = "e2e_ts_library/some_module/main.js",
)

sh_test(
    name = "module_load_test",
    srcs = ["module_load_test.sh"],
    data = [
        ":bin",
        "@bazel_tools//tools/bash/runfiles",
    ],
)
```

with `e2e/ts_library/some_module/module_load_test.sh` using the standard runfiles init code and then using `rlocation` to find launch the node process:

```
#!/bin/bash
set -e

# --- begin runfiles.bash initialization ---
...
# --- end runfiles.bash initialization ---

readonly OUT=$($(rlocation "e2e_ts_library/some_module/bin"))

if [ "$OUT" != "hello world" ]; then
  echo "Expected output 'hello world' but was $OUT"
  exit 1
fi
```

Does that work in your case?

Globegitter:
Interesting, so maybe something to raise in the bazel main repo. I have not tried the snippet you posted yet but I am sure it would fix it also. But my point is that this is not obvious in any way, that if I want to depend on a `nodejs_binary` and just execute it via `bazel run` that I need to provide some location/runflies logic. Even if it was documented, it increases the entry barrier for new developers for something that should be pretty simple.

I did open a PR in bazel for this: https://github.com/bazelbuild/bazel/pull/7774

PR #597, opened 2019-03-06
stardoc

alexeagle:


alexeagle:
Agreed, I should have commented:
it's a transition. typescript rules are the most urgent because a bunch of attributes are undocumentable today.

However for the parent workspace, we rely on repository rules to be documented, and stardoc doesn't do those yet.

I'll send followup PRs to migrate other packages as possible.

PR #603, opened 2019-03-08
Improve build and parallelize CI

gregmagolan:
CircleCI workflow time down to 6:30

No more dependency on angular/ngcontainer in CircleCI. The bazel version used in CI is the one listed in the main package.json file.

Documentation generating dependencies cleaned up and now only in `rules_nodejs_dev_dependences()` which each package calls.

Added new scripts for testing locally (some of which are also use by CI). DEVELOPING.md has instructions:

## Testing locally

This repository contains nested workspaces. To test all targets locally in the main workspace and in all nested workspaces run:

```
yarn test_all
```

To do a full clean run:

```
yarn clean_all
```

Other scripts allow you to test all or individual packages, e2e tests and examples. For example,

```
yarn test_packages_all
yarn test_packages typescript karma
yarn test_e2e_all
yarn test_e2e karma
yarn test_examples_all
yarn test_examples webapp
```


.circleci/config.yml
```diff
@@ -101,125 +173,185 @@ jobs:
...
+      - *init_environment
+      - *init_bazel
+      - *hide_node_and_yarn_local_binaries
+      - run: ./scripts/build_packages.sh jasmine
+      - persist_to_workspace:
+          root: ~/
+          paths:
+            - ./rules_nodejs/dist/npm_bazel_jasmine
+
+  build_labs_package:
```
alexeagle: next PR: I'd like to reduce number of steps to create a new package, maybe we can build all of these in one job in a loop?

scripts/clean_all.sh
```diff
@@ -0,0 +1,35 @@
...
+
+set -eu -o pipefail
+# -e: exits if a command fails
+# -u: errors if an variable is referenced before being set
+# -o pipefail: causes a pipeline to produce a failure return code if any command errors
+
+readonly RULES_NODEJS_DIR=$(cd $(dirname "$0")/..; pwd)
+cd ${RULES_NODEJS_DIR}
+
+echo_and_run() { echo "+ $@" ; "$@" ; }
```
alexeagle: why not just `set -x` to echo commands before they're run?

gregmagolan: I wanted to be selective on which ones are echoed and bracketing with `set -x` and `set +x` seemed noisier.

.circleci/config.yml
gregmagolan: yup. that would be ideal. we could shard it like that. the only complication and why I didn't in this PR was that the karma package depends on typescript package being build first.

Issue #560, opened 2019-02-01
Fine-grained dependencies not passed through to rollup_bundle()


jmsdncn:
# 🐞 bug report

### Affected Rule

`ts_library()` possibly?  In my reproduction, it is the only `bazelbuild/rules_typescript` rule being used.

### Is this a regression?

I have only checked this behavior with v0.22.1.

### Description

When using `rollup_bundle()` (from `bazelbuild/rules_nodejs`) on the output of a `ts_library()` rule, I need to either re-specify the (full, transitive) set of fine-grained dependencies used by the `ts_library()` rule, or `node_modules = "@npm//:node_modules"` (which is supposed to be deprecated, or at least discouraged).

I expected that the transitive set of fine-grained dependencies would be aggregated by `ts_library()`, and made available to `rollup_bundle()`, but that does not appear to be the case.  As it is impractical to repeat the full list of dependencies in calling `rollup_bundle()`, I seem to have to fall back on using the full `@npm//:node_modules` filegroup.

## 🔬 Minimal Reproduction

https://gist.github.com/jmsdncn/8e21f5842acb51eb131bc9fc8605a22a

## 🔥 Exception or Error

```
$ bazel build //:bundle
INFO: Invocation ID: 4adabf83-5c54-499c-9dae-038ef7841983
INFO: Analysed target //:bundle (0 packages loaded, 1 target configured).
INFO: Found 1 target...
ERROR: /usr/home/james/pg/bazel-typescript-npm-dep/BUILD:9:1: SkylarkAction bundle.es6.js failed (Exit 1) rollup failed: error executing command bazel-out/host/bin/external/build_bazel_rules_nodejs/internal/rollup/rollup --config bazel-out/freebsd-fastbuild/bin/_bundle.rollup.conf.js --output.file bazel-out/freebsd-fastbuild/bin/bundle.es6.js ... (remaining 4
argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
[!] (Plugin at pos 2 plugin) Error: Could not resolve import 'lit-html' from '/usr/home/james/.cache/bazel/_bazel_james/88a369858ad9d1bd75821a00db0f8b09/sandbox/processwrapper-sandbox/1/execroot/demo/bazel-out/freebsd-fastbuild/bin/bundle.es6/lib.js'

WARNING: Due to a breaking change in rules_nodejs 0.13.2, target //:bundle
must now declare either an explicit node_modules attribute, or
[...]

PR #612, opened 2019-03-11
Add e2e tests/examples of transitive npm deps for ts_library - nodejs_binary interop

Globegitter:
## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Tests for odd behaviour of ts_library <-> nodejs_binary (and rollup_bundle) interaction discussed and tested here: https://github.com/bazelbuild/rules_nodejs/pull/283#issuecomment-471349968

It seems that under certain (or all?) circumstances `nodejs_binary` as well as `rollup_bundle` is not collecting transitive npm dependencies.

Issue Number: N/A

## What is the new behavior?

For now this PR just adds tests but is not fixing the behaviour yet. I will also have a look at fixing the behaviour.

- [ ] Yes
- [ ] No

## Other information

cc @gregmagolan I was not sure where to best test this and if there is a better way of maybe using `nodejs_test` for this rather than `nodejs_binary` but let me know and happy to move this, clean this up etc.


Globegitter:
Related #560

Globegitter:
Should fix #560 as well as the issue discussed here: https://github.com/bazelbuild/rules_nodejs/pull/283#issuecomment-471349968

@gregmagolan would love some feedback on this.

packages/typescript/internal/build_defs.bzl
```diff
@@ -250,9 +250,11 @@ def _ts_library_impl(ctx):
         # since they don't have the required providers.
         # They were added to the action inputs for tsc_wrapped already.
         # strict_deps checking currently skips node_modules.
+        # Make sure we are including dependencies with transitive node module
+        # dependencies.
         # TODO(alexeagle): turn on strict deps checking when we have a real
         # provider for JS/DTS inputs to ts_library.
-        deps = [d for d in ctx.attr.deps if not NodeModuleInfo in d],
+        deps = [d for d in ctx.attr.deps if not NodeModuleInfo in d or (NodeModuleInfo in d and d[NodeModuleInfo].transitive == True)],
```
alexeagle: note, this line needs to go away anyway because it's causing https://github.com/bazelbuild/rules_typescript/issues/381

I'll rebase around this though, glad you've added tests so I know I don't break it

internal/common/node_module_info.bzl
```diff
@@ -35,7 +36,13 @@ def _collect_node_modules_aspect_impl(target, ctx):
 
     if hasattr(ctx.rule.attr, "tags") and "NODE_MODULE_MARKER" in ctx.rule.attr.tags:
         nm_wksp = target.label.workspace_root.split("/")[1] if target.label.workspace_root else ctx.workspace_name
-        return [NodeModuleInfo(workspace = nm_wksp)]
+        return [NodeModuleInfo(workspace = nm_wksp, transitive = False)]
```
alexeagle: I would expect transitive to have a similar meaning to how a depset has a transitive argument (the depsets to hang off this one in the tree)
Can we find another name for this attribute? Really it's like a NodeModuleInfo for something that's not a node_module. Maybe it should just be a different provider instead of overloading this one?

alexeagle:
Does this mean users no longer have to repeat deps on their rollup_bundle rule if a transitively included ts_library had the dep? That sounds great! We'll need to clean up stuff downstream.

Globegitter:
Yep exactly this is what this is aimed at fixing, as it is also necessary for a correct implementation of #283

internal/common/node_module_info.bzl
Globegitter: Yeah maybe it does make sense to have a different provider as the attribute thing did feel a bit awkard. I'll have a bit of a think and another go at it.

internal/common/node_module_info.bzl
```diff
@@ -26,6 +26,7 @@
 NodeModuleInfo = provider(
     doc = "This provider contains information about npm dependencies installed with yarn_install and npm_install rules",
     fields = {
+        "transitive": "If true this dependency has transitive npm dependencies but is not and npm package itself",
```
gregmagolan:
```suggestion
        "transitive": "If true this dependency has transitive npm dependencies but is not an npm package itself",
```

gregmagolan:
Nice. I did some initial debugging and this doesn't seem quite right and may not be hermetic. I'll do some more work on this later this evening.

Globegitter:
Yes, I have not done very extensive testing and was looking for some input first, but will also start testing this with #283 as well as with our apps. Though which part are you referring to that might not be hermetic?

gregmagolan:
> Yes, I have not done very extensive testing and was looking for some input first, but will also start testing this with #283 as well as with our apps. Though which part are you referring to that might not be hermetic?

From an initial look it seemed like the transitive npm deps like date-fns are no added to the rollup action inputs but they are still resolved from their non-sandboxed location.

The code in rollup_bundle.bzl ends up added the `.d.ts` files from deps that have `NodeModuleInfo` with `transitive = true` but not the transitive npm deps themselves..

```
    # Also include files from npm fine grained deps as inputs.
    # These deps are identified by the NodeModuleInfo provider.
    for d in ctx.attr.deps:
        if NodeModuleInfo in d:
            direct_inputs += _filter_js_inputs(d.files.to_list())
```

I'm going to play with it some more today as this would be a good one to get in and would unblock your #283 as well.

Globegitter:
Aahh yes, to be honest I was actually surprised that it already worked by just adding the transitive `NodeModuleInfo`. So I suppose it somehow follows out a symlink somewhere? I have noticed that as well sometimes in our apps, that there can be different behaviour if `node_module` exist in the user's workspace or not. And so far I am assuming that some things somehow are not preserving symlinks.

Globegitter:
Started debugging and it does indeed not seem to be hermetic. I do not find any trace of date-fns in the sandbox and the debug logs of rollup bundling show:
```
Rollup: resolving 'date-fns' from /home/markus/.cache/bazel/_bazel_markus/8464573ded3a95b161445c575d9af429/sandbox/linux-sandbox/1/execroot/e2e_typescript_3_1/bazel-out/k8-fastbuild/bin/bundle.es6/lib.js
Rollup: try to resolve 'date-fns' at '/home/markus/.cache/bazel/_bazel_markus/8464573ded3a95b161445c575d9af429/sandbox/linux-sandbox/1/execroot/e2e_typescript_3_1/bazel-out/k8-fastbuild/bin/bundle.es6/date-fns'
Rollup: allowing rollup to resolve 'date-fns' with node module resolution
Rollup: resolving './add_days/index.js' from /home/markus/.cache/bazel/_bazel_markus/8464573ded3a95b161445c575d9af429/external/npm/node_modules/date-fns/index.js
```

The last path is outside the sandbox. But need to look further into this to understand what exactly is happening.

Globegitter:
Some further progress: I found rollup itself has a `preserveSymlinks` option which defaults to `false` so I set that and I also added the `--preserve-symlinks-main` just to make sure there is nothing related to symlinks. But that has not changed anything. I have found that the node-resolve plugin has a `jail` option, which is to limit module resolution just to a specific path. That seems like a sensible safeguard to me and setting this now the build is failing indeed. Still more work to be done.

e2e/typescript_3.1/BUILD.bazel
```diff
@@ -42,3 +55,37 @@ jasmine_node_test(
...
+
+rollup_bundle(
+    name = "bundle",
+    entry_point = "main",
+    deps = [
+        ":main",
+    ],
+)
+
+rollup_bundle(
```
gregmagolan: Should change the `test` script in package.json to `bazel build ... && bazel test ...` so that these targets are built

internal/rollup/rollup.config.js
```diff
@@ -193,6 +194,7 @@ const config = {
     banner,
     format: 'TMPL_output_format',
   },
+  preserveSymlinks: true,
```
gregmagolan: Nice. Do these preserve symlinks changes work as intended to keep the resolve within the cwd()?

e2e/typescript_3.1/BUILD.bazel
Globegitter: I am pretty sure `bazel test ...` implicitly also builds all targets (unless `--build_tests_only`) is also specified. Let me verify this though.

internal/common/node_module_info.bzl
```diff
@@ -22,13 +22,36 @@ NodeModuleInfo = provider(
     },
 )
 
+NodeModuleSources = provider(
```
gregmagolan: I suspect this is over complicating things here with two different node_modules providers and a `NODE_MODULE_MARKER` tag. This may be a good opportunity to remove the tag and create a single rule that is used instead of `filegroup` in `generate_build_file.js` so we can avoid the marker and have a single provider.

@kyliau recently added the ng_apf_library() rule which is used instead of `filegroup` when an angular package format is detected. We consolidate this work into a single rule and solve the transitive deps problem as well.

internal/rollup/rollup.config.js
Globegitter: Yep, in this particular case it was really the `jail` attribute that brought the non-hermetic build to fail.

e2e/typescript_3.1/BUILD.bazel
Globegitter: Yep just verified. I had a failing `rollup_bundle` and `bazel test ...` failed on building it.

gregmagolan: Ahhhh. Good to know. Thanks for verifying.

gregmagolan:
Closing in favor of #675. Subset of changes from this PR will be ported over.

PR #614, opened 2019-03-12
Add examples from rules_typescript repo

gregmagolan:
Adds examples & e2e tests from rules_typescript repo.

Also includes:
* Cleanup of existing examples workspaces names
* Cleanup and updates to dev scripts
* Remove defaults.bzl from @npm/karma
* Removed internal/defaults.bzl from @npm/typescript

Notable missing from this PR:
`rules_typescript//internal/e2e/default_tsconfig_test.js` --- I'd like to port this is a separate PR as its a tricker one
`rules_typescript//internal/e2e/typescript_3.1` --- example of building without the npm packages. to be ported over in a separate PR along with making other examples of how to use these rules without install through npm


.clang-format
```diff
@@ -1,3 +1,7 @@
-Language:        JavaScript
-BasedOnStyle:    Google
-ColumnLimit:     100
+Language: JavaScript
+BasedOnStyle: Google
+ColumnLimit: 100
+---
+Language: Proto
+# Don't format .proto files.
```
alexeagle: that's cool, but why not? I think they are all third_party - maybe we meant to exclude third_party directory?

examples/app/.bazelrc
```diff
@@ -0,0 +1,9 @@
+# Print test logs for failed tests
```
alexeagle: we are accumulating a bunch of copy-paste code - what do you think about making these symlinks in the git repo?

examples/app/BUILD.bazel
```diff
@@ -0,0 +1,54 @@
+load("@build_bazel_rules_nodejs//:defs.bzl", "http_server", "rollup_bundle")
+load("@npm_bazel_typescript//:index.bzl", "ts_config", "ts_devserver", "ts_library")
+
+ts_library(
```
alexeagle: I was thinking that these would all go under `examples/typescript` showing how to use `packages/typescript`

otherwise if it's one-to-many, how would someone know where to find examples of using a particular package?

then the TypeScript example would have sub-packages showing the different features, like a protobuf under one subpackage. They could all be linked into the top-level example app.

WDYT?

examples/app/app.ts
```diff
@@ -0,0 +1,4 @@
+const el: HTMLDivElement = document.createElement('div');
+el.innerText = 'Hello, TypeScript';
```
alexeagle: one larger typescript example would be a bit more realistic than this toy one

examples/app/BUILD.bazel
gregmagolan: Hmmm. I think it's useful to have separate workspaces for each example. Keeps it self contained and the package.json is only for that example and a good example really shouldn't need to be updated very often. However, you're right that having them all at the top level may be too much but some examples use multiple packages such as `typescript` and `karma` and `jasmine`.

Maybe just a good `examples/README.md` that provides an overview of what each example shows?

.clang-format
gregmagolan: Sure. Separate PR?

examples/app/.bazelrc
gregmagolan: Sounds good. Separate PR?

examples/app/app.ts
gregmagolan: Agreed. I moved most of the assertions from rules_typescript to `/e2e/ts_library/...`. That's a better home for most of them. The `/e2e/app` can just be a really good example without worrying about assertions.

e2e/define_var/WORKSPACE
```diff
@@ -1,3 +1,5 @@
+workspace(name = "examples_define_var")
```
alexeagle: why bother naming it?

packages/typescript/BUILD.bazel
```diff
@@ -12,9 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# BEGIN-INTERNAL
```
alexeagle: this isn't needed - COMMON_REPLACEMENTS already mocks out the bzl_library load

packages/typescript/BUILD.bazel
```diff
@@ -38,13 +45,6 @@ genrule(
     cmd = "cp $< $@",
 )
 
-genrule(
```
alexeagle: why remove this? I'd rather have fewer INTERNAL fences, they are harder to reason about

scripts/clean_e2e.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
```
alexeagle: I'm feeling like we made a mistake not investigating the bazel integration test scheme from Ittai more. We've built a build system for the e2e/examples tests now

scripts/test_all.sh
```diff
@@ -44,7 +44,7 @@ echo_and_run bazel run //internal/node/test:has_deps_hybrid
 echo_and_run bazel run //internal/e2e/fine_grained_no_bin:index
 echo_and_run bazel run @fine_grained_deps_yarn//typescript/bin:tsc
 
-# bazel test @program_example//... # DOES NOT WORK WITH --nolegacy_external_runfiles
+# bazel test @examples_program//... # DOES NOT WORK WITH --nolegacy_external_runfiles
```
alexeagle: generally I don't think this rename is worth all the churn in the repo

what's wrong with foo_example living in examples/foo ? seems intuitive to me

e2e/ts_library/googmodule/BUILD.bazel
```diff
@@ -0,0 +1,36 @@
+load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary")
```
alexeagle: i still don't follow the reasoning for this being in e2e rather than examples. Shouldn't we have an example of closure interop?

e2e/define_var/WORKSPACE
gregmagolan: consistency is the only reason. don't feel strongly about it either way.

packages/typescript/BUILD.bazel
gregmagolan: Needed
```
js_library(
    name = "protobufjs_bootstrap_scripts",
```
in BUILD.bazel

scripts/clean_e2e.sh
gregmagolan: I think this is better than what we had and it buys us some time. The CI time is down with this setup and testing & developing locally is much better. The clean scripts are useful for local development only.

scripts/test_all.sh
gregmagolan: I prefer the workspace name matching the path order. Makes it easier to trace when you see:

```
      # We should also be able to test targets in a different workspace
      # TODO(gmagolan): move these tests into `build` job once
      # https://github.com/bazelbuild/bazel/issues/6481 is resolved
      - run: bazel test @examples_program//...
      - run: bazel test @packages_example//...
```

But don't feel strongly either way. The churn is only in 3 locations.

e2e/ts_library/googmodule/BUILD.bazel
gregmagolan: Sure. I think that's a useful example to have. The one from rules_typescript was depending on a local `devmode_consumer.bzl` which seemed more like ts_library e2e to me.

packages/typescript/BUILD.bazel
gregmagolan: Sure. I'll remove the fencing from the load statements.

gregmagolan:
Follow up PRs to this one:

* port over rules_typescript//internal/e2e/default_tsconfig_test.js
* port over rules_typescript//internal/e2e/typescript_3.1 --- example of building without the npm packages. to be ported over in a separate PR along with making other examples of how to use these rules without install through npm
* move `/internal/e2e` tests to `/e2e/`
* improve `/examples/app` to make it a more complete example

packages/typescript/BUILD.bazel
gregmagolan: for the protobuf example

alexeagle: okay but I don't think the fencing is required here?

gregmagolan: good point. removed.

PR #620, opened 2019-03-15
fix yarn_install and npm_install after changed packages

alexeagle:
Previously it did not re-copy the package contents, so the bazel workspace would be stale


gregmagolan:
Cool. One thing that will still come up until we have yarn_install use the local node_modules is that the yarn.lock file is not updated by yarn_install or npm_install.

PR #621, opened 2019-03-18
Jasmine updates

gregmagolan:
* Add pass-through index.js "main" to @bazel/jasmine to improve bootstrap pattern. Provides a boot function so user bootstrap code now only needs to call
```
require('@bazel/jasmine').boot();
```
* Fix entry_point of jasmine_runner to be `@bazel/jasmine/src/jasmine_runner.js`


packages/jasmine/index.js
```diff
@@ -0,0 +1,23 @@
+// jasmine-core is a transitive dep of jasmine
+// so look under unhoisted location first
```
alexeagle: this still needs some more commenting about why we have to do this special resolution instead of just getting the jasmine-core that node provides when we require it from here.

I think it's because we want to require() the same library that the user will later? why is it possible that a `require('jasmine-core')` in this location gives a different one than jasmine will later?

packages/jasmine/src/jasmine_node_test.bzl
```diff
@@ -39,9 +39,10 @@ def jasmine_node_test(
       srcs: JavaScript source files containing Jasmine specs
       data: Runtime dependencies which will be loaded while the test executes
       deps: Other targets which produce JavaScript, such as ts_library
-      jasmine: a label providing the jasmine dependency
       expected_exit_code: The expected exit code for the test. Defaults to 0.
       tags: bazel tags applied to test
+      jasmine: a label providing the jasmine dependency
+      entry_point_override: jasmine_runner entry point override (for internal use)
```
alexeagle: seems unused?

gregmagolan: Whoops. Old code. Thanks.

packages/jasmine/index.js
gregmagolan: Comment updated. PTAL.

PR #625, opened 2019-03-22
Add rollup-plugin-amd to rollup_bundle to support bundling *.ngfactory.js AMD files generated by ngc

gregmagolan:
This is a workaround for the following issue https://github.com/angular/angular/issues/29491


gregmagolan:
Needs a test before landing

internal/rollup/postinstall-patches.js
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+try {
+  require.resolve('shelljs');
```
alexeagle: let's not add a shelljs dependency - we are making *everyone* who uses rules_nodejs download this, right? or at least anyone who loads from the rollup rule. We can probably just do this with node APIs

internal/rollup/postinstall-patches.js
```diff
@@ -0,0 +1,45 @@
...
+set('-e');
+// print commands as being executed
+set('-v');
+// jump to project root
+cd(path.join(__dirname, '../..'));
+
+// Part of work-around for Angular ngfactory issue https://github.com/angular/angular/issues/29491.
+console.log(
+    '\n# patch: @buxlabs/amd-to-es6 to generate namespace imports instead of default imports');
+sed('-i', 'ImportDefaultSpecifier', 'ImportNamespaceSpecifier',
```
alexeagle: is there an issue filed for this, a PR against upstream, etc? how do we avoid needing to patch this in the future?

gregmagolan: I'll ask about this in an issue for amd-plugin. We may be able to land this as a configuration option for the plugin but I don't think we can change the default behavior as that would break anyone using it.

PR #627, opened 2019-03-26
Add angular example

gregmagolan:
Tests amd plugin fix for ngfactory files from https://github.com/bazelbuild/rules_nodejs/pull/625


gregmagolan:
> Hmm, I'm not sure about copy-pasting so much code to make this CI work.
> 
> Is there some way we could use the app in angular-bazel-example or angular/angular/integration/bazel as the test?
> 
> It could be "lazy" - have renovate send PRs to update those apps to snapshot builds of rules_nodejs (would need to publish snapshots first)
> Or maybe eager - in the CI for this repo we could fetch and build one of those projects, either as an external WORKSPACE, or just with some shell goop
> 
> WDYT?

Out of those two I think eager would be better as we can catch regressions sooner. On the other hand, its nice to have an example of angular in this repo but it would be another place to have to keep up to date.

alexeagle:
Let's find a way to do it where we clone angular-bazel-example, patch the workspace file and/or package.json files, and run it like we do the local e2e

gregmagolan:
Replaced by #640

PR #639, opened 2019-03-30
Clean packages from yarn cache in process_package_json step of yarn_install that have file:// URIs

gregmagolan:
Also updates /dist folders generated packages to `/dist/npm_bazel_foobar$RANDOM` and linking now handles that path. This is so that bazel sees the change to package.json when a package is re-generated and it re-runs `yarn_install`. That combined with the automatic `yarn cache clean foobar` for `file://` URIs means you can now just build a package & re-run a test like so:

```
./scripts/build_packages.sh typescript
./scripts/test_examples.sh app
```

without having to worry about cleaning the nested workspace or about the yarn cache messing up the test.

yarn cache issue for `file://` URIs: https://github.com/yarnpkg/yarn/issues/2165


scripts/clean_yarn_cache_selectively.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
gregmagolan: This script is not used by any other scripts or by CI. Just here for running manually if you'd like to clear all packages from the yarn cache.

PR #640, opened 2019-03-30
Checkout and test angular-bazel-example in CI

gregmagolan:
After checkout replacements made to package.json & WORKSPACE of angular-bazel-example using bash sed. grep is used to sanity check that the replacements took effect.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -297,6 +297,7 @@ jobs:
       - *attach_workspace
       - *init_environment
       - *init_bazel
+      - run: ./scripts/setup_examples_angular.sh
       # Some examples such as app require xvfb
       - run: xvfb-run -a ./scripts/test_examples_all.sh ${CIRCLE_NODE_INDEX} ${CIRCLE_NODE_TOTAL}
 
--- a/scripts/setup_examples_angular.sh
+++ b/scripts/setup_examples_angular.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
+# -e: exits if a command fails
+# -u: errors if an variable is referenced before being set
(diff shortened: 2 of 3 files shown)


gregmagolan:
Note: the `test_examples_all.sh` script calls `yarn test` on each example. For angular-bazel-example this does not run the e2e tests. It only runs `bazel test //src/...`. @alexeagle Should we change angular-bazel-example so that `yarn test` runs all tests or alternately can make a change here to that if there is an `e2e` script in package.json to call that for examples.

gregmagolan:
Did a negative test by taking out the amd() plugin from rollup_bundle and as expected angular-bazel-example e2e prodserver test fails:

```
1 spec, 1 failure
Finished in 10.293 seconds
[17:33:56] I/launcher - 0 instance(s) of WebDriver still running
[17:33:56] I/launcher - Running 0 instances of WebDriver
[17:33:56] I/launcher - chrome #01 failed 1 test(s)
[17:33:56] I/launcher - overall: 1 failed spec(s)
[17:33:56] E/launcher - Process exited with error code 1
INFO: Elapsed time: 17.808s, Critical Path: 15.40s
INFO: 7 processes: 6 processwrapper-sandbox, 1 worker.
//e2e:devserver_test_chromium-local                                      PASSED in 2.0s
//e2e:prodserver_test_chromium-local                                     FAILED in 11.8s
  /home/circleci/.cache/bazel/_bazel_circleci/78415e9cb21fafcabe23e7712002bc83/execroot/angular_bazel_example/bazel-out/k8-fastbuild/testlogs/e2e/prodserver_test_chromium-local/test.log

Executed 2 out of 2 tests: 1 test passes and 1 fails locally.
There were tests whose specified size is too big. Use the --test_verbose_timeout_warnings command line option to see which ones these are.
error Command failed with exit code 3.
Exited with code 1
```

PR #641, opened 2019-03-31
feat(rollup): provide sourcemaps for all output types

jbedard:
~~This adds sourcemap targets output groups for all the output formats.~~

Today there's the default target which outputs the es5 bundle.js and bundle.js.map. Then there's sub-targets such as `.es2015.js`, `.umd.js` etc. There's also `.js` which is the same as the default target but without the `.map`. But there's no way to easily fetch the .js+map of only the es2015 bundle.

~~This adds a `.js.map` target alongside every `.js` target.~~
This adds an output-group for each bundle type. The tests show an example of `filegroup`s consuming these output-groups.

### DIFF
--- a/internal/e2e/rollup/BUILD.bazel
+++ b/internal/e2e/rollup/BUILD.bazel
@@ -18,7 +18,7 @@ rollup_bundle(
 
 jasmine_node_test(
     name = "test",
-    srcs = glob(["*.spec.js"]),
+    srcs = ["rollup.spec.js"],
     configuration_env_vars = ["UPDATE_GOLDEN"],
     data = glob([
         "*_golden.js_",
@@ -33,5 +33,50 @@ jasmine_node_test(
         "//internal/e2e:check_lib",
         "@npm//jasmine",
         "@npm//unidiff",
+    ],
+)
+
+filegroup(
+    name = "bundle-outputgroups-cjs",
+    srcs = [":bundle"],
+    output_group = "cjs",
+)
+
+filegroup(
+    name = "bundle-outputgroups-umd",
+    srcs = [":bundle"],
+    output_group = "umd",
+)
+
+filegroup(
+    name = "bundle-outputgroups-es2015",
+    srcs = [":bundle"],
+    output_group = "es2015",
+)
+
+filegroup(
+    name = "bundle-outputgroups-es5_min",
+    srcs = [":bundle"],
+    output_group = "es5_min",
+)
+
+filegroup(
+    name = "bundle-outputgroups-es5_min_debug",
+    srcs = [":bundle"],
+    output_group = "es5_min_debug",
+)
+
+jasmine_node_test(
+    name = "test-outputgroups",
+    srcs = ["outputgroups.spec.js"],
+    data = [
+        ":bundle-outputgroups-cjs",
+        ":bundle-outputgroups-es2015",
+        ":bundle-outputgroups-es5_min",
+        ":bundle-outputgroups-es5_min_debug",
+        ":bundle-outputgroups-umd",
+    ],
+    deps = [
+        "@npm//jasmine",
     ],
 )
--- a/internal/rollup/rollup_bundle.bzl
+++ b/internal/rollup/rollup_bundle.bzl
@@ -474,24 +474,46 @@ def _rollup_bundle(ctx):
             code_split_es5_min_output_dir,
             code_split_es5_min_debug_output_dir,
         ]
+        output_group = OutputGroupInfo(
+            es2015 = depset([ctx.outputs.build_es2015, code_split_es2015_output_dir]),
+            es2015_min = depset([ctx.outputs.build_es2015_min, code_split_es2015_min_output_dir]),
+            es2015_min_debug = depset([ctx.outputs.build_es2015_min_debug, code_split_es2015_min_debug_output_dir]),
+            es5 = depset([ctx.outputs.build_es5, code_split_es5_output_dir]),
+            es5_min = depset([ctx.outputs.build_es5_min, code_split_es5_min_output_dir]),
+            es5_min_debug = depset([ctx.outputs.build_es5_min_debug, code_split_es5_min_debug_output_dir]),
+        )
 
     else:
         # Generate the bundles
         rollup_config = write_rollup_config(ctx)
-        run_rollup(ctx, _collect_es2015_sources(ctx), rollup_config, ctx.outputs.build_es2015)
-        run_terser(ctx, ctx.outputs.build_es2015, ctx.outputs.build_es2015_min, config_name = ctx.label.name + "es2015_min")
-        run_terser(ctx, ctx.outputs.build_es2015, ctx.outputs.build_es2015_min_debug, debug = True, config_name = ctx.label.name + "es2015_min_debug")
+        es2015_map = run_rollup(ctx, _collect_es2015_sources(ctx), rollup_config, ctx.outputs.build_es2015)
+        es2015_min_map = run_terser(ctx, ctx.outputs.build_es2015, ctx.outputs.build_es2015_min, config_name = ctx.label.name + "es2015_min", in_source_map = es2015_map)
+        es2015_min_debug_map = run_terser(ctx, ctx.outputs.build_es2015, ctx.outputs.build_es2015_min_debug, debug = True, config_name = ctx.label.name + "es2015_min_debug", in_source_map = es2015_map)
         _run_tsc(ctx, ctx.outputs.build_es2015, ctx.outputs.build_es5)
-        source_map = run_terser(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min)
-        run_terser(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min_debug, debug = True)
+        es5_min_map = run_terser(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min)
+        es5_min_debug_map = run_terser(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min_debug, debug = True)
         cjs_rollup_config = write_rollup_config(ctx, filename = "_%s_cjs.rollup.conf.js", output_format = "cjs")
-        run_rollup(ctx, _collect_es2015_sources(ctx), cjs_rollup_config, ctx.outputs.build_cjs)
+        cjs_map = run_rollup(ctx, _collect_es2015_sources(ctx), cjs_rollup_config, ctx.outputs.build_cjs)
         umd_rollup_config = write_rollup_config(ctx, filename = "_%s_umd.rollup.conf.js", output_format = "umd")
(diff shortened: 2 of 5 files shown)


internal/rollup/rollup_bundle.bzl
```diff
@@ -441,12 +441,25 @@ def _rollup_bundle(ctx):
         _generate_code_split_entry(ctx, ctx.label.name + "_chunks", ctx.outputs.build_umd)
         _generate_code_split_entry(ctx, ctx.label.name + "_chunks", ctx.outputs.build_cjs)
 
-        # There is no source map explorer output when code-splitting but we still need to satisfy the output
-        ctx.actions.expand_template(
-            output = ctx.outputs.explore_html,
-            template = ctx.file._no_explore_html,
-            substitutions = {},
-        )
+        # There is no source map or explorer output when code-splitting but we still need to satisfy the output
```
jbedard: Seems like there's no source maps when code-splitting today? I guess that could be a future task...

internal/rollup/rollup_bundle.bzl
```diff
@@ -465,18 +478,31 @@ def _rollup_bundle(ctx):
...
+
+        # es5
+        # TODO: tsc should produce a source map from the build_es2015 + source_map_es2015 inputs
+        # See: https://github.com/bazelbuild/rules_nodejs/issues/175
         _run_tsc(ctx, ctx.outputs.build_es2015, ctx.outputs.build_es5)
-        source_map = run_terser(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min)
-        run_terser(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min_debug, debug = True)
+        source_map_min = run_terser(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min)
+        run_terser(ctx, ctx.outputs.build_es5, ctx.outputs.build_es5_min_debug, debug = True, in_source_map = source_map_min)
+
```
jbedard: I guess we could just copy the `source_map_min` to `ctx.outputs.build_es5_map` just so the `build_es5_map` target exists for consistency?

internal/rollup/rollup_bundle.bzl
```diff
@@ -465,18 +478,31 @@ def _rollup_bundle(ctx):
...
         cjs_rollup_config = write_rollup_config(ctx, filename = "_%s_cjs.rollup.conf.js", output_format = "cjs")
         run_rollup(ctx, _collect_es2015_sources(ctx), cjs_rollup_config, ctx.outputs.build_cjs)
+
+        # umd
         umd_rollup_config = write_rollup_config(ctx, filename = "_%s_umd.rollup.conf.js", output_format = "umd")
         run_rollup(ctx, _collect_es2015_sources(ctx), umd_rollup_config, ctx.outputs.build_umd)
-        run_sourcemapexplorer(ctx, ctx.outputs.build_es5_min, source_map, ctx.outputs.explore_html)
-        files = [ctx.outputs.build_es5_min, source_map]
+
+        # TODO: ?
```
jbedard: Is this worth doing for every sourcemap?

internal/rollup/rollup_bundle.bzl
alexeagle: what does the rollup program produce? I bet it still creates sourcemaps, and having a size explorer for chunks is a pretty critical feature, could fix this next

internal/e2e/rollup/bundle-cjs_golden.js.map
```diff
@@ -0,0 +1 @@
...s6/internal/e2e/rollup/bar.js","bundle.es6/internal/e2e/rollup/foo.js"],"sourcesContent":["export const fum = 'Wonderland';\n","export default 'Hello';\n","export const name = 'Alice';\n","import {fum} from 'fumlib';\nimport hello from 'hello';\nimport {thing} from 'some_global_var';\n\nimport {name} from './bar';\n\nconsole.log(`${hello}, ${name} in ${fum}`);\n\n// Tests for @PURE annotations\n/*@__PURE__*/\nconsole.log('side-effect');\n\nclass Impure {\n  constructor() {\n    console.log('side-effect')\n  }\n}\n\n/*@__PURE__*/new Impure();\n\n// Test for sequences = false\nexport class A {\n  a() {\n    return document.a;\n  }\n}\nfunction inline_me() {\n  return 'abc';\n}\nconsole.error(new A().a(), inline_me(), thing);\n"],"names":["thing"],"mappings":";;;;;;;;;;;AAAO,MAAM,GAAG,GAAG,YAAY,CAAC;;ACAhC,YAAe,OAAO,CAAC;;ACAhB,MAAM,IAAI,GAAG,OAAO,CAAC;;ACM5B,OAAO,CAAC,GAAG,CAAC,CAAC,EAAE,KAAK,CAAC,EAAE,EAAE,IAAI,CAAC,IAAI,EAAE,GAAG,CAAC,CAAC,CAAC,CAAC;AAC3C,AAYA;;AAEA,AAAO,MAAM,CAAC,CAAC;EACb,CAAC,GAAG;IACF,OAAO,QAAQ,CAAC,CAAC,CAAC;GACnB;CACF;AACD,SAAS,SAAS,GAAG;EACnB,OAAO,KAAK,CAAC;CACd;AACD,OAAO,CAAC,KAAK,CAAC,IAAI,CAAC,EAAE,CAAC,CAAC,EAAE,EAAE,SAAS,EAAE,EAAEA,qBAAK,CAAC,CAAC;;;;"}
```
alexeagle: I'm questioning the volume of golden content here. Do we really want to assert on the content of all the sourcemaps? If you just want to assert that the rollup_bundle rule has some output file, we can do that just by referencing it as an input to any rule (like a trivial genrule)

internal/rollup/rollup_bundle.bzl
alexeagle: But it would have the wrong content, wouldn't it? I think we could just fix #175 next

internal/rollup/rollup_bundle.bzl
alexeagle: The user needs to be able to reason about what makes their bundle large under different cases. If two flavors of JS output could have different size characteristics, then the user would have to be able to debug them separately. But we shouldn't expose more of these sourcemapexplorer things than necessary.

Also I feel like we have gone too far with including lots of behaviors in this rule. Maybe sourcemapexplorer should be a separate rule, WDYT?

internal/rollup/rollup_bundle.bzl
```diff
@@ -664,13 +690,22 @@ ROLLUP_ATTRS = {
 
 ROLLUP_OUTPUTS = {
     "build_cjs": "%{name}.cjs.js",
+    "build_cjs_map": "%{name}.cjs.js.map",
```
alexeagle: this needs some re-organization as a starlark provider so the JS and .map can stay together as tuple through the dependent build steps...

kyliau: This seems like the ideal use case for OutputGroupInfo provider
https://docs.bazel.build/versions/master/skylark/rules.html#requesting-output-files
https://docs.bazel.build/versions/master/skylark/lib/OutputGroupInfo.html

jbedard: :+1:

Would we still want the targets for individual files? Only the `.js` ones which exist today?

jbedard: I got this working locally, but does this mean people must use the `--output_groups` flag instead of something similar to the `bundle.es2015.js` used today?

internal/e2e/check.js
```diff
@@ -9,7 +9,10 @@ function read(readPath) {
       .replace(/\r\n/g, '\n')
       .replace(/\\\\/g, '/')
       // In the sourcemap, CRLF is turned into a string encoding
-      .replace(/\\r\\n/g, '\\n');
+      .replace(/\\r\\n/g, '\\n')
+      // Replace platform specific directories in paths
+      .replace(/\/\w+-fastbuild\//g, '/__platform__-fastbuild/')
```
alexeagle: is this related? could be a separate PR?

jbedard: I was getting different folder names outputted here compared to circleci. Were the golden files previously generated by circleci so they have the circleci platform? I figured this was better then me manually editing the golden files...

kyliau:
> I think OutputGroup may be the wrong thing - it seems that it makes it impossible for a downstream rule to request a particular output? Like https://github.com/angular/angular-bazel-example/blob/master/src/BUILD.bazel#L179

If `ROLLUP_OUTPUTS` is restored then downstream targets could continue to request a specific output as before.

I think `OutputGroup` could still be useful, because it provides logical grouping of the outputs.
The downside is that downstream targets would need to declare an intermediate `filegroup` in order to retrieve the sourcemap.

```
rollup_bundle(
    name = "bundle",
    entry_point = "src/main.prod",
    deps = [
        // ...
    ],
)

filegroup(
    name = "bundle_es2015_min_with_sourcemap"
    srcs = [":bundle"],
    output_group = "es2015_min",
)

web_package(
    name = "prodapp",
    assets = _ASSETS + [
        ":bundle.min.js",
        ":bundle.min.es2015.js",
    ],
    data = [
        ":bundle_es2015_min_with_sourcemap",
    ],
    index_html = "index.html",
)
```

jbedard:
I've added a commit changing `rollup_bundle` to a macro which runs rollup as well as declaring a bunch of `filegroup` targets for a) each bundle type .js and .map b) just the .js file.

I think this is working, but I'm not sure if the macro+filegroups is the best solution, and haven't fully tested it. @alexeagle I'm not sure what you meant by "starlark provider", is that another way of doing this?

jbedard:
Update to just add an `OutputGroupInfo` to the result, with one group per bundle type. Without specifically consuming these new output-groups no one should notice this change.

Since this PR is now just this small output-group addition I've greatly simplified the diff and removed anything unnecessary.

Instead of adding new tests I've just executed the existing tests (a second time), but using the output-group instead of default info.

internal/e2e/rollup/BUILD.bazel
```diff
@@ -35,3 +35,47 @@ jasmine_node_test(
...
+
+filegroup(
+    name = "bundle-outputgroups-es5_min",
+    srcs = [":bundle"],
+    output_group = "es5_min",
+)
+
+jasmine_node_test(
+    name = "test-outputgroups",
+    srcs = glob(["*.spec.js"]),
```
kyliau: This should probably be another spec file that asserts the existence of sourcemap for each flavor in addition to the js file. Existing `rollup.spec.js` only checks sourcemap for the default output.

jbedard:
I've changed the tests to assert the presence of the `.map` files alongside `.js` when using the output groups.

kyliau:
Output groups could be requested on the command line.
For example,
```
bazel build :bundle --output_groups=es2015_min
```
See https://docs.bazel.build/versions/master/command-line-reference.html#flag--output_groups

PR #642, opened 2019-04-01
Add visible alias for the metadata entry of ts_devserver

alexeagle:
moved from https://github.com/bazelbuild/rules_typescript/pull/439

@achew22 could you help me add to the commit message here an explanation of *why* we make this change?


achew22:
How about

```
Add visible alias for the metadata entry of ts_devserver

When writing custom web servers that host your development environment, it is necessary to have the manifest of all development files that need to be served. This change reexposes the manifest under the provided visibility for consumption in such purposes.
```

PR #643, opened 2019-04-01
 feat: Add ng_apf_library rule & dev_scripts_aspect

kyliau:
This commit adds detection logic for NPM packages that are in
the Angular Package Format (APF) and generates `ng_apf_library`
targets for these packages instead of the default `filegroup` targets.

`ng_apf_library` targets provide a custom `scripts` attribute,
which is then later collected by `dev_scripts_aspect` and
exposed through the `dev_scripts` provider.

dev_scripts are automatically appended to `ts_devserver`'s `scripts`
attribute and `ts_web_test_suite`'s `srcs` so that they no longer need to be
specified manually.

For more info see design doc:
https://docs.google.com/document/d/16t7s3DplajCEbDwZTXyeYWIBHnXc22nngrXfVitFFNo

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


internal/common/dev_scripts_aspect.bzl
```diff
@@ -0,0 +1,37 @@
...
+    result = depset()
+
+    # If target is a node module it'd provide a list of `scripts`.
+    if hasattr(target, "scripts"):
+        result = depset(transitive = [result, target.scripts])
+
+    # Recursively collect transitive `dev_scripts` from the deps.
+    if hasattr(ctx.rule.attr, "deps"):
+        for dep in ctx.rule.attr.deps:
+            if hasattr(dep, "dev_scripts"):
```
alexeagle: instead of a "stringly typed" attribute, maybe we should have a Provider type?

gets into a bigger design question of what the JS provider should be. I'm also okay with adding that to our list for 8.1 now that we are getting a clearer picture of what that provider should be

internal/ng_apf_library/BUILD.bazel
```diff
@@ -0,0 +1,36 @@
...
+package(default_visibility = ["//internal:__subpackages__"])
+
+bzl_library(
+    name = "bzl",
+    srcs = glob(["*.bzl"]),
+    visibility = ["//visibility:public"],
+)
+
+filegroup(
+    name = "package_contents",
```
alexeagle: seems to be unreferenced?

I don't think we need to document this rule, since it won't be "public api" (not re-exported through the index.bzl)
can delete most of this file?

internal/ng_apf_library/ng_apf_library.bzl
```diff
@@ -0,0 +1,45 @@
...
+# limitations under the License.
+
+def _ng_apf_library_impl(ctx):
+    umds = []
+    factories = []
+
+    for file in ctx.files.srcs:
+        if file.basename.endswith(".umd.js"):
+            umds.append(file)
+        elif file.basename.endswith(".ngfactory.js"):
```
alexeagle: don't we need ngsummary.js for some tests also?

internal/ng_apf_library/ng_apf_library.bzl
```diff
@@ -0,0 +1,45 @@
...
+            factories.append(file)
+
+    return struct(
+        files = depset(
+            transitive = [src.files for src in ctx.attr.srcs] + [dep.files for dep in ctx.attr.deps],
+        ),
+        scripts = depset(umds + factories),
+    )
+
+"""Provides a replacement for the default filegroup target, with the addition of
```
alexeagle: not that we are generating stardoc for this - but I don't think it works to put a docstring as a top-level string literal before the declaration. pydoc puts it after. I think that maybe rule() itself added a `doc` attribute recently too

not important if it's not public API

internal/npm_install/ng_apf_library.js
```diff
@@ -0,0 +1,62 @@
...
+      if (set.has(sibling)) {
+        return true;
+      }
+    }
+    return false;
+  });
+}
+
+/**
+ * Prints a `ng_apf_library` rule that replaces the default filegroup target.
```
alexeagle: instead of just 'replaces' it might be nice to repeat its function "the generated target that the user will depend on, e.g. `@npm//@angular/core` will be an `ng_apf_library`

internal/npm_install/ng_apf_library.js
```diff
@@ -0,0 +1,62 @@
+/**
```
alexeagle: maybe this file would be better organized in the /internal/ng_apf_package folder? not sure if it should be grouped with "angular layering violations" code or "after installing node_modules" code

internal/npm_install/ng_apf_library.js
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Prints a `ng_apf_library` rule that replaces the default filegroup target.
+ */
+function printNgApfLibrary(pkg) {
+  const pkgDeps = pkg._dependencies.filter(dep => dep !== pkg && !dep._isNested);
+  return `
+load("@build_bazel_rules_nodejs//internal/ng_apf_library:ng_apf_library.bzl", "ng_apf_library")
+ng_apf_library(
+    name = "${pkg._name}__pkg",
```
alexeagle: a bunch of this is repeated from the filegroup version, right? did you try to reduce the duplication, or was that messy?

internal/common/dev_scripts_aspect.bzl
kyliau: I created `ScriptsProvider` and `DevScriptsProvider` - the former is used to collect scripts from direct deps, whereas the latter is used to recursively collect scripts from transitive deps.

internal/ng_apf_library/BUILD.bazel
kyliau: this is now included in `rules_nodejs_package` npm_package in the root BUILD file - I forgot to add it in the earlier commit.

internal/ng_apf_library/ng_apf_library.bzl
kyliau: moved comment to the `doc` attribute.

internal/npm_install/ng_apf_library.js
kyliau: done.

internal/npm_install/ng_apf_library.js
kyliau: I moved `ng_apf_library.js` from `npm_install` to `ng_apf_library`. Unfortunately this means the test script would have to simulate the behavior of generating the template by moving the file to the right place. Please see comments in `generate_build_file.js`, `generate_build_file.spec.js`, and `update_golden.js`.

internal/npm_install/ng_apf_library.js
kyliau: Moved `pkgDeps` back to the caller function `printPackage()` to avoid duplication. I think the rest of the code is clean enough, despite some duplication.

internal/ng_apf_library/ng_apf_library.bzl
kyliau: ah yes, `ngsummary.js` is needed for AOT test.

alexeagle:
windows CI failure looks legit:
```

==================== Test output for //internal/npm_install/test:test:
--
  | Error: ENOENT: no such file or directory, copyfile 'internal/ng_apf_library/ng_apf_library.js' -> 'internal/npm_install/ng_apf_library.js'
  | at Object.copyFileSync (fs.js:1721:3)
  | at Object.<anonymous> (D:\b\bk-windows-java8-pn7n\bazel\rules-nodejs-nodejs\internal\npm_install\test\generate_build_file.spec.js:3:15)
  | at Module._compile (internal/modules/cjs/loader.js:688:30)

``

File: internal/npm_install/generate_build_file.js [hunk 8]
```diff
@@ -690,6 +687,22 @@ filegroup(
     ],
     tags = ["NODE_MODULE_MARKER"],
 )
+`;
+}
+
+/**
+ * Given a pkg, return the skylark `filegroup` targets for the package.
```
gregmagolan:
```suggestion
 * Given a pkg, return the skylark `filegroup` and/or `ng_apf_library` targets for the package.
```

internal/npm_install/test/package/delete_build_files.js
```diff
@@ -0,0 +1,23 @@
+/**
+ * @fileoverview
+ * BUILD files in node modules have to be deleted because runfiles are not used
```
gregmagolan: Ahhhh. That makes sense and solves the mystery 👍

internal/common/dev_scripts_aspect.bzl
```diff
@@ -0,0 +1,41 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+"""Aspect to collect dev scripts from `deps` attribute.
+"""
+
+load("//internal/ng_apf_library:ng_apf_library.bzl", "ScriptsProvider")
```
alexeagle: I think it would be better for that provider to be defined outside the layering violation code. ng_apf_library should probably import this symbol from somewhere else.
(since Greg is likely adding another kind of node_module generated rule that runs browserify or something)

internal/npm_install/test/package/delete_build_files.js
kyliau: This is special to the test fixture because node_modules here are installed *outside* of the test, then added to the runfiles.

Issue #396, opened 2018-10-29
Building nodejs_image on macOS


Globegitter:
Currently it is not possible to build the `nodejs_image` on macOS, mostly because nodejs for mac is downloaded and copied into the container by bazel. See: https://github.com/bazelbuild/rules_docker/issues/458 and https://github.com/bazelbuild/rules_docker/issues/466.

I know bazel has some ways of specifying platforms for cross-compilation like `--host_platform`, `platforms` and others. I wonder if it was possible to make use of whatever the right flag/setting is to tell bazel to download the nodejs for the specified platform? I would say the number 1 use-case really is to build docker containers on macOS.

I also know this alone would not be a 100% bulletproof, e.g. when using native/compiled dependencies, but I do think it would already help quite a few people, it would definitely help our devs on macOS to be able to verify and use the docker image locally.

chrislovecnm:
Can we use a container like planter as a workaround?
https://github.com/kubernetes/test-infra/blob/master/planter/README.md

chrislovecnm:
Also this issue should be documented!!

Globegitter:
Yep using something like planter or https://github.com/nadirizr/dazel should do the trick as the host platform is then linux based so these rules would download nodejs for linux - I do think however it would not be too difficult to support this use-case (at least ignoring native dependencies to start off with), rules_go has already support for this use-case.

alexeagle:
@irengrig from the Bazel team is going to help us switch the nodejs binary to use Bazel toolchain support. That means the nodejs binary won't be a regular action input. I think it would let us solve this issue too.

chrislovecnm:
I think I got planter working on a Mac btw

dicarlo2:
@Globegitter @alexeagle FWIW - I managed to get nodejs images built on macOS with the following high level approach:
  1/ Define the nodejs binary as a toolchain and use it in nodejs_binary
  2/ Add environment variable support to yarn_install and use it to create a second workspace `@npm_linux_amd64` where yarn is run with env variables set to download pre-compiled linux binaries for native dependencies[1]
  3/ Modify nodejs_image to use the linux version of the nodejs binaries in the separate layer
  4/ Setup a second/parallel set of build rules (in our case, `ts_library` rules) that use `@npm_linux_amd64` for fine-grained dependencies rather than `@npm`.

[...]

PR #645, opened 2019-04-02
Initial Toolchain support

Globegitter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Nodejs is downloaded for each platform but does not support toolchains, so does e.g. also not support cross compilation or building of docker images on macOS.

Issue Number: First part of #396

## What is the new behavior?

Switches out nodejs to be fetched using platform specific toolchains.

- [] Yes
- [] No

## Other information

I started working on this a while ago but then got pulled away by trying to finish up other issues and also by starting to work on getting full support also for npm/yarn deps. Functionality wise it has all functionality to get nodejs as a toolchaing and support basic cross-compilation, i.e. building images on macOS that do not have any native dependencies. The platform specfic npm/yarn dependency support is incomplete and as it is turning out to be a more complex I was thinking of pulling this out of this PR and get the basic support in first and focus on npm/yarn specific support in a separate PR.


Globegitter:
This is currently still very much WIP. Will update as soon as it is in a reviewable state.

Globegitter:
Made quite a bit of progress and is functionally very close to be complete and in reviewable state. Should just be a matter of fixing the last platform specific issues from ci and cleaning things up. I will be on holiday next week so probably won't be able to finish this before.

thesayyn:
I'm looking forward to testing this. Good work

internal/node/node.bzl
```diff
@@ -133,15 +133,53 @@ def _nodejs_binary_impl(ctx):
     if hasattr(ctx.attr, "expected_exit_code"):
         expected_exit_code = ctx.attr.expected_exit_code
 
+    node_tool_info = ctx.toolchains["@build_bazel_rules_nodejs//toolchains/node:toolchain_type"].nodeinfo
+    node_tool_files = []
+    if node_tool_info.target_tool_path == "" and not node_tool_info.target_tool:
+        # If tool_path is empty and tool_target is None then there is no local
+        # node tool, we will just print a nice error message if the user
+        # attempts to do bazel run
+        ctx.actions.write(
```
gregmagolan: If this is a critical failure then using `fail("error message")` would be better so execution stops at this point

gregmagolan:
Nice. I did a once over and looks like its most of the way there functionally.

A few questions regarding usage:

- Is `node_repositories()` still called in the user's workspace?

- How would the different cases look?
  - What would be the behavior of default `node_repositories()` with toolchains?
  - How would a user select a specific version of node?
  - User wanting to use their locally installed version of node?
  - User wanting to use a version of node that is vendored into their project repo?

Globegitter:
@gregmagolan Thanks for taking a look already and apologies for the delayed response. Yep `node_repositories()` is still being called and for now should be a 100% backwards compatible (last famous words). 

So the default behaviour of `node_repositories()` is that it will setup the external nodejs repository with the downloaded nodejs binaries with the difference that it will now set-up 4 repositories: `nodejs`, `nodejs_darwin_amd64`, `nodejs_windows_amd64`, `nodejs_linux_amd64`. One of these repositories is redundant as the host os will match one of the other 3, but since these are created using a macro I can not seem how to avoid this redundancy (though I also do not see a real problem with it). Then we are registering and setting up a toolchain for all these three platforms by calling `node_configure`. 

The vendored node use-case is still supported in the same way with `node_repositores()` and will get configured equally. Though eventually I would like to deprecate `node_repositories()` for the vendored use-case as one could call `node_configure` directly with the vendored file but I have not tested that yet, and not sure I want to focus on sorting that out now already in this PR. Similarly one can now also provide a path to the nodejs binary to use a pre-installed binary on the system. I have note exposed the attribute for this yet, but will do so as I clean things up and add some tests. What would be the best way to test this btw? Would be nice to test some cross-compilation. I guess I could build a docker image on macos? WDYT?

The command to do that btw is:  `bazel run --platforms=@build_bazel_rules_nodejs//toolchains/node:linux_amd64 //app`. It does not work yet with native dependencies, that would be the next step in a separate PR.

Globegitter:
I tested this on my macbook btw and could build a docker container. It is functionally complete, just needs cleaning up, documentation and some tests.

thesayyn:
Great, I'm looking forward to this.

Globegitter:
Closed in favour of #827

PR #649, opened 2019-04-03
Support yarn_install & npm_install `bazelBin` `additionalAttributes` …

gregmagolan:
…for generated nodejs_binary targets for package.json `bin` files

For example, the `bin` `testy` under `/e2e/bazel_bin/npm/testy/package.json`:

```
{
  "name": "testy",
  "version": "0.0.1",
  "bin": {
    "testy": "./index.js"
  },
  "bazelBin": {
    "testy": {
      "additionalAttributes": {
        "configuration_env_vars": "[\"some_env\"]"
      }
    }
  }
}
```

results in a generated nodejs_binary target `@npm//node_modules/testy:testy__bin`:

```
# Wire up the `bin` entry `testy`
nodejs_binary(
    name = "testy__bin",
    entry_point = "testy/index.js",
    data = [":testy__pkg"],
    configuration_env_vars = ["some_env"],
)
```

Which also has an alias `@npm//testy/bin:testy`.


internal/npm_install/generate_build_file.js
```diff
@@ -711,12 +711,29 @@ filegroup(
 
   if (pkg._executables) {
     for (const [name, path] of pkg._executables.entries()) {
+      // Handle additionalAttributes of format:
```
alexeagle: seems like we also need to document this somewhere, maybe just start a wiki page about how to publish a package to NPM that works under Bazel?

gregmagolan: Was thinking that as well. Either a WIKI or add it to DEVELOPING.md and link from the main README.md.

PR #651, opened 2019-04-04
test(Karma): add e2e test for Karma files collected through sources_aspect

kyliau:
Karma should only load explicit node_sources and dev_scripts from
its deps and runtime_deps. There should not be a case where Karma
needs to load all files in a target.

For third-party deps, an explicit file from the package should be
provided to Karma instead of the entire package.
For example, `@npm//node_modules/rxjs:bundles/rxjs.umd.js` should
be used instead of `@npm//rxjs`.

(This effectively means devserver and Karma do not know how to
serve node modules that are not in APF format, so users have to
explicitly specify the files instead of Karma blindly loading everything).

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


packages/karma/karma.conf.js
```diff
@@ -188,8 +188,12 @@ try {
    */
   function configureFiles(conf) {
     overrideConfigValue(conf, 'files', [
+      // BEGIN BOOTSTRAP FILES
```
alexeagle: I like the idea of annotating the karma.conf to explain where things came from - are there more inputs we could associate here with where they ended up?

kyliau: yes, all templated inputs are now annotated.

kyliau:
> can you consolidate this into an existing e2e and keep the same assertions?
> we shouldn't continue to have new e2e for each

done, merged into existing karma_typescript e2e.

PR #657, opened 2019-04-05
Update bazel_skylib dependency to 0.8.0

laszlocsomor:
## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

rules_nodejs depends on [bazel_skylib 0.6.0](https://github.com/bazelbuild/bazel-skylib/releases/tag/0.6.0)

## What is the new behavior?

rules_nodejs depends on [bazel_skylib 0.8.0](https://github.com/bazelbuild/bazel-skylib/releases/tag/0.8.0)

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #658, opened 2019-04-05
Replace genrules with copy_file/write_file rules

laszlocsomor:
## PR Type

- [ ] Bugfix
- [x] Build related changes
- [ ] CI related changes

There are genrules in some BUILD files. Bazel requires Bash to build them.
Bash is an undesirable dependency on Windows.

## What is the new behavior?

Instead of using genrules, we use Starlark rules from bazel-skylib 0.8.0. They were implemented specifically for rules_nodejs. Their advantages over genrule are:

- simpler rule interface
- no Bash required on Windows
- compatible with Bazel's Bash-less test execution on Windows

See https://github.com/bazelbuild/bazel/issues/6622

- [ ] Yes
- [x] No

## Other information

This PR depends on https://github.com/bazelbuild/rules_nodejs/pull/657.


internal/npm_package/test/BUILD.bazel
```diff
@@ -1,10 +1,11 @@
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
```
alexeagle: this is a breaking change for our users. They don't currently depend on the `@bazel_skylib` repository, and I don't want to introduce such a transitive dependency because Bazel's support for transitive dependencies is so broken - e.g. how will my users get a correct WORKSPACE file with correct version of skylib as we make changes?

I think we can vendor the `write_file.bzl` into our third_party directory like we do for `paths.bzl`

laszlocsomor: Done.

laszlocsomor:
Done. Thanks!

alexeagle:
FYI we don't want merge commits since they cause non-linear history that's hard to blame, so I rebased and squashed.

laszlocsomor:
Thank you both!

PR #697, opened 2019-04-15
Fix @bazel/jasmine jasmineCore re-export so that jasmine-core transitive dep of jasmine transitive-dep is always used

gregmagolan:
Ran into yet another corner case for re-exporting the correct `jasmine-core` from `@bazel/jasmine` npm package here: https://github.com/angular/angular/pull/29913. I think I've thought it through fully now and this will always get the right version of `jasmine-core`

### DIFF
--- a/packages/jasmine/index.js
+++ b/packages/jasmine/index.js
@@ -1,4 +1,15 @@
-const jasmineCore = require('jasmine-core');
+const path = require('path');
+const fs = require('fs');
+
+// We want the jasmine-core version that is a transitive dependency of jasmine, however,
+// jasmine does not provide a re-export of its jasmine-core so we need resolve jasmine-core
+// from jasmine/node_modules/jasmine-core if it exists but from the base path of the jasmine
+// that @bazel/jasmine depends on which is path.dirname(require.resolve('jasmine/package.json'))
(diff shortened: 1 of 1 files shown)


gregmagolan:
For reference the case is. `@bazel/jasmine` dep on `jasmine` is hoisted to the root node_modules so its under `node_modules/jasmine` but root package.json has a conflicting dep on `jasmine-core` so its location is `node_modules/jasmine/node_modules/jasmine-core`.

packages/jasmine/index.js
```diff
@@ -1,4 +1,14 @@
-const jasmineCore = require('jasmine-core');
+const path = require('path');
+
+// We want the jasmine-core version that is a transitive dependency of jasmine, however,
```
kyliau: I'm thinking, if we need `jasmine-core` here, should it be part of the dependencies of this package so we don't have to guess where it gets installed?

gregmagolan: Ahhh.. Yes. That would be a much cleaner solution. @bazel/jasmine would have a dependency on the same version of jasmine-core that its jasmine dep does. Good call 😄

gregmagolan:
Replaced by #698.

PR #702, opened 2019-04-16
fix(npm_package): deps should allow files

kyliau:
**BREAKING CHANGE**: npm_package now throws for any files in `srcs` which
do not reside in the same package as the target.

The `deps` attr of `npm_package` should allow files.
A common use case would be named output of other rules, such as
`rollup_bundle`. In which case, user should be able to add
`bundle.umd.js` from `rollup_bundle` to the `deps` of `npm_package`.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/internal/npm_package/packager.js
+++ b/internal/npm_package/packager.js
@@ -86,6 +86,10 @@ function main(args) {
 
   // src like baseDir/my/path is just copied to outDir/my/path
   for (src of srcsArg.split(',').filter(s => !!s)) {
+    if (!src.startsWith(baseDir)) {
+      throw new Error(`${src} in 'srcs' does not reside in the base directory, ` +
+        `generated file should belong in 'deps' instead.`);
+    }
     copyWithReplace(src, path.join(outDir, path.relative(baseDir, src)), replacements);
   }
 
--- a/internal/npm_package/npm_package.bzl
+++ b/internal/npm_package/npm_package.bzl
@@ -88,9 +88,14 @@ def _npm_package(ctx):
             deps_sources,
             # Collect whatever is in the "data"
             dep.data_runfiles.files,
-            # For JavaScript-producing rules, gather up the devmode Node.js sources
-            dep.node_sources,
         ]
+
+        if hasattr(dep, "node_sources"):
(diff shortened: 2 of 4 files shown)


alexeagle:
We could instead fix `srcs` to strip `bazel-bin` paths when copying.

But I think this is the right thing to do - our idiom should be that `srcs` are files in your workspace, not outputs from other rules.

We should also error when `srcs` has a built output, rather than just copy it to a wrong location in the package.

internal/npm_package/test/bundle/BUILD.bazel
```diff
@@ -0,0 +1 @@
+exports_files(["bundle.min.js"])
```
alexeagle: this should be a genrule that copies the input file to an output location - we want the packager.js to copy `deps` from the bazel-bin or bazel-genfiles folders.
As you have it here, there's no reason bundle.min.js couldn't be in `srcs`

kyliau: Added genrule to generate `bundle.min.js` and added check in `packager.js` to make sure files in `srcs` reside in `baseDir`.

internal/npm_package/packager.js
```diff
@@ -86,6 +86,9 @@ function main(args) {
 
   // src like baseDir/my/path is just copied to outDir/my/path
   for (src of srcsArg.split(',').filter(s => !!s)) {
+    if (!src.startsWith(baseDir)) {
+      throw new Error(`Files in srcs must reside in base directory: ${src}`);
```
alexeagle: can we be more helpful? "looks like you meant to put this label in `deps` instead"

kyliau:
> optional: add a BREAKING CHANGE note to the commit message (ppl could have had generated files in srcs, WDYT?)

Added a BREAKING CHANGE note in the commit message, but I think the risk is minimal, because files that are not in the same package would have been copied to a wrong location, so it wouldn't have worked correctly before.

PR #703, opened 2019-04-17
feat: add ability to set jasmine seed as env variable

benlesh:
- Uses environment variable `JASMINE_SEED` to allow users to lock the randomization seed down for deterministic tests
- Adds documentation on how to use `JASMINE_SEED` environment variable

### DIFF
--- a/packages/jasmine/src/jasmine_node_test.bzl
+++ b/packages/jasmine/src/jasmine_node_test.bzl
@@ -30,6 +30,7 @@ def jasmine_node_test(
         tags = [],
         jasmine = Label("@npm//@bazel/jasmine"),
         coverage = False,
+        seed = None,
         **kwargs):
     """Runs tests in NodeJS using the Jasmine test runner.
 
@@ -44,6 +45,7 @@ def jasmine_node_test(
       tags: bazel tags applied to test
       jasmine: a label providing the jasmine dependency
       coverage: Enables code coverage collection and reporting
+      seed: Override the randomized jasmine seed with a specific seed
       **kwargs: remaining arguments are passed to the test rule
     """
     devmode_js_sources(
@@ -64,6 +66,9 @@ def jasmine_node_test(
         templated_args = templated_args + ["--coverage"]
     else:
         templated_args = templated_args + ["--nocoverage"]
+
+    if seed:
+        templated_args = templated_args + ["--seed ", str(seed)]
 
     nodejs_test(
         name = name,
--- a/packages/jasmine/src/jasmine_node_test.bzl
+++ b/packages/jasmine/src/jasmine_node_test.bzl
@@ -31,6 +31,7 @@ def jasmine_node_test(
         jasmine = Label("@npm//@bazel/jasmine"),
         coverage = False,
         seed = None,
+        randomize = True,
         **kwargs):
     """Runs tests in NodeJS using the Jasmine test runner.
 
@@ -45,7 +46,8 @@ def jasmine_node_test(
       tags: bazel tags applied to test
       jasmine: a label providing the jasmine dependency
       coverage: Enables code coverage collection and reporting
-      seed: Override the randomized jasmine seed with a specific seed
+      seed: Override the randomize jasmine seed with a specific seed
+      randomize: Whether or not to randomize the test sequence (supercedes the `seed` arg)
       **kwargs: remaining arguments are passed to the test rule
     """
     devmode_js_sources(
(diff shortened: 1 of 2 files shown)


packages/jasmine/src/jasmine_runner.js
```diff
@@ -151,7 +151,6 @@ function main(args) {
       process.exit(exitCode);
     }
 
-
```
benlesh: I think code formatting is getting me here. Sorry. I can revert this line if you care.

benlesh:
Looks like a yarn flake for `test_examples` CI.

packages/README.md
```diff
@@ -45,6 +45,14 @@ The new package should be automatically tested with `yarn test` in CI by the `te
 
 To test your package locally, you can run `yarn test_packages bar`. If you package depends on other packages you must run `yarn build_packages foo` to build a specific dependency or `yarn build_packages_all` to build all dependencies.
 
+### Providing a specific seed for deterministic testing
```
alexeagle: this README is for developing in this repo
user documentation is published at https://www.npmjs.com/package/@bazel/jasmine
you can edit that documentation here: https://github.com/bazelbuild/rules_nodejs/blob/master/packages/jasmine/src/jasmine_node_test.bzl#L36

packages/jasmine/src/jasmine_runner.js
```diff
@@ -185,6 +184,12 @@ function main(args) {
     }
   }
 
+  // If the user provides a seed as an environment variable, configure jasmine to use it.
+  const envSeed = process.env['JASMINE_SEED'];
```
alexeagle: I'm surprised that this works. Bazel generally strips out all the environment variables to keep actions hermetic - you can imagine if some environment variable doesn't have the same value on your machine as your coworker, you wouldn't get any remote cache hits since the cache key is determined from all the inputs.

https://bazelbuild.github.io/rules_nodejs/node/node.html#nodejs_test lists the `configuration_env_vars` attribute which is an allowlist for some environment variables to be passed through

benlesh: You're right, I've changed my approach and tested locally.

Now it's allowing an argument to be passed for `seed`. Documentation updated accordingly.

benlesh:
@alexeagle Latest changes were actually developed on local copies in my angular work area. Tested there.

Please have a look.

packages/jasmine/src/jasmine_runner.js
```diff
@@ -178,13 +189,21 @@ function main(args) {
...
+    // declared
     if (process.env['TEST_WORKSPACE'] === 'npm_bazel_jasmine' &&
         process.env['BAZEL_TARGET'] === '//test:sharding_test') {
       jrunner.randomizeTests(false);
     }
   }
 
+  if (seed) {
+    // Display that the seed has been overridden in bright green.
+    console.log(`\u001b[32mJasmine seed override: ${seed}\u001b[39m`);
```
alexeagle: I'm not so sure about colors here, it looks bad in a terminal that doesn't support it. I think a lot of CLI tools detect whether the terminal environment appears to support color and only conditionally print it. To be consistent I'd just leave it out here - we should have a general approach for tools run under bazel

packages/jasmine/src/jasmine_runner.js
```diff
@@ -60,31 +63,35 @@ function main(args) {
   // second is always a flag to enable coverage or not
   const coverageArg = args[1];
   const enableCoverage = coverageArg === '--coverage';
+  // third arg is an optional seed
+  const seed = args[2] === '--seed' ? args[3] : null;
```
alexeagle: this seems sad since any future flags need more special treatment. I wonder if we could just pass the residual args to jasmine's built-in command line parser?

packages/jasmine/src/jasmine_node_test.bzl
```diff
@@ -30,6 +30,7 @@ def jasmine_node_test(
         tags = [],
         jasmine = Label("@npm//@bazel/jasmine"),
         coverage = False,
+        seed = None,
```
alexeagle: this API seems reasonable to me.
I wonder if we should also expose whether or not to randomize tests

packages/jasmine/src/jasmine_runner.js
benlesh: I know very little about that. Do you have any documentation you can point me at, or a recommended approach?

packages/jasmine/src/jasmine_node_test.bzl
benlesh: Okay... I've added that feature as well.

packages/jasmine/src/jasmine_runner.js
benlesh: I'm skipping this for now. I've added two optional arguments: `randomize` and `seed` along with documentation comments.

benlesh:
@alexeagle Is there anything else you need me to do on this PR?

qzmfranklin:
@benlesh @alexeagle 

I created a PR [here](https://github.com/bazelbuild/rules_nodejs/pull/744) that adds the ability to configure arbitrary jasmine configs from json files.

Also, I was also concerned with fixing the seed to begin with. Looks like we took different paths.

I would be happy with either of our PRs merged if no further issues arise that demand tweaking the jasmine configs.

alexeagle:
I prefer #744 since we don't introduce new user-config surface area, just re-use the one Jasmine already has

PR #704, opened 2019-04-20
Symlink user node_modules in yarn_install & npm_install

gregmagolan:
This change makes yarn_install and npm_install use the `node_modules` folder in the user's directory (using a symlink to it in the external repository).

Bazel will rename all WORKSPACE, BUILD and BUILD.bazel files in fetched npm packages that contain these files to _WORKSPACE, _BUILD and _BUILD.bazel. Bazel workspaces such as @bazel/typescript will contain BUILD files but so do some may other packages. This is done to make room for the generated BUILD files that will be added to each npm package.

These generated BUILD files will be visible in the user's `node_modules` so this may surprise some users.

One new condition with this change of this is that two different yarn_install / npm_install rules can't use the same package.json or they end up using the same node_modules folder which can lead to filesystem issues. See https://github.com/bazelbuild/rules_nodejs/pull/708#issuecomment-485177627 for more details.

Another issue that user's may run into after this lands is a `node_modules` `.bazelignore` pattern only matches the root `node_modules` folder so a user with nested package.json files who has existing yarn_install & npm_install rules may run into unexpected build failures as the new `node_modules` folders that show up with this change won't be ignored by Bazel and they may contain BUILD files that should be. See https://github.com/bazelbuild/bazel/issues/8106 for more details.

This can't be landed until the `refresh()` (https://github.com/bazelbuild/bazel/pull/7752, https://github.com/bazelbuild/bazel/pull/7782) feature makes into Bazel which allows Bazel to track `node_modules` directories in user space and run or re-run the `yarn_install` or `npm_install` when the directory timestamp changes or it is deleted.

* refresh root syntax: https://docs.google.com/document/d/1V47Uk0JKzYWHJpS47gwZ3dxTAQMEaE6wHtCr8MtcEds/edit#
* refresh design doc: https://docs.google.com/document/d/1uaeQpPrSH4q46zGuXLgtzMkLVXk7Er9I-P_-ScwwGLk/edit#heading=h.xgjl2srtytjt

As this is a risky change we may want to have it be an opt-in initially before making it the default.


gregmagolan:
See #708 for investigation into the culprit of the BUILD files filesystem symlink issue

gregmagolan:
This is pretty much ready to go now as soon as `refresh()` makes it into a Bazel release. The only thing I can think of that still needs to be done is the improve the logic with renamed _BUILD files so that generated BUILD files that already exist from a past install run don't get renamed. The rename logic should check if the BUILD file in question is a generated one or one that was fetched from npm.

internal/npm_install/generate_build_file.js
```diff
@@ -287,20 +310,27 @@ def _maybe(repo_rule, name, **kwargs):
...
+        /^_BUILD\.bazel$/i.test(basename)) {
+      destFile = path.posix.join(path.dirname(file), basename.substr(1));
+    }
     const src = path.posix.join('node_modules', bwDetails.pkg._dir, file);
-    const dest = path.posix.join(workspaceSourcePath, file);
+    const dest = path.posix.join(workspaceSourcePath, destFile);
     mkdirp(path.dirname(dest));
-    console.error(`copying ${src} -> ${dest}`);
     fs.copyFileSync(src, dest);
+    console.error(src, dest);
```
kyliau: for debugging only?

internal/node/generate_build_file.js
```diff
@@ -44,9 +44,15 @@ function mkdirp(p) {
...
-  fs.writeFileSync(filePath, contents);
+/**
+ * Writes a file, first ensuring that the directory to
+ * write to exists, and flushes it to disk.
+ */
+function writeFileSync(p, content) {
+  mkdirp(path.dirname(p));
+  const fd = fs.openSync(p, 'w');
+  fs.writeSync(fd, content);
+  fs.fdatasyncSync(fd);
```
kyliau: should `fs.closeSync(fd)` be called after this?

internal/npm_install/generate_build_file.js
```diff
@@ -675,6 +682,17 @@ function filterFiles(files, exts = []) {
   // Files with spaces (\x20) or unicode characters (<\x20 && >\x7E) are not allowed in
   // Bazel runfiles. See https://github.com/bazelbuild/bazel/issues/4327
   files = files.filter(f => !f.match(/[^\x21-\x7E]/));
+  files = files.filter(f => {
+    const basename = path.basename(f);
+    // Bazel files of non-nested node_modules would have been renamed earlier with a _ prefix
+    if (/^WORKSPACE$/i.test(basename) || /^BUILD$/i.test(basename) ||
```
kyliau: nit: Could potentially test the underscore prefix in the same regex

internal/npm_install/generate_build_file.js
gregmagolan: 👍

internal/node/generate_build_file.js
gregmagolan: good catch thanks

internal/npm_install/generate_build_file.js
gregmagolan: 👍

internal/npm_install/generate_build_file.js
```diff
@@ -704,6 +711,17 @@ function filterFiles(files, exts = []) {
   // Files with spaces (\x20) or unicode characters (<\x20 && >\x7E) are not allowed in
   // Bazel runfiles. See https://github.com/bazelbuild/bazel/issues/4327
   files = files.filter(f => !f.match(/[^\x21-\x7E]/));
+  files = files.filter(f => {
+    const basename = path.basename(f);
+    // Bazel files of non-nested node_modules would have been renamed earlier with a _ prefix
+    // so we test for both
+    if (/^_?WORKSPACE$/i.test(basename) || /^_?BUILD$/i.test(basename) ||
+        /^_?BUILD\.bazel$/i.test(basename)) {
```
clydin: Three case insensitive RegExp are probably not particularly efficient.  Something similar to the following may be more efficient and more obvious as to intention.
```
switch (path.basename(f).toUpperCase()) {
  case '_WORKSPACE':
  case 'WORKSPACE':
  case '_BUILD':
  case 'BUILD':
  case '_BUILD.BAZEL':
  case 'BUILD.BAZEL':
    return true;
  default:
    return false;
}
```

gregmagolan: Good call. I got rid of the regex after some refactoring.

gregmagolan:
Note: buildkite currently failing as Bazel 0.26.0-rc is required for the build and buildkite is still on 0.25.1.

internal/npm_install/generate_build_file.js
```diff
@@ -425,40 +466,23 @@ function listFiles(rootDir, subDir = '') {
             return isDirectory ? files.concat(listFiles(rootDir, relPath)) : files.concat(relPath);
           },
           [])
+      // Files with spaces (\x20) or unicode characters (<\x20 && >\x7E) are not allowed in
+      // Bazel runfiles. See https://github.com/bazelbuild/bazel/issues/4327
+      .filter(f => !f.match(/[^\x21-\x7E]/))
```
kyliau: Using RegExp.prototype.test() here is more efficient, since match() returns an array whereas test() returns boolean value.

internal/npm_install/generate_build_file.js
```diff
@@ -518,9 +542,8 @@ function findScopes() {
 function parsePackage(p) {
   // Parse the package.json file of this package
   const packageJson = path.posix.join(p, 'package.json');
-  const pkg = isFile(packageJson) ?
-      JSON.parse(fs.readFileSync(`${p}/package.json`, {encoding: 'utf8'})) :
-      {version: '0.0.0'};
+  const pkg = isFile(packageJson) ? JSON.parse(fs.readFileSync(packageJson, {encoding: 'utf8'})) :
```
kyliau: Suggestion for a follow up fix:
The `package` object has gotten complex enough that I think it should be a Class, especially since this is Javascript, it's hard to know what custom properties are added to the object later on.

internal/npm_install/npm_install.bzl
```diff
@@ -79,34 +75,57 @@ COMMON_ATTRIBUTES = dict(dict(), **{
...
+        
+        When true, the package manager will run in the package.json folder
+        and the resulting node_modules folder will be symlinked into the
+        external repository create by this rule.
+        
+        When false, the package manager will run in the external repository
+        created by this rule and any files other than the package.json file and
+        the lock file that are required for it to run should be listed in the
+        data attribute.""",
+        default = True,
```
kyliau: Should the behavior of nested node modules be mentioned here as well?

internal/npm_install/generate_build_file.js
gregmagolan: Yea. I agree. This file has grown organically and needs refactoring. But I think the additional diff would be too great for this PR so we should refactor to classes & typescript in a follow up.

internal/npm_install/generate_build_file.js
gregmagolan: 👍

internal/npm_install/npm_install.bzl
gregmagolan: Not sure what you mean

internal/npm_install/generate_build_file.js
gregmagolan: Good catch thanks

internal/npm_install/generate_build_file.js
kyliau: Yes, I meant to say that the refactoring should be considered in a follow up PR :)

internal/npm_install/npm_install.bzl
kyliau: I thought the comment below was somehow referring to nested node modules. I was totally wrong!
```
# Do not symlink node_modules as when used in downstream repos we should not create
# node_modules folders in the @build_bazel_rules_nodejs external repository. This is
# not supported by managed_directories.
```

internal/npm_install/generate_build_file.js
clydin: Also to consider is the multiple filesystem accesses being done on this line (looks like 3 total).  Going from user to kernel space (and back again) is expensive.  Catching exceptions on the `readFileSync` call would be the most efficient in this regard and also prevents the possibility of an error due to changes in the filesystem between the calls.

.circleci/config.yml
```diff
@@ -11,7 +11,7 @@
 # If you change the `default_docker_image` version, also change the `cache_key` version
 var_1: &default_docker_image circleci/node:10.12
 var_2: &browsers_docker_image circleci/node:10.12-browsers
-var_3: &cache_key node-0.12-{{ checksum "yarn.lock" }}-{{ checksum "examples/program/yarn.lock" }}-{{ checksum "internal/npm_install/test/package/yarn.lock" }}
+var_3: &cache_key node-0.12-{{ checksum "yarn.lock" }}-{{ checksum "examples/program/yarn.lock" }}-{{ checksum "internal/npm_install/test/package/yarn.lock" }}-bust3
```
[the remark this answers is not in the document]
gregmagolan: Yes. Because the user node_modules are used and there were there were breaks during development after updating the package.json & lock files (lock files are cache keys). I needed to bust the cache a few times I as moved along.

gregmagolan:
Related Bazel issues:

* Managed directories should rerun repository rule if directory is deleted or timestamp changes: https://github.com/bazelbuild/bazel/issues/8487; tracking issue yarn_install & npm_install with managed directories can't handle deleted or manually regenerated node_modules folder: https://github.com/bazelbuild/rules_nodejs/issues/802
* Managed directories has issue with node_modules/.bin folder on Windows: https://github.com/bazelbuild/bazel/issues/8491
* Managed directories crash: https://github.com/bazelbuild/bazel/issues/8488

PR #710, opened 2019-04-21
feat(rollup): provide es5 and minified umd bundles

### LINKED ISSUE #685: rollup_bundle: provide es5 UMD bundle
Currently we derive the umd output by running rollup on the es2015 sources:
https://github.com/bazelbuild/rules_nodejs/blob/11271418a6bbd2529170270a7e61dcc5167bb16d/internal/rollup/rollup_bundle.bzl#L490

For parity with currently published zone.js, we'd like an es5 UMD bundle. Perhaps it could be produced by running a tool like babel on the es2015 output, but this would be different from other outputs. More similar would be to create it from the output of _run_tsc

/cc @jialipassion

jbedard:
Closes #685

Would be nice to have better e2e tests for all output formats imo, but I think that's a separate task. Same with fixing the es5 sourcemaps...


jbedard:
Basically just add a golden file for the for the es5 UMD?

jbedard:
Updated to add the test + golden files for `es5umd` and `min.es5umd`.

Were there any naming preferences for something other then "es5umd"?

PR #711, opened 2019-04-21
Don’t gitignore /internal/npm_install/test/golden/node_modules

gregmagolan:
This fixes a few pain points for local development.

1. golden files like `internal/npm_install/test/golden/node_modules/@angular/core/BUILD.bazel.golden` are git ignored since they are in a `node_modules` folder. This PR fixes that by adding the negate pattern `!/internal/npm_install/test/golden/node_modules` to gitignore.

2. clean scripts now recursively delete *all* node_modules folders in the repo since `.bazelignore` doesn't behave like `.gitignore` and its `node_modules` entry only tells bazel to ignore the root `/node_modules` folder. Each nested folder must be listed explicitly in `.bazelignore`.

Will file an issue about the `.bazelignore` behavior to Bazel about this soon as it will cause issues for all users of nodejs rules that have nested `node_modules` and multiple `package.json` files (especially once https://github.com/bazelbuild/rules_nodejs/pull/704 lands).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ rules_nodejs-*.tar.gz
 /internal/e2e/*/bazel-*
 /internal/npm_install/test/package/*
 node_modules
+!/internal/npm_install/test/golden/node_modules
 examples/vendored_node/node-v10.12.0-linux-x64
 examples/vendored_node/node-v10.12.0-linux-x64.tar.xz
 examples/vendored_node/yarn-v1.10.0
--- a/scripts/clean_all.sh
+++ b/scripts/clean_all.sh
@@ -11,9 +11,7 @@ cd ${RULES_NODEJS_DIR}
 echo_and_run() { echo "+ $@" ; "$@" ; }
 
 echo_and_run rm -rf ./dist
-echo_and_run rm -rf ./node_modules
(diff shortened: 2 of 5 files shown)


alexeagle:
Why not change the .gitignore file to be rooted at root instead? Making it match .bazelignore semantics?

gregmagolan:
> Why not change the .gitignore file to be rooted at root instead? Making it match .bazelignore semantics?

Because there are a lot of nested node_modules folders in nested WORKSPACES as well as the ones generated by `bazel run @nodejs//:yarn`:

```
node_repositories(
    package_json = [
        "//:package.json",
        "@examples_program//:package.json",
        "//internal/npm_install/test:package/package.json",
    ],
)
```

... and with the user node_modules symlink change every yarn_install & npm_install rule will generate one as well. Even if they are all listed explicitly in both bazelignore and gitignore it would still be an issue when switching between branches if folders got moved around.

I submitted an issue to Bazel re: the bazelignore semantics. https://github.com/bazelbuild/bazel/issues/8106. Ideally the semantics would match.

gregmagolan:
Oh... I can use `!` in `.gitignore` to accomplish this

PR #713, opened 2019-04-21
Update to buildifier 0.22.0 and fix new lint warnings

gregmagolan:
Don’t check no-effect warning since doc attribute causes a build error in Ubuntu 14.04 only. Need to update to stardoc. See https://github.com/bazelbuild/buildtools/issues/471 for more details.


internal/web_package/web_package.bzl
```diff
@@ -1,8 +1,38 @@
...
+
+"""Contains the web_package rule.
+"""
+
+def html_asset_inject(index_html, action_factory, injector, root_dirs, assets, output):
+    """Injects JS and CSS resources into the index.html.
+
+    Args:
+      index_html: the input html file
+      action_factory: TODO
```
gregmagolan: @alexeagle Can you comment with best descriptions for these?

internal/web_package/web_package.bzl
```diff
@@ -15,6 +45,18 @@ def html_asset_inject(index_html, action_factory, injector, rootDirs, assets, ou
     return output
 
 def move_files(output_name, files, action_factory, assembler, root_paths):
+    """Moves files into an output directory
+
+    Args:
+      output_name: the name of the output directory
+      files: the files to move
+      action_factory: TODO
```
gregmagolan: @alexeagle Can you comment with best descriptions for these?

internal/common/node_module_info.bzl
```diff
@@ -38,6 +38,8 @@ NodeModuleInfo = provider(
 # NodeModuleSources provider is provided by targets that are npm dependencies by the
 # `node_module_library` rule as well as other targets that have direct or transitive deps on
 # `node_module_library` targets via the `collect_node_modules_aspect` below.
+# TODO: rename to NodeModuleSourcesInfo so name doesn't trigger name-conventions warning
```
gregmagolan: @alexeagle It looks like the provider names ending with `Info` is the convention or it trips up the `name-conventions` warnings although I couldn't find that anywhere in the Bazel docs at first glance except for under the `rule` docs:

```
provides | ... Each element of the list is an `*Info` object returned by provider(), except that a legacy provider is represented by its string name instead.
```

alexeagle: sure, but let's not spend too much effort satisfying a linter, the benefits are so small

internal/npm_install/npm_install.bzl
```diff
@@ -226,6 +226,9 @@ npm_install = repository_rule(
     implementation = _npm_install_impl,
 )
 """Runs npm install during workspace setup."""
+# Adding the above docstring as `doc` attribute causes a build
```
alexeagle: I raised this with @c-parsons

stardoc also doesn't generate any docs for repository rules which is why I haven't switched us over from skydoc

internal/web_package/web_package.bzl
alexeagle: I was trying to use Law of Demeter here. We usually pass the whole ctx object but don't use it - plus the ctx object under an Aspect is different.

The doc here could be `Bazel's actions module from ctx.actions - see https://docs.bazel.build/versions/master/skylark/lib/actions.html`

internal/web_package/web_package.bzl
alexeagle: same - it's `ctx.actions`

package.json
```diff
@@ -49,7 +49,15 @@
...
         "test_packages_all": "./scripts/test_packages_all.sh",
         "test_packages": "./scripts/test_packages.sh",
-        "bazel:format": "find . -type f \\( -name \"*.bzl\" -or -name WORKSPACE -or -name BUILD -or -name BUILD.bazel \\) ! -path \"*/node_modules/*\" ! -path \"./dist/*\" ! -path \"./third_party/*\" | xargs buildifier -v --warnings=args-order,attr-cfg,attr-license,attr-non-empty,attr-output-default,attr-single-file,constant-glob,ctx-actions,ctx-args,depset-iteration,depset-union,dict-concatenation,duplicated-name,filetype,git-repository,http-archive,integer-division,load,load-on-top,native-build,native-package,out-of-order-load,output-group,package-name,package-on-top,positional-args,redefined-variable,repository-name,same-origin-load,string-iteration,unsorted-dict-items,unused-variable",
+        "// Unchecked warnings": "The following warnings are not checked as disabling them locally is broken",
+        "//   rule-impl-return": "./internal/common/typescript_mock_lib.bzl:29: rule-impl-return: Avoid using the legacy provider syntax. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#rule-impl-return)",
```
alexeagle: gross, we really need a `.bazellintrc`

internal/npm_install/npm_install.bzl
c-parsons: Stardoc does generate docs for repository rules as of the current release. repository_rule also has a "doc" parameter as of the most recent Bazel release.

internal/common/node_module_info.bzl
gregmagolan: Ok. Will rename in a separate PR. It's only used in a few places and it's better to stick to naming conventions.

internal/npm_install/npm_install.bzl
gregmagolan: Great. This repo just got updated to 0.24.1 so we can update to stardoc which will also fix this issue: https://github.com/bazelbuild/buildtools/issues/471#issuecomment-485283200

gregmagolan: Will leave that for another PR

internal/web_package/web_package.bzl
gregmagolan: Thanks

PR #714, opened 2019-04-22
Rename NodeModuleSources to NodeModulesSourcesInfo per naming convention

gregmagolan:
BREAKING CHANGE:

Although this name is an internal detail and is not used anywhere outside of this repository, since we publish multiple packages this change will still mean that an older @bazel/typescript npm package will not work with the next release of build_bazel_rules_nodejs and vice-versa.

Next release should be a minor bump with this.

There are some other breaking changes that would be good to get in all at once if we are going to bump the minor:

* remove COMPAT_VERSION & maybe VERSION as well from npm published workspaces; with the peer deps convention we don't really need version checks anymore
* remove `defs.bzl` from published npm packages since `index.bzl` is standard
* remove deprecated `preserve_symlinks` option
* make package_lock_json mandatory
* remove `jasmine_node_test` from `build_bazel_rules_nodejs`
* remove deprecated `internal/collect_es6_sources.bzl`
* remove deprecated `internal/node.bzl`

Also this but it could be a stretch so probably better to think this one through more since it will be a harder one for users to upgrade from:

* remove deprecated `node_modules` attributes (nodejs_binary, nodejs_test, rollup_bundle)? with the addition of the `node_module_library()` rule, we could make `//:node_modules` work within deps/data instead of having a separate `node_modules` attributes.

### DIFF
--- a/internal/common/node_module_info.bzl
+++ b/internal/common/node_module_info.bzl
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""NodeModuleInfo & NodeModuleSources providers and apsect to collect node_modules from deps.
+"""NodeModuleInfo & NodeModuleSourcesInfo providers and apsect to collect node_modules from deps.
 """
 
 # NodeModuleInfo provider is only provided by targets that are npm dependencies by the
@@ -35,12 +35,11 @@ NodeModuleInfo = provider(
     },
 )
 
-# NodeModuleSources provider is provided by targets that are npm dependencies by the
+# NodeModuleSourcesInfo provider is provided by targets that are npm dependencies by the
 # `node_module_library` rule as well as other targets that have direct or transitive deps on
 # `node_module_library` targets via the `collect_node_modules_aspect` below.
 # TODO: rename to NodeModuleSourcesInfo so name doesn't trigger name-conventions warning
-# buildozer: disable=name-conventions
-NodeModuleSources = provider(
+NodeModuleSourcesInfo = provider(
     doc = "Provides sources for npm dependencies installed with yarn_install and npm_install rules",
     fields = {
         "scripts": "Source files that are javascript named-UMD or named-AMD modules for use in rules such as ts_devserver",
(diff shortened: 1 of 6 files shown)


packages/typescript/internal/build_defs.bzl
```diff
@@ -14,7 +14,7 @@
 
 "TypeScript compilation"
 
-load("@build_bazel_rules_nodejs//internal/common:node_module_info.bzl", "NodeModuleInfo", "NodeModuleSources", "collect_node_modules_aspect")
+load("@build_bazel_rules_nodejs//internal/common:node_module_info.bzl", "NodeModuleInfo", "NodeModuleSourcesInfo", "collect_node_modules_aspect")
```
gregmagolan: @alexeagle Hmmmm. I wonder if we should be avoiding access internal paths in `@build_bazel_rules_nodejs` from npm published repositories such as `@npm_bazel_typescript` to avoid internal changes creating breakages like this one.

internal/common/node_module_info.bzl
```diff
@@ -35,12 +35,11 @@ NodeModuleInfo = provider(
     },
 )
 
-# NodeModuleSources provider is provided by targets that are npm dependencies by the
+# NodeModuleSourcesInfo provider is provided by targets that are npm dependencies by the
 # `node_module_library` rule as well as other targets that have direct or transitive deps on
 # `node_module_library` targets via the `collect_node_modules_aspect` below.
 # TODO: rename to NodeModuleSourcesInfo so name doesn't trigger name-conventions warning
-# buildozer: disable=name-conventions
```
gregmagolan: Can remove this disabled warning check with this

packages/typescript/internal/build_defs.bzl
gregmagolan: Maybe add a pre-commit hook that checks for this condition as well when we fix it

gregmagolan:
Not worth it at the moment. Maybe open this up again in the future and bundle it with other breaking changes.

PR #721, opened 2019-04-24
Vendor in third party npm packages to /third_party/npm...

gregmagolan:
…and automatically generate build files for them

No additional third party npm deps added with this PR.


gregmagolan:
Yes. There is not much benefit at this point and vendoring in browserify from npm was not feasible due to its size so instead what was vendored in was a ncc'd & terser'd index.min.js of browserify.

Would be good to clean up the vendoring of `third_party/github.com/gjtorikian/isBinaryFile/src/index.js` at some point which is a generated file from `third_party/github.com/gjtorikian/isBinaryFile/src/index.ts` which comes from github but not that important right now. The npm artifact for isBinaryFile has the transpiled `index.js` which we want.

PR #726, opened 2019-04-26
Return typescript legacy provider from node_modules_library with the …

gregmagolan:
…npm package's typescript declaration files.

Fixes https://github.com/bazelbuild/rules_typescript/issues/381 -- `.d.ts` files from npm deps now show up in generated tsconfig file:

```
"allowedStrictDeps": ["external/npm/node_modules/source-map/source-map.d.ts", "test/things.ts"]
```

Once released & pushed downstream the following change will need to be made in `ng_module.bzl` in angular/angular:
```
- deps = [d for d in ctx.attr.deps if not NodeModuleInfo in d],
+ deps = ctx.attr.deps,
```
in order to fix this for `ng_module` as well

### DIFF
--- a/internal/common/sources_aspect.bzl
+++ b/internal/common/sources_aspect.bzl
@@ -33,14 +33,15 @@ def _sources_aspect_impl(target, ctx):
     # get TypeScript outputs.
     if hasattr(target, "typescript"):
         node_sources = depset(transitive = [node_sources, target.typescript.es5_sources])
-    elif NodeModuleSources in target:
-        dev_scripts = depset(transitive = [dev_scripts, target[NodeModuleSources].scripts])
     elif hasattr(target, "files") and not NodeModuleInfo in target:
         # Sources from npm fine grained deps should not be included
         node_sources = depset(
             [f for f in target.files if f.path.endswith(".js")],
             transitive = [node_sources],
         )
+
+    if NodeModuleSources in target:
+        dev_scripts = depset(target[NodeModuleSources].scripts)
 
     if hasattr(ctx.rule.attr, "deps"):
         for dep in ctx.rule.attr.deps:
--- a/internal/npm_install/node_module_library.bzl
+++ b/internal/npm_install/node_module_library.bzl
@@ -19,27 +19,60 @@ load("@build_bazel_rules_nodejs//internal/common:node_module_info.bzl", "NodeMod
 
 def _node_module_library_impl(ctx):
     workspace = ctx.label.workspace_root.split("/")[1] if ctx.label.workspace_root else ctx.workspace_name
+
+    # All files in `srcs` and in `deps`
+    # TODO(gregmagolan): transitive sources should be collected an aspect to go
+    # into a NodeModuleSources.transitive_sources
     sources = depset(ctx.files.srcs, transitive = [dep.files for dep in ctx.attr.deps])
 
-    scripts = depset()
+    # scripts are a subset of sources that are javascript named-UMD or named-AMD scripts for
+    # use in rules such as ts_devserver
+    scripts = depset(ctx.files.scripts)
+
+    # TODO(gregmagolan): fix deps in generate_build_file.js and remove transitive lookup
     for src in ctx.attr.srcs:
         if NodeModuleSources in src:
             scripts = depset(transitive = [scripts, src[NodeModuleSources].scripts])
-    scripts = depset(ctx.files.scripts, transitive = [scripts])
 
-    return [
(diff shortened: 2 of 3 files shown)


gregmagolan:
Once released & pushed downstream the same change will need to be made in angular/angular `ng_module.bzl` on the `deps = [d for d in ctx.attr.deps if not NodeModuleInfo in d]` to fix https://github.com/bazelbuild/rules_typescript/issues/381 for ng_module deps

internal/npm_install/node_module_library.bzl
```diff
@@ -27,19 +27,35 @@ def _node_module_library_impl(ctx):
...
-            files = sources,
-        ),
-        NodeModuleInfo(
-            workspace = workspace,
-        ),
-        NodeModuleSources(
-            sources = sources,
-            scripts = scripts,
-            workspace = workspace,
+    declarations = [f for f in ctx.files.srcs if f.path.endswith("d.ts")]
```
alexeagle: this matches a bit too much like `pond.ts`

packages/typescript/internal/build_defs.bzl
```diff
@@ -252,13 +252,10 @@ def _ts_library_impl(ctx):
     ts_providers = compile_ts(
         ctx,
         is_library = True,
-        # Filter out the node_modules from deps passed to TypeScript compiler
-        # since they don't have the required providers.
-        # They were added to the action inputs for tsc_wrapped already.
-        # strict_deps checking currently skips node_modules.
+        # Strict_deps checking currently skips node_modules.
```
alexeagle: I think we also should remove the corresponding workaround
https://github.com/bazelbuild/rules_typescript/blob/0f7c6f624c7ac56880d466ea81d83a57c72bf963/internal/build_defs.bzl#L102-L106
which I think means this TODO will also be fixed - node modules will only be action inputs when declared as deps

internal/npm_install/node_module_library.bzl
gregmagolan: 🤦‍♂ good catch. thanks.

packages/typescript/internal/build_defs.bzl
gregmagolan: Want to remove this whole comment at this point then or wait until the upstream change is done first?

internal/common/sources_aspect.bzl
```diff
@@ -33,15 +33,16 @@ def _sources_aspect_impl(target, ctx):
...
-    elif NodeModuleSources in target:
-        dev_scripts = depset(transitive = [dev_scripts, target[NodeModuleSources].scripts])
     elif hasattr(target, "files") and not NodeModuleInfo in target:
         # Sources from npm fine grained deps should not be included
         node_sources = depset(
             [f for f in target.files if f.path.endswith(".js")],
             transitive = [node_sources],
         )
 
+    if NodeModuleSources in target:
```
gregmagolan: Necessary to move out of if/elseif clause since target can now have both "typescript" and NodeModuleSources

packages/typescript/internal/build_defs.bzl
alexeagle: Either way, let's do it

gregmagolan: Ok. I'll send a PR to rules_typescript after I merge this

internal/npm_install/generate_build_file.js
```diff
@@ -768,7 +768,7 @@ node_module_library(
     # ${pkg._dir} package contents (and contents of nested node_modules)
     srcs = [
         ":${pkg._name}__files",
-    ],${depsStarlark}
+    ],${depsStarlark}${scriptStarlark}
```
gregmagolan: Duplicating the scripts in __files and __pkg is more correct than having the transitive

```
        if NodeModuleSources in src:
            scripts = depset(transitive = [scripts, src[NodeModuleSources].scripts])
```

lookup in `node_module_library.bzl` since __files is mainly a construct for avoiding circular deps and it should not be generally used.

gregmagolan: Moving `scripts` to __pkg allows removing

```
    for src in ctx.attr.srcs:
        if NodeModuleSources in src:
            scripts = depset(transitive = [scripts, src[NodeModuleSources].scripts])
```

from node_module_library.bzl. __pkg is the standard dep for users and if a user is using __files then they are being even more fine grained with their deps and probably don't want scripts automatically ending up in ts_devserver & ts_web_test_suite

PR #736, opened 2019-05-06
No --types for rollup_bundle tsc downleveling...

gregmagolan:
& make --libs consistent between tsc and tsc-directory.

This fixes an issue observed where types files could break downleveling action if they used higher level language constructs

### DIFF
--- a/internal/rollup/rollup_bundle.bzl
+++ b/internal/rollup/rollup_bundle.bzl
@@ -224,7 +224,15 @@ def _run_rollup(ctx, sources, config, output, map_output = None):
 
 def _run_tsc(ctx, input, output):
     args = ctx.actions.args()
+
+    # No types needed since we are just downleveling.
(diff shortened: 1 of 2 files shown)


internal/rollup/rollup_bundle.bzl
```diff
@@ -224,7 +224,14 @@ def _run_rollup(ctx, sources, config, output, map_output = None):
 
 def _run_tsc(ctx, input, output):
     args = ctx.actions.args()
+
+    # No types needed since we are just downleveling.
+    # `--types` proceeded by another config argument means an empty types array
+    # for the command line parser.
+    # See https://github.com/Microsoft/TypeScript/issues/18581#issuecomment-330700612
+    args.add("--types")
```
alexeagle: add `skipLibCheck` too?

gregmagolan: good idea. done

PR #744, opened 2019-05-08
[feat] support config json file in jasmine_node_test

qzmfranklin:
This commit enables the `jasmine_node_test` rule to be configured on a
per-test basis through jasmine's configuration file.

The demanding situation for me was to turn off randomnization (default
to on in Bazel) to guarantee some test results.  This would incur adding
the following attribute to jasmine's configuration file:
    ```json
    {
        "random": false
    }
    ```

### DIFF
--- a/internal/jasmine_node_test/jasmine_runner.js
+++ b/internal/jasmine_node_test/jasmine_runner.js
@@ -63,6 +63,9 @@ function main(args) {
   process.argv.splice(2, 1)[0];
 
   const jrunner = new JasmineRunner({jasmineCore: jasmineCore});
+  if (args.length == 2) {
+    jrunner.loadConfigFile(args[1])
+  }
   fs.readFileSync(manifest, UTF8)
       .split('\n')
       .filter(l => l.length > 0)
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -162,6 +162,13 @@ def _nodejs_binary_impl(ctx):
         "TEMPLATED_repository_args": _short_path_to_manifest_path(ctx, ctx.file._repository_args.short_path),
         "TEMPLATED_script_path": script_path,
     }
+
+    files = [node, ctx.outputs.loader, ctx.file._repository_args] + ctx.files._node_runfiles
+
+    if ctx.attr.jasmine_config != None:
+        files += [ctx.file.jasmine_config]
+        substitutions["TEMPLATED_args"] += " " + ctx.attr.jasmine_config.files.to_list()[0].path
+
     ctx.actions.expand_template(
         template = ctx.file._launcher_template,
         output = ctx.outputs.script,
@@ -169,7 +176,7 @@ def _nodejs_binary_impl(ctx):
         is_executable = True,
     )
 
-    runfiles = depset([node, ctx.outputs.loader, ctx.file._repository_args] + ctx.files._node_runfiles, transitive = [sources, node_modules])
+    runfiles = depset(files, transitive = [sources, node_modules])
 
     return [DefaultInfo(
         executable = ctx.outputs.script,
@@ -196,6 +203,15 @@ _NODEJS_EXECUTABLE_ATTRS = {
         zone.js before the first `describe`.
         """,
         default = [],
+    ),
+    "jasmine_config": attr.label(
+        doc = """The path to the json configuration file of jasmine.
+        For example, you can turn off randomnization via the following flag in
+        the json config file:
+            "random": false
+        """,
+        allow_files = True,
+        single_file = True
     ),
     "configuration_env_vars": attr.string_list(
         doc = """Pass these configuration environment variables to the resulting binary.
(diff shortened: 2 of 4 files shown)


qzmfranklin:
This is something we've been using internally at LogiOcean for a few months. It served us well so we wanted to see if it could also benefit others.

Have not added tests yet (though I know it works because of tests internal to us). Would like to hear feedback before doing that.

Thanks for everything and please let me know what I could do to help you help me make it through.

internal/jasmine_node_test/jasmine_runner.js
```diff
@@ -63,6 +63,9 @@ function main(args) {
   process.argv.splice(2, 1)[0];
```
alexeagle: we take care here to pass the residual argv as-is to jasmine. Will adding a new config file argument potentially confuse any test code that reads the argv? I wish the comment here gave more info about scenarios where the tested code processes the argv...

internal/jasmine_node_test/jasmine_runner.js
```diff
@@ -63,6 +63,9 @@ function main(args) {
...
 
   const jrunner = new JasmineRunner({jasmineCore: jasmineCore});
+  if (args.length == 2) {
+    jrunner.loadConfigFile(args[1])
```
alexeagle: what happens when the user sets config options that conflict with how we configure jasmine below? if you try to add your own specs, for example? would be good to understand how the rule will behave under all the possible config options, and for those that we override, how will the user know?

internal/node/node.bzl
```diff
@@ -162,14 +162,21 @@ def _nodejs_binary_impl(ctx):
         "TEMPLATED_repository_args": _short_path_to_manifest_path(ctx, ctx.file._repository_args.short_path),
         "TEMPLATED_script_path": script_path,
     }
+
+    files = [node, ctx.outputs.loader, ctx.file._repository_args] + ctx.files._node_runfiles
```
alexeagle: files is not a very descriptive identifier here. most things in bazel are files.

internal/node/node.bzl
```diff
@@ -197,6 +204,15 @@ _NODEJS_EXECUTABLE_ATTRS = {
         """,
         default = [],
     ),
+    "jasmine_config": attr.label(
+        doc = """The path to the json configuration file of jasmine.
```
alexeagle: instead of an example, should link to their docs

qzmfranklin: Sorry, I wasn't aware of this requirement.
I will look harder at CONTRIBUTING.md (if any) next time

internal/node/node.bzl
qzmfranklin: Humm, I admit, this var name is awful.

qzmfranklin:
@alexeagle 

Thanks for your time reviewing! Really appreciated! Apology as it took a while for me to get back to this.


Drawing on your comments on using an attribute instead of a config file to spec jasmine, I think the trade-off is:

Using a config file:
- PRO: Easy to add arbitrary configs without having to patch the rules_nodejs rule itself.
- CON: Could lead to confusion if command line argument overrides config file's key-value pairs.

Using command line args: Just the opposite to using a config file.

I think it really gets down to the question below:
> How often are developers urged to config jasmine in a non-standard manner?

So far, I have only encountered `random`, as was the motivation for this PR. Back at the time, due to my limited knowledge on jasmine, I suspected that in the future more custom jasmine configs are likely to be added. To save time in the future, I cooked up the config file approach.

However, by now I am starting to think that using an attribute in the `bzl` file is a better approach. As AFAIK there is really no other demand scenarios.

What's your opinion?

qzmfranklin:
I looked around a bit and came to the following conclusion:

Instead of adding support for `jasmine.config` in `jasmine_node_test`, a better approach would be to add a `random=True` field in a wrapper and pass `random={true,false}` as part of `templated_args` (defined in `nodejs_{binary,test}`.

Close this issue.

matsko:
@qzmfranklin I need this feature as well. Do you plan on making this feature with the workaround you've posted?

qzmfranklin:
Accidentally closed. Reopen to discuss more.

qzmfranklin:
@alexeagle 

I'm sorry that my closing this PR caused confusions. Working on it now to address some of your comments. Will reply on others when I'm done.

internal/jasmine_node_test/jasmine_runner.js
qzmfranklin: This is a valid concern and request.

I wish there were clearer handling of command line args in code instead of in comment though while working on this chunk of code.

alexeagle:
saw that you commented on this on Slack again a week ago - I'll take another stab at getting this in.

alexeagle:
Hi, I wasn't able to push to this branch, so I moved your commit to a new PR, rebased on master, and fixed it up with tests. https://github.com/bazelbuild/rules_nodejs/pull/962

PR #746, opened 2019-05-08
Structural changes for npm strict deps support

gregmagolan:
These are all the changes required in this repo to get npm strict deps to work.

We have a chicken & egg problem to land this so we need to:

1) Land this PR
2) Update the rules_nodejs commit hash in rules_typescript and land https://github.com/bazelbuild/rules_typescript/pull/445
3) Update the commit hash in rules_nodejs in another PR

(3) will be a breaking change and downstream users that are depending on transitive @npm deps in their ts_library or ng_module rules will see errors that look like so:

```
main.spec.ts:1:21 - error TS2307: transitive dependency on external/npm/node_modules/typescript/lib/typescript.d.ts not allowed. Please add the BUILD target to your rule's deps.
1 import * as ts from 'typescript';
```

### DIFF
--- a/e2e/typescript_2.7/BUILD.bazel
+++ b/e2e/typescript_2.7/BUILD.bazel
@@ -29,6 +29,7 @@ ts_library(
         "@npm//@bazel/typescript",
         "@npm//@types/jasmine",
         "@npm//@types/node",
+        "@npm//typescript",
     ],
 )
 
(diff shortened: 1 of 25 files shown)


gregmagolan:
Example of compile error that turning on strict npm deps resulted in:

```
main.spec.ts:1:21 - error TS2307: transitive dependency on external/npm/node_modules/typescript/lib/typescript.d.ts not allowed. Please add the BUILD target to your rule's deps.
1 import * as ts from 'typescript';
```

PR #761, opened 2019-05-10
add browser module main priority to amd processor

Toxicable:
## PR Checklist

## PR Type

The entry point for npm_umd_bundle in npm packages is only resolved to `main`

## What is the new behavior?
We now try to resolve, `browser`, `module`, then `main`, then `index.js` and finally `${pkg.__name}.js`

Not sure, possibly
- [ ] Yes
- [] No

### DIFF
--- a/e2e/ts_devserver/BUILD.bazel
+++ b/e2e/ts_devserver/BUILD.bazel
@@ -22,7 +22,11 @@ exports_files([
 ts_library(
     name = "app",
     srcs = ["app.ts"],
-    deps = ["@npm//date-fns"],
+    deps = [
+        "@npm//date-fns",
+        "@npm//typeorm",
+        "@npm//rxjs",
+    ],
 )
 
 ts_devserver(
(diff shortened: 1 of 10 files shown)


gregmagolan:
I was thinking this would be the next improvement to the AMD processor. We may want to add the es6ify plugin with this change as well since some npm packages may use es6 or fesm5 in their distributions.

generate_build_file.js needs some refactoring at some point as well; it started out small and simple and has become rather large overtime.

Toxicable:
@gregmagolan Sounds good to me
The remaining failing test is due to not having the es6ify plugin enabled
So ill hand that over to you to handle the vendoring of that plugin when you got time for it.

gregmagolan:
I took a glace at es6ify and it's a more complicated plugin than the named-amd plugin that I previously vendored in. I'll put this on my list but probably won't have time to get to it for a few weeks.

alexeagle:
@Toxicable any plan how to get this green without es6ify plugin ? or should this go on the backlog?

Toxicable:
@alexeagle Definately needs the es6ify plugin for this one, since the test case I choose; `rxjs` with the es2015 entry point uses esm modules.
So yeah on the backlog for you guys would be the best option I think. I doubt i'd be able to vendor it correctly myself especially since @gregmagolan had a few issues with it aswell.

gregmagolan:
Replaced by #997

PR #763, opened 2019-05-14
Remove deprecated preserve_symlinks option.

Globegitter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: Removal of deprecation

One can set `preserve_symlinks=false` on `node_repositories()` to not preserve symlinks for the node module resolution.

Issue Number: N/A

## What is the new behavior?
The attribute as been removed and symlinks will always be preserved

- [x] Yes
- [ ] No

Users who explicitly set `preserve_symlinks=true` just need to remove it and for users who have `preserve_symlinks=false` there is no migration strategy currently. If we want to keep an option for users we could add some way to turn symlink preservation of we could just add something directly onto `nodejs_binary`. But not sure if that is needed.

## Other information

I have also not yet removed the test for `preserve_symlinks = False,`. If we decide we do not need to support that use-case at all I would just remove it.

The reason I want to remove this deprecated option now is, as it will make the last pieces of #645 simpler.


Toxicable:
I don't think this is a good idea, if we remove this option then we wont be able to upgrade to the next version.
We rely on this option for debugging out nodejs applications, the source maps are broken without it.

Globegitter:
Good to know. Would it be sufficient to allow setting this then on a `nodejs_binary` level? Then you could wrap all your `nodejs_binary` targets that need that in a macro?

The two ways I can imagine this, is either by making use of the `templated_args`. So if you would write
```
nodejs_binary(
  name = "...",
  templated_args = ["--node_options=--nopreserve-symlinks"],
)
```

We could deactivate it. Only problem is, node does not actually accept `nopreserve-symlinks`, so we would only use it to filter out the default flag we set. Another option would be to add a new attribute

```
nodejs_binary(
  name = "...",
  preserve_symlinks = false,
)
```

But I am not very fond of adding yet another attribute.

Another option I can think off is we just  default the `templated_args` option to `["--node_options=--preserve-symlinks"]`, so one could just set it to `[]` - but then also as soon as anyone passes in custom flags (which we have done before) then one would also have to know to set this preserve-symlinks one, which is also not great.

Or actually just thinking about it, part of the reason I wanted to remove this is because it currently lives inside the `@nodejs` repo but there is no need to touch this with toolchains and have this duplicated in all the platform specific repositories. So if we move this out into an new repository rule, something like `node_defaults()` which then creates a new repository `@node_defaults` or similar then we would only have this once and it would also not need to concern itself with toolchains and we can use it to make any node flags default for all binaries. `rules_k8s` has something similar: https://github.com/ecosia/rules_k8s#k8s_defaults

We could probably even make this backwards compatible, by setting it up in the `node_repositories()` all.

I am leaning towards the last solution. WDYT @Toxicable?

cc @gregmagolan @alexeagle

Toxicable:
Any of those solutions would work for me, I don't mind putting in a little bit of effort to make it work.
However, another solution i'd like to propose is that we fix the sourcemaps issues so it dosen't rely on this flag.
That way we wouldn't have a need for this flag at all.

Toxicable:
So i've jsut been trying out rollup_bundle again and as it turns out with `preserve_symlinks = False` on it fails to resolve deps.
Therefore im now very much for making this change :P
But just with a path for how we can continue to be able to debug our nodejs apps at the same time (as before)

alexeagle:
I think we should try to remove the option, and fix the broken sourcemaps.

When I first added the preserve_symlinks I tried one of the other approaches. We could dig through the history to understand that, I sadly don't remember.

Easiest answer is to remove it and then fix the issues reported by users. Since we have been printing a noisy deprecation warning for many releases, I don't see a problem removing it.

@Globegitter what's up with the broken circleci step? just need to delete `e2e/node_loader_no_preserve_symlinks` I guess?

Globegitter:
@Toxicable Are you on the latest version of rules_nodejs? `rollup_bundle` should now preserve symlinks irregardless of the node setting: https://github.com/bazelbuild/rules_nodejs/blob/master/internal/rollup/rollup.config.js#L206

Globegitter:
@alexeagle yep deleted the tests now, so should be ready to merge.

Toxicable:
on 0.30.0 `rollup_bundle` still fails for me with the error below
But when I flip `preserve_symlinks` to `True` it works
```
RROR: /home/f.wiles/workspace/WORKSPACE/apps/bazel-test-app/src/BUILD.bazel:47:1: Bundling JavaScript apps/bazel-test-app/src/bundle.es2015.js [rollup] failed (Exit 1) rollup failed: error ex
ecuting command bazel-out/host/bin/external/build_bazel_rules_nodejs/internal/rollup/rollup --config bazel-out/k8-fastbuild/bin/apps/bazel-test-app/src/_bundle.rollup.conf.js --output.fi
le ... (remaining 5 argument(s) skipped)

[!] (commonjs plugin) Error: Could not resolve import 'lodash' from '/home/f.wiles/.cache/bazel/_bazel_f.wiles/4e95cbad76bfb4ed000904165db39268/execroot/WORKSPACE/bazel-out/k8-fastbuild/bin/ap
ps/bazel-test-app/src/bundle.es6/apps/bazel-test-app/src/main.prod.js'
Error: Could not resolve import 'lodash' from '/home/f.wiles/.cache/bazel/_bazel_f.wiles/4e95cbad76bfb4ed000904165db39268/execroot/WORKSPACE/bazel-out/k8-fastbuild/bin/apps/bazel-test-app/src/
bundle.es6/apps/bazel-test-app/src/main.prod.js'
    at Object.notResolved (/home/f.wiles/.cache/bazel/_bazel_f.wiles/4e95cbad76bfb4ed000904165db39268/execroot/WORKSPACE/bazel-out/k8-fastbuild/bin/apps/bazel-test-app/src/_bundle.rollup.conf.js:163:11)
    at /home/f.wiles/.cache/bazel/_bazel_f.wiles/4e95cbad76bfb4ed000904165db39268/external/build_bazel_rules_nodejs_rollup_deps/node_modules/rollup-plugin-commonjs/src/utils.js:20:68

```

alexeagle:
@Toxicable this PR will make preserve_symlinks always True, so it sounds like that does fix your rollup_bundle so it's okay to merge this?

Toxicable:
@alexeagle Correct.
However, that will then break our sourcemaps, or rather to be more specific, it will cause our source maps to become readonly when debugging nodejs apps in VS Code.
Compared to the current behaviour with `preserve_symlinks = False` where VS Code follows a path to the bazel symlinked version of the file, which is still editable but it's still suboptimnal since it kills the TS language features and git.

Globegitter:
@Toxicable Why do you need source maps to be editable?

Toxicable:
@Globegitter Sorry, I misworded that a bit.
When we set `preserve_symlinks = False` the source file resolved by VS Code is the bazel symlinked version of the source file, not the real source file.

I think I explain it a bit clearer here: https://github.com/bazelbuild/rules_typescript/issues/254

alexeagle:
@Toxicable I still don't understand if you're objecting to this being merged.
I'd like to remove complexity while we can still make these breaking changes.

Toxicable:
My objection to this being merged is because of the source map issues: https://github.com/bazelbuild/rules_nodejs/issues/823
This is why i've been pushing so hard on that the last few days.

If we flipped this flag then debugging Nodejs apps, which is the majority of our use case at the moment would be near impossible.
However, at the same time, this flag is blocking us from pushing forward with using Bazel for our frontend builds.
But if we solved the source map issues it wouldn't be a problem.

Toxicable:
We've worked around needing this option and continue to have source map support with this hack here; https://github.com/bazelbuild/rules_nodejs/issues/823#issuecomment-501468171

alexeagle:
@soldair is working on a `--require` script that would prevent symlink resolves going outside the execroot/runfiles root. This may be a more principled answer, then maybe we allow users to continue to choose how to run node. There may be cases where they have to have it set one way or the other for their program to work?

alexeagle:
current thought is we should just undeprecate the option and let users control it

PR #765, opened 2019-05-15
Now places data relative to package.json

jroylance:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently, Bazel will place data dependencies relative to the project root. 

Issue Number: N/A

## What is the new behavior?

After this change, Bazel will place data dependencies relative to project.json

- [X] Yes
- [ ] No

Ensure that all data is referred to relative to package.json, rather than the project root. 

## Other information


JaredNeil:
@alexeagle, This is a fix for the problem with postinstall scripts I mentioned to you at ng-conf. @jroylance is an intern on our team for the summer.

Previously the `npm_install` and `yarn_install` would break the relative paths between `package.json` and any `data` files if the `package.json` was not at the root of the workspace. This change maintains the relative paths no matter where the `package.json` is.

Unfortunately this probably does count as a breaking change.
One new limitation that this introduces is that all `data` to a `npm_install` must be in or below the package that `package.json` is in, but I think that fits more closely with the intent of a `package.json` since reaching out of the package to find or run files seems very fragile.

alexeagle:
Thanks, this makes sense to me. Could you fix the two broken circleci failures?

Note that we'll soon land #704 which will mostly obviate the need for this - we'll expect that most users don't pass `data[]` to `yarn_install`/`npm_install` since the install will run in the project directory.

alexeagle:
@jroylance now that `data` is unneeded, do you still plan to work on this?

JaredNeil:
No, I don't think it's worth doing any more if installation is actually happening in the source tree. I'm fine with just closing the PR.

Issue #764, opened 2019-05-15
karma_web_test should resolve .umd.js.map files


Toxicable:
# 🚀 feature request

### Relevant Rules
`karma_web_test`

### Describe the solution you'd like
*.umd.js.map files should be automatiically resolved by `karma_web_test`

### Describe alternatives you've considered
manually ading them

gregmagolan:
The change needed is https://github.com/bazelbuild/rules_nodejs/blob/master/internal/npm_install/generate_build_file.js#L750. Just need to add `.umd.js.map` to the array.

PR #722, opened 2019-04-24
add check for source map urls

Toxicable:
## PR Checklist
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
- [x] Refactoring (no functional changes, no api changes)

## What is the current behavior?
There is no test for source map paths

https://github.com/bazelbuild/rules_typescript/issues/321
https://github.com/bazelbuild/rules_typescript/issues/254

## What is the new behavior?
Unchanged, this is only a test

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #767, opened 2019-05-16
add .umd.js.map files to Ng Apf files

Toxicable:
## PR Checklist

## PR Type

Issue Number: N/A

- [ ] Yes
- [ ] No

## Other information
fixes https://github.com/bazelbuild/rules_nodejs/issues/764#issuecomment-492902720

### DIFF
--- a/internal/npm_install/generate_build_file.js
+++ b/internal/npm_install/generate_build_file.js
@@ -738,7 +738,7 @@ function isNgApfPackage(pkg) {
  */
 function getNgApfScripts(pkg) {
   return isNgApfPackage(pkg) ?
-      filterFiles(pkg._files, ['.umd.js', '.ngfactory.js', '.ngsummary.js']) :
+      filterFiles(pkg._files, ['.umd.js', '.umd.js.map', '.ngfactory.js', '.ngsummary.js']) :
       [];
 }
 
--- a/internal/npm_install/test/golden/node_modules/@angular/core/BUILD.bazel.golden
+++ b/internal/npm_install/test/golden/node_modules/@angular/core/BUILD.bazel.golden
@@ -632,7 +632,9 @@ node_module_library(
     srcs = [":core__files"],
     scripts = [
         ":bundles/core-testing.umd.js",
+        ":bundles/core-testing.umd.js.map",
         ":bundles/core.umd.js",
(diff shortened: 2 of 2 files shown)


gregmagolan:
Oh, right. That's doesn't work because the map files end up in the ts_devserver bundle via scripts but actually need to be in data. Thanks for putting this up.

@Toxicable Does adding the maps manually to `data` get you the stack traces you expect?

They should make it there automatically but the question is what is the cleanest path?
1) adds a `scripts_maps` the NodeModuleSources provider and have those picked up by `sources_aspect` into `dev_maps`
2) makes `scripts` in NodeModuleSources both .js & the .map files and have `sources_aspect` put `.js` files into `dev_scripts` and `.map` files into `dev_maps`
3) makes `scripts` in NodeModuleSources both .js & the .map & `dev_scripts` have both & have ts_devserver & ts_web_test_suite sort between .js & .map files in `dev_scripts`

(1) seems least desirable since `sources` in NodeModuleSources is already a mixed bag of files so no reason that `scripts` can be both .js files and .map so I think it's probably either (2) or (3)

Thought?

/cc @alexeagle @kyliau

alexeagle:
We need a failing test that demonstrates a broken source map. Somehow we need a node program that looks like the actual client, and uses source-map-support to assert that the stack traces map back to the right .ts lines

Toxicable:
I started on a PR which would act like a editor to check the paths produced by source maps
https://github.com/bazelbuild/rules_nodejs/pull/722

This could be extended to include colunm and line checks

alexeagle:
@Toxicable could you re-base on https://github.com/bazelbuild/rules_nodejs/pull/848 and fix the test?

gregmagolan:
> @Toxicable could you re-base on #848 and fix the test?

#848 merged :)

Toxicable:
@gregmagolan @alexeagle  Done

Toxicable:
CI failing due to:
>12 06 2019 19:23:14.661:WARN [middleware:karma]: Invalid file type (map), defaulting to js.

Looks like something is picking up the map files, and trying to test them

Toxicable:
probably needs to be rewritten with the changes to `generate_build_file.js`, will do if there's interest in the related ticket

PR #480, opened 2018-12-21
feat: add support for Label in nodejs_binary#entry_point

manekinekko:
Closes #32

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

nodejs_binary#entry_point was a string.

## What is the new behavior?

nodejs_binary#entry_point is now a Label.

## Does this PR introduce a breaking change?

- [x] Yes
- [ ] No

The `entry_point` accepts a target's name as an entry point. If the target is a rule, it should produce the JavaScript entry file that will be passed to the nodejs_binary rule). For example:

 ```
        filegroup(
            name = "entry_file",
            srcs = ["workspace/path/to/entry/file"]
        )
        nodejs_binary(
          name = "my_binary",
          ...
          entry_point = ":entry_file",
        )
```

If the entry JavaScript file belongs to the same package (as the BUILD file), you can simply reference it by its relative name to the package directory:

```
        nodejs_binary(
          name = "my_binary",
          ...
          entry_point = ":file.js",
        )
```

## ⚠️ Downstream Rules That Need To Be Updated ⚠️
- [ ] rules_typescript: https://github.com/sfeir-open-source/rules_typescript/commit/9538466
- [ ] rules_sass: https://github.com/manekinekko/rules_sass/commit/9862dfc

PR #777, opened 2019-05-22
Add support for Label in nodejs_binary and rollup_bundle entry_point

### LINKED ISSUE #32: Support for entry points relative to current BUILD file directory

gregmagolan:
Replaces #480

Closes #32

⚠️ Depends on upstream sass change: https://github.com/bazelbuild/rules_sass/pull/87
⚠️ Depends on upstream rules_typescript change: https://github.com/bazelbuild/rules_typescript/pull/453
⚠️ Depends on downstream angular-bazel-example change: https://github.com/angular/angular-bazel-example/pull/450

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

nodejs_binary#entry_point was a string.

## What is the new behavior?

nodejs_binary#entry_point is now a Label.

- [x] Yes
- [ ] No

The `entry_point` accepts a target's name as an entry point. If the target is a rule, it should produce the JavaScript entry file that will be passed to the nodejs_binary rule). For example:

 ```
        filegroup(
            name = "entry_file",
            srcs = ["workspace/path/to/entry/file"]
        )
        nodejs_binary(
          name = "my_binary",
          ...
          entry_point = ":entry_file",
        )
```

If the entry JavaScript file belongs to the same package (as the BUILD file), you can simply reference it by its relative name to the package directory:

```
        nodejs_binary(
          name = "my_binary",
          ...
          entry_point = ":file.js",
        )
```

### DIFF
--- a/internal/e2e/rollup/BUILD.bazel
+++ b/internal/e2e/rollup/BUILD.bazel
@@ -3,11 +3,8 @@ load("//:defs.bzl", "rollup_bundle")
 
 rollup_bundle(
     name = "bundle",
-    srcs = [
-        "bar.js",
-        "foo.js",
-    ],
-    entry_point = "internal/e2e/rollup/foo.js",
+    srcs = ["bar.js"],
+    entry_point = ":foo.js",
     globals = {"some_global_var": "runtime_name_of_global_var"},
     license_banner = ":license.txt",
     deps = [
--- a/internal/e2e/rollup_code_splitting/BUILD.bazel
+++ b/internal/e2e/rollup_code_splitting/BUILD.bazel
@@ -6,8 +6,8 @@ rollup_bundle(
         ["*.js"],
         exclude = ["*.spec.js"],
     ),
-    additional_entry_points = ["internal/e2e/rollup_code_splitting/additional_entry.js"],
-    entry_point = "internal/e2e/rollup_code_splitting/main1.js",
+    additional_entry_points = ["build_bazel_rules_nodejs/internal/e2e/rollup_code_splitting/additional_entry.js"],
+    entry_point = ":main1.js",
     license_banner = ":license.txt",
 )
 
@@ -18,10 +18,11 @@ jasmine_node_test(
         "main1.spec.js",
     ],
     data = glob([
-        "*_golden.js_",
+        "goldens/*",
     ]) + [
         ":bundle",
         ":bundle.min.js",
+        ":bundle.min.es2015.js",
     ],
     deps = [
         "//internal/e2e:check_lib",
--- a/internal/jasmine_node_test/jasmine_node_test.bzl
+++ b/internal/jasmine_node_test/jasmine_node_test.bzl
@@ -56,10 +56,11 @@ def jasmine_node_test(
         tags = tags,
     )
 
+    jasmine_runner_label = Label("//internal/jasmine_node_test:jasmine_runner.js")
+
     all_data = data + srcs + deps
-    all_data += [Label("//internal/jasmine_node_test:jasmine_runner.js")]
     all_data += [":%s_devmode_srcs.MF" % name]
-    entry_point = "build_bazel_rules_nodejs/internal/jasmine_node_test/jasmine_runner.js"
+    entry_point = jasmine_runner_label.relative(":jasmine_runner.js")
 
     # If the target specified templated_args, pass it through.
     templated_args = kwargs.pop("templated_args", []) + ["$(location :%s_devmode_srcs.MF)" % name]
--- a/internal/jasmine_node_test/jasmine_node_test.bzl
+++ b/internal/jasmine_node_test/jasmine_node_test.bzl
@@ -56,11 +56,9 @@ def jasmine_node_test(
         tags = tags,
     )
 
-    jasmine_runner_label = Label("//internal/jasmine_node_test:jasmine_runner.js")
-
     all_data = data + srcs + deps
     all_data += [":%s_devmode_srcs.MF" % name]
-    entry_point = jasmine_runner_label.relative(":jasmine_runner.js")
+    entry_point = Label("//internal/jasmine_node_test:jasmine_runner.js")
 
     # If the target specified templated_args, pass it through.
     templated_args = kwargs.pop("templated_args", []) + ["$(location :%s_devmode_srcs.MF)" % name]
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -89,6 +89,13 @@ def _write_loader_script(ctx):
     if len(ctx.attr.entry_point.files) != 1:
         fail("labels in entry_point must contain exactly one file")
 
+    entry_point_path = expand_path_into_runfiles(ctx, ctx.file.entry_point.short_path)
+
+    # If the entry point specified is a typescript file then set the entry
+    # point to the corresponding .js file
+    if entry_point_path.endswith(".ts"):
+        entry_point_path = entry_point_path[:-3] + ".js"
+
     ctx.actions.expand_template(
         template = ctx.file._loader_template,
         output = ctx.outputs.loader,
@@ -97,7 +104,7 @@ def _write_loader_script(ctx):
             "TEMPLATED_bootstrap": "\n  " + ",\n  ".join(
                 ["\"" + d + "\"" for d in ctx.attr.bootstrap],
             ),
-            "TEMPLATED_entry_point": expand_path_into_runfiles(ctx, ctx.file.entry_point.short_path),
+            "TEMPLATED_entry_point": entry_point_path,
             "TEMPLATED_gen_dir": ctx.genfiles_dir.path,
             "TEMPLATED_install_source_map_support": str(ctx.attr.install_source_map_support).lower(),
             "TEMPLATED_module_roots": "\n  " + ",\n  ".join(module_mappings),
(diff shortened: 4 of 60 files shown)

### CODE REVIEW

File: internal/e2e/rollup/BUILD.bazel [hunk 1]
```diff
@@ -3,11 +3,8 @@ load("//:defs.bzl", "rollup_bundle")
 
 rollup_bundle(
     name = "bundle",
-    srcs = [
-        "bar.js",
-        "foo.js",
-    ],
-    entry_point = "internal/e2e/rollup/foo.js",
+    srcs = ["bar.js"],
+    entry_point = ":foo.js",
```
alexeagle: interesting that you don't have to specify foo.js in the srcs. It seems like it reduces one way users can fall off the happy path. on the other hand it means you have to consider entry_point when enumerating the inputs to a rule.
Should we give it more thought?

internal/e2e/rollup_code_splitting/BUILD.bazel
```diff
@@ -6,8 +6,8 @@ rollup_bundle(
         ["*.js"],
         exclude = ["*.spec.js"],
     ),
-    additional_entry_points = ["internal/e2e/rollup_code_splitting/additional_entry.js"],
-    entry_point = "internal/e2e/rollup_code_splitting/main1.js",
+    additional_entry_points = ["build_bazel_rules_nodejs/internal/e2e/rollup_code_splitting/additional_entry"],
```
alexeagle: do we have to change addl_entry_points at the same time?

internal/jasmine_node_test/jasmine_node_test.bzl
```diff
@@ -56,10 +56,11 @@ def jasmine_node_test(
         tags = tags,
     )
 
+    jasmine_runner_label = Label("//internal/jasmine_node_test:jasmine_runner.js")
+
     all_data = data + srcs + deps
-    all_data += [Label("//internal/jasmine_node_test:jasmine_runner.js")]
     all_data += [":%s_devmode_srcs.MF" % name]
-    entry_point = "build_bazel_rules_nodejs/internal/jasmine_node_test/jasmine_runner.js"
+    entry_point = jasmine_runner_label.relative(":jasmine_runner.js")
```
alexeagle: this is surprising, why does it need a relative calculation when the jasmine_runner_label already points to the right file?

internal/node/node.bzl
```diff
@@ -209,12 +223,66 @@ _NODEJS_EXECUTABLE_ATTRS = {
         allow_files = True,
         aspects = [sources_aspect, module_mappings_runtime_aspect, collect_node_modules_aspect],
     ),
-    "entry_point": attr.string(
+    "entry_point": attr.label(
         doc = """The script which should be executed first, usually containing a main function.
-        This attribute expects a string starting with the workspace name, so that it's not ambiguous
-        in cases where a script with the same name appears in another directory or external workspace.
+        The `entry_point` accepts a target's name as an entry point.
```
alexeagle: "target" is confusing here, ppl don't think of their input .js files as being targets

internal/node/node.bzl
```diff
@@ -209,12 +223,66 @@ _NODEJS_EXECUTABLE_ATTRS = {
...
+        )
+
+        nodejs_binary(
+            name = "bin",
+            data = [":main"]
+            entry_point = ":main.ts",
+        )
+        ```
+
+        The rule will use the corresponding es5 `.js` output of the ts_library rule as the entry point.
```
alexeagle: es5 -> devmode? we already changed default devmode of ts_library to es2015

internal/node/node.bzl
```diff
@@ -209,12 +223,66 @@ _NODEJS_EXECUTABLE_ATTRS = {
...
+
+        If the entry point target is a rule, it should produce the JavaScript entry file that will be passed to the nodejs_binary rule).
```
alexeagle: need to mention that it may only produce one output?

internal/rollup/rollup_bundle.bzl
```diff
@@ -597,11 +607,55 @@ ROLLUP_ATTRS = {
         It is sufficient to load one of these SystemJS boilerplate/entry point
         files as a script in your HTML to load your application""",
     ),
-    "entry_point": attr.string(
+    "entry_point": attr.label(
         doc = """The starting point of the application, passed as the `--input` flag to rollup.
-        This should be a path relative to the workspace root.
+        The `entry_point` accepts a target's name as an entry point.
```
alexeagle: target confusing again

package.bzl
```diff
@@ -35,11 +35,11 @@ def rules_nodejs_dev_dependencies():
     """
 
     # Dependencies for generating documentation
+    # TODO(manekinekko): switch to https://github.com/bazelbuild/rules_sass/ when the changes have been released
```
alexeagle: changes -> should point to a PR /commit so we know how to go check if it's released yet

packages/jasmine/test/BUILD.bazel
```diff
@@ -3,46 +3,84 @@ load("@npm_bazel_jasmine//:index.bzl", "jasmine_node_test")
 jasmine_node_test(
     name = "underscore_spec_test",
     srcs = ["foo_spec.js"],
-    jasmine = "@npm//jasmine",
-    deps = ["//:jasmine_runner"],
+    # The jasmine label here is specific to local testing since
```
alexeagle: nit: lot of duplication - maybe a local macro in this package to make the tests shorter? OTOH it's easier to read without the indirection

at minimum I'd say "see comment above" on the later ones, or maybe extract jasmine as a variable so the comment can go there

packages/labs/webpack/src/cli.ts
```diff
@@ -41,10 +41,14 @@ function main(config: webpack.Configuration): 0|1 {
   return exitCode;
 }
 
-if (require.main === module) {
+export function main() {
```
alexeagle: why this refactoring? I'd rather not have process.exitCode happening if main() is called from a test

internal/e2e/rollup/BUILD.bazel
gregmagolan: Hmmm. That's a good question. One strike against doing it this way is that if the user specifies `:foo.ts` as the entry_point they still have to add the ts_library `:foo_lib` to deps. I'm not sure what is best. On the other hand, the user adding `:foo.js` to entry_point and to srcs doesn't break anything with the current approach.

internal/e2e/rollup_code_splitting/BUILD.bazel
gregmagolan: nope. not sure how that change got in. backed it out. the `.js` is optional there but no reason to change it.

internal/node/node.bzl
gregmagolan: removed that line

internal/node/node.bzl
gregmagolan: right 👍

internal/rollup/rollup_bundle.bzl
gregmagolan: 👍

packages/labs/webpack/src/cli.ts
gregmagolan: Artifact of dev before `cli.ts` was a valid entry_point. Will revert. Thanks

package.bzl
gregmagolan: 👍

internal/node/node.bzl
gregmagolan: 👍

internal/jasmine_node_test/jasmine_node_test.bzl
gregmagolan: yup. there is no reason for that 🤷‍♂

packages/jasmine/test/BUILD.bazel
gregmagolan: good idea. extracted duplication to .bzl file

internal/e2e/rollup/BUILD.bazel
gregmagolan: There is some smell however with not having the file in `srcs` as seen here: https://github.com/bazelbuild/rules_nodejs/pull/777/files#diff-6e8a939ed5da5eca956cd2f53701b77e. Probably better to force it to be specified

PR #778, opened 2019-05-22
Forward all args to originalResolveFilename

Globegitter:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

If I call `require.resolve("../test.js", { paths: ["/some/path/in/sandbox"] })` the options are not forwarded correctly because the `_resolveFilename` method takes 4 parameters.

## What is the new behavior?

All expected parameters are forwarded, so are the options now and paths will be respected.

- [ ] Yes
- [x] No

## Other information

I ran into this as the latest version of one of our npm dependencies is making use of this case and that fix fixes it.

### DIFF
--- a/internal/node/node_loader.js
+++ b/internal/node/node_loader.js
@@ -351,7 +351,7 @@ module.constructor._resolveFilename = function(request, parent) {
   // Built-in modules, relative, absolute imports and npm dependencies
   // can be resolved using request
   try {
-    const resolved = originalResolveFilename(request, parent);
+    const resolved = originalResolveFilename.apply(null, arguments);
     if (resolved === request || request.startsWith('.') || request.startsWith('/') ||
         request.match(/^[A-Z]\:[\\\/]/i)) {
       if (DEBUG)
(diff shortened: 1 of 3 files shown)


internal/node/node_loader.js
```diff
@@ -433,15 +432,15 @@ module.constructor._resolveFilename = function(request, parent) {
...
-        resolveRunfiles(undefined, NODE_MODULES_ROOT, request), parent);
+    const resolved = originalResolveFilename(resolveRunfiles(undefined, NODE_MODULES_ROOT, request), parent, isMain, options);
     if (DEBUG)
       console.error(
           `node_loader: resolved ${request} within node_modules (${NODE_MODULES_ROOT}) to ` +
           `${resolved} from ${parentFilename}`
       );
     return resolved;
   } catch (e) {
+    console.error(e);
```
alexeagle: do you mean to leave this here? seems like it could be very spammy even for a successful execution where some other resolution succeeds

Globegitter: Whoops sorry. That was left from debugging. Will remove.

PR #780, opened 2019-05-22
Add test coverage for manual_build_file_contents... 

gregmagolan:
via an example of how to make a target for node_modules/.bin/*

Example also shows use of migration instructions for users that were depending on `node_modules/.bin/BUILD` targets: https://github.com/bazelbuild/rules_nodejs/wiki#node_modulesbinbuild-file-no-longer-generated.

### DIFF
--- a/examples/bazel_managed_deps/deps.spec.js
+++ b/examples/bazel_managed_deps/deps.spec.js
@@ -10,4 +10,15 @@ describe('dependencies', () => {
      () => {
        require('jasmine-core');
      });
+
+  it('.bin files should be in runfiles via @npm//:bin_files data dep', () => {
+    try {
+      expect(require.resolve('.bin/jasmine').endsWith('npm/node_modules/.bin/jasmine')).toBe(true);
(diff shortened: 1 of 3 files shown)


examples/bazel_managed_deps/deps.spec.js
```diff
@@ -10,4 +10,15 @@ describe('dependencies', () => {
...
      });
+
+  it('.bin files should be in runfiles via @npm//:bin_files data dep', () => {
+    try {
+      expect(require.resolve('.bin/jasmine').endsWith('npm/node_modules/.bin/jasmine')).toBe(true);
+      expect(require.resolve('.bin/tsc').endsWith('npm/node_modules/.bin/tsc')).toBe(true);
+      expect(require.resolve('.bin/tsserver').endsWith('npm/node_modules/.bin/tsserver'))
+          .toBe(true);
+    } catch (_) {
+      fail('.bin file not resolved');
```
alexeagle: nit: should say which require failed, or just make three assertions

gregmagolan: 👍

PR #782, opened 2019-05-23
Add `query --experimental_allow_incremental_repository_updates`...

gregmagolan:
to common.bazelrc as its needed for `bazel query` to work when `managed_directories` is used.

`run --experimental_allow_incremental_repository_updates` is not needed


alexeagle:
I added this bit to the migration wiki as well

PR #783, opened 2019-05-23
Remove bazelWorkspaces version & compatVersion...

gregmagolan:
...as we now have a single version policy.

Also fixes & adds test for bazelWorkspaces.rootPath

`bazelWorkspaces` `version` & `compatVersion` attributes and versions checks are no longer needed since there should never be more than one version of a bazel workspace installed from npm. There should be no transitive deps on any `@bazel/foo` npm packages or other npm packages that install bazel workspaces. Only peerDeps should be used in npm packages to express npm deps for npm packages that install bazel workspaces.


gregmagolan:
> that's cool, I never understood why to set a compat_version in a package.json anyway, or how to know what's the right value

Yes. It was a bad idea to try to solve the multiple version problem and it wasn't clear how it would be set correctly. The better solution is to just avoid multiple versions using peerDeps.

PR #784, opened 2019-05-23
Further node resolve filename fix

Globegitter:
## PR Checklist

## PR Type

Another fix of very corner case module importing of a usage that I found in one of our dependencies: https://github.com/nuxt/nuxt.js/blob/91c3642e64166a939683b856e5b34ef42d12bb0e/packages/core/src/resolver.js#L26

I did also submit a possible fix to the project itself: https://github.com/nuxt/nuxt.js/pull/5796 but not sure if it will get accepted at this stage.

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

```
     Module._resolveFilename(npm_module, {
        paths: ["..],
      })
```

does not work because of the checks done in the custom `_resolveFilename` method

## What is the new behavior?

Node.js seems to allow a call of: 
```
     Module._resolveFilename(npm_module, {
        paths: ["..],
      })
```

which seems to behave very similar to `require.resolve(npm_module, {paths: [".."]})` but it does work different internally as the first invocation sets the paths of the parent rather than setting options. So now we check if `parent.paths` has been set and special case that import so it can work.

- [ ] Yes
- [x] No

## Other information

Not a 100% sure if we should really support this but I guess it does work with vanilla nodejs, so it may be worth it.


internal/e2e/fine_grained_no_bin/index.js
```diff
@@ -1,3 +1,5 @@
 const path = require("path");
+const Module = require("module");
 
 console.log('hello ' + require.resolve("./test.js", { paths: [ path.join(__dirname, 'test') ] }));
+console.log('hello ' + Module._resolveFilename("test", { paths: [ path.join(__dirname, 'test', 'node_modules') ] }));
```
alexeagle: I guess the assertion here is just that it doesn't throw? we don't care what we get back from _resolveFilename in any test?

Globegitter: Yeah true, I did not think about it, because I thought we don't need to test nodejs' resolving mechanism but it is probably still worth doing - let me add a proper test for it.

alexeagle: ping me when you've added a test?

Globegitter: @alexeagle Done.

Globegitter:
@alexeagle any idea why these changes (as well as the ones from https://github.com/bazelbuild/rules_nodejs/pull/778) would fail on windows only? As far as I can see the code is platform independent and I also do not have a windows laptop to debug.

PR #791, opened 2019-05-24
Remove sh_* wrappers from nodejs_* rules

### LINKED ISSUE #51: jasmine_node_test should not need an sh_binary wrapper macro for nodejs_test

### LINKED ISSUE #733: nodejs_test seems to be ignoring the timeout attribute
nodejs_test seems to be ignoring the `timeout` input and defaults to `moderate` with a 5 minute timeout
<pre>
  <code>
nodejs_test(
    name="test",
    data = [
        "//example:test.js",
    ],
    entry_point = "broken_timeout/example/test.js",
    timeout="short",
)
  </code>
</pre>
test.js
<pre>
  <code>
console.log('doing the test');

setTimeout(() => process.exit(0), 120 * 1000);
  </code>
</pre>

When I run this test it should timeout after a minute but instead is passes after two minutes

<pre>
  <code>
bazel test //example:test
INFO: Analysed target //example:test (0 packages loaded, 0 targets configured).
INFO: Found 1 test target...
Target //example:test up-to-date:
  bazel-bin/example/test_bin.sh
  bazel-bin/example/test
INFO: Elapsed time: 121.168s, Critical Path: 120.88s
INFO: 2 processes: 2 linux-sandbox.
INFO: Build completed successfully, 2 total actions
//example:test                                                           PASSED in 120.4s

INFO: Build completed successfully, 2 total actions
  </code>
</pre>

Here is a zip file containing a simple bazel project demonstrating the problem
[broken_timeout_example.zip](https://github.com/bazelbuild/rules_nodejs/files/3133328/broken_timeout_example.zip)

**Operating System:**
<pre>
  <code>
Ubuntu 18.04.2 LTS
  </code>
</pre>
<pre>
  <code>
bazel version
Build label: 0.24.1.1
Build target: bazel-out/k8-fastbuild/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Fri Apr 5 22:38:01 2019 (1554503881)
Build timestamp: 1554503881
  </code>
</pre>
**Rules version (SHA):**
<pre>
  <code>
1f6d878a9ea3a095291f66e3d1a0f6b4641f5159
  </code>
</pre>

alexeagle:
These were a clumsy workaround for Windows but cause a number of problems.

Fixes #51
Fixes #733

Note to reviewers: this will require a bunch of manual testing on Windows I think.

### DIFF
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -206,7 +206,7 @@ def _nodejs_binary_impl(ctx):
             transitive_files = runfiles,
             files = node_tool_files + [
                         ctx.outputs.loader,
-                    ] + ctx.files._source_map_support_files +
+                    ] + ctx.files._source_map_support_files + ctx.files._bash_runfiles +
 
                     # We need this call to the list of Files.
                     # Calling the .to_list() method may have some perfs hits,
@@ -376,6 +376,9 @@ _NODEJS_EXECUTABLE_ATTRS = {
         `--node_options=--preserve-symlinks`
         """,
     ),
+    "_bash_runfiles": attr.label(
+        default = Label("@bazel_tools//tools/bash/runfiles"),
+    ),
     "_launcher_template": attr.label(
         default = Label("//internal/node:node_launcher.sh"),
         allow_single_file = True,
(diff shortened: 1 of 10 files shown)


internal/node/node.bzl
```diff
@@ -390,69 +390,39 @@ remote debugger.
...
-      testonly: applied to nodejs_binary and wrapper binary
       **kwargs: passed to the nodejs_binary
     """
-    all_data = data + ["@bazel_tools//tools/bash/runfiles"]
-
     nodejs_binary(
-        name = "%s_bin" % name,
-        data = all_data,
+        name = name,
+        data = data + ["@bazel_tools//tools/bash/runfiles"],
```
Globegitter: Any reason we could not add this as an implicit dependency to the rules to completely get rid of the macro?

Globegitter:
@alexeagle Removing the wrapper has an additional benefit of making it easier to access any providers `nodejs_binary` exports, like the one introduced in https://github.com/bazelbuild/rules_nodejs/pull/863. Right now one has to access the `_bin` target which is not entirely clear without reading the source code.

Globegitter:
@alexeagle Removing the wrapper has an additional benefit of making it easier to access any providers `nodejs_binary` exports, like the one introduced in https://github.com/bazelbuild/rules_nodejs/pull/863. Right now one has to access the `_bin` target which is not entirely clear without reading the source code.

Globegitter:
Also another benefit I am just noticing of having the bash runfiles as an implicit dependency, right now the nodejs provider I have written has to include the bash files in its `sources` attribute rather than in the `node_runfiles` where it really belongs to. It is a small file so it will not make much of a difference in regards to caching I presume but it would still be nice to be able to differentiate it.

internal/node/node.bzl
alexeagle: yes certainly, I was being lazy

alexeagle:
okay as expected this works everywhere but Windows. I need to setup a new windows testing environment...

alexeagle:
@meteorcloudy picked this up and finished in #1078 thank you!!

PR #798, opened 2019-05-28
Remove runfiles.

Globegitter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes
Simplifying the setup/build of `nodejs_binary` and `nodejs_test`, which simplifies work for https://github.com/bazelbuild/rules_nodejs/pull/645 and https://github.com/bazelbuild/rules_nodejs/pull/283

Every nodejs_binary gets the `_node_runfiles` filegroup added, which adds the real node binary as well as npm, yarn and some other libraries. Further the implicit node binary dependency is a wrapper script which adds the bazel managed node to the PATH.

## What is the new behavior?
For `nodejs_binary` the wrapper script target was exchanged with the real node binary and the functionality of the wrapper script, setting the `PATH` was added to the `node_loader.js`. The runfiles are entirely removed as an implicit dependency as binaries can run fine without these now.

These means we only depend on one single file to execute our `nodejs_binary`, that should make things a little faster as less things will need to be symlinked but also reduce size of docker containers for most and it will simplify things for the toolchain PR as now it only has to worry about providing one single file target.

- [x] Yes
- [ ] No

People that depended on the fact that yarn or node was included as an implicit dependency would now have to add `@nodejs//:node_runfiles` explicitely to their data dependencies.


internal/node/node_loader.js
```diff
@@ -25,6 +25,14 @@
 var path = require('path');
 var fs = require('fs');
 
+const isWindows = /^win/i.test(process.platform);
+// Ensure that node is added to the path for any subprocess calls
+if (isWindows) {
+  process.env.PATH = `${path.dirname(process.execPath)};${process.env.PATH}`
```
gregmagolan: nit: less duplication with
```
process.env.PATH = `${path.dirname(process.execPath)}${isWindows ? ';' : ':'}${process.env.PATH}`;
```

jbedard: Or to avoid all the quotes and `${...}`...
```
[path.dirname(process.execPath), process.env.PATH].join(isWindows ? ';' : ':')
```

Globegitter:
@gregmagolan 
About 1, yeah I left the original script there on purpose as the same target for that. For 2 yes RBE still works and it is actually being tested in buildkite CI. I did have a failure about that to start off with.

Yeah I was not sure why anyone would depend on the runfiles downstream - but that is why I left them, because I don't know. But I would be very happy to remove them and we can always bring it back if someone needs them for some reason.

Globegitter:
Done - I kept it in 2 commits. One to address the comment. The other to remove the runfiles.

Globegitter:
@alexeagle squashed.

alexeagle:
I rebased around the merge conflict and force-pushed to your branch, FYI. will merge now if green

Globegitter:
@alexeagle anything blocking this now from getting merged?

PR #803, opened 2019-05-29
fix: license SPDX should be `Apache-2.0` instead of `Apache 2.0`

alan-agius4:
The Apache license version 2.0 SPDX short identifier is Apache-2.0

//cc @clydin 

More info: https://spdx.org/licenses/


gregmagolan:
@alexeagle @me Should also fix this in https://github.com/angular/angular-bazel-example

PR #804, opened 2019-05-29
@bazel/create --packageManager and --typescript options

alexeagle:


alexeagle:
I'll just add a feature to run the package manager for you to ensure the lock file is there before you run bazel

PR #805, opened 2019-05-30
Update README to recommend @bazel/create for quickstart

alexeagle:


gregmagolan:
nit: you can now remove `README.md` from the release script:

```
for (const f in ['README.md', 'packages/create/index.js']) {
```

alexeagle:
I checked that - but the long-form with the version and sha256 still appear below in the alternate install instructions

PR #810, opened 2019-05-31
Update examples/vendored_node to reflect best practice of specifying node version

gregmagolan:
The docs for node_repositories() reflect this now and the reason is that node_repositories() will set node options depending on version used


gregmagolan:
> How will someone know if the version here is wrong?

If they are vendoring in a version of node prior to 10.2.0 and don't specify that's it an older version then the build will fail and node will complain that `--preserve-symlinks-main` is not a valid argument. Probably a rare case but could happen.

PR #812, opened 2019-06-01
feat: auto add type def

### LINKED ISSUE #760: [feat] auto add @npm//@types/pkg if it exists when @npm//pkg is in deps
# 🚀 feature request
### Relevant Rules
ts_library

### Description
A simple addition that automatically resolves the types package for those external packages that don't come with types themselves.

When a package is in the deps array, for example: `@npm//express`
The user should not have to define `@npm//@types/express` - ts_library should add it for them.
If the types package does not exist, then it should not fail and allow the TS compiler to fail later on if the user has `noImplicitAny` turned on - for packages that don't ship with types.

Manually adding the typing packages

santoshyadav198613:
Fixes #760

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #760

- [ ] Yes
- [x] No

### DIFF
--- a/internal/npm_install/generate_build_file.js
+++ b/internal/npm_install/generate_build_file.js
@@ -966,7 +966,11 @@ nodejs_binary(
  * Given a pkg, return the skylark aliases for the package.
  */
 function printPackageAliases(pkg) {
-  return `
+
+  const pkgJson = path.posix.join(pkg._dir, 'package.json');
+
+  let content = '';
+  content = `
 # Generated target alias for npm package "${pkg._dir}"
 ${printJson(pkg)}
 alias(
@@ -989,6 +993,22 @@ alias(
   actual = "//node_modules/${pkg._dir}:${pkg._name}__typings"
 )
 `;
+
+if (pkg.dependencies != null) {
+  const typingPackage = pkg.dependencies[`@types/${pkg.name}`];
+  if (typingPackage) {
+    content += `
+  node_module_library(
+    name = "${pkg._name}",
+    deps = [
+      "//node_modules/@types/${pkg.name}"
+    ]
+  )
+  `
+  }
+
+}
+return content;
 }
 
 /**


internal/npm_install/generate_build_file.js
```diff
@@ -989,6 +1020,19 @@ alias(
   actual = "//node_modules/${pkg._dir}:${pkg._name}__typings"
 )
 `;
+
+if (types){
```
Toxicable: Instead of looking through the node_modules file system for types why don't we look through the package.json instead.
Looking at `node_modules` you might accidentally pickup a package that wasn't installed by the user.
So maybe something like:
```
const typingPackage = pkgJson.dependencies[`@types/${pkg.name}`]
if(typingPackage) {
  // since the typingPackage exists in the pkgJson then add it as deps since it'll be created as a package at some point
}

```

internal/npm_install/generate_build_file.js
```diff
@@ -966,7 +967,11 @@ nodejs_binary(
  * Given a pkg, return the skylark aliases for the package.
  */
 function printPackageAliases(pkg) {
-  return `
+
+  const pkgJson = path.posix.join(pkg._dir, 'package.json');
```
Toxicable: You shouldn't need to re-read the packageJson off the disk, `pkg` already has all the information you need

santoshyadav198613:
Can you check now, the failure is related to test case size too big. Please let me know if this is related to my PR.

alexeagle:
why so many reformatting changes? it's distracting - clang-format should have made it consistent with what we had before.

I'd expect another file changed to illustrate how the behavior is different. We have some golden files like https://github.com/bazelbuild/rules_nodejs/blob/master/internal/npm_install/test/golden/BUILD.bazel.golden - one of these at least should be changed by this code to serve as a test for your changes.

santoshyadav198613:
Thanks @alexeagle , will make the changes. I had applied clang format, but looks like it was not correct.

internal/npm_install/generate_build_file.js
Toxicable: You shouldn't need these changes anymore.

santoshyadav198613:
> There should be a change to the golden files with this addition.
> Could you run the testing suit and accept the changes required to make it pass

Hi @Toxicable ,

I tried running yarn test_all but can not see any files as modified. Also while pushing changes getting below error:
$ find . -type f \( -name "*.bzl" -or -name WORKSPACE -or -name BUILD -or -name BUILD.bazel \) ! -path "*/node_modules/*" ! -path "./dist/*" ! -path "./third_party/*" | xargs buildifier -v --warnings=attr-cfg,attr-license,attr-non-empty,attr-output-default,attr-single-file,confusing-name,constant-glob,ctx-actions,ctx-args,depset-iteration,depset-union,dict-concatenation,duplicated-name,filetype,function-docstring,git-repository,http-archive,integer-division,load,load-on-top,module-docstring,name-conventions,native-build,native-package,out-of-order-load,output-group,package-name,package-on-top,positional-args,redefined-variable,repository-name,return-value,same-origin-load,string-iteration,unreachable,unsorted-dict-items,unused-variable --mode=check
/e/rules_nodejs/rules_nodejs/node_modules/.bin/buildifier: line 1: /../@bazel/buildifier-win32_x64/buildifier.exe: No such file or directory
error Command failed with exit code 123.

PR #814, opened 2019-06-01
Fix compatibility for Bazel 0.27

laurentlb:
With this change, the code can build with the Bazel flag `incompatible_string_join_requires_strings`, which will be enabled by default in Bazel 0.27.0.

When the flag is enabled, the `string.join` function expects string elements (instead of implicitly doing a conversion to string).

## PR Type

- [x] Bugfix
- [x] Build related changes
- [ ] CI related changes


laurentlb:
Yes, sorry. The flag in Bazel is `incompatible_string_join_requires_strings` and it's explained here: https://github.com/bazelbuild/bazel/issues/7802

laurentlb:
Added to bazelrc to avoid regressions.

.bazelrc
```diff
@@ -3,3 +3,5 @@ import %workspace%/common.bazelrc
 
 # Mock versioning command to test the --stamp behavior
 build --workspace_status_command="echo BUILD_SCM_VERSION 1.2.3"
+
+common --incompatible_string_join_requires_strings
```
alexeagle: ooh, I've never seen "common" - does that cover `query` too? @gregmagolan

PR #820, opened 2019-06-05
Remove es5.min rollup_bundle output from runfiles

alexeagle:
Having it there causes it to always be built, whether requested or not.
This partially unblocks using rollup_bundle in dev mode.

BREAKING CHANGE:
if you expect the rollup_bundle es5.min file to be a runtime dependency of a rule, you must now include it in the data[] of that rule

### DIFF
--- a/internal/rollup/rollup_bundle.bzl
+++ b/internal/rollup/rollup_bundle.bzl
@@ -537,7 +537,11 @@ def _rollup_bundle(ctx):
         )
 
     return [
-        DefaultInfo(files = depset(files), runfiles = ctx.runfiles(files)),
+        DefaultInfo(
+            files = depset(files),
+            # NB: we don't include any runfiles here since they would always be built
+            # regardless if they are requested or not
+        ),
         output_group,
     ]
 


alexeagle:
testing out the implications by patching this change into angular-bazel-example

alexeagle:
yeah looks okay when manually tested with the prodapp in angular-bazel-example. I guess we have explicit dependencies on the outputs we need at runtime now.

internal/rollup/rollup_bundle.bzl
```diff
@@ -537,7 +537,11 @@ def _rollup_bundle(ctx):
         )
 
     return [
-        DefaultInfo(files = depset(files), runfiles = ctx.runfiles(files)),
+        DefaultInfo(
+            files = depset(files),
+            # NB: we don't include any runfiles here since they would always be built
```
jbedard: IDK if this comment is really required. Almost seems obvious that es5min is not a "runtime depdency/requirement" of the default es5min files 🤷‍♂

alexeagle: I mostly added this to remind us not to re-introduce the problem, since I didn't add a test that asserts that default actions don't run

(the only test I could think of is similar to my repro where the default file fails to build, but that would be hard to build into the rollup_bundle rule)
(maybe if we could assert on the action graph of a build, that would be the right way to assert what we want here)

PR #827, opened 2019-06-06
Initial toolchain support

Globegitter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Nodejs is downloaded for each platform but does not support toolchains, so does e.g. also not support cross compilation or building of docker images on macOS.

Issue Number: First part of #396

## What is the new behavior?

Switches out nodejs to be fetched using platform specific toolchains.

- [ ] Yes
- [x] No


Globegitter:
CI error seems unrelated:

```
ERROR: Analysis of target '//:test' failed; build aborted: no such package '@remote_coverage_tools//': java.io.IOException: Error downloading [https://mirror.bazel.build/bazel_coverage_output_generator/releases/coverage_output_generator-v1.0.zip] to /home/circleci/.cache/bazel/_bazel_circleci/7f4dae65be8a29efc03e2c4af7f6dd90/external/remote_coverage_tools/coverage_output_generator-v1.0.zip: connect timed out
```

Toxicable:
We've been seeing this flake a lot in our CI internally recently.
Kinda good to see it's not isolated.

Globegitter:
The only thing missing here now is docs as well as some specific tests. Not sure how to test the cross Compilation though. But the way one could build a docker image on mac (without native dependencies) is via: `bazel run --platforms=@build_bazel_rules_nodejs//toolchains/node:linux_amd64 //app`

Globegitter:
Added some docs now to the main README. I also had more of a look at starlark testing and I think we could utilise this (see https://docs.bazel.build/versions/master/skylark/testing.html#using-a-test-target) to also add some tests. But given the impact of this PR, I wonder if we could just try and merge this asap and then I can do a follow-up PR just for tests and maybe even adding a new doc for advanced use-cases like creating a custom toolchain etc.

alexeagle:
I'm fine with augmenting the docs in a separate change but I'd like to have at least a basic regression test in place given the pace of development in the repo

internal/node/node.bzl
```diff
@@ -161,25 +160,41 @@ def _nodejs_binary_impl(ctx):
...
-        is_executable = True,
-    )
+    node_tool_info = ctx.toolchains["@build_bazel_rules_nodejs//toolchains/node:toolchain_type"].nodeinfo
+    node_tool_files = []
+    if node_tool_info.target_tool_path == "" and not node_tool_info.target_tool:
+        # If tool_path is empty and tool_target is None then there is no local
+        # node tool, we will just print a nice error message if the user
+        # attempts to do bazel run
+        ctx.actions.write(
+            content = ("echo node toolchain was not properly configured so %s cannot be executed." % ctx.attr.name),
```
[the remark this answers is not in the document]
Globegitter: That is a very good question indeed, which I was wondering as well. I copied it from https://github.com/bazelbuild/rules_k8s/blob/master/k8s/object.bzl#L203

But yeah happy to change that.

Globegitter:
We currently do not have any docker tests. I was thinking that as well but then we add a circular dependency on rules_docker as they already depend on rules_nodejs. I guess in theory it is fine as it is just a dev dependency. WDYT @alexeagle 

But yeah I will also have a look into the starlark tests and see if we can utilise those for these purposes. The other thing we could do is just get the sha from all the different node binaries and compare that in the test.

Globegitter:
Interesting, also seeing some toolchain references in the docs here: https://github.com/bazelbuild/bazel-skylib/blob/master/docs/unittest_doc.md like `unittest_toolchain` and `register_unittest_toolchains` so maybe there is unittest support for toolchains. Will investigate more.

Globegitter:
Turns out that is just toolchain setup for unittest itself and not to be used for unit testing. Having played around with it a bit, what it seems we could test via the unit testing is that with the different platform flags we are getting the nodejs binary from different external repositories. Would that be enough as a first round of tests? Then we can still see about doing a full e2e test scenario as a follow-up. Maybe it is also possible to hear from other rules maintainers how they test this functionality.

thesayyn:
When this will be merged to upstream?

toolchains/node/node_configure.bzl
```diff
@@ -0,0 +1,75 @@
...
+# limitations under the License.
+"""
+Defines a repository rule for configuring the node binary.
+"""
+
+def _impl(repository_ctx):
+    if repository_ctx.attr.target_tool and repository_ctx.attr.target_tool_path:
+        fail("Can only set one of target_tool or target_tool_path but both where set.")
+
+    host_os = repository_ctx.os.name.lower()
```
gregmagolan: host_os unused

toolchains/node/node_configure.bzl
```diff
@@ -0,0 +1,75 @@
...
+    substitutions = {}
+    default_tool_path = repository_ctx.attr.target_tool_path or repository_ctx.which("node") or ""
```
gregmagolan: A bit hard to follow at first glance as `substitutions["%{TOOL_ATTRS}"]` gets overriden when `target_tool` is set.
I would go with
```
if repository_ctx.attr.target_tool:
    substitutions = { "%{TOOL_ATTRS}": "    target_tool = \"%s\"\n" % repository_ctx.attr.target_tool }
else:
    default_tool_path = repository_ctx.attr.target_tool_path or repository_ctx.which("node") or ""
    substitutions = { "%{TOOL_ATTRS}": "    target_tool_path = \"%s\"\n" % default_tool_path }
```

gregmagolan: Makes it easier to see that they are mutually exclusive

toolchains/node/node_configure.bzl
```diff
@@ -0,0 +1,75 @@
...
+
+def node_configure(node_repository_names = None, **kwargs):
+    """Creates an external repository with a node_toolchain target properly configured.
+
+    Args:
+      node_repository_names: list of strings
+      **kwargs: all the args of the _node_configure rules
+    """
+
+    native.register_toolchains(
```
gregmagolan: Seems like the call to `native.register_toolchains(` would fit better in `node_repositories` itself that is aware of all the OS names at and if a user uses `node_configure` then the toolchains in this list won't be registered twice.

Something like this perhaps in node_repositories():
```
    # This needs to be setup so toolchains can access nodejs for all different versions
    for os_arch_name in OS_ARCH_NAMES:
        os_name = "_".join(os_arch_name)
        node_repository_name = "nodejs_%s" % os_name
        _maybe(
            _nodejs_repo,
            name = node_repository_name,
            package_json = package_json,
            node_version = node_version,
            yarn_version = yarn_version,
            vendored_node = vendored_node,
            vendored_yarn = vendored_yarn,
            node_repositories = node_repositories,
            yarn_repositories = yarn_repositories,
            node_urls = node_urls,
            yarn_urls = yarn_urls,
            preserve_symlinks = preserve_symlinks,
        )
        native.register_toolchains("@build_bazel_rules_nodejs//toolchains/node:node_%s_toolchain" % os_arch_name[0])
        node_toolchain_configure(
            name = "%s_config" % node_repository_name,
            target_tool = "@%s//:node_bin" % node_repository_name,
        )
```

which means the `node_configure` macro could go and it would just be `node_configure = repository_rule(` and also the toolchain name for osx would have to change to:

```
toolchain(
    name = "node_darwin_toolchain",
    target_compatible_with = [
        "@bazel_tools//platforms:osx",
        "@bazel_tools//platforms:x86_64",
    ],
    toolchain = "@nodejs_darwin_amd64_config//:toolchain",
    toolchain_type = ":toolchain_type",
)
```
?

toolchains/node/node_configure.bzl
```diff
@@ -0,0 +1,75 @@
...
+            mandatory = False,
+            allow_single_file = True,
+        ),
+        "target_tool_path": attr.string(
+            doc = "Absolute path to a pre-installed nodejs binary for the target os.",
+            mandatory = False,
+        ),
+    },
+)
+
```
gregmagolan: Comment from macro would move to here if making the change below.
"""
Creates an external repository with a node_toolchain //:toolchain target properly configured.
"""

toolchains/node/node_toolchain.bzl
```diff
@@ -0,0 +1,50 @@
...
+NodeInfo = provider(
+    doc = "Information about how to invoke the node binary.",
+    fields = {
+        "target_tool": "A hermetically downloaded nodejs executable target.",
+        "target_tool_path": "Path to an existing nodejs executable",
+    },
+)
+
+def _node_toolchain_impl(ctx):
+    if not ctx.attr.target_tool and not ctx.attr.target_tool_path:
```
gregmagolan: Also add,
```
    if repository_ctx.attr.target_tool and repository_ctx.attr.target_tool_path:
        fail("Can only set one of target_tool or target_tool_path but both where set.")
```
incase this is used without node_configure

toolchains/node/node_toolchain.bzl
```diff
@@ -0,0 +1,50 @@
...
+            doc = "Target for a downloaded nodejs binary for the target os.",
+            mandatory = False,
+            allow_single_file = True,
+        ),
+        "target_tool_path": attr.string(
+            doc = "Absolute path to a pre-installed nodejs binary for the target os.",
+            mandatory = False,
+        ),
+    },
+)
```
gregmagolan: nit: add a brief rule doc below the rule

internal/node/node_repositories.bzl
```diff
@@ -20,9 +20,10 @@ See https://docs.bazel.build/versions/master/skylark/repository_rules.html
 
 load("//internal/common:check_bazel_version.bzl", "check_bazel_version")
 load("//internal/common:check_version.bzl", "check_version")
-load("//internal/common:os_name.bzl", "os_name")
+load("//internal/common:os_name.bzl", "OS_NAMES", "os_name")
 load("//internal/npm_install:npm_install.bzl", "yarn_install")
 load("//third_party/github.com/bazelbuild/bazel-skylib:lib/paths.bzl", "paths")
+load("//toolchains/node:node_configure.bzl", node_toolchain_configure = "node_configure")
```
gregmagolan: Seems like a better name for the .bzl file and repository_rule is `node_toolchain_configure` so it is less ambigious as to what it does?

toolchains/node/node_toolchain.bzl
```diff
@@ -0,0 +1,50 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""
+This module implements the node toolchain rule.
+"""
+
+NodeInfo = provider(
+    doc = "Information about how to invoke the node binary.",
```
gregmagolan: nit: I think its better to refer to it as the nodejs "executable" instead of "binary" in these doc comments so as not to get confused with `nodejs_binary`

toolchains/node/node_toolchain.bzl
```diff
@@ -0,0 +1,50 @@
...
+            target_tool = ctx.attr.target_tool,
+        ),
+    )
+    return [toolchain_info]
+
+node_toolchain = rule(
+    implementation = _node_toolchain_impl,
+    attrs = {
+        "target_tool": attr.label(
+            doc = "Target for a downloaded nodejs binary for the target os.",
```
gregmagolan: nit: nodejs "executable" and also consolidate doc comment here and in NodeInfo above so its consistent

toolchains/node/node_toolchain.bzl
```diff
@@ -0,0 +1,50 @@
...
+            mandatory = False,
+            allow_single_file = True,
+        ),
+        "target_tool_path": attr.string(
+            doc = "Absolute path to a pre-installed nodejs binary for the target os.",
```
gregmagolan: nit: nodejs "executable" and also consolidate doc comment here and in NodeInfo above so its consistent

toolchains/node/node_configure.bzl
```diff
@@ -0,0 +1,75 @@
...
+        Label("@build_bazel_rules_nodejs//toolchains/node:BUILD.tpl"),
+        substitutions,
+        False,
+    )
+
+_node_configure = repository_rule(
+    implementation = _impl,
+    attrs = {
+        "target_tool": attr.label(
+            doc = "Target for a downloaded nodejs binary for the target os.",
```
gregmagolan: nit: nodejs "executable" and also consolidate doc comment here and in NodeInfo above so its consistent

toolchains/node/node_configure.bzl
```diff
@@ -0,0 +1,75 @@
...
+            mandatory = False,
+            allow_single_file = True,
+        ),
+        "target_tool_path": attr.string(
+            doc = "Absolute path to a pre-installed nodejs binary for the target os.",
```
gregmagolan: nit: nodejs "executable" and also consolidate doc comment here and in NodeInfo above so its consistent

internal/npm_install/npm_install.bzl
```diff
@@ -104,7 +104,7 @@ def _create_build_files(repository_ctx, node, lock_file):
         repository_ctx.attr.name,
         str(lock_file),
         ",".join(repository_ctx.attr.included_files),
-    ])
+    ], quiet = False)
```
gregmagolan: for debugging?

internal/node/node_repositories.bzl
```diff
@@ -567,6 +587,9 @@ def node_repositories(
         minimum_bazel_version = "0.21.0",
     )
 
+    # This "nodejs" repo is just for convinience so one does not have to target @nodejs_<os_name>//...
+    # At macro time we can unfortunately not figure out the host os so that means there will be a
+    # duplicate set of external repositories
```
gregmagolan: I suppose this means that for an initial clean build the nodejs archive will get downloaded twice? Small price to pay I supposed for toolchain support.

internal/node/node_repositories.bzl
```diff
@@ -436,12 +443,25 @@ if %errorlevel% neq 0 exit /b %errorlevel%
...
             "TEMPLATED_node_bin_actual": node_exec_label,
-            "TEMPLATED_node_dir": NODE_DIR,
             "TEMPLATED_npm_actual": npm_node_repositories_entry,
+            "TEMPLATED_vendored_node": "true" if repository_ctx.attr.vendored_node else "false",
             "TEMPLATED_yarn_actual": yarn_node_repositories_entry,
         },
     )
-    result = repository_ctx.execute([node_entry, "generate_build_file.js"])
+    host_os = os_name(repository_ctx)
+    if host_os in repository_ctx.attr.name or repository_ctx.attr.name == "nodejs":
```
gregmagolan: A little safer if you use `"_%s_" % host_os`:
```
if ("_%s_" % os_name(repository_ctx)) in repository_ctx.attr.name or repository_ctx.attr.name == "nodejs":
```

internal/node/node_repositories.bzl
```diff
@@ -436,12 +443,25 @@ if %errorlevel% neq 0 exit /b %errorlevel%
...
+        # We have to use the relative path here otherwise bazel reports a cycle
+        result = repository_ctx.execute([node_entry, "generate_build_file.js"])
+    else:
+        # NOTE: Ideally we would not need this logic here and could just depend on the @nodejs//:node_bin alias but it is not possible.
+        # See: https://github.com/bazelbuild/bazel/issues/8674
+        node_path = "node.exe" if is_windows_os else "bin/node"
```
gregmagolan: Took me a second to see what was going on here. This is a check if the nodejs repo being setup is the same as the host since we have to run `node generate_build_files.js` so if the nodejs repo is not the same as the host, the @nodejs host repo node executable is used. Can add a bit more explanation?

Another PR we should remove generate_build_file.js altogether if possible and just write a BUILD file using `repository_ctx.template` directly instead of generating ` generate_build_file.js`. Not sure why that was put in in the first place but I don't see anything that is does that starlark couldn't handle.

internal/node/node_repositories.bzl
```diff
@@ -188,7 +191,11 @@ def _prepare_node(repository_ctx):
     Args:
       repository_ctx: The repository rule context
     """
-    is_windows = os_name(repository_ctx).find("windows") != -1
+    is_windows_os = os_name(repository_ctx).find("windows") != -1
```
gregmagolan: this could just be `os_name(repository_ctx) == "windows"`

internal/node/node_repositories.bzl
```diff
@@ -188,7 +191,11 @@ def _prepare_node(repository_ctx):
...
+
+    # TODO: Maybe we want to encode the OS as a specific attribute rather than do it based on naming?
+    is_windows_repository = repository_ctx.attr.name.find("windows") != -1
+    is_windows = is_windows_os or is_windows_repository
```
gregmagolan: this doesn't seem right. what matters is if the repository is windows when preparing it not if the host os is right? I'll do some testing on Windows tomorrow.

nit: also `is_windows_repository = repository_ctx.attr.name.find("_windows_") != -1` is a bit safer.

internal/node/test/BUILD.bazel
```diff
@@ -1,5 +1,8 @@
 load("//:defs.bzl", "nodejs_binary")
 load("//internal/js_library:js_library.bzl", "js_library")
+load(":nodejs_toolchain_test.bzl", "nodejs_binary_test_suite")
+
+nodejs_binary_test_suite()
```
gregmagolan: 😎

internal/npm_install/npm_install.bzl
```diff
@@ -166,9 +166,10 @@ def _npm_install_impl(repository_ctx):
 
     _check_min_bazel_version("npm_install", repository_ctx)
 
-    is_windows = os_name(repository_ctx).find("windows") != -1
-    node = repository_ctx.path(get_node_label(repository_ctx))
-    npm = get_npm_label(repository_ctx)
+    os = os_name(repository_ctx)
+    is_windows = os.find("windows") != -1
```
gregmagolan: nit: is_windows = os = "windows"

internal/node/node_labels.bzl
```diff
@@ -17,37 +17,37 @@
 Labels are different on windows and linux/OSX.
 """
 
-def get_node_label(repository_ctx):
-    if repository_ctx.os.name.lower().find("windows") != -1:
-        label = Label("@nodejs//:bin/node.cmd")
+def get_node_label(os_name):
+    if os_name.find("windows") != -1:
```
gregmagolan: nit: os_name == "windows"

README.md
```diff
@@ -438,6 +438,12 @@ Note: the arguments passed to `bazel run` after `--` are forwarded to the execut
 
 [bazel instructions]: https://docs.bazel.build/versions/master/install.html
 
+### Toolchains
+
+Aka cross-compilation aka building docker images on mac/windows. When you add `node_repositories()` to your `WORKSPAE` file it will setup node for all currently supported platforms, Linux, macOS and Windows. If you want to build an app or a docker image for another platform you have to pass in the `--platforms` flag to Bazel. E.g. `bazel build --platforms=@build_bazel_rules_nodejs//toolchains/node:linux_amd64 //app` will ensure that the linux nodejs binary is downloaded and provided to the nodejs_binary.
```
gregmagolan: nit: spelling `WORKSPACE`

Maybe put the bit about cross compilation & building docker images for another platform to a subheading such as #### Cross-compilation.

Also, I don't think "aka" quite fits as toolchains serve another purpose which is to de-couple the nodejs toolchain definition from the usage in nodejs_binary (https://docs.bazel.build/versions/master/toolchains.html#overview). This also enables RBE when building on osx & windows. A better description could be that toolchains "allow" cross compilation and building docker images for non-host platforms?

README.md
```diff
@@ -438,6 +438,12 @@ Note: the arguments passed to `bazel run` after `--` are forwarded to the execut
...
+
+Note: The toolchain currently only provides a platform-specific nodejs binary. Any native modules will still be fetched/built, by npm/yarn, for your host platform, so they will not work on the target platform. Support for cross-compilation with native dependencies will follow.
```
gregmagolan: 😎

internal/node/node.bzl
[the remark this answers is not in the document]
gregmagolan: 👍 to fail. should not happen anyway as you're checking for that condition in `node_configure`

internal/common/os_name.bzl
```diff
@@ -26,10 +34,10 @@ def os_name(repository_ctx):
...
-        return "darwin_amd64"
+        return OS_NAMES[0]
     elif os_name.find("windows") != -1:
-        return "windows_amd64"
+        return OS_NAMES[1]
     elif os_name.startswith("linux"):
-        return "linux_amd64"
+        return OS_NAMES[2]
     else:
         fail("Unsupported operating system: " + os_name)
```
gregmagolan: Maybe the reverse functions belong in here?

```
def is_windows(os_name):
    return os_name == OS_NAMES[1]
```

etc... so that we're not assuming os_name.find("windows") is the correct check in various spots

internal/node/node_repositories.bzl
Globegitter: As long as the sha is provided (which it is per default) we will only download it once, but yes there is duplicate work with the tar extraction etc and if the `--experimental_repository_cache_hardlinks` flag is not used then it will also take up twice the space on disk. But yeah I also thought it is a worthy tradeoff and I could not think of any other way how we currently could do this.

 If bazel was to support a WORKSPACE alias as requested here https://github.com/bazelbuild/bazel/issues/3219 then I think we could have another repository rule which would generate a `.bzl` file with the correct alias, that the user then would then import as an alternative.

Or maybe actually now that I am thinking about it we could do something similar also without an alias. I will have to play around with it if what I am thinking can actually work out and would improve the situation.

gregmagolan: Nice. I didn't realize `download_and_extract` uses the repository cache. Cool 👍  That elimates most of the duplication of effort.

toolchains/node/node_toolchain.bzl
```diff
@@ -0,0 +1,50 @@
...
+    doc = "Information about how to invoke the node binary.",
+    fields = {
+        "target_tool": "A hermetically downloaded nodejs executable target.",
+        "target_tool_path": "Path to an existing nodejs executable",
+    },
+)
+
+def _node_toolchain_impl(ctx):
+    if not ctx.attr.target_tool and not ctx.attr.target_tool_path:
+        print("No nodejs binary was found or built, executing run for rules_nodejs targets might not work.")
```
gregmagolan: Should this be a fail? I can't think of a case where this would be desirable.

internal/node/node_repositories.bzl
Globegitter: Yeah, now thinking about it I think you are right. We care about which os the repository is for, so cross compilation on windows might indeed be broken with this. We only care about the host os when we run `repository_ctx.execute` as well as for the alias repository `nodejs`. So maybe we should change this to `(is_windows_os and repository_ctx.attr.name == "nodejs") or ...`, or as my comment says above we could also start passing in an OS attribute to be a bit more generic. WDYT?

internal/node/node_repositories.bzl
Globegitter: It is already within the toolchains directory so I thought that would be clear from there. But can still rename if you prefer. Let me know.

internal/node/node_repositories.bzl
Globegitter: Yep agreed would be much nicer if we could get rid of the `generate_build_file.js`. Extended the comment.

toolchains/node/node_toolchain.bzl
Globegitter: Further below we are falling back to looking for `node` on the PATH if nothing was provided. Again that approach is taken from rules_k8s. But maybe we do not want that?

Globegitter:
@gregmagolan Currently `target_tool_path` is actually not plumbed through to `node_repositories`, which is probably what we would want to do and it is not that much effort, we should probably just agree on the API for it. As how can we let the user specify a different path for a different platform? Should accept some form of dict, where we map os to path? And should we also provide the user with a way an easy way to specify take node on PATH? Also do we then want to start supporting taking npm and yarn from path? Or do we still want do download that in all cases? 

Given all that my thinking was to really make this officially supported only in a follow-up PR.

internal/common/os_name.bzl
Globegitter: I am not sure I follow. Could you provide some more detail?

Globegitter:
Seems CI is passing again - the building_karma_package step fails on a dependency download.

internal/node/node_repositories.bzl
gregmagolan: I like the explicit naming so its clear that its configuring a `node_toolchain`. Will avoid the `node_toolchain_configure = "node_configure"` on `node_repositories.bzl`. I'll push this along with a few last touches.

internal/node/node_repositories.bzl
gregmagolan: Nice fix with the alias

internal/common/os_name.bzl
gregmagolan: I cleaned it up. Opted for is_windows_os(rctx) to use in the various places needed so they don't need to check if .find("windows") anymore or use os_name(rctx)

toolchains/node/node_toolchain.bzl
gregmagolan: Ahhh. I see. I node_toolchain_configure you look for "node":
```
def _impl(repository_ctx):
    if repository_ctx.attr.target_tool and repository_ctx.attr.target_tool_path:
        fail("Can only set one of target_tool or target_tool_path but both where set.")

    if repository_ctx.attr.target_tool:
        substitutions = {"%{TOOL_ATTRS}": "    target_tool = \"%s\"\n" % repository_ctx.attr.target_tool}
    else:
        default_tool_path = repository_ctx.attr.target_tool_path or repository_ctx.which("node") or ""
        substitutions = {"%{TOOL_ATTRS}": "    target_tool_path = \"%s\"\n" % default_tool_path}
```
but then it would be set at this point if it was found... and if not then it found fail further down at this point:
```
    if node_tool_info.target_tool_path == "" and not node_tool_info.target_tool:
        # If tool_path is empty and tool_target is None then there is no local
        # node tool, we will just print a nice error message if the user
        # attempts to do bazel run
        fail("The node toolchain was not properly configured so %s cannot be executed. Make sure that target_tool_path or target_tool is set." % ctx.attr.name)
```

gregmagolan: Cleaned this up

gregmagolan:
Replaced by #898

PR #833, opened 2019-06-08
feat(rollup): run terser in threads

Toxicable:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/internal/rollup/package.json
+++ b/internal/rollup/package.json
@@ -12,6 +12,7 @@
         "shelljs": "0.8.3",
         "source-map-explorer": "^1.7.0",
         "terser": "^3.16.1",
+        "p-limit": "^2.2.0",
         "tmp": "0.0.33",
         "typescript": "3.3.3333"
     },
--- a/internal/rollup/rollup_bundle.bzl
+++ b/internal/rollup/rollup_bundle.bzl
@@ -334,6 +334,9 @@ def _run_terser(ctx, input, output, map_output, debug = False, comments = True, 
         args.add("--debug")
         args.add("--beautify")
 
+    # we need this arg to enable worker_threads
+    args.add("--node_options=--experimental-worker")
+
     ctx.actions.run(
         progress_message = "Optimizing JavaScript %s [terser]" % output.short_path,
         executable = ctx.executable._terser_wrapped,
--- a/internal/rollup/rollup_bundle.bzl
+++ b/internal/rollup/rollup_bundle.bzl
@@ -334,7 +334,8 @@ def _run_terser(ctx, input, output, map_output, debug = False, comments = True, 
         args.add("--debug")
         args.add("--beautify")
 
-    # we need this arg to enable worker_threads
+    # we need this arg to enable worker_threads on Node.js v10
+    # TODO: Turn this off for Node.js v12?
     args.add("--node_options=--experimental-worker")
 
     ctx.actions.run(
--- a/internal/rollup/rollup_bundle.bzl
+++ b/internal/rollup/rollup_bundle.bzl
@@ -335,7 +335,7 @@ def _run_terser(ctx, input, output, map_output, debug = False, comments = True, 
         args.add("--beautify")
 
     # we need this arg to enable worker_threads on Node.js v10
-    # TODO: Turn this off for Node.js v12?
+    # TODO: Turn this off for Node.js v12
     args.add("--node_options=--experimental-worker")
 
     ctx.actions.run(
--- a/internal/rollup/terser-wrapped.js
+++ b/internal/rollup/terser-wrapped.js
@@ -20,47 +20,81 @@
 // each .js file in that folder and outputs to a specified output
 // folder. If the input is a file then terser is just run
 // one that individual file.
-
 const fs = require('fs');
 const path = require('path');
-const tmp = require('tmp');
-const child_process = require('child_process');
 
-const DEBUG = false;
+// worker_threads is only avilable on version > 10.5.0
+const [major, minor] = process.version.replace('v', '').split('.');
+const WORKER_THREADS_AVILABLE = major > 10 || major == 10 && minor > 5
+let worker_threads;
 
-// capture the inputs and output options
-const argv = require('minimist')(process.argv.slice(2));
-const inputs = argv._;
-const output = argv.output || argv.o;
-const debug = argv.debug;
-const configFile = argv['config-file'];
-// delete the properties extracted above as the remaining
-// arguments are forwarded to terser in execFileSync below
-delete argv._;
-delete argv.output;
-delete argv.o;
-delete argv.debug;
-delete argv['config-file'];
-
-if (DEBUG)
-  console.error(`
-terser: running with
-  cwd: ${process.cwd()}
-  argv: ${process.argv.slice(2).join(' ')}
-  inputs: ${JSON.stringify(inputs)}
-  output: ${output}
-  debug: ${debug}
-`);
-
-if (inputs.length != 1) {
-  throw new Error(`Only one input file supported: ${inputs}`);
+if (WORKER_THREADS_AVILABLE) {
+  worker_threads = require('worker_threads');
+} else {
+  console.warn(
+      `WARNING: worker_threads not are not avilable on your version of Nodejs: ${process.version}
+    Running Terser in serial mode
+    Upgrade your Nodejs version to 10.5.0 minimum for worker_threads to be avilable`)
 }
 
-const input = inputs[0];
+const DEBUG = false;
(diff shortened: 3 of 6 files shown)


internal/rollup/terser-wrapped.js
```diff
@@ -20,47 +20,81 @@
...
-
 const fs = require('fs');
 const path = require('path');
-const tmp = require('tmp');
-const child_process = require('child_process');
 
-const DEBUG = false;
+// worker_threads is only avilable on version > 10.5.0
+const [major, minor] = process.version.replace('v', '').split('.');
+const WORKER_THREADS_AVILABLE = major > 10 || major == 10 && minor > 5
```
alexeagle: nit: why not just try/catch on the `require` call?

internal/rollup/terser-wrapped.js
```diff
@@ -20,47 +20,81 @@
...
-
-if (inputs.length != 1) {
-  throw new Error(`Only one input file supported: ${inputs}`);
+if (WORKER_THREADS_AVILABLE) {
+  worker_threads = require('worker_threads');
+} else {
+  console.warn(
+      `WARNING: worker_threads not are not avilable on your version of Nodejs: ${process.version}
+    Running Terser in serial mode
+    Upgrade your Nodejs version to 10.5.0 minimum for worker_threads to be avilable`)
```
alexeagle: we should be more specific here, to update the /WORKSPACE file, otherwise users will update the nodejs on their machine and it still won't work

internal/rollup/terser-wrapped.js
```diff
@@ -74,51 +108,92 @@ function runterser(inputFile, outputFile, sourceMapFile) {
...
+}
 
-  fs.writeFileSync(config, JSON.stringify(terserConfig));
+/**
+ *
+ * @param {{inputFile: string, outputFile: string, sourceMapFile: string}[]} files
+ * @param {boolean} debug
+ */
+async function runTerserParallel(files, debug) {
+  // TODO: do some concurrency limiting here
```
alexeagle: do we have to finish the TODO before we merge/release this?

internal/rollup/terser-wrapped.js
```diff
@@ -20,47 +20,81 @@
...
-`);
-
-if (inputs.length != 1) {
-  throw new Error(`Only one input file supported: ${inputs}`);
+if (WORKER_THREADS_AVILABLE) {
+  worker_threads = require('worker_threads');
+} else {
+  console.warn(
+      `WARNING: worker_threads not are not avilable on your version of Nodejs: ${process.version}
+    Running Terser in serial mode
```
alexeagle: "serial mode" is a bit obtuse for a novice user, how about just "optimizing one file at a time"

internal/rollup/terser-wrapped.js
clydin: should also be `minor >= 5`

internal/rollup/rollup_bundle.bzl
```diff
@@ -334,6 +334,9 @@ def _run_terser(ctx, input, output, map_output, debug = False, comments = True,
         args.add("--debug")
         args.add("--beautify")
 
+    # we need this arg to enable worker_threads
+    args.add("--node_options=--experimental-worker")
```
clydin: If on Node 12+ this shouldn't be added

internal/rollup/terser-wrapped.js
clydin: We should.  context switching will destroy performance gains.
`os.cpus().length - 1` is typically a good cap.

internal/rollup/terser-wrapped.js
Toxicable: Total guess here, but I would assume that doing arithmetic is faster than recovering from an exception.
However, in hindsight, it's probably the smallest amount of performance difference here.

internal/rollup/rollup_bundle.bzl
Toxicable: Fair, not sure how to get the nodejs version in in skylark though.
@alexeagle You mentioned there's a way to do it the other day?

internal/rollup/package.json
```diff
@@ -12,6 +12,7 @@
         "shelljs": "0.8.3",
         "source-map-explorer": "^1.7.0",
         "terser": "^3.16.1",
+        "p-limit": "^2.2.0",
```
Toxicable: This dosen't actually work, how do we add internal npm modules?

internal/rollup/terser-wrapped.js
```diff
@@ -20,47 +20,84 @@
...
+      fs.mkdirSync(output);
+    }
+    const dir = fs.readdirSync(input);
+    const files = dir.filter(f => f.endsWith('.js')).map(f => {
+      const inputFile = path.join(input, path.basename(f));
+      const outputFile = path.join(output, path.basename(f));
+      return {inputFile, outputFile, sourceMapFile: outputFile + '.map'};
+    })
+
+    // TODO: allow config to force serial workers here
```
Toxicable: @Globegitter Suggested this one.
It would be a good addition but probably not apart of this first cut.

internal/rollup/terser-wrapped.js
```diff
@@ -74,51 +111,99 @@ function runterser(inputFile, outputFile, sourceMapFile) {
...
-    require.resolve('build_bazel_rules_nodejs_rollup_deps/node_modules/terser/bin/uglifyjs'),
-    inputFile, '--output', outputFile, '--config-file', config
-  ];
+/**
+ *
+ * @param {{inputFile: string, outputFile: string, sourceMapFile: string}[]} files
+ * @param {boolean} debug
+ */
+async function runTerserParallel(files, debug) {
+  // TODO: make this configurable
```
Toxicable: Same here, would be a good candidate for config, but we'll worry about it later I think

internal/rollup/terser-wrapped.js
```diff
@@ -74,51 +111,99 @@ function runterser(inputFile, outputFile, sourceMapFile) {
...
-      isWindows ? 'node.exe' : 'node', args,
-      {stdio: [process.stdin, process.stdout, process.stderr]});
+  const terserOptions = buildTerserConfig(sourceMapFile, debug)
+  const inputCode = (await fs.promises.readFile(inputFile)).toString();
+  const result = Terser.minify(inputCode, terserOptions);
+
+  if (result.error) {
+    // TODO: figure out error reporting here
+    // scroll up a little from this: https://github.com/terser-js/terser#minify-options
+    throw result.error;
```
Toxicable: Im not certain of the format of this error object, we need to do some trail and error I think to figure out what kind of errors we get from terser.
Im not familiar enough with it to know how to cause an error :P

internal/rollup/rollup_bundle.bzl
Globegitter: Once we have toolchains it would actually be possible to add the node version to the toolchain and then one could get it in starlark. Not sure though how it would be possible currently though.

clydin: If problematic, node 12 currently won't error with the option present.  So this could be addressed moving forward.

internal/rollup/terser-wrapped.js
clydin: I think a worker pool would also make sense here.  There's no need to tear down and bring up new threads and re-initialize terser for each file.

internal/rollup/terser-wrapped.js
```diff
@@ -74,51 +108,92 @@ function runterser(inputFile, outputFile, sourceMapFile) {
...
+ * @param {Function} Terser
+ * @param {boolean} debug
+ * @returns {Promise<void>}
+ */
+async function runTerser(inputFile, outputFile, sourceMapFile, Terser, debug) {
+  if (DEBUG) console.error(`Minifying ${inputFile} -> ${outputFile} (sourceMap ${sourceMapFile})`);
 
-  if (DEBUG) console.error(`Running node ${args.join(' ')}`);
+  const terserOptions = buildTerserConfig(sourceMapFile, debug)
+  const inputCode = (await fs.promises.readFile(inputFile)).toString();
```
clydin: Are character encoding and BOM presence relevant here?

Toxicable: I really don't know. We could add a `trim()` call which would remove the BOM if it is actually encoded as that.
I don't know the impact of stripping the BOM and re-writing it as plain UTF8 once we're done

Toxicable:
replaced with https://github.com/bazelbuild/rules_nodejs/pull/1172

PR #839, opened 2019-06-10
I825 no additional entry points

Jesse-Good:
## PR Checklist

- [x ] Tests for the changes have been added (for bug fixes / features)
- [x ] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ x] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Build related changes
- [ ] CI related changes

rollup_bundle build rule requires additional_entry_points explicitly listed.

Issue Number: 825

## What is the new behavior?

rollup_bundle adds boolean parameter auto_additional_entry_points, which can be used instead of additional_entry_points. auto_additional_entry_points is True by default.

- [ x] Yes
- [ ] No

Existing invocations of rollup_bundle() that omit additional_entry_points may need to set auto_additional_entry_points to False.


internal/e2e/rollup_code_splitting/BUILD.bazel
```diff
@@ -11,18 +11,32 @@ rollup_bundle(
     license_banner = ":license.txt",
 )
 
+rollup_bundle(
```
alexeagle: hmm, I think we want to just change most of the tests to remove addl_entry_points and test the auto behavior everywhere, except maybe for one test that's left on the legacy behavior just to make sure we don't break it during the transition.

internal/web_package/test/BUILD.bazel
```diff
@@ -5,6 +5,7 @@ package(default_visibility = ["//visibility:public"])
 
 rollup_bundle(
     name = "bundle",
+    auto_additional_entry_points = False,
```
alexeagle: in particular I think we want to have it True here - I'd like to understand if that's breaking by changing the tests to illustrate just what that breakage is

internal/e2e/rollup_code_splitting/BUILD.bazel
Jesse-Good: Ok, here's another branch where I just convert the test to auto, deleting test coverage for (non auto) additional entry points. I have to delete the tests for additional_entry.js to get this to work. A new chunk-X.js is created (it's working) but I think I can only write a test for it if put X (the hash) into the test. Do you think that's OK? 

https://github.com/bazelbuild/rules_nodejs/compare/master...Jesse-Good:i825-autoEntryPoints2?expand=1

Jesse-Good:
> The new test doesn't demonstrate that code splitting is happening - are there actually additional chunks produced?

Yes, chunk is being produced but with name chunk-X.js. Is it safe to write a test against the hashed X value?

internal/e2e/rollup_code_splitting/BUILD.bazel
Jesse-Good: PTAL. I've updated this branch.

internal/web_package/test/BUILD.bazel
Jesse-Good: //internal/web_package/test2:test fails because the <script> tags look different. There's a bunch of <script type="module" and nomodule in the actual output but the test is not expecting that. 
[test.log](https://github.com/bazelbuild/rules_nodejs/files/3298439/test.log)

alexeagle: paired on this in person - we now understand that the golden file changes to the HTML are desirable - enabling code splitting had the side effect of enabling differential loading

internal/rollup/rollup_bundle.bzl
```diff
@@ -611,10 +611,19 @@ ROLLUP_ATTRS = {
         It is sufficient to load one of these SystemJS boilerplate/entry point
         files as a script in your HTML to load your application""",
     ),
+    "auto_additional_entry_points": attr.bool(
```
alexeagle: decided to rename it "enable_code_splitting" since this is orthogonal to the number of entry points

internal/rollup/rollup_bundle.bzl
```diff
@@ -611,10 +611,19 @@ ROLLUP_ATTRS = {
...
+        doc = """Set this to True instead of specifying additional_entry_points to
```
alexeagle: add a link to https://rollupjs.org/guide/en#code-splitting

and "code splitting is always enabled when additional_entry_points is non-empty"

internal/e2e/rollup_code_splitting/BUILD.bazel
Jesse-Good: Ok, I've updated :bundle to enable code splitting. And I added a new test to cover the previous behavior with specifying additional_entry_points.

Jesse-Good:
PTAL, merged from origin and responded to comments.

internal/rollup/rollup.config.js
```diff
@@ -206,17 +202,16 @@ const config = {
   preserveSymlinks: true,
 }
 
-if (enableCodeSplitting) {
-  config.experimentalCodeSplitting = true;
-  config.experimentalDynamicImport = true;
```
gregmagolan: Nice. Looks like the feature is no longer experimental since I worked on this 👍

internal/e2e/rollup_code_splitting/BUILD.bazel
```diff
@@ -7,7 +7,19 @@ rollup_bundle(
...
+)
+
+rollup_bundle(
+    name = "bundle_multi_entry",
+    srcs = glob(
+        ["*.js"],
+        exclude = ["*.spec.js"],
+    ),
     additional_entry_points = ["build_bazel_rules_nodejs/internal/e2e/rollup_code_splitting/additional_entry.js"],
+    enable_code_splitting = False,
```
gregmagolan: This would implied to be true because `additional_entry_points` is specified?

alexeagle: yeah this should have been an error, will fix

alexeagle:
I'll try it in angular-bazel-example before next release - want to get Jesse's PR in under his authorship before I do more with it

Issue #809, opened 2019-05-31
CI fails with Python toolchains (Bazel 0.27)


brandjon:
Bazel 0.27 enables [`--incompatible_use_python_toolchains`](https://github.com/bazelbuild/bazel/issues/7899). This [breaks](https://buildkite.com/bazel/bazelisk-plus-incompatible-flags/builds/126#3ed74187-9bd0-4bf7-8d1c-0dcfc3d5eb70) `rules_nodejs`'s BuildKite pipeline.

The error message indicates that this is because you have Python 3 targets, and the autodetecting toolchain can't find the `python3` command. Our CI workers have `python3` installed under `/usr/local/bin`, but the `PATH` is apparently `.:/bin:/usr/bin`.

This is likely because you have `--incompatible_strict_action_env` in your bazelrc. I've filed bazelbuild/bazel#8536 to track this problem. In the meantime, you may be able to work around this by setting `--action_env` in your bazelrc.

brandjon:
This problem [does not reproduce](https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1009#c865b486-d892-48cf-83c6-ce12a63d6c3f) in the most recent runs due to an unrelated breakage.

brandjon:
I've confirmed that in the case of a `py_test` (but not necessarily `py_binary`!), the action environment (`--action_env`) is respected. The error indicates that the `PATH` is

```
.:/bin:/usr/bin
```

The `/bin:/usr/bin` part comes from the [default value](https://github.com/bazelbuild/bazel/blob/9ca0744b92035f236c98580a491722bda6a68fc3/src/main/java/com/google/devtools/build/lib/bazel/rules/BazelRuleClassProvider.java#L453) for `PATH` under a strict action env on a non-windows platform. The preceding `.` part comes from [test-setup.sh](https://github.com/bazelbuild/bazel/blob/9ca0744b92035f236c98580a491722bda6a68fc3/tools/test/test-setup.sh#L224).

It seems that a solution would be to add `--action_env=PATH` to the bazelrc. I've confirmed this on a mac worker with an image similar to our CI runners. The alternative requires adding an explicit Python toolchain that knows how to find Python 3 on mac. It's unclear whether that kind of toolchain definition belongs in rules_nodejs or our CI workers.

brandjon:
Ping @alexeagle for visibility, as this affects the 0.27 Bazel release candidate.

alexeagle:
[...]

PR #842, opened 2019-06-11
test: disable our one py_test

alexeagle:
This should unblock Bazel 0.27 RC which requires a python toolchain

See #809


alexeagle:
The flag is in bazel but you can't set it with the `common` pseudo-command unless it's valid for every other command. Let's try with just `build` and `test`

PR #860, opened 2019-06-13
fix for angular/angular-bazel-example/issues/364

gregmagolan:
This removes the need for downstream users that use rxjs with Bazel to maintain
their own rxjs_shims.js file. rxjs used commonly enough that the shims file can
be moved to `@build_bazel_rules_nodejs//umd_shims:rxjs-shims.umd.js` and
referenced from there downstream.

Upcoming rxjs 7 release may be publishing separate UMD bundles for each deep import which
would solve the underlying issue by not requiring any shims file.

Fixes https://github.com/angular/angular-bazel-example/issues/364 (kindof)

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -65,6 +65,7 @@ npm_package(
         "//internal/npm_install:package_contents",
         "//internal/npm_package:package_contents",
         "//internal/web_package:package_contents",
+        "//umd_shims:package_contents",
     ],
 )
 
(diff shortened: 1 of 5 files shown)


alexeagle:
decided over chat that we can just publish a super-low-tech `rxjs-requirejs` package to npm

This has the advantage that we can use package.json peerDeps to guide users about which versions of rxjs the compat shim should be used with

in theory the README there could explain how to use rxjs with requirejs in the way that we do, but maybe no one cares

PR #862, opened 2019-06-13
fix(typescript): fix check_rules_typescript_version function

gregmagolan:
@alexeagle Should we keep this function or just remove it? Not likely to be used by anyone at the moment as its not against the right version. If we keep it seems we should also add one to each npm package.

Example usage:

```
load("@npm_bazel_typescript//:version.bzl", "check_rules_typescript_version")
check_rules_typescript_version(minimum_version_string = "0.31.1")
```

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

`check_rules_typescript_version` is out of date and not checking against the correct version

## What is the new behavior?

`check_rules_typescript_version` fixed

- [ ] Yes
- [x] No

### DIFF
--- a/e2e/typescript_2.7/WORKSPACE
+++ b/e2e/typescript_2.7/WORKSPACE
@@ -37,3 +37,8 @@ install_bazel_dependencies()
 load("@npm_bazel_typescript//:index.bzl", "ts_setup_workspace")
 
 ts_setup_workspace()
+
+load("@npm_bazel_typescript//:version.bzl", "check_rules_typescript_version")
+
+# Smoke test for check_rules_typescript_version
+check_rules_typescript_version(minimum_version_string = "0.31.1")
--- a/e2e/typescript_2.8/WORKSPACE
+++ b/e2e/typescript_2.8/WORKSPACE
@@ -37,3 +37,8 @@ install_bazel_dependencies()
 load("@npm_bazel_typescript//:index.bzl", "ts_setup_workspace")
 
 ts_setup_workspace()
+
+load("@npm_bazel_typescript//:version.bzl", "check_rules_typescript_version")
+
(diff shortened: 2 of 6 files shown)


alexeagle:
I think we added this function in order to fix the error message if you have skewed versions. We can drop it but should think about what bad state users can get into, and what error they'll get.
In other places like angular/compiler-cli we have a JS conditional that throws if a peer dep is out-of-range. Is that what we need here?
Maybe it would be enough for `npm install` to report peerDep warnings if you have wrong `@bazel/typescript`?

gregmagolan:
Reporting failing peerDeps from npm install and yarn install would be good. Tho warnings are typically ignored anyway which is why there is no warning function in starlark. Not sure the best way to approach that.

alexeagle:
Promoting peerDeps warnings to error seems really useful generally as an npm workflow. we could put that in but why doesn't the `yarn` or `npm` program do that? seems like their job; maybe they had a good reason not to add it

Toxicable:
>Promoting peerDeps warnings to error seems really useful generally as an npm workflow

If we're doing this, then it will need a opt out flag.
We currently have a bunch of peedDep warning becasue the 3rd party lib hasn't updated to angular v8 yet
But it's totally fine since there wern't any breaking changes between 7->8

gregmagolan:
> > Promoting peerDeps warnings to error seems really useful generally as an npm workflow
> 
> If we're doing this, then it will need a opt out flag.
> We currently have a bunch of peedDep warning becasue the 3rd party lib hasn't updated to angular v8 yet
> But it's totally fine since there wern't any breaking changes between 7->8

Yes. I think pretty much any medium to large project has peerDep warnings that they ignore. Sometimes for good reasons and sometimes just because nothing is broken.

PR #863, opened 2019-06-13
Add NodeJSRuntimeInfo provider

Globegitter:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

When using `nodejs_image` of `rules_docker` there currently was no way for it to put the npm modules and source files into different layers so even if just changes are being made to the source files incremental rebuilds will have to repack all of node_modules which can be quite slow. In our app a rebuild could take up to 10s.

## What is the new behavior?
We now expose a provider that other rules can access the needed `nodejs_binary` runfiles separately, split by its toolchain (i.e. the node binary), node_modules, source files and all other necessary/implicit runfiles. This would allow for `nodejs_image´ to create dedicated layers and in my tests it brought incremental builds, where just source files changed, down to ~3s.

- [ ] Yes
- [x] No

## Other information

This replaces #283 as it has gotten stale by now but also updates the provider to be compatible with the upcoming toolchain support.

### DIFF
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -208,6 +208,7 @@ def _nodejs_binary_impl(ctx):
             # https://github.com/bazelbuild/bazel/issues/6908
             runfiles = ctx.runfiles(
                 transitive_files = runfiles,
+                collect_data = True,
             ),
         ),
         NodeJSRuntimeInfo(
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -25,6 +25,19 @@ load("//internal/common:expand_into_runfiles.bzl", "expand_location_into_runfile
 load("//internal/common:module_mappings.bzl", "module_mappings_runtime_aspect")
 load("//internal/common:sources_aspect.bzl", "sources_aspect")
 
+NodeJSRuntimeInfo = provider(
+    doc = """All files needed to run the NodeJS binary, separated by its source files, node_modules, node_runfiles, and
+    the toolchain, i.e. the node binary itself.
+    This can be useful for example when creating a docker image to be able to put npm packages and data files on
+    separate layers for better caching.""",
+    fields = {
+        "node_modules": "A depset of node modules file objects, i.e. third party dependencies",
+        "node_runfiles": "A depset of files needed for each nodejs_binary to run",
+        "sources": "A depset of non node-module source file objects",
+        "toolchain": "A depset of non node-module source file objects",
+    },
+)
+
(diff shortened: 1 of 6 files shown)


alexeagle:
Cool, I agree we should fix this
One nit: we should have a plan for where Providers are loaded from. Things in /internal should be private API where I think rule authors should probably use all the Providers so they should be public API?
We can discuss in the meeting today how many files they should go in.

Globegitter:
Ah just saw that comment now, so missed out on talking with @gregmagolan about that. But yes it would be good to find some more public place for these. We also depend on the `NodeModuleSources` provider in some internal rules.

I also started to adding starlark unit tests, which I think are the perfect way to test this. Just need to finish them up and make sure they also pass on windows.

Globegitter:
I am a fan of these starlark unit-tests. They uncovered some inefficiencies - that we where including all node_modules in sources as well. So we where passing some files around twice. That should be fixed now. Just hoping it will not break anything else.

alexeagle:
some broken logic doing `path.join` somewhere...
`Cannot find module '/home/circleci/.cache/bazel/_bazel_circleci/02fc451c9ce3986cffaddce25971cb5b/sandbox/processwrapper-sandbox/82/execroot/build_bazel_rules_nodejs/bazel-out/k8-fastbuild/bin/internal/e2e/rollup/test.runfiles/npm/node_modules/home/circleci/.cache/bazel/_bazel_circleci/02fc451c9ce3986cffaddce25971cb5b/sandbox/processwrapper-sandbox/82/execroot/build_bazel_rules_nodejs/bazel-out/k8-fastbuild/bin/internal/e2e/rollup/test.runfiles/build_bazel_rules_nodejs/internal/e2e/rollup/bundle.min.js'`

Globegitter:
Hmm yeah I will investigate.

Globegitter:
@alexeagle Found a fix and also cleaned up the way runfiles are created and added a test to make sure we have the same contents as before. So this PR is also coming with some slight performance improvements, at least theoretical ones.

Globegitter:
Ok still some of the tests failing - will investigate.

Globegitter:
So, should be all fixed now @alexeagle

Globegitter:
Now that we have toolchain support in we can actually update this provider to include the real toolchain. Will do so shortly.

alexeagle:
I think maybe https://github.com/bazelbuild/rules_docker/pull/1346 made this obsolete? We now get node_modules in a lower layer of the docker image?

PR #861, opened 2019-06-13
build: update to bazel 0.27.0rc5

gregmagolan:
Testing against Bazel 0.27.0rc5 as I saw an unexpected error locally:

```
Building package typescript //:npm_package
checking deps: jasmine
all deps built
+ bazel build --workspace_status_command=../../scripts/current_version.sh //:npm_package
Starting local Bazel server and connecting to it...
DEBUG: Rule 'build_bazel_rules_typescript' indicated that a canonical reproducible form can be obtained by modifying arguments shallow_since = "1559861420 -0700"
DEBUG: Call stack for the definition of repository 'build_bazel_rules_typescript' which is a git_repository (rule definition at /private/var/tmp/_bazel_greg/b90b17756fe48c3e5b8d42d3ed61ee61/external/bazel_tools/tools/build_defs/repo/git.bzl:252:18):
 - /Users/greg/google/gregmagolan/rules_nodejs/packages/typescript/WORKSPACE:44:1
INFO: SHA256 (https://github.com/manekinekko/rules_sass/archive/9862dfc96a4a1f66fe171ef5e043b29853e8445b.zip) = 6035f25dbcd1e5575a681f5d7b9d9821dde09fcf73ccea053d1bc1c9523b1a2a
DEBUG: Rule 'io_bazel_rules_sass' indicated that a canonical reproducible form can be obtained by modifying arguments sha256 = "6035f25dbcd1e5575a681f5d7b9d9821dde09fcf73ccea053d1bc1c9523b1a2a"
DEBUG: Call stack for the definition of repository 'io_bazel_rules_sass' which is a http_archive (rule definition at /private/var/tmp/_bazel_greg/b90b17756fe48c3e5b8d42d3ed61ee61/external/bazel_tools/tools/build_defs/repo/http.bzl:237:16):
 - /private/var/tmp/_bazel_greg/b90b17756fe48c3e5b8d42d3ed61ee61/external/build_bazel_rules_nodejs/package.bzl:40:5
 - /Users/greg/google/gregmagolan/rules_nodejs/packages/typescript/WORKSPACE:31:1
ERROR: /Users/greg/google/gregmagolan/rules_nodejs/packages/typescript/BUILD.bazel:71:20: **kwargs arguments are not allowed in BUILD files. Pass the arguments in explicitly.
ERROR: error loading package '': Package '' contains errors
INFO: Elapsed time: 17.784s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (1 packages loaded)
```

PR #866, opened 2019-06-15
Update to Bazel 0.27.0rc5

gregmagolan:
Replaces https://github.com/bazelbuild/rules_nodejs/pull/861

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,9 +9,9 @@
 
 ## IMPORTANT
 # If you change the `default_docker_image` version, also change the `cache_key` version
-var_1: &default_docker_image circleci/node:10.12
-var_2: &browsers_docker_image circleci/node:10.12-browsers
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -9,9 +9,9 @@
 
 ## IMPORTANT
 # If you change the `default_docker_image` version, also change the `cache_key` version
-var_1: &default_docker_image circleci/node:10.12
-var_2: &browsers_docker_image circleci/node:10.12-browsers
-var_3: &cache_key node-0.12-{{ checksum "yarn.lock" }}-{{ checksum "examples/program/yarn.lock" }}
+var_1: &default_docker_image circleci/node:10.14
```
alexeagle: did we check that the 0.27 release notes covered this? should comment there as good citizens if not

alexeagle: It's probably the Ubuntu 14 deprecation that's in the release notes alerayd

PR #873, opened 2019-06-17
chore: use v2 Bazel Bash runfiles init

alexeagle:


e2e/bazel_bin/test.sh
```diff
@@ -1,31 +1,16 @@
...
-  source "$(grep -m1 "^bazel_tools/tools/bash/runfiles/runfiles.bash " \
-            "$RUNFILES_MANIFEST_FILE" | cut -d ' ' -f 2-)"
-else
-  echo >&2 "ERROR: cannot find @bazel_tools//tools/bash/runfiles:runfiles.bash"
-  exit 1
-fi
-# --- end runfiles.bash initialization ---
+# --- begin runfiles.bash initialization v2 ---
+# Copy-pasted from the Bazel Bash runfiles library v2.
+set -uo pipefail; f=bazel_tools/tools/bash/runfiles/runfiles.bash
```
laszlocsomor: Remove `set -uo pipefail; ` to relax the error handling, like the old versions did.

The old runfiles init boilerplate has `set -euo pipefail` [1] but `e2e/bazel_bin/test.sh` (and the other changed files) omitted that line.

[1] https://github.com/bazelbuild/bazel/blob/616481a6b4ffa1818c04bedefe138880b45fe7fa/tools/bash/runfiles/runfiles.bash#L40

[2] https://github.com/bazelbuild/bazel/blob/934045fb0d80f3b9d89238b19dbec06863c47b33/tools/bash/runfiles/runfiles.bash#L56

alexeagle:
@laszlocsomor this fails on Mac only.

When I debugged on mac and added more printlns I see it is trying to read `/dev/null/bazel_tools/tools/bash/runfiles/runfiles.bash` - does it seem like a bug in the runfiles library? anyone else successfully using the v2 on mac?

laszlocsomor:
> When I debugged on mac and added more printlns I see it is trying to read `/dev/null/bazel_tools/tools/bash/runfiles/runfiles.bash` - does it seem like a bug in the runfiles library?

It's not, the init code and runfiles library do this intentionally, in case some envvar is unset.

This is how the init code tests whether RUNFILES_DIR is set and valid: `${RUNFILES_DIR:-/dev/null}/bazel_tools/...`. If it's set, the combined path is valid. Otherwise the default value `/dev/null` is taken yielding a non-existent path `/dev/null/bazel_tools/...` (which is expected) that the code fails to source, and so the init code tries a different path.

laszlocsomor:
Btw, you don't _have_ to use the v2 init code. The v1 init code still works. The v2 init code is shorter and easier to copy-paste, so in new code you add fewer boilerplate lines.

alexeagle:
@laszlocsomor oh! I see now, we have `set -e` above the runfiles init, so that makes the `source /dev/null/path` fail. Interesting that bash on ubuntu doesn't have the same behavior.

I'll move the `set -e` after the runfiles init. It feels like this might be a common problem since scripts often set -e at the top, and paste the runfiles init later.

laszlocsomor:
Seeing that you just pushed a commit, I wanted to say again that using the v2 init code is _not_ necessary, the v1 code is still fully functional.

alexeagle:
Wow, you're watching closely! Yeah I'm considering dropping this, as you say there's not much benefit, but thought I'd spend a few minutes getting it green first.

alexeagle:
Greg did this in #1531

Issue #823, opened 2019-06-05
Debugging node process started by Bazel from VSCode


alexeagle:
I don't know if this is a real issue, just copying it from the JIRA board. 
Not sure how to reproduce. Maybe @vikerman or @Toxicable knows.

Toxicable:
The cause of this issue is this here: https://github.com/bazelbuild/rules_typescript/issues/254
When you have `preserve_symlinks = False` set

Otherwise if you have `preserve_symlinks = True` ( the default setting) then you get this one here: https://github.com/bazelbuild/rules_typescript/issues/321

## ts_library

`preserve_symlinks =True` https://github.com/bazelbuild/rules_typescript/issues/254
When debugging a nodejs program in VS Code and you hit a a breakpoint by using the `debugger` keyword, VS Code will open a the symlinked version of your source file which sits in bazels output dir. 
When this happens VS Code will trying to parse all of the TS in this folder causing all TS language support to stop working.
You also cannot use the breakpoint UI in VS Code to set breakpoints, you must use the `debugger` keyword.

`preserve_symlinks = False` https://github.com/bazelbuild/rules_typescript/issues/321
When this is set you must once again use  the `debugger` keyword to hit a breakpoint, but when you do the file that VS Code opens is readonly and cannot be edited.

## jasmine_node_test 

Also another issue that I havn't documented is when you try debugging a source file from under `jasmine_node_test` VS Code takes you to the JS version of the file, but instead of the line of code being executed the source map points to something like line 3, character 13000, it's only pointing to the AMD header and due to this is effectively useless.

## Proposed Solution

The solution I propose to solve this is that we modify the source maps from `ts_library` to point to the real source files in the users directory.
This could be done by adding a post-processing step to tsc_wrapped which goes through each of the produced sourcemap file and processes it's urls to point back to the original repository

Toxicable:
One downside I see to this approach is that solving this problem with the ts plugin/transformer approach means it only solves it for files compiled under `ts_library`.
Which will cover 99% of use cases i'd expect, but im not certain if thats enough.

Toxicable:
Ok after some more investigation the above solution is not a great solution.
We need to find a value for `config["compilerOptions"]["sourceRoot"]` can allow the source map program to traverse from the runfiles back to the source files.

[...]

PR #874, opened 2019-06-18
Jasmine builtin coverage

Toxicable:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -157,7 +157,7 @@ def _nodejs_binary_impl(ctx):
         if k in ctx.var.keys():
             env_vars += "export %s=\"%s\"\n" % (k, ctx.var[k])
 
-    if ctx.attr.coverage:   
+    if ctx.attr.coverage:
         env_vars += "export NODE_V8_COVERAGE=$TEST_TMPDIR/$BUILD_TIMESTAMP\n"
 
     expected_exit_code = 0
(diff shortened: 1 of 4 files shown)


Toxicable:
First attempt at adding coverage to `nodejs_binary` as a builting feature, rather than just `jasmine_node_test`.
In it's current state this will not work.

Using NODE_V8_COVERAGE nodejs(or v8?) will only write the coverage report once the process exits. Since the `jasmine_runner.js` is effectively apart of the users code, it cannot access the coverage report.

One solution here would be to add another program to `node_launcher.sh` that only runs when coverage is enable that collects up the coverage report and turns it into a human readable report for the dev or stores it away for bazel to use.

Toxicable:
Put in a feature request here: https://github.com/nodejs/node/issues/28283
Which would solve the above issues.

Toxicable:
Ok so the solution here is still to use the inspector api, with a few extra flags.
So no need to add the coverage flag to `nodejs_binary`
However, this is now blocked on https://github.com/bazelbuild/rules_nodejs/issues/823
To resolve better source maps

Toxicable:
replaced with https://github.com/bazelbuild/rules_nodejs/pull/1135

PR #875, opened 2019-06-19
fix(typescript): exclude typescript lib declarations in...

gregmagolan:
...node_module_library transitive_declarations as the including of these .d.ts  is controlled by compilerOptions libs (except for typescript.d.ts and tsserverlibrary.d.ts which are imported standardly)
. Including all of the can cause typings conflicts.

Unblocks https://github.com/angular/angular/pull/31019


gregmagolan:
Fix for
```
packages/service-worker/worker/src/service-worker.d.ts:30:15 - error TS2300: Duplicate identifier 'Client'.

30 declare class Client {
                 ~~~~~~

  external/npm/node_modules/typescript/lib/lib.webworker.d.ts:694:11
    694 interface Client {
                  ~~~~~~
    'Client' was also declared here.
  external/npm/node_modules/typescript/lib/lib.webworker.d.ts:701:13
    701 declare var Client: {
                    ~~~~~~
    and here.
```
included and test coverage as well

PR #897, opened 2019-07-03
test: switch back to master branch of angular-bazel-example integration test/example

gregmagolan:


scripts/setup_examples_angular.sh
```diff
@@ -28,8 +28,7 @@ printf "\n\nSetting up /examples/angular\n"
   # Clean example
   echo_and_run cd ${EXAMPLES_DIR}
   rm -rf angular
-  # TODO(gregmagolan): switch back to upstream
-  echo_and_run git clone --depth 1 --single-branch --branch node_modules-fix-proto https://github.com/gregmagolan/angular-bazel-example.git angular
+  echo_and_run git clone --depth 1 --single-branch --branch master https://github.com/gregmagolan/angular-bazel-example.git angular
```
alexeagle: why is `gregmagolan` still in the path?

gregmagolan: 🤦‍♂

PR #908, opened 2019-07-11
feat(stylus): add initial stylus rule

alexeagle:
Thanks for suggesting this in https://github.com/angular/angular/issues/19058 @glebmachine


alexeagle:
I live-coded this, here's a recording:
https://drive.google.com/drive/folders/1I71w8JSGcZk1nUX2oXvT8NLT47Onb1t0

I'll post it more widely too

packages/stylus/index.bzl
```diff
@@ -0,0 +1,43 @@
...
+    attrs = {
+        "src": attr.label(
+            mandatory = True,
+            allow_single_file = True,
+        ),
+        "deps": attr.label_list(
+            allow_files = True,
+        ),
+        "_compiler": attr.label(
+            default = Label("@npm//stylus/bin:stylus"),
```
gregmagolan: Should be `"compiler"` so its not a private attribute to allow users to change to `@foobar//stylus/bin:stylus`

packages/stylus/BUILD.bazel
```diff
@@ -0,0 +1 @@
+# Marker that this directory is a bazel package
```
gregmagolan: `npm_package` rule and also add to `scripts/packages.sh`? At one point I tried adding the release script to CI but bazel was crashing in CircleCI and haven't track it down yet.

gregmagolan:
Also add `stylus` to `.bazelignore` to fix cirle `test` job & buildkite

packages/stylus/index.bzl
```diff
@@ -0,0 +1,43 @@
...
+
+    # We want foo.styl to produce foo.css
+    output_name = src.basename[:-5] + ".css"
+    css_output = ctx.actions.declare_file(output_name)
+    map_output = ctx.actions.declare_file(output_name + ".map")
+    ctx.actions.run(
+        outputs = [css_output, map_output],
+        inputs = [src] + ctx.files.deps,
+        executable = ctx.executable._compiler,
+        arguments = [
```
alan-agius4: Probably we should enable `compress` and `resolve-url`.

Also, we should probably a way to pass the `include`

More info: http://stylus-lang.com/docs/executable.html

alexeagle: adding a test for `resolve-url` shows an interesting consequence, TAL. Same problem as sourcemaps...

How would users want to use `include`? seems like that should just come from the path of the `deps`?

alan-agius4: I was thinking more something similar to the `sass_binary` and how they handle Sass include paths.

They have an additional attribute for this, I also think that with this approach it might be might be more performant  since imports resolutions will be limited to the fine grained include paths installed of the entire `deps`.

https://github.com/bazelbuild/rules_sass#implicit-output-targets

In future also, we might want to make `compress` and `sourcemap` configurable by the user.

packages/stylus/test/stylus_binary_spec.js
```diff
@@ -0,0 +1,25 @@
...
+const content = require('fs').readFileSync(cssPath, {encoding: 'utf-8'});
+if (content.indexOf('body #logo') < 0) {
+  console.error('Expected the css file to be transformed');
+  process.exitCode = 1;
+}
+if (content.indexOf('width:10px') < 0) {
+  console.error('Expected the css file to be transformed');
+  process.exitCode = 1;
+}
+if (content.match(/\r|\n/)) {
```
alan-agius4: Small nit: `.test` should be faster since you want a boolean check

packages/stylus/docs/install.md
```diff
@@ -0,0 +1,31 @@
...
+**WARNING: this is beta-quality software. Breaking changes are likely. Not recommended for production use without expert support.**
+
+The Stylus rules run the stylus CSS preprocessor with Bazel.
+
+## Installation
+
+Add the `@bazel/stylus` npm package to your `devDependencies` in `package.json`.
+
+Your `WORKSPACE` should declare a `yarn_install` or `npm_install` rule named `npm`.
+It should then install the rules found in the npm packages using the `install_bazel_dependencies' function.
```
alan-agius4: There is a missing backtick after install_bazel_dependencies

packages/stylus/index.bzl
alexeagle: Asked Jeremy, he says this attribute only exists on sass_binary for legacy reasons and it would be better not to include it for new code.

Added user-config for compress and sourcemap.

alexeagle:
rebased on new layout scheme for packages, PTAL

packages/stylus/src/BUILD.bazel
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+
+package(default_visibility = ["//visibility:public"])
+
+bzl_library(
+    name = "bzl",
+    testonly = True,
```
gregmagolan: nit: not necessary unless depending on rules_webtesting:web bzl_library target

packages/stylus/src/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "keywords": [
+        "stylus",
+        "bazel"
+    ],
+    "bazelWorkspaces": {
+        "npm_bazel_stylus": {
+            "rootPath": "."
+        }
+    },
+    "scripts": {
```
gregmagolan: "scripts" no longer necessary in this package.json after packages refactor

packages/stylus/src/index.bzl
```diff
@@ -0,0 +1,68 @@
...
+        "src": attr.label(
+            doc = "A single .styl Stylus file to transform",
+            mandatory = True,
+            allow_single_file = True,
+        ),
+        "compiler": attr.label(
+            doc = """Label that points to the stylus binary to run.
+            If you install your npm packages to a workspace named something other than "npm",
+            you may need to set this to `@my_npm_name//stylus/bin:stylus`""",
+            default = Label("@npm//stylus/bin:stylus"),
```
gregmagolan: Don't use `@npm//stylus` here as it may be a user's version from their package.json. To get to a version of stylus that has been tested against the rule entry point or bin target should be in `@npm//@bazel/stylus`.

packages/stylus/index.bzl
alan-agius4: What is the new alternative/suggested way without include paths?

As if users want to import let's say `bootstrap` in their scss file. They will need to do a full import without include paths?

```css
@import '../../node_modules/bootstrap/scss/functions'
```

Some context https://github.com/angular/angular/issues/31619#issuecomment-512707874

alexeagle: I would expect that to be written `@import 'bootstrap/scss/functions'` - meaning the node_modules should always be in the sass include path. I didn't check if that actually works though.

alan-agius4: Afaik preprocess include paths are empty by default.

alexeagle: yeah so I think we should add a test for this, and hard-code correct paths. I'm not sure users should have control over something that affects paths

alan-agius4: Yeah some tests here are mandatory. I mean we can start with hard coded paths and evaluate the option to add paths publicly if needs be.

PR #920, opened 2019-07-16
Refactor packages to root workspace

gregmagolan:
Removes all of the /packages nested workspaces and have them all build & tested from the root WORKSPACE.

Also updates all packages docs to stardoc.


.bazelrc
```diff
@@ -12,6 +12,13 @@ build --host_force_python=PY2
 build --incompatible_use_python_toolchains
 common --incompatible_string_join_requires_strings
 
+# TODO(gregmagolan): remove the following flags once @io_bazel_rules_go is compatible with Bazel 0.27.0
+build --noincompatible_require_ctx_in_configure_features
+build --noincompatible_no_support_tools_in_action_inputs
+
+# Faster TypeScript compiles
```
alexeagle: I thought we would remove all of these now that recent bazel has the strategy_list_selection flag that selects worker mode automatically in actions that support it?

BUILD.bazel
```diff
@@ -11,6 +11,7 @@ exports_files([
 
 bzl_library(
     name = "bzl",
+    testonly = True,
```
alexeagle: hmm, why?

internal/npm_package/packager.js
```diff
@@ -97,43 +97,49 @@ function main(args) {
 
   // src like baseDir/my/path is just copied to outDir/my/path
   for (src of srcsArg.split(',').filter(s => !!s)) {
-    if (!src.startsWith(baseDir)) {
-      throw new Error(`${src} in 'srcs' does not reside in the base directory, ` +
-        `generated file should belong in 'deps' instead.`);
+    if (src.startsWith('external/')) {
```
alexeagle: unit tests for this file that could be updated?

internal/npm_package/packager.js
```diff
@@ -97,43 +97,49 @@ function main(args) {
...
+      // The `deps` attribute should typically be outputs of other rules.
+      // However, things like .d.ts sources of a ts_library or data attributes
+      // of ts_library will result in source files that appear in the deps
+      // so we have to allow this.
+      return path.join(outDir, path.relative(baseDir, f));
+    }
   }
 
   // deps like bazel-bin/baseDir/my/path is copied to outDir/my/path
   // Don't include external directories in the package, these should be installed
```
alexeagle: is this comment out-of-date now?

packages/jasmine/docs/BUILD.bazel
```diff
@@ -1,12 +1,51 @@
...
+
+load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_test")
+load("@io_bazel_skydoc//stardoc:stardoc.bzl", "stardoc")
+
+package(default_visibility = ["//visibility:public"])
 
 exports_files(["install.md"])
 
-skylark_doc(
+stardoc(
```
alexeagle: any particular reason to change from skydoc in the same PR?

scripts/packages.sh
```diff
@@ -5,6 +5,4 @@ set -eu -o pipefail
 # -u: errors if an variable is referenced before being set
 # -o pipefail: causes a pipeline to produce a failure return code if any command errors
 
-# Define packages so that dependencies come first as locally
-# they will be built in the order provided
-export readonly PACKAGES=( hide-bazel-files jasmine typescript karma labs )
+export readonly PACKAGES=( create hide-bazel-files jasmine typescript karma labs )
```
alexeagle: what would be needed to get rid of this file?
the release script could use a `bazel query` to iterate the things we care about

WORKSPACE
```diff
@@ -42,19 +42,43 @@ Try running `yarn bazel` instead.
...
 )
 
+local_repository(
+    name = "npm_bazel_labs",
+    path = "packages/labs/src",
+)
+
+local_repository(
+    name = "npm_bazel_typescript",
+    path = "packages/typescript/src",
```
alexeagle: I'm still a bit confused, it seems like we still have nested workspaces, so what's the magic that lets us simplify the scripts and CI? just that there's a directory above the workspace so the packages can refer to each other using build_bazel_rules_nodejs labels, then that top dir uses the nested workspace labels?

BUILD.bazel
gregmagolan: io_bazel_rules_webtesting has testonly on its bzl_library targets which feeds downstream to all rules using it; its not needed for all bzl_library targets in this repo but figured consistency is good

packages/jasmine/docs/BUILD.bazel
gregmagolan: wanted to use the same pattern for /doc /src /test for all of the packages and was simpler to have them all use stardoc; there is still one skylark_doc target left at the root

WORKSPACE
gregmagolan: main reason is to support the `http_archive` use can for users that don't want to use `npm_bazel_foobar` without `@bazel/foobar`. 

```
http_archive(name = "npm_bazel_foobar", url="...", ...)
```

The tests and docs can be outside of that workspacer but the rules themselves and their dependencies on other `npm_bazel_other` rules need to be in a `npm_bazel_foobar` namespace for this usecase.

.bazelrc
gregmagolan: 👍 Got pulled in from one of the old nested workspaces but you're right that it can be removed

BUILD.bazel
alexeagle: nothing enforces that we do this (the rule that depends on these is not itself testonly), and these files are not used by tests. In theory a bzl_library could be used downstream by users who depend on our starlark API (not actually required to load a .bzl file today)
I don't see how we'll keep this consistent in the future, and I don't really care about being consistent with rules_webtesting if the sources aren't together.
I think I'd rather not have these

gregmagolan: There was an issue I ran into with not having them in a few places. I'll reproduce

gregmagolan: Here's the one under `/packages/karma/src/BUILD.bazel`:
```
bzl_library(
    name = "bzl",
    srcs = glob(["*.bzl"]),
    deps = [
        "@build_bazel_rules_nodejs//internal/common:bzl",
        "@build_bazel_rules_nodejs//internal/js_library:bzl",
        "@io_bazel_rules_webtesting//web",
    ],
)
```
depends on "@io_bazel_rules_webtesting//web" which has testonly = True so it needs it as well.

gregmagolan: And that feeds into the `stardoc` rule which depends on it:
```
stardoc(
    name = "docs",
    testonly = True,
    out = "index.md",
    input = "@npm_bazel_karma//:index.bzl",
    deps = [
        "@npm_bazel_karma//:bzl",
        # We need to restate local workspace dependencies here in `//foo:bzl`
        # format to work-around a bug in stardoc where .bzl files from
        # `@build_bazel_rules_nodejs//foo:bzl` style deps are not found
        # by the doc generator:
        # ```
        # Exception in thread "main" java.lang.IllegalStateException: File external/npm_bazel_karma/karma_node_test.bzl imported '@build_bazel_rules_nodejs//internal/common:devmode_js_sources.bzl', yet internal/common/devmode_js_sources.bzl was not found, even at roots [.].
        # ```
        "//internal/common:bzl",
        "//internal/js_library:bzl",
    ],
)
```

gregmagolan: The error for the above being:
```
ERROR: /private/var/tmp/_bazel_greg/35306ad70b8da8f7bc31590523de35c0/external/npm_bazel_karma/BUILD.bazel:54:1: in bzl_library rule @npm_bazel_karma//:bzl: non-test target '@npm_bazel_karma//:bzl' depends on testonly target '@io_bazel_rules_webtesting//web:web' and doesn't have testonly attribute set
ERROR: Analysis of target '//packages/karma:generate_README' failed; build aborted: Analysis of target '@npm_bazel_karma//:bzl' failed; build aborted
```

gregmagolan: Removed all the testonly's except for the two places they are necessary

WORKSPACE
gregmagolan: Also, all of the test targets for the packages are the root workspace `build_bazel_rules_nodejs` (under `/packages/foobar/test` by convetion) so `bazel test ...` will test everything and also run all of the build targets in the nested `npm_bazel_other` workspaces as those are deps of the test targets.

The nested workspaces themselves are under `/packages/foobar/src` by convention and there is a `/packages/foobar/src/defaults.bzl` which provides macros for the rules provided to use in the downstream http_archive or local development use case. Could probably use a better name than `defaults.bzl`?

All local tests and packages that depend on other packages use the `defaults.bzl` overrides while e2e tests & examples and downstream consumers of the npm packages will use the `index.bzl` rules.

internal/npm_package/packager.js
gregmagolan: Yes. Good catch. external repo files may now be included if the external repos are explicitly listed in `vendor_external`. This is handled in `npm_package.bzl`:
```
# Takes a depset of files and returns a corresponding list of file paths without any files
# that aren't part of the specified package path. Also include files from external repositories
# that explicitly specified in the vendor_external list.
def _filter_out_external_files(ctx, files, package_path):
    result = []
    for file in files:
        if file.short_path.startswith(package_path):
            result.append(file.path)
        else:
            for v in ctx.attr.vendor_external:
                if file.short_path.startswith("../%s/" % v):
                    result.append(file.path)
    return result
```
I hijacked that attribute you put in originally to support vendoring in the rules_typescript code.

gregmagolan: Necessary to support the `npm_bazel_foobar` local_repository pattern where the rules themselves are. The npm_package for karma, for example, makes use of this:
```
npm_package(
    name = "npm_package",
    srcs = [
        "@npm_bazel_karma//:package_contents",
    ],
    replacements = COMMON_REPLACEMENTS,
    vendor_external = [
        "npm_bazel_karma",
    ],
    deps = [
        ":copy_LICENSE",
        ":generate_README",
        "@npm_bazel_karma//:bazel_karma_lib",
    ],
)
```

scripts/packages.sh
gregmagolan: I think it could be done with a bazel query fairly easily now that the npm_package targets are in the root workspace.

internal/npm_package/packager.js
gregmagolan: Good call. Spec & test updated to include external repo files in srcs & deps.

Had to use ts_library for this instead of mock_ts_library since behavior of the two differs slightly in file paths.

PR #921, opened 2019-07-17
feat(less): initial less_binary rule

alexeagle:


packages/less/index.bzl
```diff
@@ -0,0 +1,48 @@
...
+
+    # We want foo.less to produce foo.css
+    output_name = src.basename[:-5] + ".css"
+    css_output = ctx.actions.declare_file(output_name)
+    map_output = ctx.actions.declare_file(output_name + ".map")
+    ctx.actions.run(
+        outputs = [css_output, map_output],
+        inputs = [src] + ctx.files.deps,
+        executable = ctx.executable.compiler,
+        arguments = [
```
alan-agius4: We should also include the `include-path`, See: http://lesscss.org/usage/#less-options

And comment as https://github.com/bazelbuild/rules_nodejs/pull/908/files/25a81ec2939ec8c2ed14d90a46548eaab0715232..f79e28293b4f3f351b86c75855e05c32bee4d7c5#r304255135

> I was thinking more something similar to the sass_binary and how they handle Sass include paths.
> 
> They have an additional attribute for this, I also think that with this approach it might be might be more performant since imports resolutions will be limited to the fine grained include paths installed of the entire deps.
> 
> https://github.com/bazelbuild/rules_sass#implicit-output-targets
> 
> In future also, we might want to make compress and sourcemap configurable by the user.

packages/less/test/less_binary_spec.js
```diff
@@ -0,0 +1,6 @@
+const cssPath = require.resolve('npm_bazel_less/test/foo.css');
+const content = require('fs').readFileSync(cssPath, {encoding: 'utf-8'});
+if (content.indexOf('.link {\n  color: #428bca') < 0) {
```
alan-agius4: Nit: can use `includes` 

```js
content.includes('.link {\n  color: #428bca')
```

packages/less/index.bzl
alexeagle: Looking at rules_sass, there isn't an example of how this is used. I'll ask @jelbourn to explain this today if he understands it.

Is it just so you can say `import(./thing)` when it's really `./some/deeper/path/to/thing`?

alan-agius4: Yeah that’s usage of it.

gregmagolan:
Replaced by https://github.com/bazelbuild/rules_nodejs/pull/1017

Issue #927, opened 2019-07-19
Angular package format detection doesn't work for Ivy-only packages


alexeagle:
```
"dependencies": {
"@angular/core": "angular/core-builds#master-ivy-aot",
}
```
doesn't contain metadata.json files so our detection logic here
https://github.com/bazelbuild/rules_nodejs/blob/818cfb2291a8fa9f683d16e8dcc24af7b0eb2870/internal/npm_install/generate_build_file.js#L695-L707
fails

kyliau:
context: https://github.com/angular/core-builds/tree/master-ivy-aot
It looks like there is no "special" file here that we could use to detect APF.
Idea: for snapshots, we could include a magic file in repo to indicate angular package format.

kyliau:
For third party packages published to npm we recommend running ngcc and bundling metadata.json so that wouldn't be a problem in the short term.

alexeagle:
Nope, metadata.json is a View-Engine-specific artifact.

codebreach:
Is `rules_nodejs` compatible with IVY?

Sorry I couldn't find any information elsewhere.

The `angular/components` and `angular/angular` repos seem to use `angular/bazel` with their own internal rules but `angular/bazel` is deprecated so just wanted to understand what is the correct way of doing this now?

FWIW, I have been trying to port our app using Ivy to bazel following the example here and I am getting weird errors relating to certain custom elements not being defined (they are). 

```
 error TS-998001: [ngtsc] 'sdk-label' is not a known element:
1. If 'sdk-label' is an Angular component, then verify that it is part of this module.
2. If 'sdk-label' is a Web Component then add 'CUSTOM_ELEMENTS_SCHEMA' to the '@NgModule.schemas' of this component to suppress this message.

3     <sdk-label [label]="label" [isRequired]="required"></sdk-label>
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  projects/draftkit/src/lib/input-field/input-with-label/input-with-label.component.ts:12:16
    12   templateUrl: "./input-with-label.component.html",
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    Error occurs in the template of component InputWithLabelComponent.
```

`sdk-label` is correctly defined and exported from a different `NgModule` which is in the `deps` of this `NgModule`

[...]

PR #929, opened 2019-07-20
Use ANGULAR_PACKAGE magic file to detect APF

kyliau:
`metadata.json` are no longer produced in Ivy mode, and this breaks the
detection logic for Angular Package Format (APF).
As a stop gap solution for the snapshot builds, use ANGULAR_PACKAGE
magic file in the package to detect APF.

Closes https://github.com/bazelbuild/rules_nodejs/issues/927

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/internal/npm_install/generate_build_file.js
+++ b/internal/npm_install/generate_build_file.js
@@ -694,6 +694,9 @@ function filterFiles(files, exts = []) {
  */
 function isNgApfPackage(pkg) {
   const set = new Set(pkg._files);
+  if (set.has('ANGULAR_PACKAGE')) {
+    return true;
+  }
   const metadataExt = /\.metadata\.json$/;
   return pkg._files.some((file) => {
     if (metadataExt.test(file)) {


gregmagolan:
Seems like this would work in most cases tho I can see there being some issues for packages that have "ng-update" such as rxjs but are not in APF:
rxjs has:
```
  "ng-update": {
    "migrations": "./migrations/collection.json"
  },
```

alexeagle:
@gregmagolan am I right that in that case we would put the wrong UMD files into concatjs? And that would break the application? And the user wouldn't even have a remedy to avoid it?

gregmagolan:
> @gregmagolan am I right that in that case we would put the wrong UMD files into concatjs? And that would break the application? And the user wouldn't even have a remedy to avoid it?

That would be possible as if a package has "ng-update" and its a dep then all of its `*.umd.js` files will get added automatically to ts_devserver & ts_web_test_suite. But if any of these umd files are unnamed then it will break things and there will be no remedy.

I think if the goal is to detect APF then we should be looking specifically for its characteristics in https://docs.google.com/document/d/1CZC2rcpxffTDfRDs6p1cfbmKNLA6x5O-NtkJglDaBVs/preview#heading=h.k0mh3o8u5hx:

```
node_modules/@angular/core
├── README.md
|
├── esm2015
|   ├── core
|   |   ├── ....js
|   ├── testing
|   |   ├── ....js
│   ├── core.js (ESM/ES2015)
│   ├── core.js.map
|   ├── testing.js
│   └── testing.js.map
|
├── esm5
│   ├── core.js (ESM/ES5)
│   ├── core.js.map
|   ├── testing.js
│   └── testing.js.map
|
├── fesm2015
│   ├── core.js (ESM/ES2015)
│   ├── core.js.map
|   ├── testing.js
│   └── testing.js.map
|
├── fesm5
│   ├── core.js (ESM/ES5)
│   ├── core.js.map
```

gregmagolan:
The more general solution since we really don't care about APF but we are looking for UMD bundles. We can look through all .umd.js files in a package and detect if they are named-UMD files and only include the ones that are. This wouldn't be too expensive and only needs to run once when the repository rule runs.

kyliau:
Some packages, like [material](https://unpkg.com/@angular/material@8.1.1/package.json), only has esm5 and esm2015. 

If we include all `.umd.js` files in a package, we'll have to make sure there's no collision in module name.

gregmagolan:
> Some packages, like [material](https://unpkg.com/@angular/material@8.1.1/package.json), only has esm5 and esm2015.
> 
> If we include all `.umd.js` files in a package, we'll have to make sure there's no collision in module name.

Yes. We should probably check for collisions in the general non-APF case if we implement it.

Right now we add all umd files and also ngfactory & ngsummary files for APF:

```
/**
 * If the package is in the Angular package format returns list
 * of package files that end with `.umd.js`, `.ngfactory.js` and `.ngsummary.js`.
 */
function getNgApfScripts(pkg) {
  return isNgApfPackage(pkg) ?
      filterFiles(pkg._files, ['.umd.js', '.ngfactory.js', '.ngsummary.js']) :
      [];
}
```

gregmagolan:
Seems like we could narrow it down a bit with a few checks to detect APF with Ivy:

1) package.json has "ng-update"
2) there is a /bundles folder
3) include all .umd.js files /bundles (also check that they are named-UMD and make sure names don't collide)
4) also include all .ngfactory & .ngsummary since this are required for ViewEngine

Not doing (3) could lead to issues since rxjs, for example, has "ng-update" and a bundles folder with .umd.js files there _but_ on older versions of rxjs those bundles were non named.

alexeagle:
We decided to go back to the original approach of changing angular to put an apf marker in package json

kyliau:
Angular PR: https://github.com/angular/angular/pull/31900

alexeagle:
will need a different PR to detect the magic file

PR #942, opened 2019-07-28
Docs

alexeagle:
- copied some files from skydoc, these should maybe be in third_party but the docs/ directory is what gets served so they have to be there. maybe checking in symlink might work.
- dunno if the search feature works
- should have some instructions for developers to run the Jekyll site locally to have some idea how it renders

I'm pushing to my master branch so I hoped https://alexeagle.github.io/rules_nodejs/ would show a preview, I think it's just very delayed? maybe it will show by the time you look at it


Globegitter:
I think I am seeing the updated docs - but the styling looks a bit off:

image

alexeagle:
@Globegitter it was using absolute URL which work on localhost:4000 but the site is hosted under a subdirectory. Fixed by using relative URLs... PTAL

This might be a good chance for us to go over all the docs for formatting issues and clarity too.

alexeagle:
need to re-open this so it can be from my master branch, since that's the place where previews work

PR #946, opened 2019-07-30
fix: fencing for npm packages

gregmagolan:
and other minor cleanups


gregmagolan:
> hmm we didn't have any test coverage to catch this leak :(

I'll add some in a follow up PR

PR #952, opened 2019-07-31
Bazel integration testing

gregmagolan:
Adds a new `bazel_integration_test` rule which enables testing of nested workspaces from the root workspace as part of `bazel test ...`. The nested workspace under test is tested with the generated build_bazel_rules_nodejs release .tar.gz (`//:release`) and all generated npm packages (`//packages/foobar:npm_package`) that it depends on.

NB: buildkite docker images are missing shared libs that are required for ts_web_test_suite & protractor_test_suite chrome & firefox browsers:
```
    # BazelCI docker images are missing shares libs to run a subset browser tests:
    # mac: firefox does not work, chrome works
    # ubuntu: firefox and chrome do not work --- there are 0 tests to run
    # windows: firefox works, chrome does not work
```
Will follow up with BazelCI team on those. We are at least testing chrome on bazelci Mac, firefox on bazelci Windows & both on CircleCI linux.

NB: Uses https://github.com/gregmagolan/angular-bazel-example/commit/eed23d16d4c0b4f13c615303de90a4940f2a037a so allow `e2e_angular_bazel_example` test to run on bazelci mac (since firefox does not work there). nothing to test for `e2e_angular_bazel_example` on bazelci ubuntu there since all tests are browser tests. this is temporary until the NB2 is resolved with buildkite.

NB: Could not get e2e_angular_bazel_example to test on Windows due to what seems like memory issues. Will leave that for later.

FUTURE:
* leave notes of tuning CI memory usage for bazel-in-bazel tests in a README somewhere
* unit test for test_runner.js


.bazelci/presubmit.yml
```diff
@@ -21,6 +21,11 @@ tasks:
     - "--test_tag_filters=-fix-bazelci-ubuntu"
     test_targets:
     - "//..."
+  ubuntu1604_angular_bazel_example:
+    name: ubuntu1604_angular_bazel_example
+    platform: ubuntu1604
+    test_targets:
+    - "//:e2e_angular_bazel_example"
```
alexeagle: why is it separate? what happens when we move angular_bazel_example to the examples/ directory - then will it be part of the normal test targets?

WORKSPACE
```diff
@@ -304,5 +306,72 @@ rbe_autoconfig(
 
 load("@build_bazel_integration_testing//tools:repositories.bzl", "bazel_binaries")
```
alexeagle: do we still use this?

BUILD.bazel
```diff
@@ -81,3 +84,52 @@ pkg_tar(
     strip_prefix = "./rules_nodejs_package",
     tags = ["manual"],
 )
+
+BAZEL_VERSION = "0.28.1"
+
+BAZEL_BINARY = "@build_bazel_bazel_%s//:bazel_binary" % BAZEL_VERSION.replace(".", "_")
+
+NPM_PACKAGES = {
```
alexeagle: my docs change introduces a similar map. At some point we'll need a globally shared variable for this like in package.bzl or something

BUILD.bazel
```diff
@@ -81,3 +84,52 @@ pkg_tar(
...
+        "//:release": "build_bazel_rules_nodejs",
+    },
+    workspace_files = "@e2e_jasmine//:all_files",
+    workspace_under_test = "e2e_jasmine",
+)
+
+bazel_integration_test(
+    name = "e2e_angular_bazel_example",
+    size = "large",
+    timeout = "long",
```
alexeagle: I'm excited to have `flaky_test_attempts=2` around all these tests that rely on package installs

WORKSPACE
```diff
@@ -304,5 +306,72 @@ rbe_autoconfig(
...
+#
+# Setup bazel_integration_test repositories
+#
+
+local_repository(
+    name = "e2e_jasmine",
+    path = "e2e/jasmine",
+)
+
+git_repository_under_test(
```
alexeagle: curious: would this be any easier if we moved the code now? Looks like you had to do some mildly crazy stuff to get this working

internal/bazel_integration_test/BUILD.bazel
```diff
@@ -0,0 +1,36 @@
...
+load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("//:defs.bzl", "nodejs_binary")
+
+package(default_visibility = ["//visibility:public"])
+
+nodejs_binary(
+    name = "test_runner",
+    data = ["@npm//tmp"],
+    entry_point = ":test_runner.js",
+    install_source_map_support = False,
```
alexeagle: do users still have to do this? I thought we fixed it by vendoring the source-map-support

WORKSPACE
gregmagolan: yes. it would probably be easier in this case. although the mildly crazy stuff could still come in handle in the future for running integration tests against other repos.

internal/bazel_integration_test/BUILD.bazel
gregmagolan: good question. I don't remember the details but will TAL.

.bazelci/presubmit.yml
gregmagolan: running into memory issues running such a large test target with bazel in bazel so had to separate it out for now. may need to limit the memory & number of parallel actions for the nested bazel. still investigating.

WORKSPACE
gregmagolan: yes. we're using the `bazel_binaries()`:
```
BAZEL_VERSION = "0.28.1"
BAZEL_BINARY = "@build_bazel_bazel_%s//:bazel_binary" % BAZEL_VERSION.replace(".", "_")
```
but is only a small portion of `@build_bazel_integration_testing` which we could either vendor in or reproduce and improve on. It doesn't do much except download and extract a bazel release in a repository rule and then setup a filegroup with the extracted files.

The `test_runner.js` has to make assumptions about the name of the bazel executable:
```
const bazelBinary =
    require.resolve(`${config.bazelBinaryWorkspace}/bazel${isWindows ? '.exe' : ''}`);
```
which could be improved on.

WORKSPACE
gregmagolan: In the external repo case I think I could get it to work with less crazy stuff if the external repo was not in runfiles. In that case, test_runner would just copy all the files from the external repo but these would not be inputs to the test so if the external repo was updated the test would not automatically re-run. At least in the angular-bazel-example case this may be better as adding `workspace_files = "@e2e_angular_bazel_example//:bazel_integration_test_files",` meant having to add `io_bazel_rules_docker`, `io_bazel_rules_k8s` and mock `npm_angular_bazel` repositories to the root WORKSPACE even tho they are not used since `@e2e_angular_bazel_example//:BUILD.bazel` references those repos.

gregmagolan: It's actually really problematic to have internal and external details in BUILD files in external repositories. I think Bazel could improve this by either adding built-in build file fencing or maybe even better allow a parallel `BUILD.private` file which is meant for internal only build targets so that the repositories references and used for those targets don't leak into the public/external BUILD file targets.

WORKSPACE
```diff
@@ -304,5 +306,87 @@ rbe_autoconfig(
...
+)
+
+local_repository(
+    name = "e2e_webpack",
+    path = "e2e/webpack",
+)
+
+git_repository_under_test(
+    name = "e2e_angular_bazel_example",
+    branch = "add-buildkite-filter-flags",
```
alexeagle: hmm, if we test against floating version, our CI could be broken by a commit to the remote repo. We probably should have a hash and let renovate advance it?

internal/bazel_integration_test/bazel_integration_test.bzl
```diff
@@ -0,0 +1,122 @@
...
+    else:
+        return ctx.workspace_name + "/" + p
+
+def _bazel_integration_test(ctx):
+    # Serialize configuration file for test runner
+    config = ctx.actions.declare_file("%s.config.js" % ctx.attr.name)
+    ctx.actions.write(
+        output = config,
+        content = """// bazel_integration_test runner config generated by bazel_integration_test rule
+module.exports = {{
```
alexeagle: maybe this could be a lot more readable if you create a struct of the data, then just use struct.to_json() to create the content?

internal/bazel_integration_test/test_runner.js
```diff
@@ -0,0 +1,250 @@
+/**
```
alexeagle: seems like we should let Ittai and co. know about the additional features we implemented. Not sure if we would want to upstream this or not though

.bazelci/presubmit.yml
```diff
@@ -61,9 +81,54 @@ tasks:
...
+    - echo 'startup --host_jvm_args=-Xms256m --host_jvm_args=-Xmx1536m' >> .bazelrc
+    build_targets:
+    # Split up the build & test as test takes a long time and has very
+    # little output so running build first makes CI output friendlier.
+    # The build step can also use up more memory which may be required
+    # to build the release package with this configuration.
+    - "//e2e:e2e_angular_bazel_example"
+    test_flags:
+    - "--local_resources=792,1.0,1.0"
+    # test_args will be passed to the nested bazel process
```
alexeagle: if you have anything in your head that would help us later when tuning the resources under CI, would be nice to have that in a README somewhere

.bazelci/presubmit.yml
```diff
@@ -45,6 +55,16 @@ tasks:
...
+  # ubuntu1804_angular_bazel_example:
+    # name: ubuntu1804_angular_bazel_example
+    # platform: ubuntu1804
+    # bazel-in-bazel angular_bazel_example fails on ubuntu as the bazelci
+    # ubuntu docker image is missing shares libs required for the versions
+    # of chrome and firefox fetched by rules_webtesting. On circleci
+    # we have the same issue but we can work around it using apt-get.
+    # on bazelci apt-get fails with permission denied and there is no sudo
+    # command to switch to root.
+    # TODO(gregmagolan): figure out how to install the shared libs on bazelci
```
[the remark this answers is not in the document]
gregmagolan: @philwo ^

.bazelci/presubmit.yml
```diff
@@ -61,9 +81,54 @@ tasks:
     - "@bazel_workspace_b//:bin"
     - "@bazel_workspace_b//subdir:bin"
     build_targets:
-    - "//..."
+    - "//internal/..."
```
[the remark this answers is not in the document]
gregmagolan: good call. tags would be better.

.bazelignore
```diff
@@ -3,7 +3,15 @@ internal/npm_install/test/node_modules
...
+e2e/karma
+e2e/karma_stack_trace
+e2e/karma_typescript
+e2e/symlinked_node_modules_npm
+e2e/symlinked_node_modules_yarn
+e2e/ts_auto_deps
+e2e/ts_devserver
+e2e/ts_library
+e2e/tsconfig_extends
 # Examples are tested by running examples/test_example.sh outside of Bazel
```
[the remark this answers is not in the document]
gregmagolan: For now... I'll update these comments once everything is tested with bazel_integration_test

e2e/jasmine/package.json
```diff
@@ -1,15 +1,12 @@
...
   "dependencies": {
-    "@bazel/jasmine": "file:../../release/npm_bazel_jasmine",
+    "@bazel/jasmine": "latest",
     "zone.js": "0.8.29"
   },
   "//": "Include an incompatible jasmine as a devDependency to verify that jasmine_node_test works regardless",
   "devDependencies": {
     "jasmine": "2.99.0",
     "jasmine-core": "2.5.2"
-  },
```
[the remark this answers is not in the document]
gregmagolan: sure. I'll add the `test` script back in so `yarn test` or `npm run test` works.

WORKSPACE
gregmagolan: that's what I was thinking would be best. tho renovate doesn't know what git_repository_under_test. we could alias git_repository_under_test as git_repository but there are other git_repositories in this WORKSPACE which would have a little more work to do after fetch... tho its not a terrible amount. In short, git_repository_under_test adds the `//:bazel_integration_test_files` targets to the repository which is a simple operation.

This is currently depending on my fork https://github.com/gregmagolan/angular-bazel-example/commit/eed23d16d4c0b4f13c615303de90a4940f2a037a so that I could test on buildkite OSX so we'd need to either push that commit upstream or fix the underlying buildkite docker image and get the shared libs that firefox needs on OSX and that firefox and chrome need on Ubuntu.

internal/bazel_integration_test/test_runner.js
gregmagolan: I think this is rather different that what they are doing and its also tuned to do package.json replacements which is only relevant for rules_nodejs. But when this lands we should give him a brief maybe there are parts that would useful for him.

e2e/jasmine/package.json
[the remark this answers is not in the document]
gregmagolan: On that note, the release script should also update the `build_bazel_rules_nodejs` http archive in the e2e tests & examples as it is no longer a local_repository but it points to the last release. This is probably better for publishing examples to their own repos. Not sure about what is best for e2e tests. `bazel test ...` from the root workspace will now test everything so what would you expect if you `cd e2e/jasmine && bazel test ...`?

npm packages @bazel/foobar are now `"latest"` since we don't want the nested repos to depend on build artifacts from the main repository (unless done automatically in bazel_integration_test). It would be inconsistent to run against `local_repository(path = "../..")` if the npm based repositories are also not based on your local mods.

.bazelci/presubmit.yml
gregmagolan: It's a long painful process tuning so that it passes consistently and still has enough cores to run quickly. There are differences on CircleCI and buildkite ubuntu, mac & Windows as well. I'll summarize what I found in a README.

@philwo Is there documentation somewhere on the memory & cores available on the different bazelci platforms? I didn't see it in the docs & did not see it anywhere obvious in the buildlkite UI.

internal/bazel_integration_test/test_runner.js
achew22: @gregmagolan, this file (and associated `.bzl` files) are fucking gold. Great work! As the maintainer of the golang verison of this, I'm jealous I do this. I would be interested in helping you upstreaming this if you're open to the idea. Especially if you're interested in trying to do as much as possible in skylark so that it's reusable between all the existing implementations.

This seems like something that would be generally helpful to anyone writing any repo rules, and even more so because it is well abstracted (it's own test rule) and doesn't require you to do anything more than make a workspace file.

internal/bazel_integration_test/bazel_integration_test.bzl
```diff
@@ -0,0 +1,122 @@
...
+    "workspace_files": attr.label(),
+    "workspace_under_test": attr.string(),
+    "_test_runner": attr.label(
+        executable = True,
+        cfg = "host",
+        default = Label("@build_bazel_rules_nodejs//internal/bazel_integration_test:test_runner"),
+    ),
+}
+
+bazel_integration_test = rule(
```
achew22: You might consider wrapping this in [a macro](https://github.com/bazelbuild/bazel-integration-testing/blob/master/go/bazel_integration_test.bzl#L44) and creating 1 test for each of the various versions of Bazel that are officially supported by rules_nodejs. It's good to get coverage for older Bazels, but it does come at the cost of a slower build. 

If you're not interested in testing anything other than the latest 3 versions

```
load("@com_github_bazelbuild_bazel_integration_testing//tools:common.bzl", "GET_LATEST_BAZEL_VERSIONS")
```

`GET_LATEST_BAZEL_VERSIONS` also takes an optional param, that defaults to 3, that is the number of minor versions to test and reduces the list to the latest patch version of each. So, for example, if 0.30.3, 0.30.2, 0.30.1, 0.30.0, 0.29.0, 0.28.1, and 0.28.0 are around, it'll give you 0.30.3, 0.29.0, and 0.28.1.

.bazelci/presubmit.yml
```diff
@@ -18,9 +18,39 @@ tasks:
...
+    - echo 'startup --host_jvm_args=-Xms256m --host_jvm_args=-Xmx1536m' >> .bazelrc
+    build_targets:
+    - "//e2e/..."
+    test_flags:
+    - "--local_resources=792,1.0,1.0"
+    # test_args will be passed to the nested bazel process
+    - "--test_arg=--local_resources=13288,7.0,1.0"
+    - "--test_arg=--test_tag_filters=-no-bazelci,-no-bazelci-ubuntu"
+    test_targets:
+    - "//e2e/..."
```
alexeagle: is it meant to be disjoint with the one above? I was expecting a `test_tag_filters=e2e` to be mutually exclusive with `test_tag_filters=-e2e`

WORKSPACE
```diff
@@ -321,5 +323,84 @@ rbe_autoconfig(
...
+bazel_binaries(versions = ["0.28.1"])
+
+#
+# Support creating Docker images for our node apps #
+# Not used but needed for @e2e_angular_bazel_example//:bazel_integration_test_files target
+#
+
+http_archive(
+    name = "io_bazel_rules_docker",
+    sha256 = "aed1c249d4ec8f703edddf35cbe9dfaca0b5f5ea6e4cd9e83e99f3b0d1136c3d",
```
alexeagle: ugh, it's going to be hard to avoid version skew between this repo and angular-bazel-example if these deps aren't shared via a `package.bzl#devDeps` or something.
Will be fixed by moving angular-bazel-example here, assuming we actually follow through on that

WORKSPACE
```diff
@@ -321,5 +323,84 @@ rbe_autoconfig(
...
+        "angular-bazel-example",
+    ]),
+    kind = "deployment",
+)
+
+#
+# Setup bazel_integration_test repositories
+#
+
+[local_repository(
```
alexeagle: we could do this comprehension for the `packages/foo` workspaces too

e2e/BUILD.bazel
```diff
@@ -0,0 +1,143 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("@build_bazel_rules_nodejs//internal/bazel_integration_test:bazel_integration_test.bzl", "bazel_integration_test")
+
+BAZEL_VERSION = "0.28.1"
+
+BAZEL_BINARY = "@build_bazel_bazel_%s//:bazel_binary" % BAZEL_VERSION.replace(".", "_")
+
+NPM_PACKAGES = {
```
alexeagle: this seems like non-local info, ought to come from packages/index.bzl or something

e2e/karma/dummy_test.sh
```diff
@@ -0,0 +1,2 @@
+echo "Just a dummy test so that we have a test target for //... on bazelci ubuntu with bazel_integration_test"
```
alexeagle: I don't get this - is there nothing to `bazel test` in this repo? nothing we need to assert?

internal/bazel_integration_test/bazel_integration_test.bzl
alexeagle: we thought about that, there are currently no old bazels we support but when the first LTS one comes out we should test that we support it following the schedule Bazel has for support themselves

internal/bazel_integration_test/test_runner.js
```diff
@@ -0,0 +1,250 @@
...
+  if (packageJsonReplacementKeys.length) {
+    for (const packageJsonKey of packageJsonReplacementKeys) {
+      const regex = new RegExp(`\"${packageJsonKey}\"\\s*\:\\s*\"[^"]+`)
+      const replacement = `"${packageJsonKey}": "${config.packageJsonRepacements[packageJsonKey]}`;
+      packageJsonContents = packageJsonContents.replace(regex, replacement);
+    }
+  }
+
+  if (DEBUG)
+    console.log(`\n\npackage.json file with replacements:\n========================================================================================\n${
```
alexeagle: if you really want the log formatted with these section breaks, maybe a `debug()` helper function would be better

e2e/karma/dummy_test.sh
gregmagolan: Nothing when you filter out the chrome & firefox tests and `bazel test ...` errors out if it doesn't find any matching tests

WORKSPACE
gregmagolan: 👍

.bazelci/presubmit.yml
gregmagolan: 👍

e2e/BUILD.bazel
gregmagolan: agreed. moved to packages/index.bzl and also used in WORKSPACE to compress the npm_package_foobar defs

internal/bazel_integration_test/test_runner.js
gregmagolan: 👍

internal/bazel_integration_test/BUILD.bazel
gregmagolan: you are correct. it is not needed anymore.

PR #955, opened 2019-08-01
test: move /e2e/ts_library ts_library tests to /packages/typescript/test

gregmagolan:
No functional changes to tests. Just moving them to the root workspace so they are tested with `bazel test ...`


packages/protractor/test/protractor-2/BUILD.bazel
```diff
@@ -41,7 +34,7 @@ ts_library(
     name = "ts_spec",
     testonly = True,
     srcs = ["test.spec.ts"],
-    tsconfig = ":tsconfig-test",
+    tsconfig = "//packages/typescript/test:tsconfig-test",
```
alexeagle: this seems undesirable dependency between packages, why not have a copy of this tsconfig here if it's needed in protractor?

packages/typescript/test/foobar/BUILD.bazel
```diff
@@ -1,6 +1,6 @@
-load("@npm_bazel_typescript//:index.bzl", "ts_library")
+load("@npm_bazel_typescript//:index.from_src.bzl", "ts_library")
 
-package(default_visibility = ["//visibility:public"])
+package(default_visibility = [])
```
alexeagle: isn't this the same as no `package` declaration?

WORKSPACE
```diff
@@ -110,6 +110,16 @@ node_repositories(
 
 yarn_install(
     name = "npm",
+    manual_build_file_contents = """
+filegroup(
```
alexeagle: this is really surprising; add a comment saying why this is needed?

packages/typescript/test/foobar/BUILD.bazel
gregmagolan: 👍 not sure how that go in there. thanks

packages/protractor/test/protractor-2/BUILD.bazel
gregmagolan: 👍

WORKSPACE
gregmagolan: 👍

PR #957, opened 2019-08-01
revamp docsite

alexeagle:
Moved readme bits out of README.md to the docsite, readme is now just a short intro, quickstart, and testimonials

- we now have all rules documented in one place. Will be https://bazelbuild.github.io/rules_nodejs/ - during this PR review you can preview at http://alexeagle.github.io/rules_nodejs
- the individual markdown files are also quite useful, eg. https://github.com/alexeagle/rules_nodejs/blob/master/docs/TypeScript.md
- these docs will get versioned with the repo so it's easy to get https://github.com/alexeagle/rules_nodejs/blob/1.0.0/docs/TypeScript.md for v1.0.0 for example
- Migrated away from skydoc altogether, cuts our WORKSPACE dep on bazel sources
- We need our own templates to format the docs to avoid dumb <table> issues. It's not that much code.
- release process and instructions are unchanged

will cleanup commit description and squash before merging


package.json
```diff
@@ -86,7 +86,7 @@
         "bazel:lint-fix": "yarn bazel:format --lint=fix",
         "format": "git-clang-format",
         "format-all": "clang-format --glob='{internal/**/,examples/**/}*.{js,ts}' -i",
-        "skydoc": "bazel build //docs && unzip -o -d docs dist/bin/docs/docs-skydoc.zip",
+        "skydoc": "bazel build //docs && cp -f dist/bin/docs/*.md docs",
```
gregmagolan: `rm -rf docs` before `cp`?

alexeagle: not that simple, some of the markdown files in the docs/ folder are generated but some are not.

alexeagle:
merged in 778c32c

PR #962, opened 2019-08-04
Jasmine custom config

alexeagle:
Finally chose an approach to fix this.

- Accept a config file like in #744 
- now the user can have their spec files named anything, doesn't need to follow our conventions

### DIFF
--- a/packages/jasmine/src/jasmine_node_test.bzl
+++ b/packages/jasmine/src/jasmine_node_test.bzl
@@ -28,6 +28,7 @@ def jasmine_node_test(
         deps = [],
         expected_exit_code = 0,
         tags = [],
+        config_file = None,
         coverage = False,
         jasmine = "@npm//@bazel/jasmine",
         jasmine_entry_point = "@npm//:node_modules/@bazel/jasmine/jasmine_runner.js",
@@ -43,6 +44,15 @@ def jasmine_node_test(
       deps: Other targets which produce JavaScript, such as ts_library
       expected_exit_code: The expected exit code for the test.
       tags: Bazel tags applied to test
+      config_file: (experimental) label of a file containing Jasmine JSON config.
+
+        Note that not all configuration options are honored, and
+        we expect some strange feature interations.
+        For example, if you list spec_files, they will be tested
+        but not instrumented for code coverage.
+
+        See https://jasmine.github.io/setup/nodejs.html#configuration
+
       coverage: Enables code coverage collection and reporting.
       jasmine: A label providing the `@bazel/jasmine` npm dependency.
       jasmine_entry_point: A label providing the `@bazel/jasmine` entry point.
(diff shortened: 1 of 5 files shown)


alexeagle:
@benlesh I hope you enjoy using this! :)

alexeagle:
ugh, really on Windows the `spec_files: "**/test_config_*.js"` doesn't work??

packages/jasmine/src/jasmine_node_test.bzl
```diff
@@ -61,12 +67,21 @@ def jasmine_node_test(
...
+        templated_args.append("--nocoverage")
+
+    if config_file:
+        # Calculate a label relative to the user's BUILD file
+        pkg = Label("%s//%s:__pkg__" % (native.repository_name(), native.package_name()))
+        all_data.append(pkg.relative(config_file))
+        templated_args.append("$(location %s)" % config_file)
     else:
-        templated_args = templated_args + ["--nocoverage"]
+        templated_args.append("--noconfig")
```
gregmagolan: Doesn't look like config_file makes it into runfiles unless I'm missing something?

packages/jasmine/src/jasmine_runner.js
```diff
@@ -51,24 +51,35 @@ Error.stackTraceLimit = Infinity;
...
+  const configFile = readArg();
 
   // the relative directory the coverage reporter uses to find anf filter the files
-  const cwd = process.cwd()
+  const cwd = process.cwd();
 
   const jrunner = new JasmineRunner({jasmineCore: jasmineCore});
+  if (configFile !== '--noconfig') {
+    jrunner.loadConfigFile(require.resolve(configFile));
+  }
```
gregmagolan: sourceFiles will be incorrect for coverage when enabled as it only considers `IS_TEST_FILE`. Maybe leave a TODO for that although I don't see an easy way to fix it?

I've been thinking for a while about adding a `specs` and `non_specs` attributes to jasmine_node_test (as well as protractor & karma since they have the same spec filtering issue (https://github.com/bazelbuild/rules_nodejs/issues/513#issuecomment-478135251). This would keep compat and mean that `srcs` is still just strictly source files (as per bazel idiom) and `srcs` and `deps` would still get the automatic filter applies. For users that want fine grained control over the sources and specs then could switch to `specs` and `non_specs` (not sure about naming) which would accept source files and handle the javascript provider.

The above would also mean changing jasmine_node_test to a proper rule and removing the macro which is just a wrapper around nodejs_test.

packages/jasmine/src/jasmine_node_test.bzl
alexeagle: it goes into `all_data` which puts it in the `data` attr of the nodejs_test - so yeah it should be

packages/jasmine/src/jasmine_runner.js
alexeagle: I added more notes to the docs for the new attribute and marked it experimental.

Yeah the fact that it's just a macro made this change harder too, but it didn't feel like the right time to dive into refactoring...

PR #965, opened 2019-08-05
Frozen-lockfile flag for yarn_install

bjovanovic84:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

yarn_install rule modifies yarn.lock file, not suitable for use in CI.

Issue Number: 941

## What is the new behavior?
Added a new attribute `frozen_lockfile` in yarn_install, so it can be set to `True` to prevent changes to yarn.lock file.

- [ ] Yes
- [x] No


bjovanovic84:
@googlebot I signed it!

alexeagle:
hmm, how would this work? Would you set it to `True` in the WORKSPACE file, and therefore apply both to CI and local dev?
If so, then changes to package.json would require running `yarn` outside of bazel to first update the lockfile, since we would fail if Bazel runs yarn install and there are updates needed.
If that's desirable, then I think we should default this flag to True rather than expect developers to discover it?

Or were you thinking that it would be `True` on CI but `False` locally, how would you accomplish that?

bjovanovic84:
> hmm, how would this work? Would you set it to `True` in the WORKSPACE file, and therefore apply both to CI and local dev?
> If so, then changes to package.json would require running `yarn` outside of bazel to first update the lockfile, since we would fail if Bazel runs yarn install and there are updates needed.
> If that's desirable, then I think we should default this flag to True rather than expect developers to discover it?
> 
> Or were you thinking that it would be `True` on CI but `False` locally, how would you accomplish that?

We set it to true in the workspace, as we want reproducible dependencies. We do run yarn outside of bazel (yarn add) and that's a deliberate action, otherwise the lock file would be changing frequently without any code changes.
I set the default to False in order not to change the current behavior. My opinion is that it should be True by default, as having False kind of breaks hermeticity. Nobody's really updating package.json manually, so having bazel use the flag by default would be my expectation (other people from my firm agree with this as well).

Toxicable:
> Nobody's really updating package.json manually, so having bazel use the flag by default would be my expectation (other people from my firm agree with this as well).

We manually update our package.json.  

While I support this feature in CI, it needs to be implemented in a way that local dev dosen't suffer.

PR #997, opened 2019-08-10
Add browser module main priority to generated umd bundles

gregmagolan:
* Use babelfiy transform esm->cjs in npm_umd_bundle

`es6ify` was not trivial to vendor in as it uses https://github.com/google/traceur-compiler internally which does compile down into a vendorable format with ncc. `babelify` on the other hand uses `@babel/core` which does compile down into a vendorable format with ncc.

Replaces https://github.com/bazelbuild/rules_nodejs/pull/761

//cc @Toxicable @alexeagle


gregmagolan:
Some npm packages such at typeorm are going to require custom work for the user as optional dynamic requires such as
```
        try {
            this.sqlite = require("react-native-sqlite-storage");
        }
        catch (e) {
            throw new DriverPackageNotInstalledError("React-Native", "react-native-sqlite-storage");
        }
```
need to ignored with `-u react-native-sqlite-storage` passed to browserify if they are not used

Toxicable:
@gregmagolan Dam, that's unfortunate about the dynamic requires.
Im not certain of the underlying process but I know with webpack if it tried to bundle a dynamic require, it simply ignores it and throws a warning.

But otherwise, nice work you've done here!
Nice work in vendoring babel, im sure it'll come in hand for other rules aswell

internal/npm_install/npm_umd_bundle.bzl
```diff
@@ -65,6 +66,9 @@ NPM_UMD_ATTRS = {
         mandatory = True,
         allow_single_file = True,
     ),
+    "excluded": attr.string_list(
+        doc = """List of excluded packages that should not be bundled""",
```
alexeagle: needs an example or point to some more doc about why I might need to exclude a package.
Also probably good to point to the CLI option `-u` where we end up plugging this in, since that's authoritative on what are the semantics

internal/npm_install/test/BUILD.bazel
```diff
@@ -54,14 +54,32 @@ npm_umd_bundle(
...
+    package = "@npm//typeorm",
+)
+
 jasmine_node_test(
     name = "browserify_test",
     srcs = ["browserify.spec.js"],
     data = [
         ":core-util-is.umd.js",
         ":minimist.umd.js",
+        ":rxjs.umd.js",
```
alexeagle: this is getting gnarly dependency graph because a change to any bundle re-runs all the tests. doesn't really matter for us I think other than setting a bad example for others :)

package.json
```diff
@@ -15,13 +15,12 @@
         "@bazel/ibazel": "^0.10.3",
         "@commitlint/cli": "^8.0.0",
         "@commitlint/config-conventional": "^8.0.0",
+        "date-fns": "1.30.1",
         "@types/hammerjs": "2.0.35",
         "@types/jasmine": "~3.3.13",
         "@types/node": "7.0.18",
         "@types/webpack": "~4.4.35",
-        "@zeit/ncc": "^0.20.4",
```
alexeagle: where did this go??

scripts/vendor_browserify.sh
```diff
@@ -1,22 +1,427 @@
...
+
+    echo "Vendoring ${PKG} to ${DST_DIR}"
+    rm -rf ${DST_DIR}
+    mkdir -p ${DST_DIR}
+
+    echo "- compiling ${PKG} with ncc"
+    capture_and_run ${THIRD_PARTY_DIR}/node_modules/.bin/ncc build ${SRC_DIR}/bin/cmd.js -o ${DST_DIR}
+
+    echo "- local mod: revert https://github.com/browserify/browserify/pull/1801"
+    capture_and_run sedi 's#parent.id !== self._mdeps.top.id#parent.id#' ${DST_DIR}/index.js
```
alexeagle: in theory this won't work anymore - a PR is against the sources but now you mirror the NPM package, so if the package was authored in typescript we'd be applying a patch against the output JS instead.
other than that I don't have a good reason why we shouldn't mirror from npm rather than github...

scripts/vendor_browserify.sh
```diff
@@ -1,22 +1,427 @@
...
+""" > ${DST_DIR}/BUILD.bazel
+  )
+
+  ##############################################################################
+  # Vendor ieee754
+  # (required by browserify)
+  ##############################################################################
+  (
+    CMDS=""
+    readonly PKG="ieee754"
```
alexeagle: it looks like this is intended to loop through the packages, but then the logic is all copy-pasta anyway - do you mean to refactor this script before merging?

internal/npm_install/generate_build_file.js
```diff
@@ -581,6 +581,96 @@ function cleanupEntryPointPath(p) {
...
+ * This uses seveal mainFileNames in priority to find a correct usable file
+ * @param {any} pkg 
+ * @returns {string | undefined}
+ */
+function resolvePkgMainFile(pkg) {
+  // es2015 is another option for mainFile here
+  // but its very uncommon and im not sure what priority it takes
+  //
+  // this list is ordered, we try resolve `browser` first, then `module` and finally fall back to `main`
+  const mainFileNames = ['browser', 'module', 'main']
```
alexeagle: I guess we could add an attribute to let `npm_umd_module` user configure explicitly what key to use? not needed yet it seems

internal/npm_install/test/BUILD.bazel
gregmagolan: its a small test :)

package.json
gregmagolan: I put all of the vendoring-only deps in /third_party/package.json. Keeps the main deps smaller and also ensures that we're not accidentally using pacakages from @npm that should be vendored such as `@npm//babel/core` (ran into that issue when working on this)

internal/npm_install/generate_build_file.js
gregmagolan: 👍 future PR if needed

scripts/vendor_browserify.sh
gregmagolan: I'll refactor and cleanup a bit

scripts/vendor_browserify.sh
gregmagolan: still works. I double checked. sources are `.js`. PR got merged but is not part of the latest release yet (`16.5.0`). released code is still
```
                if (opts.preserveSymlinks && parent.id !== self._mdeps.top.id) {
                    return cb(err, path.resolve(file), pkg, file)
                }
```

gregmagolan:
> how much bigger is the .tgz distro?

about double: 592KB -> 1.35MB. I can reduce the size a bit before merging by minifying babel/core & not including the unminified babel/core & babelify. `@babel/core` after ncc and unminified is 2.45MB so that is where most of the size increase is coming from.

gregmagolan:
> > how much bigger is the .tgz distro?
> 
> about double: 592KB -> 1.35MB. I can reduce the size a bit before merging by minifying babel/core & not including the unminified babel/core & babelify. `@babel/core` after ncc and unminified is 2.45MB so that is where most of the size increase is coming from.

with minified babelify & babel/core in the release package only it is down to
```
-r-xr-xr-x 1 circleci circleci 1062311 Aug 12 20:58 dist/bin/release.tar.gz
```
added this output to circleci test job output

scripts/vendor_browserify.sh
gregmagolan: done

internal/npm_install/npm_umd_bundle.bzl
gregmagolan: good point. added more details here

gregmagolan:
> I think this makes sense to me as a step towards behaving like other bundlers that understand all the fields and do some transforms on the module syntax

Maybe we use webpack instead here in the future as it can handle the dynamic requires which browserify can't?

.circleci/config.yml
```diff
@@ -154,6 +154,9 @@ jobs:
 
       - run: bazel test ... --test_tag_filters=-e2e,-examples
 
+      # check size of release.tar.gz
+      - run: ls -l dist/bin/release.tar.gz
```
alexeagle: we could save it as a circleci artifact instead - then we could hook up buildsize.org if we care to see the size changes reported in PRs

Issue #227, opened 2018-06-12
Unable to use http cluster with bazel


kalbasit:
We currently have an HTTP server build using nodejs cluster. When we call [`fork`](https://nodejs.org/docs/latest-v8.x/api/cluster.html#cluster_cluster_fork_env) it blows up because it's looking for the binary at the wrong place.

```sh-session
$ git clone https://github.com/kalbasit/bazel-reproducible.git
$ cd bazel-reproducible/http-cluster-does-not-work
$ bazel run //server:http
WARNING: ignoring _JAVA_OPTIONS in environment.
INFO: Analysed target //server:http (0 packages loaded).
INFO: Found 1 target...
Target //server:http up-to-date:
  bazel-bin/server/http_bin.sh
  bazel-bin/server/http
INFO: Elapsed time: 0.728s, Critical Path: 0.04s
INFO: 0 processes.
INFO: Build completed successfully, 1 total action

INFO: Running command line: bazel-bin/server/http
WARNING: source-map-support module not installed.
   Stack traces from languages like TypeScript will point to generated .js files.

[server-1528846565034] Starting master
[server-1528846565034] listening on :8080
[server-1528846565034] Cluster options: {
    "count": 8,
    "name": "server-1528846565034"
}
[server-1528846565034] starting worker thread
[server-1528846565034] starting worker thread
[server-1528846565034] starting worker thread
[server-1528846565034] starting worker thread
[server-1528846565034] starting worker thread
[server-1528846565034] starting worker thread
[server-1528846565034] starting worker thread
module.js:549
    throw err;
    ^

Error: Cannot find module '/home/kalbasit/.cache/bazel/_bazel_kalbasit/727d60ccb89e565dc5ed6e14ea3a46c6/execroot/com_github_kalbasit_bazel/bazel-out/k8-fastbuild/bin/server/http.runfiles/com_github_kalbasit_bazel/com_github_kalbasit_bazel/server/http.js'
    at Function.Module._resolveFilename (module.js:547:15)
    at Function.Module._load (module.js:474:25)
    at Function.Module.runMain (module.js:693:10)
    at startup (bootstrap_node.js:188:16)
    at bootstrap_node.js:609:3
[server-1528846565034] starting worker thread
[server-1528846565034] Worker 32525 died
[server-1528846565034] starting worker thread
module.js:549
    throw err;
    ^
```

It's trying to find the module at 

[...]

PR #1003, opened 2019-08-13
feat(builtin): split node_loader into node_patcher

Toxicable:
## PR Checklist

## PR Type

Unable to use bazels module resolution under a `child_process` or `worker_threads`

Issue Number: #312 #227

## What is the new behavior?
This change splits up node_launcher into `node_loader.js` + `node_patcher.js`
`node_loader.js` - job is to run any bootstrap modules, setup source map support and finally call the users main entry point.
`node_patcher` - job is to only patch the module resolution logic of `require` so it can resolve modules under bazels runfile layout.
Part of this change also adds logic so that any worker (child process or thread) also has their require patched correctly so that they can require modules just the same as in the main thread.
Meaning we'll modify `node_launcher.sh` to call `node` with `--require=${target}node_patcher.js`.
In my testing, this change alone will enable works for `child_process` on nodejs 10.x and greater, however, we will need nodejs 12.x to enable this fix for `worker_threads`.
With this in mind, we'll make a test for either case but allow the `worker_threads` one to fail untill 12.x is LTS and we can enable it as the default.

- [x] Yes
- [ ] No
Even those i'd consider it an "internal" api, if anyone is relying on the current behaviour of node_loader then this change will break it.


internal/node/node_loader.js
```diff
@@ -71,408 +74,11 @@ node_loader: running ${TARGET} with
...
-      files.forEach(f => {
-        runfilesManifest[path.posix.join(test, f)] = path.posix.join(tree, f);
-      })
-      return runfilesManifest[res] || runfilesManifest[res + '.js'];
-    }
-    segments.pop();
-  }
-}
+// we patch require here
+module.constructor._load(patchModulePath)
```
soldair: why this not just the provided `require`? 
it's not patched yet and if it was you would probably want to honor changes other loaders may have made.
I can think of a few cases where you might need to work around other patches to the module system but i'd like to know which =)

internal/node/node_loader.js
```diff
@@ -71,408 +74,11 @@ node_loader: running ${TARGET} with
...
+// then we also patch process.execArgs sothat any child process also does this patch
+process.execArgv = process.execArgv.concat([`--require=${patchModulePath}`]);
```
soldair: Will nodejs will already have a `--require=...loader.js` arg in `execArgv`? 
So when a child process or worker executes it again i would expect this logic to run node with `node --require=loader.js --require=patcher.js`  and add another `--require=patcher.js
`
the extra --require=...patcher.js are a no op because loader requires it but the failure mode here is using up the processes argv space unexpectedly.

internal/node/node_patcher.js
```diff
@@ -0,0 +1,477 @@
...
+ */
+'use strict';
+var path = require('path');
+var fs = require('fs');
+
+console.log('PATCHER');
+
+const isWindows = /^win/i.test(process.platform);
+// Ensure that node is added to the path for any subprocess calls
+process.env.PATH = [path.dirname(process.execPath), process.env.PATH].join(isWindows ? ';' : ':');
```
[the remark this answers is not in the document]
Toxicable: Ah good call

internal/node/node_loader.js
Toxicable: Yeah, I tried this with `require` and and for some reason it just wouldn't work, no error or anything. But that shouldn't matter too much since this isn't what we want to go with anyway, was just testing some ideas

internal/node/node_loader.js
Toxicable: Yeah, this isn't right at all, I was just testing what would happen if you modify the execArgv from inside the program - turns out it's ignored by node in the child process (maybe it should be marked as `readonly` then?)
Either way, we'll move this into the `node_launcher.sh` script

Toxicable:
@soldair Thanks for the review, i've updated the description with a better explaination of the goals here. Which should answer some of the questions you had around here, will try get some time tonight to make some changes so it reflects the description :P

Toxicable:
The only failure now is because of this logic
```
if [[ $patcher = /* ]]
then
  NODE_OPTIONS+=( "--require=$patcher" )
else
  NODE_OPTIONS+=( "--require=./$patcher" )
fi
```
Is there a proper way to check for absolute/relative or to normalize this path in node_laucher?

internal/node/node_patcher.js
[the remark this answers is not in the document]
Toxicable: This was a copy paste error, it shouldn't be in the patcher, since it gets done in the loader

internal/node/test/BUILD.bazel
```diff
@@ -68,6 +68,12 @@ nodejs_binary(
     entry_point = ":module-name.js",
 )
 
+nodejs_binary(
+    name = "workers",
+    data = [":worker-main.js", "worker-thread.js"],
+    entry_point = ":worker-main.js",
+)
+
```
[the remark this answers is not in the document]
Toxicable: I've extended that test a bit so it does a `require('lib')` where the lib is a bazel defined module, if the patcher hasn't been run then doing this will throw

internal/node/node_loader.js
Toxicable: replaced with the `--require` implementation

internal/node/node_loader.js
Toxicable: migrated to the `--require` from launcher implementation, so we don't need this anymore

Toxicable:
@soldair Those look like some pretty easy changes to make. Although I don't think ill get any time on this week till the weekend to make those changes. If you want to pick it up however im always on slack.

Toxicable:
I've addressed those points but it'll still fail some tests since I don't quite know how to resolve the issue I mentioned above about the patcher path

alexeagle:
Note, with the upcoming linker approach, the node_modules directory on disk when the action starts should already be in the state all node programs expect, so we may not need this.

(remains to be seen whether all features of node_loader.js can be replaced by the linker, eg. runfiles, but I think it's eminently possible)

Toxicable:
Sounds good to me. If we're able to  remove th need for node_patcher.js then im happy

PR #1004, opened 2019-08-13
docs: add instructions for patching the tgz release

alexeagle:
And other minor tweaks

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -96,33 +96,101 @@ Bazel is generally only a build tool, and is unaware of your version control sys
 However, when publishing releases, you typically want to embed version information in the resulting distribution.
 Bazel supports this natively, using the following approach:
 
-1) Your `tools/bazel.rc` should pass the `workspace_status_command` argument to `bazel build`.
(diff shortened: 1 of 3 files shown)


docs/index.md
```diff
@@ -96,34 +96,102 @@ Bazel is generally only a build tool, and is unaware of your version control sys
...
+# Build them in one command to maximize parallelism
+$BAZEL build --config=release $NPM_PACKAGE_LABELS
+# publish one package at a time to make it easier to spot any errors or warnings
+for pkg in $NPM_PACKAGE_LABELS ; do
+  $BAZEL run -- ${pkg}.${NPM_COMMAND} --access public --tag latest
+done
+```
+
+> WARNING: Bazel can't track changes to git tags. That means it won't rebuild a target if only the result of the workspace_status_command has changed. So changes to the version information may not be reflected if you re-build the package or bundle, and nothing in the package or bundle has changed.
+,cu
```
gregmagolan: ‘,cu’ typo

PR #1005, opened 2019-08-13
refactor: reduce some duplication in npm_package rules

alexeagle:
Would like to reduce the boilerplate as we make more and more packages

### DIFF
--- a/tools/defaults.bzl
+++ b/tools/defaults.bzl
@@ -1,0 +1,44 @@
+"""Wrappers around build rules
+
+These set common default attributes and behaviors for our local repo
+"""
+
+load(
+    "@build_bazel_rules_nodejs//:defs.bzl",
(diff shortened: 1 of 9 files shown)


tools/defaults.bzl
```diff
@@ -0,0 +1,44 @@
...
+        srcs = ["@build_bazel_rules_nodejs//:LICENSE"],
+        outs = ["LICENSE"],
+        cmd = "cp $< $@",
+    )
+
+    deps = kwargs.pop("deps", [])
+    deps.append(":copy_LICENSE")
+
+    # Make every package visible to tests
+    visibility = kwargs.pop("visibility", [
```
gregmagolan: that's a nice pattern vs adding it to params

alexeagle: yeah I learned that from @Ubehebe in https://github.com/bazelbuild/rules_nodejs/pull/398

PR #1011, opened 2019-08-14
docs(typescript): document how to increase heap

alexeagle:
Motivated by https://github.com/bazelbuild/rules_typescript/issues/314#issuecomment-520619564

/cc @JaredNeil

### DIFF
--- a/packages/typescript/docs/install.md
+++ b/packages/typescript/docs/install.md
@@ -102,6 +102,35 @@ nodejs_binary(
 See https://github.com/bazelbuild/rules_nodejs#dependencies for more information on
 managing npm dependencies with Bazel.
 
+## Customizing the TypeScript compiler binary
+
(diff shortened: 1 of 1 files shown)


Toxicable:
Might be worth noting here if you're hitting the default heap size limit then you should probably look at splitting up that compilation unit into ismaller ones before doing this?

JaredNeil:
Yeah, we tried that with some of our larger `ng_module` targets and it increased the build time by ~3x, so we had to go back to massive targets until we can figure out why small targets were so slow.

JaredNeil:
Oh, defining a custom compiler is also needed if you want to use tsickle since the default compiler doesn't depend on "@npm//tsickle", so it [throws an error](https://github.com/bazelbuild/rules_typescript/blob/master/internal/tsc_wrapped/tsc_wrapped.ts#L435).

Issue #1015, opened 2019-08-15
node_loader.js error message doesn't match Node's own error message


bweston92:
# 🐞 bug report

### Description

The loader script `rules_nodejs` uses a custom error message for failed package resolution. Some users/packages match against the error message that Node's official loader script provides.

## 🔬 Minimal Reproduction

https://github.com/bweston92/debug-gatsby-bazel-rules-nodejs

**Rules version (SHA):**

<pre>
  <code>
f98fa199c034850eb5be750011f306deacf645ea
  </code>
</pre>

**Anything else relevant?**

Discussion on Gatsby: https://github.com/gatsbyjs/gatsby/pull/16573

Message in Slack from @alexeagle 
> seems like a rules_nodejs bug to me. commented on that issue - I think we just need to make our error message look more identical to the one vanilla nodejs would print and that gatsby expects.

alexeagle:
Was fixed in #1016

PR #1016, opened 2019-08-15
Reformat the error message for Node loader.js

bweston92:
The format of Node's own error from loader.js is https://github.com/nodejs/node/blob/a49b20d3245dd2a4d890e28582f3c013c07c3136/lib/internal/modules/cjs/loader.js#L264

## PR Checklist

## PR Type

- [x] Bugfix describe:

Issue Number: #1015

## What is the new behavior?

The error message reflects the same as Node's error message here: https://github.com/nodejs/node/blob/a49b20d3245dd2a4d890e28582f3c013c07c3136/lib/internal/modules/cjs/loader.js#L264

- [x] Yes
- [ ] No

Use the new error message format.

### DIFF
--- a/internal/node/node_loader.js
+++ b/internal/node/node_loader.js
@@ -467,16 +467,21 @@ module.constructor._resolveFilename = function(request, parent, isMain, options)
     failedResolutions.push(`node_modules attribute (${NODE_MODULES_ROOT}) - ${e.toString()}`);
   }
 
-  const error = new Error(
-      `Cannot find module '${request}'. ` +
-      'Please verify that the package.json has a valid "main" entry'
-  );
+  // Print the same error message that vanilla nodejs does.
+  // See https://github.com/bazelbuild/rules_nodejs/issues/1015
+  let moduleNotFoundError = `Cannot find module '${request}'. ` +
+      'Please verify that the package.json has a valid "main" entry';
+  if (DEBUG) {
+    moduleNotFoundError += `\nrequired in target ${TARGET} by '${parentFilename}'\n  looked in:\n` +
+        failedResolutions.map(r => `    ${r}`).join('\n') + '\n';
+  }
+  const error = new Error(moduleNotFoundError);
   error.code = 'MODULE_NOT_FOUND';
   // todo - error.path = ?;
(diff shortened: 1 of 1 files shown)


alexeagle:
Sorry, we had a bug following the release yesterday which broke CI, would you mind re-basing this on master to get it green?

internal/node/node_loader.js
```diff
@@ -468,9 +468,15 @@ module.constructor._resolveFilename = function(request, parent, isMain, options)
...
-      `${TARGET} cannot find module '${request}' required by '${parentFilename}'\n  looked in:\n` +
-      failedResolutions.map(r => `    ${r}`).join('\n') + '\n');
+      `Cannot find module '${request}'. ` +
+      'Please verify that the package.json has a valid "main" entry'
+  );
   error.code = 'MODULE_NOT_FOUND';
+  // todo - error.path = ?;
+  error.requestPath = parentFilename;
+  error.bazelTarget = TARGET;
+  error.failedResolutions = failedResolutions;
```
alexeagle: will something up the stack catch this error and print it? I think we still want to help the user resolve the problem, in theory the failed resolutions help with that

bweston92:
Unless someone wraps require statements in a try catch and handles the
error with a function that knows we decorate the error with these keys
“bazelTarget” and “failedResolutions” there isn’t a lot we can do.

I understand we want to be more verbose but the that is a trade off, others
are matching on Node’s error so these packages either need to all change or
these simply we change these rules.

I will rebase shortly as afk.

alexeagle:
Could you use the `DEBUG` constant?
`if (DEBUG) addMoreLinesToTheErrorMessage`

that lets us use the long-form error when we are figuring out bugs

bweston92:
I did think about that, will amend when I rebase.

internal/node/node_loader.js
alexeagle: resolved it myself

bweston92: whoops thanks for that @alexeagle I have been busy with some stuff.

bweston92:
Is that error a flake? Seems like only failing on the MacOS build...

PR #1017, opened 2019-08-15
feat(less): new less package

jvanallen93:
## Other information
Feature was requested in [angular/angular#19058](https://github.com/angular/angular/issues/19058#issuecomment-521240505) and myself. 

I saw @alexeagle made a pull request for Less support and one of the comments was support for `include-paths`. [According to Alex](https://github.com/bazelbuild/rules_nodejs/pull/908#discussion_r307002316), it seem that feature isn't needed as it'll be eventually deprecated as well in the Sass rules eventually so I didn't include it in this implementation. 

I didn't include `compress` as an option either because, [according to Less's documentation](http://lesscss.org/usage/#less-options-compress-deprecated-), it is currently deprecated.


e2e/less/yarn.lock
```diff
@@ -0,0 +1,408 @@
...
+
+
+"@bazel/less@file:../../dist/bin/packages/less/npm_package":
+  version "1.2.3"
+  dependencies:
+    less "^3.9.0"
+
+ajv@^6.5.5:
+  version "6.10.2"
+  resolved "https://nexus-binrepo.sapphirepri.com/nexus/content/groups/npm-all/ajv/-/ajv-6.10.2.tgz#d3cea04d6b017b2894ad69040fec8b623eb4bd52"
```
alexeagle: oops, this file needs to be regenerated without your custom registry mirror - this fails on CI

e2e/less/package.json
```diff
@@ -0,0 +1,10 @@
+{
+    "name": "e2e-less",
+    "private": true,
+    "devDependencies": {
+        "@bazel/less": "file:../../dist/bin/packages/less/npm_package"
```
alexeagle: this can be "latest" - after we push the first time, the e2e directory will be self-contained

e2e/less/yarn.lock
jvanallen93: I knew this was going too smoothly. Nice catch!

e2e/less/package.json
jvanallen93: Gotcha. So is it expected for the e2e builds to fail until this package is deployed?

packages/less/src/index.bzl
```diff
@@ -0,0 +1,57 @@
...
+    attrs = {
+        "src": attr.label(
+            mandatory = True,
+            allow_single_file = True,
+        ),
+        "compiler": attr.label(
+            doc = """Label that points to the lessc binary to run.
+            If you install your npm packages to a workspace named something other than "npm",
+            you may need to set this to `@my_npm_name//less/bin:lessc`""",
+            default = Label("@npm//less/bin:lessc"),
```
gregmagolan: Should be `default = Label("@npm//@bazel/less/bin:less"),` so by default it uses the index.js file in the npm package and requires the version of less that @bazel/less depends on and not a user version by mistake

PR #1020, opened 2019-08-16
refactor: remove unused transitive_es5_sources (& es6) from legacy typescript provider

gregmagolan:


gregmagolan:
This was just testing to see if either of these was needed. Looks like transitive_es6_sources is required here (although not used at all in g3) still but will get removed in the upcoming JS providers refactor.

PR #1031, opened 2019-08-18
feat(builtin): introduce dynamic dependencies concept

alexeagle:
This lets our generated nodejs_binary targets for npm packages express dynamic runtime-only dependencies
This is needed for example so that rollup binary can require the rollup-plugin-* packages


internal/npm_install/generate_build_file.js
```diff
@@ -436,6 +438,27 @@ function hasRootBuildFile(pkg, rootPath) {
   return false;
 }
 
+function addDynamicDependencies(pkgs, dynamic_deps = DYNAMIC_DEPS) {
+  pkgs.forEach(p => {
+    function match(name) {
+      // Automatically include dynamic dependency on plugins of the form pkg-plugin-foo
+      if (name.startsWith(`${p._name}-plugin-`)) return true;
```
gregmagolan: Add automatic and wildcard support to doc string?

gregmagolan:
This is great! 🏖

alexeagle:
@soldair points out that our assumption that we can teach Bazel the npm dependency graph is misguided - the only principled answer is that anything in node_modules can require() anything else.

We get performance gains out of fine-grained npm deps but we should push harder to get the same performance without requiring users to take steps like this to make the dependency graph completely cover all the require() edges.

PR #1040, opened 2019-08-20
feat(typescript): add JSModuleInfo provider to ts_library outputs

gregmagolan:
@alexeagle and I worked on this over the last week to prototype how a new set of JS providers would behave in all of the nodejs rules. This also includes the start of new simplified rollup and terser rules based on our discussions with @jbedard.

Also depends on provider changes to ts_library found here: https://github.com/gregmagolan/rules_typescript/commit/125d40b070cdd6fc2d465115976b8b67f5afa6c0

Open to comments and suggestions. We're aiming to get new JS providers as a breaking change for 1.0.

//cc @jbedard @Toxicable @mgechev @filipesilva @soldair @josephperrott @alan-agius4 @Globegitter

### DIFF
--- a/providers.bzl
+++ b/providers.bzl
@@ -25,8 +25,10 @@ load(
 load(
     "//internal/providers:js_providers.bzl",
     _JSEcmaScriptModuleInfo = "JSEcmaScriptModuleInfo",
+    _JSModuleInfo = "JSModuleInfo",
     _JSNamedModuleInfo = "JSNamedModuleInfo",
     _js_ecma_script_module_info = "js_ecma_script_module_info",
+    _js_module_info = "js_module_info",
     _js_named_module_info = "js_named_module_info",
 )
 load(
@@ -42,6 +44,8 @@ load(
 
 provide_declarations = _provide_declarations
 DeclarationInfo = _DeclarationInfo
+JSModuleInfo = _JSModuleInfo
+js_module_info = _js_module_info
 JSNamedModuleInfo = _JSNamedModuleInfo
 js_named_module_info = _js_named_module_info
 JSEcmaScriptModuleInfo = _JSEcmaScriptModuleInfo
--- a/internal/npm_install/node_module_library.bzl
+++ b/internal/npm_install/node_module_library.bzl
@@ -15,7 +15,7 @@
 """Contains the node_module_library which is used by yarn_install & npm_install.
 """
 
-load("@build_bazel_rules_nodejs//:providers.bzl", "DeclarationInfo", "NpmPackageInfo", "js_named_module_info")
+load("@build_bazel_rules_nodejs//:providers.bzl", "DeclarationInfo", "NpmPackageInfo", "js_module_info", "js_named_module_info")
 
 def _node_module_library_impl(ctx):
     workspace = ctx.label.workspace_root.split("/")[1] if ctx.label.workspace_root else ctx.workspace_name
@@ -69,6 +69,11 @@ def _node_module_library_impl(ctx):
             DeclarationInfo(
                 declarations = declarations,
                 transitive_declarations = transitive_declarations,
+            ),
+            js_module_info(
+                module_format = "",
+                sources = direct_sources,
+                deps = ctx.attr.deps,
             ),
             js_named_module_info(
                 sources = depset(ctx.files.named_module_srcs),
--- a/internal/pkg_npm/pkg_npm.bzl
+++ b/internal/pkg_npm/pkg_npm.bzl
@@ -6,7 +6,7 @@ If all users of your library code use Bazel, they should just add your library
 to the `deps` of one of their targets.
 """
 
-load("//:providers.bzl", "DeclarationInfo", "JSNamedModuleInfo", "NodeContextInfo")
+load("//:providers.bzl", "DeclarationInfo", "JSModuleInfo", "NodeContextInfo")
 load("//internal/common:path_utils.bzl", "strip_external")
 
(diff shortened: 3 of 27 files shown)


packages/terser/src/index.bzl
```diff
@@ -0,0 +1,126 @@
+"User configuration to run the terser binary under bazel"
+
+load("@build_bazel_rules_nodejs//:providers.bzl", "JSModuleInfo", "collect_js_modules")
+
+TERSER_ATTRS = {
+    "srcs": attr.label_list(
```
jbedard: I commented earlier on changing this to just `src`, which I think makes the `collect_js_modules` unnecessary which might be nice. Any downside to doing that?

packages/terser/src/index.bzl
```diff
@@ -0,0 +1,126 @@
...
+        outputs = outputs,
+        executable = ctx.executable.terser_bin,
+        arguments = [args],
+        progress_message = "Optimizing JavaScript %s [terser]" % (output.short_path),
+    )
+    return outputs
+
+def _terser(ctx):
+    modules = collect_js_modules(ctx)
+    outputs = run_terser(ctx, modules.sources, ctx.outputs.optimized)
```
jbedard: What if terser is invoked without using `JSModuleInfo`? I thought we'd be running terser on each provider individually, something such as what was started at https://github.com/jbedard/rules_nodejs/commit/58c8d83813a5972ceb811d56745f669e0e2eb102#diff-86d64554a52a8e537d18c6619fcedc10R129

packages/terser/src/index.bzl
gregmagolan: I think `src` will likely be a better shape since it ends up being a single output file one way or the other

providers.bzl
```diff
@@ -0,0 +1,129 @@
...
+Note that this means a given build either produces debug or non-debug output.
+If users really need to produce both in a single build, they'll need two rules with
+differing 'debug' attributes.
+"""
+
+JSModuleInfo = provider(
+    doc = """TODO""",
+    fields = {
+        "module_format": "a string like [amd, cjs, esm, iife, umd]",
+        "sources": "depset of JavaScript files",
```
jbedard: Where do sourcemaps live? I thought there would be a `sourcemaps` property beside this (not sure where I got that thought though 🤔).

gregmagolan: Thinking the would live together now as they always want to travel together except in a few leaf cases such as bundling when you want to filter them out. Having them separated in the provider was leading to a lot of extra boilerplate in ever rule & aspect the provider was used. Trying this out for size.

packages/terser/src/index.bzl
gregmagolan: That seems like a strange and unexpected shape as well since terser would then have multiple outputs and would need to opinionated in the output name. With a single output from either `JSModuleInfo` (if it exists) or from DefaultInfo files (fallback) then it would be just be named `{name}.js` by default. To use other providers you would put a JS provider filter rule before terser that would take a provider such as `JSNamedModuleInfo` and output it as `JSModuleInfo` for terser to consume. Something similar is already done in g3 for ts_devserver which does not know how to find the named js on its own.

providers.bzl
jbedard: That sounds good, we should update this field description though.

packages/terser/src/index.bzl
jbedard: Yeah the output names would definitely be a problem, and I guess putting a single `{name}.js` file into both providers like this solves that.

Would it ever be possible for a user to expect `DefaultInfo` to be used instead of `JSModuleInfo`, and that causing unexpected output in `DefaultInfo`? I guess just adding another step (putting things into a `filegroup`?) would fix that though.

WORKSPACE
```diff
@@ -66,13 +76,16 @@ load("//:defs.bzl", "npm_install", "yarn_install")
 
 yarn_install(
     name = "npm",
+    # Declare the implicit dependency from a tool to the things it will `require`
+    dynamic_deps = {
```
alexeagle: not needed, since the dynamic-deps feature does this pattern automatically

packages/terser/src/index.bzl
gregmagolan: Initial terser rule was landed on master and uses some bazel magic @alexeagle discovered to get named outputs for both the `{name}.js` and `{name}.js.map` if sourcemap is enabled. The landed version uses DefaultInfo only and I rebased on this branch for it to handle the JSModuleInfo provider being worked on here as well.

Default behavior is use JSModuleInfo if found otherwise fallback to DefaultInfo, however, we're leaning towards not using apsects to collect JSTransitiveModuleInfo so JSModuleInfo will always provide all direct & transitive deps. This means that if you really want to force the usage of DefaultInfo you can the `src` through a `filegroup` which will filter it out and because no aspect is run it won't be found.

gregmagolan:
JSEcmaScriptModuleInfo provider factored out of this PR and landed in #1201 as the load bearing minimal change required to get ts_library outputs working with the new rollup_bundle for 0.38.0

I'll be factoring out JSNamedModuleInfo and JSModuleInfo out of this as separate PRs and the NodeModuleSources and NodeModuleInfo providers cleanup as a separate PR as well

gregmagolan:
JSNamedModule provider & related load bearing changes factored out into https://github.com/bazelbuild/rules_nodejs/pull/1215

gregmagolan:
Dup of https://github.com/bazelbuild/rules_nodejs/pull/1843

PR #1043, opened 2019-08-21
feat: add templated_args_file to allow long agrs to be written to a file

mattem:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Agrs placed in templated_args can be too long for the OS to pass to the binary. This is happening internally for us when passing a large list of files to a nodejs_binary for processing.

## What is the new behavior?
By specifying the `templated_args_file` attr on `nodejs_binary`, the rule will write the args to the file specified and pass that as an argument. Args prefixed with `--node_options=` are ignored and not set in the arg file.

- [ ] Yes
- [x] No


mattem:
So overflow into the param file automatically based on args length, and allow setting to always use?

alexeagle:
could we always use a params file? I think this is possible since the only code that parses the argv is our node_launcher?

mattem:
That seems like a breaking change? I would have thought passing args to the binary is quite common, and moving that to a file for a small set of args seems overkill? I quite like the explicitness of specifying an args file rather than having bazel or the rule decide.

alexeagle:
For other actions, it's not breaking - Bazel decides to spill the arguments to a file, and the action it spawns is code we control which handles the params file and expands the options before passing it to the program we are running. I think with node_launcher.sh that might be tricky but still possible?

my motivation is to avoid making the API surface more complex and adding more code paths to get into the binary. It would be simpler for everything to work the same way, and users not to fall off the cliff you did and then have to discover this feature is the way back to safety

mattem:
I'm not sure I follow here, perhaps I've misunderstood how the params file is used by Bazel in actions. I thought bazel would spill the args into the file and then pass the file as a param to the underlying binary, which in turn would need to know how to a, recognize it's been passed a file and  b, read that file and parse the lines as args. If it doesn't have that logic, it wouldn't work?

Having the wrapper script unpack the params file and pass them to the binary just causes the same issue again, it would overflow. Unless you're saying that node_launch.sh would patch in the args from the params file to `process.argv`? In which case I agree, that wouldn't be breaking and seems a good direction to take this :)

alexeagle:
I understand better. The launcher can intercept the params file, but can't call node with the whole argv because that causes the same problem you have now. Only inside the node program can we read and unpack the params file.
I don't think the wrapper has the ability to pass the params into the node process such that they appear in `process.argv` without actually having been passed as command-line args.
We *could* read the params file in the node_loader.js however, and set the `process.argv` there.
Let me sync with @soldair about whether that's a good idea. We've been trying to reduce the responsibilities of the loader so that programs running under bazel are mostly identical to how they're run without it

alexeagle:
If you have time to add a repro to this PR, that would help when experimenting with other possible solutions.

mattem:
> Only inside the node program can we read and unpack the params file.

Yes, exactly. A lot of our apps that do large file list processing are expecting to be passed an args file, or will assume an arg file from an arg prefixed with `@`.

If the goal is to keep running programs the same under bazel as they are run without it, then the method in this PR already kinda does that, no? It is up to the binary to unpack and process the params file.

> If you have time to add a repro to this PR, that would help when experimenting with other possible solutions.

Yup, will do that today

Issue #1045, opened 2017-12-13
Add a React example authored in TSX


linuxerwang:
Is it possible for rules_typescript to compile ReactJS? I tried a bit and found that it can generate the bundle.js but in the devserver it's not running.

Here is the generated bundle.js:

```js
(function (factory) {
    if (typeof module === "object" && typeof module.exports === "object") {
        var v = factory(require, exports);
        if (v !== undefined) module.exports = v;
    }
    else if (typeof define === "function" && define.amd) {
        define("__main__/streamline/fe/src/entry", ["require", "exports", "react", "react-dom"], factory);
    }
})(function (require, exports) {
    "use strict";
    Object.defineProperty(exports, "__esModule", { value: true });
    var React = require("react");
    var ReactDOM = require("react-dom");
    var App = function () {
        return (React.createElement("div", null,
            React.createElement("p", null, "Hello world!")));
    };
    ReactDOM.render(React.createElement(App, null), document.getElementById('app'));
});
```

Looks like the anonymous factory function was never called. Both requirejs and amd was installed with yarn.

Any help? Thanks.

Issue #933, opened 2019-07-24
react and csstips use module syntax rollup doesn't understand


kellycampbell:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `rollup_bundle`

### Is this a regression?

Unknown

### Description

In a rolled up js file, the browser has an error where a * import was used:

`Uncaught TypeError: undefined is not a function`

My original ts code: 
```
import * as csstips from 'csstips';

console.log("Setting up page with csstips", csstips);
csstips.normalize();
csstips.setupPage('#root');
```

but the code where I used csstips has this in the es2015.js file:

```
console.log("Setting up page with csstips", csstips);
undefined();
undefined('#root');
```

My BUILD rules are like this:

```
ts_library(
    name = "app_lib",
    srcs = glob([
        "**/*.ts", 
        "**/*.tsx",
    ]),
    deps = [
        "@npm//csstips",
        "@npm//csx",
        "@npm//react",
        "@npm//react-dom",
        "@npm//tslib",
        "@npm//typestyle",
        "@npm//@types",
    ],
    tsconfig = "//:tsconfig.json",
)

rollup_bundle(
    name = "app_bundle",
    deps = [
        ":app_lib",
    ],
    entry_point = ":app.tsx",
    globals = {
        "react": "React",
        "react-dom": "ReactDOM",
    },
)
```

kellycampbell:
I found a working solution:

```
import csstips from 'csstips';
```

surlyengineer:
Having a similar problem with `import * as React from 'react'`, e.g.

kellycampbell:
I use `import React from 'react';`

My tsconfig has the following flags which create the __importStar and __importDefault:

```
        "esModuleInterop": true,
        "allowSyntheticDefaultImports": true,
```

kellycampbell:
On my original problem, it seems to be something specific to this csstips library, and I suspect also may be due to a bug in either rollup or typescript.

Another library from the same author as csstips works ok with the import *:

Works with `rollup_bundle` but doesn't work with `ts_library`:
```
import csstips from 'csstips';
import * as csx from 'csx';
```

Works with `ts_library` but not with `rollup_bundle`:
```
import * as csstips from 'csstips';
import * as csx from 'csx';
```

The result of all this is with React and using that csstips library I can do either server-side-rendering without hydration from the rollup bundle, or no SSR and including the bundle normally, but not both SSR + hydrate ("isomorphic").

[...]

Issue #555, opened 2019-02-21
How to bundle react.js into the bundle with rollup_bundle?


solomatov:
Is there any standard way to embed react.js and other libraries into the resulting bundle like webpack does? Currently, I bundle react separately via globals.

What about other libraries?

It would be nice to provide information on this in the documentation.

stephanwlee:
Hi rules_nodejs, thanks for the awesome rules. If I may interpret what @solomatov and other issues like #532 are saying, I think the rules_nodejs should have some minimal example using React cause it is not straight-forward. 

In case it is not clear, let me share why it is not so straight-forward. Technically, React is just like other npm packages and you can use rules_nodejs to npm install and bundle application using rollup. React, however, is a bit different than some npm packages because it is isomorphic and uses `process.env.NODE_ENV`. The NODE_ENV, of course, is a Node global and is not present in the browser. When building such library, people often use [Babel plugin](https://babeljs.io/docs/en/babel-plugin-transform-inline-environment-variables) or [rollup-plugin-replace](https://github.com/rollup/rollup-plugin-replace) to replace the statement as a string literal. 

Now, I wish rules_nodejs to include an example that uses React to dogfood how difficult customizing `rollup_bundle` can be. Few problems I faced:
1. naively, I wished to pass my custom rollup configuration but that is not possible. You must fork the rollup_bundle using `internal/rollup` which is not ideal
2. cannot use a node_modules in a custom rollup config not present in this package.json https://github.com/bazelbuild/rules_nodejs/blob/15cf52364b7bd3f5ab2849fc3a6c3ae248c931f7/internal/rollup/package.json#L4-L15. I ended up creating my own nodejs_binary for rollup (I understand why this would be desired but keeping track of dependencies of bundler vs. app feel a bit too much -- bifurcating dependencies in devDependencies vs. dependencies, imo, should be enough).
3. related to (2) but https://github.com/bazelbuild/rules_nodejs/blob/e115dea562ca60654b496df677b0b4327c62eba8/internal/rollup/rollup_bundle.bzl#L54-L55 is probably out-dated and is not straight-forward. https://github.com/angular/angular/blob/master/packages/bazel/src/ng_rollup_bundle.bzl

Perhaps I am missing context or information. I would appreciate it if you can point to a working example that makes my point invalid. Thanks a lot!

tony-scio:
[...]

PR #1076, opened 2019-08-30
feat(example): add react example

codesuki:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes
Example for react

Issue Number: N/A

- [ ] Yes
- [ ] No

## Other information
There are 3 big blockers for react to be easily usable.
https://github.com/bazelbuild/rules_nodejs/issues/1045
https://github.com/bazelbuild/rules_nodejs/issues/555
https://github.com/bazelbuild/rules_nodejs/issues/933

I added workarounds for both so the example works with devserver and prodserver (bundle)


alexeagle:
wow @codesuki I :heart: so much. Thanks!

I cloned this locally and I'm trying to get the devserver to work. Both at your commit and the one I've pushed to your branch, I can
`require(['react'])` in the devtools, but `require(['react']).default` is undefined, so the generated js for greeting.tsx
```
$ cat dist/bin/greeting.js 
(function (factory) {
    if (typeof module === "object" && typeof module.exports === "object") {
        var v = factory(require, exports);
        if (v !== undefined) module.exports = v;
    }
    else if (typeof define === "function" && define.amd) {
        define("examples_reactjs/greeting", ["require", "exports", "react"], factory);
    }
})(function (require, exports) {
    "use strict";
    Object.defineProperty(exports, "__esModule", { value: true });
    const react_1 = require("react");
    const greeting = react_1.default.createElement("h1", null, "Hello, world");
```

fails there at that last line.

Does it have anything to do with the `tsxFactory` option to `tsc`? I'll keep debugging but LMK if you already have an answer. I assume the devserver worked for you?

codesuki:
Hey, thanks for improving the PR! It's much cleaner now. I didn't know what to do about the tests.

I just left for Hawaii vacation so I cannot check the code now. 

I guess you get a null pointer exception? 
When I prepared the PR both the dev server and the prod server (bundle) worked. I don't remember if my output is also had 'react.default'. I'll confirm once I am back.

86:
`devserver` works for me. I got different output from @alexeagle. TypeScript version is not same?
```
$ cat dist/bin/greeting.js                                                                                                                                                                                                                                              git:react-example*
(function (factory) {
    if (typeof module === "object" && typeof module.exports === "object") {
        var v = factory(require, exports);
        if (v !== undefined) module.exports = v;
    }
    else if (typeof define === "function" && define.amd) {
        define("examples_reactjs/greeting", ["require", "exports", "react"], factory);
    }
})(function (require, exports) {
    "use strict";
    Object.defineProperty(exports, "__esModule", { value: true });
    const React = require("react");
    const greeting = React.createElement("h1", null, "Hello, world");
    exports.default = greeting;
});
```

alexeagle:
Okay, problem here was
- for devmode to work properly, we should use `import * as React from 'react'` so that we have output JS with `react_1.default.createElement` rather than `react_1.createElement`
- that means that the rollup-plugin-commonjs wants to find the createElement symbol in the react/index.js file - but it can't. https://github.com/rollup/rollup-plugin-node-resolve/issues/58 describes the problem

I've added another commit here to make the test pass - but we don't really want the `namedExports: {
        'bazel-out/k8-fastbuild/bin/bundle.es6/external/npm/node_modules/react/index.js':
            ['createElement'],` mumble in the rollup.config.js

Let's say you're not using Bazel. It seems that bundling React into your app with Rollup requires this workaround where you list some symbols in the rollup.config. Seems not desirable.
So I suspect we should change this example to not bundle React into the app, instead we should load it as a script at runtime?

codesuki:
Does the bundle work when using `import * as React from 'react'`? 
I ran into this issue before: https://github.com/bazelbuild/rules_nodejs/issues/933 

Ah, the named exports you mention fixes this right? I agree, having to adjust rollup.config.js is cumbersome. 

Anyhow, what is the benefit of using `import *` over `import React from ...`? 
i.e. what is the difference between `react_1.default.createElement` and `react_1.createElement`?

I haven't done any web dev for several years now so I can't say anything useful to whether or not react should be bundled or loaded at runtime. Correct me if I say something wrong.
In general, I think loading it as an external dep would allow the browser to cache it between pages if the same version from some CDN is referenced, on the other hand, bundling could enable dead code elimination which I am not sure rollup actually does. 

For our actual production case, we don't mind either way I think. (I know this is just about the example here)

sahilprasad:
@codesuki @alexeagle thanks for your work on this!

Any thoughts on how this may work with a CRA application?

jkrems:
Hey - this is really cool! Some (JS/node) stylistic notes:

1. It may be more cleaner to teach the rollup rule about `--environment=NODE_ENV:production`. Potentially even setting it by default. If we don't, we'll always risk that we bundle suboptimal 3rd party code. The "it throws" is the happy path - the worse case is that the check is protected by defensive code and it's just silently running verbose dev code in production.
2. Afaik `import React from 'react'` is the canonical import and at least in rollup it does work. The big issue is that react ships CommonJS, not ES modules. So `import *` isn't really correct since the only namespace member from an ESM perspective is `default`, there are no named exports.
3. In the spec, it may make sense to use a clean sandbox for the DOM APIs instead of putting them into the global scope. Here's an example how it might look like:

```js
'use strict';

const domino = require('domino');
const fs = require('fs');
const vm = require('vm');

describe('react webapp', () => {
  let sandbox;
  beforeAll(() => {
    // Domino gives us enough of the DOM API that we can run our JavaScript in node rather than the
    // browser. That makes this test a lot faster
    const html = fs.readFileSync(require.resolve('./index.html'));
    const window = domino.createWindow(html, '/');
    // Make all Domino types available as types in the sandbox.
    sandbox = vm.createContext({
      window,
      document: window.document,
      navigator: window.navigator,
      process: {
        env: 'production',
      },
    });
  });

  it('works', () => {
    const bundlePath = require.resolve('./bundle.es2015');
    const bundleContents = fs.readFileSync(bundlePath, 'utf8');
    vm.runInContext(bundleContents, sandbox, { filename: bundlePath });

    expect(sandbox.document.body.textContent.trim()).toEqual('Hello from React!');
  });
});
```

Toxicable:
Lets wait till the new rollup rule has landed, there's a couple of different rollup configs/plugins I'd like to try out here which may make things a bit easier, which will be possible with the new rule.

@jkrems On your point 3. while I fundamentally agree that sandboxing code is a good idea, I don't think it fits unit tests.
None of the React examples i've seen do it, it'll make unit tests slower and it's bad UX since there's a bunch of API's not known to most devs.
I don't think we should recommend that approach for unit testing in this example.

bjchambers:
We tried following this approach and it worked OK for getting react working with a `ts_devserver`, but when trying to use `react-router` it has problems. Both `react-router` and `react-router-dom` have a check for the type of module they are loaded with (to make sure it's the same). Even if both are loaded in the `ts_devserver` using `__umd`, they report as different.

It would be helpful if the example included some of these common libraries, since they seem to be sensitive to how things are connected. Specifically, when looking in the browser:

react-router.umd.js includes
```
if (process.env.NODE_ENV !== "production") {
  if (typeof window !== "undefined") {
    var global = window;
    var key = "__react_router_build__";
    var buildNames = {
      cjs: "CommonJS",
      esm: "ES modules",
      umd: "UMD"
    };

    if (global[key] && global[key] !== "esm") {
      var initialBuildName = buildNames[global[key]];
      var secondaryBuildName = buildNames["esm"]; // TODO: Add link to article that explains in detail how to avoid
      // loading 2 different builds.

      throw new Error("You are loading the " + secondaryBuildName + " build of React Router " + ("on a page that is already running the " + initialBuildName + " ") + "build, so things won't work right.");
    }

    global[key] = "esm";
  }
}
```

and `react-router-dom.umd.js` includes:
```
{
  if (typeof window !== "undefined") {
    var global = window;
    var key = "__react_router_build__";
    var buildNames = {
      cjs: "CommonJS",
      esm: "ES modules",
      umd: "UMD"
    };

    if (global[key] && global[key] !== "cjs") {
      var initialBuildName = buildNames[global[key]];
      var secondaryBuildName = buildNames["cjs"]; // TODO: Add link to article that explains in detail how to avoid
      // loading 2 different builds.

      throw new Error("You are loading the " + secondaryBuildName + " build of React Router " + ("on a page that is already running the " + initialBuildName + " ") + "build, so things won't work right.");
    }

    global[key] = "cjs";
  }
}
```

jkrems:
> On your point 3. while I fundamentally agree that sandboxing code is a good idea, I don't think it fits unit tests.

What I wrote is pretty much how Jest works. It hides it behind some nice facade but this is pretty close to the exact behavior (down to the node APIs used). It's not very common to do this setup in individual test files, that's definitely true. :) But if your test is using `document` or  `window` (as opposed to shallow renders via sth like enzyme) you absolutely shouldn't be setting those on the "normal" global object.

EmmanuelDemey:
We have just checkouted your branch and we have a runtime error. 

image

mrmeku:
@alexeagle I think this can be closed now

AnatoleLucet:
@mrmeku why is that?

johnedmonds:
FYI `bazel-out/k8-fastbuild/bin/bundle.es6/external/npm/node_modules/react/index.js` doesn't work on macos, the path would be `bazel-out/darwin-fastbuild/bin/bundle.es6/external/npm/node_modules/react/index.js`

PR #1080, opened 2019-09-02
feat: add default DEBUG and VERBOSE_LOGS configuration_env_vars to nodejs_binary

gregmagolan:
- all rules updated to use DEBUG and VERBOSE_LOGS environment variables
- added golden_debug attribute golden_file_test to support the case where a rule has different output if DEBUG is set; for example
```
golden_file_test(
name = "test",
actual = "out.min.js",
golden = "output.golden.js_",
golden_debug = "output.debug.golden.js_",
)
```

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -106,6 +106,16 @@ tasks:
     - "--test_tag_filters=-e2e,-examples,-fix-bazelci-ubuntu"
     test_targets:
     - "//..."
+  ubuntu1804_debug:
+    name: ubuntu1804_debug
+    platform: ubuntu1804
+    test_flags:
+    - "--define=VERBOSE_LOGS=1"
+    - "--define=DEBUG=1"
+    - "--test_tag_filters=-e2e,-examples,-fix-bazelci-ubuntu"
+    test_targets:
+    - "//..."
+    - "//packages/terser/test/debug:test_define_DEBUG"
   ubuntu1804_e2e:
     name: ubuntu1804_e2e
     platform: ubuntu1804
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -152,7 +152,7 @@ jobs:
       - *init_bazel
       - *hide_node_and_yarn_local_binaries
 
-      - run: bazel test ... --test_tag_filters=-e2e,-examples
+      - run: bazel test --test_tag_filters=-e2e,-examples ...
 
       - store_artifacts:
           path: dist/bin/release.tar.gz
--- a/DEVELOPING.md
+++ b/DEVELOPING.md
@@ -24,6 +24,10 @@ To do a full clean run:
 yarn clean_all
 ```
 
+## Debugging
+
+See `Debugging` section under `/docs/index.md`.
+
 ## Releasing
 
 Start from a clean checkout at master/HEAD.
--- a/common.bazelrc
+++ b/common.bazelrc
@@ -26,7 +26,32 @@ test --test_output=errors
 
 # Support for debugging NodeJS tests
 # Add the Bazel option `--config=debug` to enable this
-test:debug --test_arg=--node_options=--inspect-brk --test_output=streamed --test_strategy=exclusive --test_timeout=9999 --nocache_test_results
+# --test_output=streamed
+#     Stream stdout/stderr output from each test in real-time.
+#     See https://docs.bazel.build/versions/master/user-manual.html#flag--test_output for more details.
+# --test_strategy=exclusive
+#     Run one test at a time.
+# --test_timeout=9999
+#     Prevent long running tests from timing out
+#     See https://docs.bazel.build/versions/master/user-manual.html#flag--test_timeout for more details.
+# --nocache_test_results
+#     Always run tests
+# --node_options=--inspect-brk
+#     Pass the --inspect-brk option to all tests which enables the node inspector agent.
+#     See https://nodejs.org/de/docs/guides/debugging-getting-started/#command-line-options for more details.
+# --define=VERBOSE_LOGS=1
+#     Rules will output verbose logs if the VERBOSE_LOGS environment variable is set. `VERBOSE_LOGS` will be passed to
+#     `nodejs_binary` and `nodejs_test` via the default value of the `default_env_vars` attribute of those rules.
+# --define=DEBUG=1
+#     Rules may change their build outputs if the DEBUG environment variable is set. For example,
+#     mininfiers such as terser may make their output more human readable when this is set. `DEBUG` will be passed to
+#     `nodejs_binary` and `nodejs_test` via the default value of the `default_env_vars` attribute of those rules.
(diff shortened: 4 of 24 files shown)


internal/rollup/terser-wrapped.js
```diff
@@ -59,20 +59,20 @@ if (inputs.length != 1) {
...
     'sourceMap': {'filename': sourceMapFile},
     'compress': {
       'pure_getters': true,
       'passes': 3,
       'global_defs': {'ngDevMode': false, 'ngI18nClosureMode': false},
-      'keep_fnames': !debug,
-      'reduce_funcs': !debug,
-      'reduce_vars': !debug,
-      'sequences': !debug,
+      'keep_fnames': !DEBUG,
```
alexeagle: seems wrong - DEBUG was set from 'DEBUG_LOGS' - shouldn't this one be set from 'DEBUG'?

gregmagolan: good catch thanks

gregmagolan:
I'll add some docs to the main readme for these settings. I think we can recommend that user's add
```
test:debug --test_arg=--node_options=--inspect-brk --test_output=streamed --test_strategy=exclusive --test_timeout=9999 --nocache_test_results --define=VERBOSE_LOGS=1
run:debug --define=VERBOSE_LOGS=1
build:debug --define=DEBUG=1
```
to the bazelrc in the docs and explain what each flag does including the two define flags. User's can then choose to exclude certain flags such as `build:debug --define=DEBUG=1` if they don't want `--config=debug` to affect their build.

Also, I think VERBOSE_LOGS may be better than DEBUG_LOGS as per your slack comment.

.bazelci/presubmit.yml
```diff
@@ -106,6 +106,16 @@ tasks:
     - "--test_tag_filters=-e2e,-examples,-fix-bazelci-ubuntu"
     test_targets:
     - "//..."
+  ubuntu1804_debug:
+    name: ubuntu1804_debug
+    platform: ubuntu1804
+    test_flags:
+    - "--define=VERBOSE_LOGS=1"
```
alexeagle: hmm, I'm not sure it's worth adding CI load for this. let's just turn on these options in all our builds? I think spammy is probably good so we can understand CI failures for all jobs?

.bazelci/presubmit.yml
```diff
@@ -106,6 +106,16 @@ tasks:
...
+    - "--define=DEBUG=1"
+    - "--test_tag_filters=-e2e,-examples,-fix-bazelci-ubuntu"
+    test_targets:
+    - "//..."
+    - "//packages/terser/test/debug:test_define_DEBUG"
```
alexeagle: I guess this is the one bit of coverage we miss if we don't have a mix of debug/non-debug jobs. Not convinced it's needed since the `debug` attribute covers like 80% of the feature

.circleci/config.yml
```diff
@@ -152,7 +152,22 @@ jobs:
...
+  test_debug:
+    <<: *job_defaults
+    resource_class: xlarge
+    steps:
+      - *attach_workspace
+      - *init_environment
+      - *init_bazel
+      - *hide_node_and_yarn_local_binaries
+
+      - run: bazel test --define=DEBUG=1 --define=VERBOSE_LOGS=1 --test_tag_filters=-e2e,-examples ... //packages/terser/test/debug:test_define_DEBUG
```
alexeagle: so much CI load!
aren't we at the point of only adding things to buildkite now?

DEVELOPING.md
```diff
@@ -24,6 +24,70 @@ To do a full clean run:
 yarn clean_all
 ```
 
+## Debugging
+
+Adding following options to your `.bazelrc` file to add support for debugging NodeJS tests and to make the build and test
+output more verbose.
+
+```
+# Support for debugging NodeJS tests
```
alexeagle: maybe just point to common.bazelrc so we don't have to duplicate it in markdown

internal/golden_file_test/bin.js
```diff
@@ -22,7 +24,8 @@ ${prettyDiff}
       
 Update the golden file:
 
-            bazel run ${process.env['BAZEL_TARGET'].replace(/_bin$/, '')}.accept
+            bazel run ${debug ? '--define=DEBUG=1 ' : ''}${
```
alexeagle: yeah I guess that's the principled thing... kinda annoying though

internal/node/node.bzl
```diff
@@ -269,6 +269,19 @@ _NODEJS_EXECUTABLE_ATTRS = {
         allow_files = True,
         aspects = [sources_aspect, module_mappings_runtime_aspect, collect_node_modules_aspect],
     ),
+    "default_env_vars": attr.string_list(
+        doc = """Default environment variables that are added to `configuration_env_vars`.
+
+        This is separate from the default of `configuration_env_vars` so that a user can set `configuration_env_vars`
```
alexeagle: I think the indent here is wrong and will make docs look wrong

.circleci/config.yml
gregmagolan: yes. I think we can probably start taking CircleCI down. only a few assertions left there that are not covered on buildkite.

gregmagolan: I'll remove this job and rest can be removed separately

DEVELOPING.md
gregmagolan: 👍

internal/node/node.bzl
gregmagolan: Right. There a mix everywhere. Will fix here. Should make a PR to just clean them up everywhere.

.bazelci/presubmit.yml
gregmagolan: Since `- "--define=DEBUG=1"` I think we need to run all tests without it set and with it set. No other way to know if the rules that change their output based on this break things.

DEVELOPING.md
```diff
@@ -24,6 +24,42 @@ To do a full clean run:
 yarn clean_all
 ```
 
+## Debugging
```
alexeagle: I don't think this is the right file - users aren't directed to look at anything outside docs/

common.bazelrc
```diff
@@ -26,7 +26,30 @@ test --test_output=errors
...
+# --test_timeout=9999
+#     Prevent long running tests from timing out
+#     See https://docs.bazel.build/versions/master/user-manual.html#flag--test_timeout for more details.
+# --nocache_test_results
+#     Always run tests
+# --node_options=--inspect-brk
+#     Pass the --inspect-brk option to all tests which enables the node inspector agent.
+#     See https://nodejs.org/de/docs/guides/debugging-getting-started/#command-line-options for more details.
+# --define=VERBOSE_LOGS=1
+#     Rules will output verbose logs if the VERBOSE_LOGS environment variable is set.
```
alexeagle: nit: maybe also note that they need to use `default_ctx_vars` or whatever

gregmagolan: 👍

DEVELOPING.md
gregmagolan: 👍 moved to /docs/install.md

PR #1083, opened 2019-09-03
Forward environment variables needed for compilation of native dependencies.

uri-canva:
Just a draft to check if this would be interesting to maintainers.

Some packages have native dependencies that need certain environment variables to be set correctly to be able to build. `JAVA_HOME` is required for https://github.com/joeferner/node-java.

I'm not sure this can be easily upstreamed: we cannot make this configurable as an attribute as `environ` has to be declared in the rule. It could be made configurable if we made a macro that declared the rule with the right `environ`, but that would introduce an extra indirection to using the rule, and it would make `rules_nodejs` a bit unconventional compared to other rule sets.

If we don't make it configurable it is a breaking change, as we're now forwarding `PATH` instead of leaving it as the system path only, plus we are forwarding a lot of environment variables which would otherwise be unspecified and thus autodetected, for example `JAVA_HOME` through `/usr/libexec/java_home` on macOS or similar mechanisms for other settings / OSes.

Additionally it will cause the repository rule to be run every time any of those environment variable changes, regardless of whether the user needs them or not in `rules_nodejs`. This is especially bad if the user calls bazel from their IDE, as usually calling bazel from the IDE will pass a completely different set of environment variables compared to calling it from the command line.

Open to any solution to all these issues. Maybe add a separate variant of `npm_install` and `yarn_install` that does this? Like `native_yarn_install` or something like that.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

## Other information

### DIFF
--- a/internal/npm_install/npm_install.bzl
+++ b/internal/npm_install/npm_install.bzl
@@ -151,6 +151,16 @@ data attribute.
     ),
 })
 
+ENVIRONMENT_KEYS = [
+    "CC",
+    "CFLAGS",
+    "CXX",
+    "CXXFLAGS",
+    "JAVA_HOME",
+    "LDFLAGS",
+    "PATH",
+]
+
 def _create_build_files(repository_ctx, rule_type, node, lock_file):
     error_on_build_files = repository_ctx.attr.symlink_node_modules and not repository_ctx.attr.always_hide_bazel_files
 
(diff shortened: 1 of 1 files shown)


uri-canva:
Will fill out the rest of the PR description once some of the uncertainties are sorted out.

> please, look at the "Scope of the project" section in the README.md file

There doesn't seem to be one.

PR #1085, opened 2019-09-03
generate es2015 module correctly when not splitting

asuffield:
The default in write_rollup_config is iife, so this mode currently emits iife instead, then passes it to tsc (!).


jbedard:
@asuffield is this still necessary? Can you provide an example of what issue this causes?

asuffield:
Hmm. Well it's been a while, the relevant mode did not work at all when I had to fix it, but much has changed since then and I'm not sure of the current status.

asuffield:
The code has changed so much that this particular PR is long since irrelevant. I can't even find where the bug would have moved to now.

PR #1088, opened 2019-09-04
fix(builtin): fix localWorkspacePath logic

### LINKED ISSUE #1087: Manifest path logic gets wrong localWorkspacePath, breaks require
When we resolve from a RUNFILES_MANIFEST_FILE (generally Windows only), then we load the runfiles manifest and do some path logic on it to power later resolutions.

A manifest typically looks like

```
bazel_tools/tools/bash/runfiles/runfiles.bash C:/users/alexeagle/_bazel_alexeagle/7yoikqcr/external/bazel_tools/tools/bash/runfiles/runfiles.bash
build_bazel_rules_nodejs/packages/terser/test/inline_sourcemap/out.min.js C:/users/alexeagle/_bazel_alexeagle/7yoikqcr/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/terser/test/inline_sourcemap/out.min.js
build_bazel_rules_nodejs/packages/terser/test/inline_sourcemap/out.min.js.map C:/users/alexeagle/_bazel_alexeagle/7yoikqcr/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/terser/test/inline_sourcemap/out.min.js.map
build_bazel_rules_nodejs/packages/terser/test/inline_sourcemap/spec.js C:/users/alexeagle/documents/github/rules_nodejs/packages/terser/test/inline_sourcemap/spec.js

```

We iterate through these entries and skip to the last one, thanks to this logic:
https://github.com/bazelbuild/rules_nodejs/blob/df37fca0113f9e7e7536185cb907e2e6ca9e33f2/internal/node/node_loader.js#L126-L129
so we print
```
[test_loader.js] using localWorkspacePath C:/users/alexeagle/documents/github/rules_nodejs/
```
and things work.

---

However, if the manifest has a new entry that appears higher, and is under the output_base but not under the bin_dir (like the linker.js I was adding in #1079) then the manifest starts like this

```
bazel_tools/tools/bash/runfiles/runfiles.bash C:/users/alexeagle/_bazel_alexeagle/7yoikqcr/external/bazel_tools/tools/bash/runfiles/runfiles.bash
build_bazel_rules_nodejs/internal/node/node_loader.js C:/users/alexeagle/_bazel_alexeagle/7yoikqcr/external/build_bazel_rules_nodejs/internal/node/node_loader.js
```

The second entry isn't matched by the short-circuit logic, so it gets used. then we print
`[test_loader.js] using localWorkspacePath C:/users/alexeagle/_bazel_alexeagle/7yoikqcr/external/build_bazel_rules_nodejs/`
and later module resolutions fail.

alexeagle:
If the manifest has an entry in the output_base but not in the binDir or genDir, we get a wrong path for later resolutions

Fixes #1087

### DIFF
--- a/internal/node/node_loader.js
+++ b/internal/node/node_loader.js
@@ -107,6 +107,7 @@ function loadRunfilesManifest(manifestPath) {
   const runfilesManifest = Object.create(null);
   const reverseRunfilesManifest = Object.create(null);
   const input = fs.readFileSync(manifestPath, {encoding: 'utf-8'});
+  const outputBase = manifestPath.substring(0, manifestPath.indexOf('/execroot/'));
 
   // Absolute path that refers to the local workspace path. We need to determine the absolute
   // path to the local workspace because it allows us to support absolute path resolving
@@ -124,7 +125,7 @@ function loadRunfilesManifest(manifestPath) {
     // runfile refers to a different workspace, or the current runfile resolves to a file
     // in the bazel-out directory (bin/genfiles directory).
     if (localWorkspacePath || !runfilesPath.startsWith(USER_WORKSPACE_NAME) ||
-        realPath.includes(BIN_DIR) || realPath.includes(GEN_DIR)) {
+        realPath.startsWith(outputBase)) {
       continue;
     }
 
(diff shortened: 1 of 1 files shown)


gregmagolan:
mac_cross_compile buildkite flaked but passed on 2nd run

PR #1090, opened 2019-09-04
refactor(builtin): improve node toolchain

alexeagle:
- Support the toolchains attribute of rules like genrule, so they can get the NODE_PATH
- Expose files and a path to node, rather than whether the node binary is locally installed or came from Bazel fetching it

### DIFF
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -119,11 +119,12 @@ def _write_loader_script(ctx):
         is_executable = True,
     )
 
-def _short_path_to_manifest_path(ctx, short_path):
-    if short_path.startswith("../"):
-        return short_path[3:]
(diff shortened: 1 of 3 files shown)


internal/node/node.bzl
```diff
@@ -119,11 +119,12 @@ def _write_loader_script(ctx):
         is_executable = True,
     )
 
-def _short_path_to_manifest_path(ctx, short_path):
-    if short_path.startswith("../"):
-        return short_path[3:]
+# Avoid writing non-normalized paths (workspace/../other_workspace/path)
+def _to_manifest_path(ctx, file):
```
gregmagolan: note to self: there are a few other places that we have _short_path_to_manifest_path and a few that are named _file_to_manifest_path

alexeagle:
Thanks for the help @pcloudy

PR #1092, opened 2019-09-04
refactor: replace usages of expand_path_into_runfiles to _to_manifest_path

gregmagolan:
Now that node entry_point is a label we can do this.

This also makes _to_manifest_path naming consistent in all the rules.

Note: expand_location_into_runfiles is still used for templated_args in nodejs_binary & nodejs_test


gregmagolan:
> opt: could move _to_manifest_path to a widely-consumed util.bzl

let's follow up on that later

PR #1103, opened 2019-09-06
fix(npm_install): dynamic_deps attribute not working for scoped packages

devversion:
Currently it's not possible to specify a dynamic dependency for a scoped package. e.g. the default `dynamic_deps` value that adds `tsickle` to `@bazel/typescript` doesn't work because it never matches.

This is because the dynamic dependency logic currently checks against the `pkg._name`, which is just the trailing path segment of the package path. This is incorrect and confusing as it omits the package scope. e.g. `@bazel/typescript` => `typescript`.

### DIFF
--- a/internal/npm_install/generate_build_file.js
+++ b/internal/npm_install/generate_build_file.js
@@ -444,9 +444,9 @@ function addDynamicDependencies(pkgs, dynamic_deps = DYNAMIC_DEPS) {
   pkgs.forEach(p => {
     function match(name) {
       // Automatically include dynamic dependency on plugins of the form pkg-plugin-foo
-      if (name.startsWith(`${p._name}-plugin-`)) return true;
+      if (name.startsWith(`${p._moduleName}-plugin-`)) return true;
 
-      const value = dynamic_deps[p._name];
+      const value = dynamic_deps[p._moduleName];
       if (name === value) return true;
 
       // Support wildcard match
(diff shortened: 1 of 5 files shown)


internal/npm_install/generate_build_file.js
```diff
@@ -485,13 +485,13 @@ function findPackages(p = 'node_modules') {
...
       f => pkgs.push(parsePackage(f), ...findPackages(path.posix.join(f, 'node_modules'))));
 
-  addDynamicDependencies(pkgs);
-
   const scopes = listing.filter(f => f.startsWith('@'))
                      .map(f => path.posix.join(p, f))
                      .filter(f => isDirectory(f));
   scopes.forEach(f => pkgs.push(...findPackages(f)));
 
+  addDynamicDependencies(pkgs);
```
devversion: @alexeagle Related to that: Should we consider in a follow-up to just call this _once_  when the packages are determined? It looks like it re-iterates through all packages a couple of times right now?

alexeagle: Yes, agreed

@soldair is looking at this code right now, getting some better tests around it, converting to TS. So I think we should revisit performance mistakes like that after he's done

devversion: sounds good. thanks for clarifying that!

PR #1105, opened 2019-09-06
build: update to bazel 0.29.0 & add patch support to git_repository_under_test

gregmagolan:


alexeagle:
Again it feels like `git_repository_under_test` is a lot of maintenance burden and if we had `examples/angular` we wouldn't have a use for it...

Shouldn't we update all the way to 1.0.0-rc1 ?

gregmagolan:
> Again it feels like `git_repository_under_test` is a lot of maintenance burden and if we had `examples/angular` we wouldn't have a use for it...
> 
> Shouldn't we update all the way to 1.0.0-rc1 ?

Yes. Probably not worth keeping it around if we have examples/angular. I'll close this and we should update to 1.0.0-rc.1 (need an npm package) in another PR.

PR #1112, opened 2019-09-09
feat(builtin): detect APF node module format if ANGULAR_PACKAGE file found

gregmagolan:
Ivy angular npm package will contain ANGULAR_PACKAGE files as identifiers. See https://github.com/angular/angular/pull/31900 & https://github.com/bazelbuild/rules_nodejs/issues/927.


internal/npm_install/test/golden/@angular/core/BUILD.bazel.golden
```diff
@@ -4,6 +4,8 @@ load("@build_bazel_rules_nodejs//internal/npm_install:node_module_library.bzl",
 filegroup(
     name = "core__files",
     srcs = [
+        "//:node_modules/@angular/core/ANGULAR_PACKAGE",
```
alexeagle: this change suggests we don't really want a golden file for this package... how much of this delta do I care about? I don't think any of this is related to the fix

PR #1113, opened 2019-09-09
fix(builtin): linker test should run program as an action

alexeagle:
We were accidentally using runfiles resolution to read all user inputs, because we ran as an sh_test


internal/linker/link_node_modules.ts
```diff
@@ -0,0 +1,225 @@
...
+    // We assume here that the path is already linked to the correct target.
+    // Could add some logic that asserts it here, but we want to avoid an extra
+    // filesystem access so we should only do it under some kind of strict mode.
+    return;
+  }
+  log_verbose(`symlink( ${path} -> ${target} )`);
+  // Use junction on Windows since symlinks require elevated permissions.
+  // We only link to directories so junctions work for us.
+  fs.symlinkSync(target, path, 'junction');
+  if (VERBOSE_LOGS) {
```
gregmagolan: maybe `if (VERBOSE_LOGS || DEBUG)` so if you set DEBUG env you're checking for this? Maybe throw in that case as well since seems like failure you should find out about if you set DEBUG?

packages/typescript/src/index.from_src.bzl
```diff
@@ -32,3 +33,37 @@ def ts_library(**kwargs):
         compiler = "@build_bazel_rules_typescript//internal:tsc_wrapped_bin",
         **kwargs
     )
+
+# In rules_nodejs "builtin" package, we are creating the toolchain for building
+# tsc-wrapped and executing ts_library, so we cannot depend on them.
+# However, we still want to be able to write our tooling in TypeScript.
+# This macro lets us check in the resulting .js files, and still ensure that they are
+# compiled from the .ts by using a golden file test.
+def checked_in_ts_library(name, checked_in_js, **kwargs):
```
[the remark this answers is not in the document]
alexeagle: it's the golden_file_test below

internal/linker/link_node_modules.ts
alexeagle: I didn't want to throw because I think it's probably wrong for the program to have different behavior when you turn on VERBOSE_LOGS - seems like we want a strict mode in general which is maybe different from how much logging we do

internal/linker/link_node_modules.ts
```diff
@@ -0,0 +1,219 @@
...
+ * This replaces the need for custom module resolution logic inside the process.
+ */
+import * as fs from 'fs';
+import * as path from 'path';
+
+// Run Bazel with --define=VERBOSE_LOGS=1 to enable this logging
+const VERBOSE_LOGS = !!process.env['VERBOSE_LOGS'];
+
+function log_verbose(...m: string[]) {
+  // This is a template file so we use __filename to output the actual filename
```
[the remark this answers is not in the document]
alexeagle: was already fixed, somehow you were on a snapshot from last night

internal/linker/link_node_modules.ts
gregmagolan: I was thinking just throw if DEBUG was set not on VERBOSE_LOGS. Program can have different behaviors with DEBUG.

gregmagolan: But maybe a strict mode would make more sense instead of overloading the DEBUG semantics

PR #1117, opened 2019-09-11
chore: parallelize link creation in linker

soldair:
## PR Checklist

tests and docs have not been changed as the goal is to keep behavior the same but change execution of the linker to async.

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

this is a performance enhancement of the linker

## What is the new behavior?

the linker bin now creates links in parallel

- [ ] Yes
- [x] No


alexeagle:
Update the golden file:

            bazel run //internal/linker:linker_lib_check_compiled.accept

soldair:
this depends on the fs.promises api which is flagged as experimental in node 10.
node 10 users get a printed warning. 
`(node:82088) ExperimentalWarning: The fs.promises API is experimental`

I know we intend on moving to node 12 which should remove this warning but if we want i can use util.promisify to wrap the fs functions i need which would avoid printing this notice.

alexeagle:
We're a month away from node 12 LTS so I think we should not add any code to support node 10

internal/linker/link_node_modules.ts
```diff
@@ -189,37 +207,49 @@ export function main(args: string[], runfiles: Runfiles) {
...
 
   // Symlinks to packages need to reach back to the workspace/runfiles directory
   const workspaceRelative = path.relative('.', workspaceDir);
   const runfilesRelative = runfiles.dir ? path.relative('.', runfiles.dir) : undefined;
 
   // Now add symlinks to each of our first-party packages so they appear under the node_modules tree
-  for (const m of Object.keys(modules)) {
+  const links = []
+
+  const linkModule = async (name:string,modulePath:string)=>{
```
alexeagle: it doesn't look like the formatter ran over this.
It's configured as a pre-commit hook, can you check whether it's running for you?
(IIRC we figured it wasn't worth a CI check for formatting since it's nicely built into the commit hook)

internal/linker/link_node_modules.ts
```diff
@@ -189,37 +207,49 @@ export function main(args: string[], runfiles: Runfiles) {
...
-      target = path.join(workspaceRelative, toWorkspaceDir(modules[m]));
+      target = path.join(workspaceRelative, toWorkspaceDir(modulePath));
     }
-    symlink(target, m);
+
+    await symlink(target, name);
   }
 
+  for (const m of Object.keys(modules)) {
+    links.push(linkModule(m,modules[m]))
```
alexeagle: hmm where is our linter for missing semicolon?
It matters because clang-format reflows lines and isn't ASI aware

soldair:
i think linting is resolved.

PR #1119, opened 2019-09-11
Default rollup bundle config specifies invalid input option

devversion:
Apparently the default rollup bundle configuration specifies an invalid option called `resolveBazel`. This causes an error to be thrown as rollup complains about invalid/unknown options.

This is not an issue for the default consumers of the rollup bundle configuration as those specify `--silent`, but it will be an issue if someone uses its own rollup rule that uses this configuration. Specifically I hit this scenario on windows where the arguments for the rollup rule were too long and the "--silent" option got accidentally omitted..

### DIFF
--- a/internal/rollup/rollup.config.js
+++ b/internal/rollup/rollup.config.js
@@ -150,7 +150,6 @@ const inputs = [TMPL_inputs];
 const multipleInputs = inputs.length > 1;
 
 const config = {
-  resolveBazel,
   onwarn: (warning) => {
     // Always fail on warnings, assuming we don't know which are harmless.
     // We can add exclusions here based on warning.code, if we discover some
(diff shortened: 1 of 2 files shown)


alexeagle:
That's funny, I tried to remove this yesterday too while working on the react example. I think we should figure out the CI failure here, even though the new rollup rule landed yesterday and should be in a release soon.

devversion:
@alexeagle It looks like some tests for the rollup config just added the `resolveBazel` function to the config so that it is exported for testing 😁

I'm not sure how we can solve it as rollup expects the config to be the default export (meaning that we cannot provide other exports without adding new properties to the config --> which fails)

devversion:
@alexeagle Updated the test.

PR #1122, opened 2019-09-11
Terser on directories

alexeagle:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


packages/terser/src/index.js
```diff
@@ -1,8 +1,63 @@
...
+
+  fs.readdirSync(input).forEach(f => {
+    if (f.endsWith('.js')) {
+      const inputFile = path.join(input, path.basename(f));
+      const outputFile = path.join(output, path.basename(f));
+      // We don't want to implement a command-line parser for terser
+      // so we invoke its CLI as child processes, just altering the input/output
+      // arguments. See discussion: https://github.com/bazelbuild/rules_nodejs/issues/822
+      // TODO: under Node 12 it should use the worker threads API to saturate all local cores
+      child_process.fork(
```
Toxicable: We do this the same as for a single file when under threads, just need to modify process.argv inside the thread

alexeagle: okay, good to know, I think I'll try to land this super-simple and let you do the concurrency limiting and hook it up with worker_threads API?

Toxicable: Yeah sounds like a plan

packages/terser/src/index.js
```diff
@@ -1,8 +1,76 @@
...
+      // arguments. See discussion: https://github.com/bazelbuild/rules_nodejs/issues/822
+      // FIXME: this causes unlimited concurrency, which will definitely eat all the RAM on your
+      // machine;
+      //        we need to limit the concurrency with something like the p-limit package.
+      // TODO: under Node 12 it should use the worker threads API to saturate all local cores
+      child_process.fork(
+          // __dirname will be the path to npm_bazel_terser (Bazel's name for our src/ directory)
+          // and our node_modules are installed in the ../npm directory since they're external
+          // Note that the fork API doesn't do any module resolution.
+          path.join(__dirname, '../npm/node_modules/terser/bin/uglifyjs'),
```
Toxicable: couldn't you do something like: `const path = path.join(require.resolve('terser'), 'bin/uglifyjs')`?

alexeagle: Actually it was busted when tested from e2e anyway.

I think the simplest thing is since fork doesn't do module resolve, it should be `require.resolve(the_same_thing_as_below)`

Fixed, thx

PR #1123, opened 2019-09-12
chore: parallelize terser runs

soldair:
this is missing needed negative tests to make sure error output is useful and exit code is correct.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/terser/src/index.js
+++ b/packages/terser/src/index.js
@@ -44,8 +44,11 @@ function terserDirectory(input) {
     fs.mkdirSync(output);
   }
 
+
+  // we should allow users to configure this.
+  // tracking issue: https://github.com/bazelbuild/rules_nodejs/issues/1132
+  const maxConcurrency = (os.cpus().length - 1 || 1);
   let work = [];
-  let concurrency = (os.cpus().length - 1 || 1);
   let active = 0;
   let errors = [];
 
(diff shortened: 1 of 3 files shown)


packages/terser/src/index.js
```diff
@@ -43,20 +44,49 @@ function terserDirectory(input) {
     fs.mkdirSync(output);
   }
 
+  let work = [];
+  let concurrency = (os.cpus().length-1||1);
+  let active = 0;
+  let errors = [];
+
+  function exec([inputFile,outputFile]){
+    active++;
```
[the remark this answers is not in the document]
soldair: :100: added this in the force push. the new version uses spawn is wrapped in a promise and catches both rejections and potential synchronous exceptions from the actual spawn.

packages/terser/src/index.js
```diff
@@ -43,20 +44,49 @@ function terserDirectory(input) {
...
+      done();
+    })
+  }
+
+  function done(){
+    if(work.length) exec(work.shift()); 
+    else if(!active) {
+      if(errors.length) {
+        console.error(JSON.stringify(errors,null,'  '));
+        process.exit(1);
```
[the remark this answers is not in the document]
soldair: updated. i absolutely agree. 
It doesn't make it impossible to test it just means the smallest unit is an integration test of running the whole script. It's certainly not best practice even though one off scripts get a lot of grey area.

packages/terser/src/index.js
```diff
@@ -43,20 +44,52 @@ function terserDirectory(input) {
...
+    active++;
+    let args =
+        [require.resolve('terser/bin/uglifyjs'), inputFile, '--output', outputFile, ...residual];
+
+    spawn(process.execPath, args).then((data) => {
+      if (data.code) {
+        errors.push({file: inputFile, out: data.out + '', err: data.error + '', code: data.code})
+      }
+
+      console.log('finished: ', inputFile, '\n', data.out + '');
```
alexeagle: we should be silent on success - maybe use log_verbose if you want to leave this in so that users bug reports include this

soldair: but how will they depend on my observable outputs! (jk)
sounds good.

packages/terser/src/index.js
```diff
@@ -43,31 +57,91 @@ function terserDirectory(input) {
...
+        .then(
+            (data) => {
+              if (data.code) {
+                errors.push(inputFile)
+                // NOTE: Even though a terser process has errored we continue here to collect all of
+                // the errors. this behavior is another candidate for user configuration because
+                // there is value in stopping at the first error in some use cases.
+
+                log_error(
+                    'errored: ' + inputFile + '\n', 'OUT: ' + data.out + '\n',
```
alexeagle: more readable with a template literal (backticks plus interpolation)

packages/terser/src/index.js
```diff
@@ -43,31 +57,91 @@ function terserDirectory(input) {
...
+    const err = [];
+    const out = [];
+    let proc = child_process.spawn(cmd, args);
+
+    proc.stdout.on('data', (buf) => {
+      out.push(buf);
+    });
+    proc.stderr.on('data', (buf) => {err.push(buf)})
+    proc.on('exit', (code) => {
+      resolve({out: Buffer.concat(out), err: err.length ? Buffer.concat(err) : false, code: code});
```
alexeagle: nit: `code: code` could be shorthand `code`

alexeagle: I'm a bit surprised we never reject

soldair: failures are an expected output not an exceptional one so we just transfer the data reliably

packages/terser/src/index.js
soldair: i agree. 
Template literals get really messy if you every need to have white space for formatting that doesn't impact the output error message. Like in this case when the error message makes the line length too long and you need to format it for readability again the other way. You have to use multiple concatenated template literals for the middle ground i guess

Issue #931, opened 2019-07-22
Demonstrate how to gather code coverage for web tests


alexeagle:
Should serve to answer https://stackoverflow.com/questions/56556944/how-to-measure-coverage-when-running-typescript-tests-using-karma-and-bazel and also probably will reproduce the issues described there.

iirina:
Unfortunately there are no updates in bazel coverage land and there is no future work planned at this time. For this particular use case I can recommend using [aspects](https://docs.bazel.build/versions/master/skylark/aspects.html) to collect the transitive source files. An aspect can be attached to `karma_web_test_suite`. The aspect can collect sources and propagate through all dependencies attributes. There is [an advance example](https://docs.bazel.build/versions/master/skylark/aspects.html#advanced-example) which uses an aspect to count the files in the transitive path. Let me know if this is reasonable or if there are further questions.

iirina:
Thanks @alexeagle. The test runner creates `COVERAGE_OUTPUT_FILE` as an output for coverage invocations and `COVERAGE_DIR` for coverage generated files, but the variables are not exposed to Skylark. We can look into what it takes to expose them.

Toxicable:
We'd also like to apply the same coverage stratergy for nodejs test coverage, only it uses a slightly different instrumentation method, but transforms are easy.

@iirina I don't know if we need those variables exposed to skylark, although im sure we'll need them there eventually, but just having them avilable to the program being instrumented I think would get us there.
In my testing `process.env.COVERAGE_DIR` (in a JS program) was never defiend when running under `bazel coverage` - is there some flag I need to enable?

If you could take a quick look over: https://github.com/bazelbuild/rules_nodejs/pull/1135/files  
It's roughly how we'd collect coverage in nodejs and also has a script for processing, merging and writing reports

sgammon:
@alexeagle / @iirina with the release of bazel 2.1.0 and the associated coverage fixes, could i give this issue a bump?

just like the poster on SO says, i've got it to a point where i can run coverage and generate lcov reports. i just need to figure out where to put the report so i can trigger it being merged with the combinedreport

Toxicable:
@sgammon I've integrated `bazel coverage` into rules_nodejs and @bazel/jasmine in #1135 
Take a look there, specifically at coverage_runner.js on where and how to handle the coverage data.

sgammon:
@Toxicable thank you!! i'll take a look

Kritz1994:
Hi,
[...]

Issue #867, opened 2019-06-15
source map support in coverage reports


Toxicable:
# 🚀 feature request

### Relevant Rules
`jasmine_node_test`

### Description
We should still use `v8-coverage` to collect coverage reports without having to fork the process.
However, c8 and transitively v8-to-istanbul provide better reporting due to having source map support and more up to date istanbul deps.

Toxicable:
This however is blocked on #823 
Since the source map lib needs to resolve the souce map back to the source file from the executed file.

Also blocked on and https://github.com/istanbuljs/v8-to-istanbul/issues/31

Toxicable:
Aside from the blocking issues it's minor changes needed on our side so just going to  document that here for now:

```
        const Report = require('c8/lib/report');
        fs.mkdirpSync(path.join(covDir, 'tmp'));
        // we have to write it in a top level result object since thats what the lib expecets
        fs.writeFileSync(path.join(covDir, 'tmp', 'coverage-final.json'), JSON.stringify({result: sourceCoverge}));

        const report = Report({
          reportsDirectory: covDir,
          tempDirectory: path.join(covDir, 'tmp'),
          reporter: ['text', 'text-summary', 'html'],
          exclude: ['**/node_modules/**'],
        });
```

Toxicable:
I think the best way to handle this would be to add a `coverage` flag to `nodejs_binary` which would then internally set the `NODE_V8_COVERAGE` to a dir where the collection will be written.
We need to do this at a `nodejs_binary` level so that we don't have to use the `inspector` api and so that we don't have to use a fork later on.

We'll still need to handle the reporting at the end of `jasmine_node_test`.

However, if we switched to `c8` for reporting and transitively `v8-to-istanbul` we'd need the source maps situation to be fixed, since they are unable to resolve our source files in the current setup.
As mentioned above.
Which is the same issue here: https://github.com/bazelbuild/rules_nodejs/issues/823

thesayyn:
Hello @Toxicable  is there any update on this

Toxicable:
@thesayyn Yeah so I know what needs to be done to make this change
However due to #823 the result will be broken since our source maps don't map back to sources correctly and after talking with @alexeagle I don't know if there is actually a solution to that problem.

So untill we figure out source maps properly, this is blocked

sjoerdvisscher:
Hi, @Toxicable, the issues you point to here are now closed. Does this mean this issue has a chance to progress?

[...]

Issue #638, opened 2019-03-29
jasmine_node_test bazel coverage support 


Toxicable:
# 🚀 feature request

### Relevant Rules
jasmine_node_test

### Describe the solution you'd like
Add support for the `bazel coverage` command

We currently support coverage via the `coverage = True` flag which outputs a text summary of the coverage collected by that rule. https://github.com/bazelbuild/rules_nodejs/commit/9abc78ffb596f1a78d0671cd51ec9b398dbc3f67
This however does not merge coverage between targets

The solution is to use the `bazel coverage` command to aggregate and merge coverage reports from the various targets throughout a repository.

Currently sort of blocked by https://github.com/bazelbuild/bazel/issues/5885
There is no general way to utilise `bazel coverage` but it is possible to hack around it.

### Describe alternatives you've considered
This is the only option for coverage under jasmine throughout a repository

Toxicable:
I just found `COVERAGE_OUTPUT_FILE` which is an env var set by bazel that that tells you were to write coverage files

Toxicable:
Another related link.
Looks like each build rule and return a provider for coverage to indicate which files should be instrumented.
https://docs.bazel.build/versions/0.23.0/skylark/rules.html#code-coverage-instrumentation

We should use this instead of parsing through the manifest file to know what files we should collect coverage for

PR #1135, opened 2019-09-13
move cov to nodejs test

Toxicable:
This moves coverage support from jasmine_node_test to nodejs_binary, meaning any test can collect coverage.
It also 
This does not report any coverage to the user, it only collects coverage and stores it in V8's format somewhere in a tmpdir that's not reasonable accessible.
This also allows support for worker threads and process coverage collection.

The next step from here would be to do more integration work with `bazel coverage` but this lays down the ground work for collecting coverage correctly .

## PR Checklist

## PR Type

Code coverage can only be collected via the jasmine_node_test rule

Issue Number: N/A

## What is the new behavior?
Code coverage can be collected on run rule using nodejs_test

- [x] Yes

removes the `coverage` attribute, instead relying on the command being invoked with `bazel coverage`.

## Other information

fixes https://github.com/bazelbuild/rules_nodejs/issues/1218


Toxicable:
replaced with https://github.com/bazelbuild/rules_nodejs/pull/874

internal/node/node.bzl
```diff
@@ -155,6 +155,10 @@ def _nodejs_binary_impl(ctx):
         if k in ctx.var.keys():
             env_vars += "export %s=\"%s\"\n" % (k, ctx.var[k])
 
+    if ctx.attr.coverage:
+        # TODO: use proper path for tmpdir
+        env_vars += "export NODE_V8_COVERAGE=/tmp/$RANDOM\n"
```
alexeagle: coverage only applies to test right? bazel gives you `$TEST_TMPDIR` but you should wait until you're inside the test action to try to read the environment

internal/node/node.bzl
```diff
@@ -259,6 +265,11 @@ _NODEJS_EXECUTABLE_ATTRS = {
         Note, this can lead to different outputs produced by this rule.""",
         default = [],
     ),
+    "coverage": attr.bool(
+        doc = """Weather to collect code coverage information
```
alexeagle: feels like something that should be requested for an execution, not configured on the target. did we ever find a way to sense whether we are running under `bazel coverage`?

packages/jasmine/src/jasmine_runner.js
```diff
@@ -86,17 +80,6 @@ function main(args) {
                        // Filter out files from node_modules
                        .filter(f => !IS_NODE_MODULE.test(f))
```
alexeagle: what about this feature where we only gathered coverage on production code rather than on tests? do we not need that anymore?

internal/node/node.bzl
Toxicable: While only doing coverage under `bazel test` makes sense, we often run out test targets with `bazel run` when we're developing them, simply becuase it's the same command we use to run our nodejs targets.
However, if there's no way to get access to a tmpdir under `bazel run` then we'll have to do that, we'll have to ensure this is documented

Toxicable: Also note even with the `bazel coverage` integration, we'd still need this set since the v8 coverage format is only an intermediate format

internal/node/node.bzl
Toxicable: Agreed, there are some environment vars that are set under `bazel coverage`, looking through the links here https://github.com/bazelbuild/rules_nodejs/issues/638

packages/jasmine/src/jasmine_runner.js
Toxicable: ah yes, missed that in, added it back in the coverage processor

internal/node/node.bzl
Toxicable: since we're only doing coverage with `bazel coverage` as per the below comment, I changed this to be `        env_vars += "export NODE_V8_COVERAGE=$TEST_TMPDIR/$RANDOM\n"` - not sure if $RANDOM is correct but it's close if not

Toxicable:
@soldair Should be good for a review now.
I've reduced the scope on this to only be the brekaing changes, ill make another PR that adds in the coverage processing script.

Toxicable:
@alexeagle On this one do you want at least feature parity with `jasmine_node_test#coverage=True`
Before we merge this?

One path we could take:
- keep  `v8-coverage` to do that but that means we don't fix #1218
- remove or deprecate the `coverage` attribute
- enabled the flag `--coverage` for `jasmine_runner.js` to be based off `ctx.configuration.coverage_enabled` instead of the attribute.

This only solve the issue about the breaking change of removing (or maybe deprecating) the `coverage` attribute.
Further work needs to be done to remove `v8-coverage` explained in detail more here: https://github.com/bazelbuild/rules_nodejs/issues/931#issuecomment-539666196
Also relates here: https://github.com/bazelbuild/rules_nodejs/issues/867

TLDR: 
- need Skylark support for aggrigating coverage reports accross targets
- need source map support to swap `v8-coverage` for `c8` or even `v8-to-istanbul`

Toxicable:
Took another look at this today
I was looking further at using `bazel coverage` for coverage aggregation.

Main issue is that `COVERAGE_DIR` isn't defiend when running `bazel coverage`
I think if that's defined then we can set `NODE_V8_COVERAGE=$COVERAGE_DIR`
Doing this would cause v8 to write it's coverage to this dir, then if we defined out `LCOV_MERGE`, it could collect all this v8 format coverage, convert it and aggregate it.


Here's a bunch of links to related topics/code that I found which I'll probably look at again next time.

https://github.com/bazelbuild/bazel/blob/f708b0f847ba6c1d2ccfd33c283a2dd81e0803e6/site/docs/skylark/rules.md#code-coverage-instrumentation
https://github.com/dropbox/dbx_build_tools/blob/3f68abae62135488e135af145a0b48e9030488ff/build_tools/py/py.bzl#L841
https://github.com/bazelbuild/bazel/issues/10660
https://docs.google.com/document/d/1WRQTXQBV-m1_HDkAPvRSfGIrDW4pkI7ATPWtNdC8k9E/edit#heading=h.pt2fqnujvoh2
https://groups.google.com/forum/#!topic/bazel-discuss/LC97zBrMG84
https://github.com/bazelbuild/bazel/blob/master/tools/test/collect_cc_coverage.sh
https://github.com/bazelbuild/bazel/blob/master/tools/test/collect_coverage.sh
https://github.com/bazelbuild/bazel/issues/6293
https://github.com/bazelbuild/bazel/issues/8670
https://github.com/bazelbuild/bazel/issues/5885

Toxicable:
Progress update; TLDR: closer but still not quite there

Along with a huge help from @armooo 
We were able to somewhat get `bazel coverage` working.

In it's current state:
Bazel runs the program with https://github.com/bazelbuild/bazel/blob/master/tools/test/collect_coverage.sh
Which gives us the env vars we need to collect coverage from targets.
The test target runs as expected and v8 writes it's coverage information.
Then collect_coverage tries to run out lcov_merger but runfiles fails and it exits without anything helpful.

I've added a new binary: `lcov_merger` which will be responsible for colelcting v8 coverage from a target, merging and converting to lcov for bazel to use later on.

Assuming we can solve the issues above we should be able to use `--combined_report=lcov` to produce a single coverage report at `bazel-out/_coverage/_coverage_report.dat`

We need to use the `coverage_common.instrumented_files_info` provider which i've added to `nodejs_binary` and `ts_library`, it will need to be added to any rule that provides source files which need to be instrumented

Once we have `lcov_merger` running we may run into some source map issues.

If you want to test it out yourself run `npx bazel coverage packages/jasmine/test:coverage_test` on this branch

Toxicable:
This looks like the issue causing the runfiles to fail https://github.com/bazelbuild/bazel/issues/4033

Toxicable:
I've changed tac on tthis one a bit, i've moved the logic into the jasmine runner for combining and converting to lcov.
It's less portable doing it this way, since we'd need to implement it for each runner, but at least this provides a example.

However; it requires some hacks to c8 and v8-to-istanbul, around source maps, to get it working i've modified them manually.
So ill need to work on fixing those issues and maybe upstreaming something there.

Aside from that this now work with `bazel coverage`, however all the outpout from `collect_coverage.sh` is pretty annoying, might look seeing if we can turn off the `set -x` that it does.

Next piece is to figure out how with bazel we can get it to extract the coverage data out, it looked like we can use `--combined-report=lcov` but that resulted in an error 
```
Internal error thrown during build. Printing stack trace: java.lang.NullPointerException
        at com.google.devtools.build.lib.bazel.coverage.BazelCoverageReportModule$1.getArgs(BazelCoverageReportModule.java:115)
        at com.google.devtools.build.lib.bazel.coverage.CoverageReportActionBuilder.generateCoverageReportAction(CoverageReportActionBuilder.java:246)
        at com.google.devtools.build.lib.bazel.coverage.CoverageReportActionBuilder.createCoverageActionsWrapper(CoverageReportActionBuilder.java:193)
        at com.google.devtools.build.lib.bazel.coverage.BazelCoverageReportModule$1.createCoverageReportActionsWrapper(BazelCoverageReportModule.java:97)
```

Toxicable:
Managed to work around the runfiles issues I mentioned above by putting the code into @bazel/jasmine, we can refactor it back into lcov_merger later on once we resolve those runfiles issues.
Just waiting on this to be published to npm https://github.com/istanbuljs/v8-to-istanbul/releases/tag/v4.1.0

Otherwise it's ready for review

packages/jasmine/src/coverage_runner.js
```diff
@@ -0,0 +1,99 @@
...
+
+  const exitCode = result.status;
+
+  // 0 indicates success
+  // so if it's anything tthen exit here since bazel only collects coverage on success
+  if (exitCode !== 0) {
+    process.exit(exitCode);
+  }
+
+  if (!coverageEnabled) {
```
gregmagolan: Could shorten this to just
```
if (exitCode !== 0 || !coverageEnabled) {
    if (result.error) {
      console.error(result.error);
    }
   process.exit(exitCode || 1);
}
```
The `|| 1` is to cover the case of `result.status === null` if the subprocess terminates due to signal (https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options).

Logging out result.error if it exists is helpful as well as otherwise the failure is silent.

gregmagolan: `coverageEnabled` is not defined anywhere?

packages/jasmine/src/coverage_runner.js
```diff
@@ -0,0 +1,99 @@
...
+                       // we can only cover js files
+                       .filter(f => ['.js', '.jsx', '.ts', '.tsx'].includes(path.extname(f)))
+                       // the actual coverage written by v8 will only inclue urls to .js
+                       // so map inputs to their outputs
+                       .map(f => {
+                         const p = path.parse(f);
+                         return path.format({...p, base: null, ext: '.js'});
+                       });
+
+  // onyl require in c8 when we're actually going to do some coverage
```
gregmagolan: "onyl" => "only"

packages/jasmine/src/coverage_runner.js
```diff
@@ -0,0 +1,99 @@
...
+  // so we give it a dir that it can write to
+  // later on we'll move and rename it into output_file as bazel expects
+  const tmpdir = process.env['TEST_TMPDIR'];
+  const c8OutputDir = path.join(tmpdir, crypto.randomBytes(4).toString('hex'));
+  fs.mkdirSync(c8OutputDir);
+
+  const includes = instrumentedSourceFiles
+                       // we can only cover js files
+                       .filter(f => ['.js', '.jsx', '.ts', '.tsx'].includes(path.extname(f)))
+                       // the actual coverage written by v8 will only inclue urls to .js
```
gregmagolan: "inclue" => "include"

packages/jasmine/src/coverage_runner.js
```diff
@@ -0,0 +1,99 @@
...
+  // c8 will name the output report file lcov.info
+  // so we give it a dir that it can write to
+  // later on we'll move and rename it into output_file as bazel expects
+  const tmpdir = process.env['TEST_TMPDIR'];
+  const c8OutputDir = path.join(tmpdir, crypto.randomBytes(4).toString('hex'));
+  fs.mkdirSync(c8OutputDir);
+
+  const includes = instrumentedSourceFiles
+                       // we can only cover js files
+                       .filter(f => ['.js', '.jsx', '.ts', '.tsx'].includes(path.extname(f)))
```
gregmagolan: `.mjs`  as well?

Seems like this in an ideal world this list would be configurable somewhere.

packages/jasmine/src/coverage_runner.js
```diff
@@ -0,0 +1,99 @@
...
+  await new c8
+      .Report({
+        include: includes,
+        exclude: [],
+        reportsDirectory: c8OutputDir,
+        // tempDirectory as actually the dir that c8 will read from for the v8 json files
+        tempDirectory: coverageDir,
+        resolve: '',
+        // TODO: maybe add an attribute to allow more reporters
+        // or maybe an env var?
```
gregmagolan: Would more reporters be useful since we're only interested in the lcov data for bazel coverage?

packages/jasmine/src/jasmine_node_test.bzl
```diff
@@ -90,7 +86,7 @@ def jasmine_node_test(
     nodejs_test(
         name = name,
         data = all_data,
-        entry_point = jasmine_entry_point,
+        entry_point = coverage_entry_point,
```
gregmagolan: Assuming we can't tell in the macro context that coverage is enabled to select `jasmine_entry_point` if its not?

packages/jasmine/src/jasmine_node_test.bzl
```diff
@@ -29,9 +29,9 @@ def jasmine_node_test(
         expected_exit_code = 0,
         tags = [],
         config_file = None,
-        coverage = False,
         jasmine = "@npm//@bazel/jasmine",
         jasmine_entry_point = "@npm//:node_modules/@bazel/jasmine/jasmine_runner.js",
+        coverage_entry_point = "@npm//:node_modules/@bazel/jasmine/coverage_runner.js",
         **kwargs):
     """Runs tests in NodeJS using the Jasmine test runner.
```
gregmagolan: For an easier upgrade path should add something like:
```
if kwargs.pop("coverage", False):
   fail("comment on how use the new coverage method")

File: packages/typescript/src/internal/build_defs.bzl [hunk 11]
```diff
@@ -284,7 +284,11 @@ def _ts_library_impl(ctx):
         ),
         # TODO: Add remaining shared JS provider from design doc
         # (JSModuleInfo) and remove legacy "typescript" provider
-        # once it is no longer needed.
+        # once it is no longer needed.,
```
gregmagolan: nit: stray comma in here

packages/typescript/src/internal/build_defs.bzl
```diff
@@ -284,7 +284,11 @@ def _ts_library_impl(ctx):
...
+
+        # indicates that the files listed in this invocation under srcs should be instrumented by coverage
+        # also transitively collect from deps
+        coverage_common.instrumented_files_info(ctx, source_attributes = ["srcs"], dependency_attributes = ["deps"]),
```
gregmagolan: Will need this in ng_module and any other rules that transform source files to `.js` looks like?

packages/jasmine/src/coverage_runner.js
```diff
@@ -0,0 +1,99 @@
...
+const path = require('path');
+const crypto = require('crypto');
+const childProcess = require('child_process');
+
+/**
+ * This is designed to collect the coverage of one target, since in nodejs
+ * and using NODE_V8_COVERAGE it may produce more than one coverage file, however bazel expects
+ * there to be only one lcov file. So this collects up the v8 coverage json's merges them and
+ * converts them to lcov for bazel to pick up later.
+ * TODO: The functionality in this script could be put into _lcov_merger then it would apply to
```
gregmagolan: 😎Seems like we should land this functionality in the _lcov_merger binary with a nodejs_binary if possible. The trouble I see there is that that nodejs_binary would have a `@npm//c8` dependency then which would be a dep if coverage is enabled or not. Is that the blocker for that?

packages/jasmine/src/coverage_runner.js
Toxicable: I've found it helpful to have either `text` it `text-summary` to instantly report on the unit testing suit you're working on so you can aim to cover more paths.
From what I understand going the bazel route will be pretty slow and wont work with ibazel

packages/jasmine/src/coverage_runner.js
Toxicable: You can make the dep on c8 dynamic like i've done with coverage_runner by just using `require` isntead of import. Then then the user gets an error they'd have to install c8 and provide it to the nodejs_binary, the same as how the current implementation works

packages/typescript/src/internal/build_defs.bzl
Toxicable: Yup! any rule that provides fules which should be instrumented will need this.
I don't know how it'll work for pure JS files, any ideas there?

packages/jasmine/src/coverage_runner.js
Toxicable: but I agree that nodejs_test is the right place to put this. I'd need your help to figure out the runfiles issues though

packages/jasmine/src/coverage_runner.js
Toxicable: I must have refactored that out by accident when moving from lcov_merger to coverage_runner. fixed!

packages/jasmine/src/coverage_runner.js
Toxicable: Agreed, this is defined in serveral different places now, would be great to have a single definition

Toxicable: Hmmm in the code below we convert all ext's to .js but .mjs is a valid ext, will need to make that more complex

Toxicable: im not actually sure how to handle this now, I've added a few comments in the code here but it's really this part:
```
// TODO: how do we know what source files produce .mjs?
```

packages/jasmine/src/jasmine_node_test.bzl
Toxicable: we can only ttell that we're working under coverage with either an `env` var or access to `ctx` neither of which we have in this macro, not sure how else to pick what entry point to use here

packages/jasmine/src/coverage_runner.js
Toxicable: Although swapped to `process.exit(exitCode == null ? 1 : exitCode)` since 0 is falsey :P

sgammon:
@Toxicable is there a way to test this locally?

Toxicable:
@sgammon yeah sure
pull down the repository
do a `yarn install` then run `npx bazelisk coverage packages/jasmine/test:coverage`
You'll then see some output from bazel that points to the resulting lcov file

packages/jasmine/src/coverage_runner.js
Toxicable: I've added back the lcov_merger_js and added some debug info if you wanna take a look at the runfiels issues i've been having

sgammon:
@Toxicable did i misunderstand that unified coverage works out of the box with `jasmine_node_test` as is? or does this PR add it, or is that future work?

sgammon:
@Toxicable forgive me, i'm just trying to figure out where this fits i

Toxicable:
This PR setups up @bazel/jasmine (maybe more generically nodejs_test) to produce an lcov coverage report for bazel on a per target basis.

With this information you can use bazel's `--combine_report` flag to combine the reports from multiple targets into a single one that spans your repository.
The output format of this will be lcov, so some work would be needed to convert that into something more usable if you tooling dosen't already support lcov.
However, last time I tried the `--combine_report` flag it threw a runttime error, looks like an upstream bug in bazel.

There is definately more work to be done here, but this lays out the patterns for how we should be handling coverage with bazel.
Areas that still need to be addressed with this merged:
- workspace source file source map support support
- per target feedback under ibazel
- lcov -> some other format of reports from combine_report guidance or tooling

sgammon:
@Toxicable i've got `--combined_report=lcov` working in my environment, after the changes in Bazel `2.1.0` (i have to disable a few things, like Elemental2, which still seem to interfere, but it works).

`lcov` output is fine, in fact it's preferable in my case since it's way more merge-friendly than other formats.

it doesn't seem to work out of the box for `--combined_report` in my environment when it comes to `rules_nodejs`, but that could certainly be me. is there anything i have to do, theoretically, besides pass `coverage=True` (before this PR), or just run the tests (after this PR)?

common.bazelrc
```diff
@@ -76,6 +76,9 @@ common --experimental_allow_incremental_repository_updates
 build --incompatible_strict_action_env
 run --incompatible_strict_action_env
 
+# when running `bazel coverage` ensure that the test targets are instrumented
```
gregmagolan: can you expand on this? what changes/breaks if it is missing? is this something users will need to add downstream?

packages/typescript/src/internal/build_defs.bzl
gregmagolan: That's a good point. js_library may be useful in this regard. Also, in my pkg_npm PR I'm able to remove the `module_from_src` which was previously required by the linker as the rule can determine if the srcs are source files or generated files and ensure they are all uniform. Provides a lightweight linkable package with `module_name`.... although`module_name` is a bad name and should be renamed to `pacakge_name` as it is named in `pkg_npm` in that PR.

gregmagolan: https://github.com/bazelbuild/rules_nodejs/pull/1715/files#diff-850cec1f6d2867d724158a5607f79fee

gregmagolan:
This is looking good. I tried it out your test target
```
bazel coverage packages/jasmine/test:coverage_test
```
and got coverage data:
```
rules_nodejs-2 (nodejs_test_coverage) $ cat /private/var/tmp/_bazel_greg/2c05a014295d8bb5c9c367df13018b3c/execroot/build_bazel_rules_nodejs/bazel-out/darwin-fastbuild/testlogs/packages/jasmine/test/coverage_test/coverage.datTN:
SF:../../../packages/jasmine/test/coverage_source.ts
FN:1,isString
FNF:1
FNH:1
FNDA:2,isString
DA:1,2
DA:2,2
DA:3,1
DA:4,1
DA:5,1
DA:6,1
DA:7,2
LF:7
LH:7
BRDA:7,0,0,1
BRDA:7,1,0,1
BRDA:7,2,0,0
BRDA:7,3,0,0
BRDA:7,4,0,1
BRDA:1,5,0,2
BRDA:2,6,0,1
BRF:7
BRH:5
end_of_record
```
It also looks like `--combined_report=lcov` works with bazel 2.1.0 as with:
```
bazel coverage packages/jasmine/test:coverage_test --combined_report=lcov
```
I see
```
INFO: LCOV coverage report is located at /private/var/tmp/_bazel_greg/2c05a014295d8bb5c9c367df13018b3c/execroot/build_bazel_rules_nodejs/bazel-out/_coverage/_coverage_report.dat
 and execpath is bazel-out/_coverage/_coverage_report.dat
```
which contains
```
rules_nodejs-2 (nodejs_test_coverage) $ cat /private/var/tmp/_bazel_greg/2c05a014295d8bb5c9c367df13018b3c/execroot/build_bazel_rules_nodejs/bazel-out/_coverage/_coverage_report.dat
SF:../../../packages/jasmine/test/coverage_source.ts
FN:1,isString
FNDA:2,isString
FNF:1
FNH:1
BRDA:1,5,0,-
BRDA:2,6,0,-
BRDA:7,0,0,-
BRF:3
BRH:0
DA:1,2
DA:2,2
DA:3,1
DA:4,1
DA:5,1
DA:6,1
DA:7,2
LH:7
LF:7
end_of_record
```

packages/typescript/src/internal/build_defs.bzl
gregmagolan: I removed this line to see how it would affect the coverage output and it was not affected. Only removing the `instrumented_files_info ` from `node.bzl` broke the coverage with
```
test/coverage_test.sh: line 98: COVERAGE_DIR: unbound variable
```
assuming on the line generated by
```
env_vars += "export NODE_V8_COVERAGE=$COVERAGE_DIR\n"
```

gregmagolan:
On further thought, the logic in coverage_runner that first runs jasmine and then outputs the coverage report with `c8` is not specific to jasmine_node_test at all and could be put into a node script that is always run by the node_test launcher if `process.env.COVERAGE_DIR` is set. Lets discuss at the sync meeting.

gregmagolan:
Ok. This is getting closer now. Moved the lcov merge to the launcher so it is run on all nodejs_binary targets that have coverage turned on.

A few things left from my perspective:

1) Understand how `coverage_common.instrumented_files_info` works and why it seems it is not needed in `ts_library`

2) Figure out what the file extension filter & mappings should be in the lcov merge script

3) Make it easy to configure c8 reporters so you can get a text output if desired. Either through an attribute on nodejs_binary, an env var, or some other method

Toxicable:
1. Ill do some investigation around with this one, see if I can get some better answers
2. Mappings: I'm not sure how to solve this one.
Maybe the solution here is to just assume everything transforms files into `*js`?
Or maybe we have a provider that tells us what maps to what?
For filtering: we could filter ext's down with `instrumented_files_info` since it also has a ext property we could leaverage, I don't think this is a great idea though since sometimes `.html` is valid
Maybe this is a blacklist, rather than whitelist? and we just ignore the ones that break it
3. For this one I vote an env var then you can use it like `npx ibazel run :test --define=C8_REPORTERS=text
maybe think about adding an attribute in the future

internal/node/launcher.sh
```diff
@@ -222,13 +220,40 @@ export BAZEL_PATCH_ROOT=$(dirname $PWD)
 # fail before we could assert that we expected that failure.
 readonly EXPECTED_EXIT_CODE="TEMPLATED_expected_exit_code"
 if [ "${EXPECTED_EXIT_CODE}" -eq "0" ]; then
-  # Replace the current process (bash) with a node process.
-  # This means that stdin, stdout, signals, etc will be transparently
-  # handled by the node process.
-  # If we had merely forked a child process here, we'd be responsible
-  # for forwarding those OS interactions.
-  exec "${node}" "${LAUNCHER_NODE_OPTIONS[@]:-}" "${USER_NODE_OPTIONS[@]:-}" "${MAIN}" ${ARGS[@]+"${ARGS[@]}"}
```
Toxicable: by not using exec do we still handle "stdin, stdout, signals, etc" still?

Toxicable: also if we're ditching `exec` could we compbine this with the one belonw, Line 260

gregmagolan: Yes. I think combining the two is reasonable. I don't think there is much value in running coverage when expected exit code != 0 but maybe someone will find that useful and it would simplify the code

gregmagolan: Good point re: the exec. We should look back at why the exec was there (trace back to the commit where it was added and see what it was fixing as I don't recall) and if there is anything special needed to pipe the io streams & signals. The non-exec non-zero exit code path had no special handling but that path is likely not used at all outside of this repo.

Toxicable: This looks like where it was introduced https://github.com/bazelbuild/rules_nodejs/commit/c124496339f26e1d4d5bda1fe827220222d9b177

Toxicable: Looks like w'ell need to go back to the trap approah to handle SIGTERM, SIGINT, SIGQUIT. will throw that together and see what it looks like with that other refactor in

gregmagolan: Thanks @Toxicable . Was just looking at that commit and came to the same conclusion.

common.bazelrc
Toxicable: added a better description, what do ya think?
```
# When running `bazel coverage` --instrument_test_targets needs to be set in order to 
# collect coverage information from test targets
coverage --instrument_test_targets
` ``

File: internal/node/launcher.sh (hunk 15 above)
Comment (Toxicable) [reply internal/node/launcher.sh]: Modified it based off your changes and the ones in that MR, please check it well, my bash skills are not amazing

**Comment by Toxicable** [comment]
Just pushed up a few changes, please take another review @gregmagolan 

1. I tried looking at the implementation of coverage_common but it's defiend in Java here https://github.com/bazelbuild/bazel/blob/7891d3b4be8549f762f32308a4085f802e7608e5/src/main/java/com/google/devtools/build/lib/skylarkbuildapi/test/CoverageCommonApi.java
But my wild guess is that we don't need to specify it on the `ts_library` since all the files are accessible through the `data` attribute?
I think we should leave it off since it appears to work, but if we find a case where we need to add it, we can add it back then

2. My feeling here is that until we have a better solution we just map everything to .js, once we figure out what use cases this dosen't satisfy we could look at options here.  

3. I think leaving this as is for now is going to be fine, we could revisit the API for that on a follow up.

**Comment by Toxicable** [comment]
@googlebot I consent.

**Comment by gregmagolan** [comment]
@Toxicable Can you plz rebase this to head? We can land this now as master is on the 2.0.0-next.0 track now.

**Comment by Toxicable** [comment]
@gregmagolan done!

**Comment by Toxicable** [comment]
CI failure looks like we broke something, not sure
```
packages/typescript/test/lit_plugin/bad.ts:9:4 - error TS2322: [lit] Unknown tag <unknown-element>. Did you mean <lit-element>?
  Check that you've imported the element, and that it's declared on the HTMLElementTagNameMap. If it can't be imported, consider adding it to the 'globalTags' plugin configuration or disabling the 'no-unknown-tag' rule.

9   <unknown-element></unknown-element>
     ~~~~~~~~~~~~~~~
//packages/typescript/test/lit_plugin:lit_plugin:1:1 - error TS2322: Expected a compilation error matching "TS2322: \\[lit\\] Unknown tag \"unknown-element\". Did you mean 'lit-element'?"
```

gregmagolan:
Turns out that the failures you saw were from upgrading npm packages such as `terser` and `ts-lit-plugin` and `rollup*` to newer versions. Behavior of these npm packages changed slightly which causes some tests to fail. You must have turned your yarn.lock and regenerated it from scratch and the packages were marked `^x.x.x`.

@alexeagle We should turn on renovate for npm packages as that would catch npm package upgrade failures. I tried configuring renovate to make npm update PRs a little while ago but still not working.

gregmagolan:
In any case, this should go green now.

PR #1139, opened 2019-09-16
Add npm_package_bin rule

alexeagle:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


e2e/stylus/BUILD.bazel
```diff
@@ -1,9 +1,21 @@
...
+    srcs = ["test.styl"],
+    outs = [
+        "test.css",
+        "test.css.map",
+    ],
+    args = [
+        "test.styl",
+        "--out",
+        # Output paths must use $(location) since Bazel puts them in a platform-dependent output directory
+        "$(location test.css)",
```
gregmagolan: As a user, can I use a relative path here such as `../test.css`? Not that it makes sense in this example but in general is that possible?

gregmagolan: You linked the docs which answers the question :) https://docs.bazel.build/versions/master/be/general.html#genrule.srcs

internal/node/npm_package_bin.bzl
```diff
@@ -0,0 +1,68 @@
...
+        package: an npm package whose binary to run, like "terser". Assumes your node_modules are installed in a workspace called "npm"
+        package_bin: the "bin" entry from `package` that should be run. By default package_bin is the same string as `package`
+        tool: a label for a binary to run, like `@npm//terser/bin:terser`. This is the longer form of package/package_bin
+    """
+    if not tool:
+        if not package:
+            fail("You must supply either the tool or package attribute")
+        if not package_bin:
+            package_bin = package
+        tool = "@npm//%s/bin:%s" % (package, package_bin)
```
gregmagolan: Seems like this could allow a bit more flexibility by allow user to override the `@npm` workspace without specifying the whole tool? But maybe having the tool override is sufficient

internal/npm_install/generate_build_file.js
```diff
@@ -1024,6 +1042,26 @@ nodejs_binary(
...
+    result = `load("@build_bazel_rules_nodejs//:defs.bzl", "npm_package_bin")
+
+`;
+  }
+  for (const name of executables.keys()) {
+    result = `${result}
+
+# Generated helper macro to call ${name}
+def ${name}(**kwargs):
+    npm_package_bin(package = "${pkg._dir}", package_bin = "${name}", **kwargs)
```
gregmagolan: I think specific `tool` here instead of `package` and `package_bin` as the `@npm` workspace may not be `@npm` for this npm_install/yarn_install

packages/less/BUILD.bazel
```diff
@@ -33,6 +33,8 @@ npm_package(
     srcs = [
         "@npm_bazel_less//:package_contents",
     ],
+    # less package is deprecated
+    tags = ["do-not-publish"],
```
gregmagolan: Update release script to key off of this tag?

gregmagolan:
This is really nice and clean. Nice work! Naming makes sense to me.

I think another PR we can have inform npm_install & yarn_install via the @bazel/typescript  package.json how to add a ts_library macro to index.bzl with the appropriate @npm_foo override as discussed here: https://github.com/bazelbuild/rules_nodejs/issues/1149#issuecomment-532770498. Ultimately it would be really nice if users can load all npm delivered bazel rules from `@npm_foo//@bazel/bar:index.bzl` or `@npm_foo//fum:index.bzl`.

e2e/stylus/BUILD.bazel
alexeagle: yeah, this will be tough for users I think, because if you just write "test.css" it will write to the input folder. if you're lucky you get an error about that folder being readonly in the sandbox, more likely it writes it and you get an error that bazel says the declared output wasn't produced.
But this is standard genrule behavior so not strictly our job to improve that?

internal/node/npm_package_bin.bzl
alexeagle: hmm, we could easily have another attribute for the package workspace. but that does make the API surface of this rule bigger just to preserve some syntax sugar, which seems probably wrong to me.

internal/npm_install/generate_build_file.js
alexeagle: yes good call

packages/less/BUILD.bazel
alexeagle: it already does, that's how we avoided publishing `@bazel/rollup` and `@bazel/terser` to this point.

PR #1140, opened 2019-09-16
build: update to latest rules_webtesting commit to pickup windows fixes

gregmagolan:
Updates to latest rules_webtesting commit to pick up two Windows fixes.

1) https://github.com/bazelbuild/rules_webtesting/commit/e082ffd85190a9573b2367ef2ee7dadc87dc21c7
2) https://github.com/bazelbuild/rules_webtesting/commit/e7e838126b64c9a51dc06078e2f491b9a07b144e

We are using a patch to pull in (1) at the moment so this PR also removes the patch.

Not having (2) is blocking the rules_nodejs update in angular/angular: https://github.com/angular/angular/pull/32151

Other options are:

1) wait for rules_webtesting cuts the next release so we can update to the released version
2) explicitly pull this rules_webtesting commit into https://github.com/angular/angular/pull/32151 for now so it can land

### DIFF
--- a/rules_webtesting.patch
+++ b/rules_webtesting.patch
@@ -11,3 +11,161 @@ index 384d018..2cbbb4a 100644
          },
      }),
      visibility = ["//browsers:__subpackages__"],
+-- 
+2.20.1
+
+From a7c1f6b825640a95e687adc884a442fde23ccd95 Mon Sep 17 00:00:00 2001
+From: Yun Peng <pcloudy@google.com>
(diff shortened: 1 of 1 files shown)


gregmagolan:
UPDATE: depending on a rules_webtesting commit doesn't work as that requires rules_go to built its metadata binary from source. we'll need to update to the next rules_webtesting release or, alternately, include both windows fixes in the patch we already have here.

gregmagolan:
Note: I verified manually the patch is applied correctly just to be safe

PR #1141, opened 2019-09-17
chore: rename duplicate CircleCI variable

### LINKED ISSUE #1138: CircleCI config.yml is both valid and invalid
This is a "appears to be a bug but somehow works anyway" report.

The current master CircleCI `config.yml`:

https://github.com/bazelbuild/rules_nodejs/blob/master/.circleci/config.yml

Contains two different entries at the top, used to define blocks for right later reuse, both under the key var_8. For reasons I cannot explain, when the repository is forked and the fork is attached to CircleCI, it detects this invalidity (duplicate keys) and refuses to run. Yet the CircleCI on **this** repository has been successfully running in spite of that problem for months.

(Wild guess: there's something interesting behind the scenes at CircleCI around backward compatibility to somehow make this work.)

In any event, this appears to be inadvertent, and at least in some sense a bug. It takes a trivial one line change on that file to fix it, if someone can confirm this is not crazy I will be happy to make the trivial PR.

kylecordes:
Strangely, CircleCI works on this repo but does not work on a cloned
repo because of a duplicate variable name. Fixes #1138.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -88,7 +88,7 @@ var_10: &restore_cache
     keys:
       - *cache_key
 
-var_8: &attach_workspace
+var_11: &attach_workspace
   attach_workspace:
     at: ~/
 


gregmagolan:
Thanks @kylecordes !

PR #1145, opened 2019-09-17
docs: improve node_repositories documentation

alexeagle:


docs/Built-ins.md
```diff
@@ -13,6 +13,181 @@ These rules are available without any npm installation, via the `WORKSPACE` inst
...
+- Using a custom version:
+You can pass in a custom list of NodeJS and/or Yarn repositories and URLs for node_resositories to use.
+- Using a local version:
+To avoid downloads, you can check in vendored copies of NodeJS and/or Yarn and set vendored_node and or vendored_yarn
+to point to those before calling node_repositories.
+
+This rule exposes the `@nodejs` workspace containing some rules the user can call later:
+
+- Run node: `bazel run @nodejs//:node path/to/program.js`
+- Install dependencies using npm: `bazel run @nodejs//:npm install`
```
gregmagolan: does this need to be `bazel run @nodejs//:npm -- install`?

docs/Built-ins.md
```diff
@@ -13,6 +13,181 @@ These rules are available without any npm installation, via the `WORKSPACE` inst
...
+
+### Usage
+
+```
+node_repositories(name, node_repositories, node_urls, node_version, package_json, preserve_symlinks, vendored_node, vendored_yarn, yarn_repositories, yarn_urls, yarn_version)
+```
+
+
+
+#### `name`
```
gregmagolan: This is used for toolchains; I don't think a user would ever want to specify this without blowing things up as the `host` is determined from the `name`:
```
    if repository_ctx.name == "nodejs":
        host = os_name(repository_ctx)
    else:
        host = repository_ctx.name.split("nodejs_", 1)[1]
```

gregmagolan:
There's a bit or doc in `/docs/install.md` that duplicates part of this and I think belongs more with node_repositories:
```
To use the Yarn package manager, which we recommend for its built-in
verification command, you can run:

```sh
$ bazel run @nodejs//:yarn
```

If you use npm instead, run:

```sh
$ bazel run @nodejs//:npm install
```

The `@nodejs//:yarn` and `@nodejs//:npm` targets will run yarn/npm on all of the
package.json contexts listed `package_json` attribute of the `node_repositories`
repository rule in your WORKSPACE file (`node_repositories(package_json = [...])`).

If there are multiple package.json contexts in this rule but you would like to
run the bazel managed yarn or npm on a single context this can be done
using the following targets:

```sh
$ bazel run @nodejs//:bin/yarn -- <arguments passed to yarn>
```

If you use npm instead, run:

```sh
$ bazel run @nodejs//:bin/npm -- <arguments passed to npm>
```

Note: on **Windows** the targets are `@nodejs//:bin/yarn.cmd` and `@nodejs//:bin/npm.cmd`.

This will run yarn/npm in the current working directory. To add a package with the `yarn add` command,
for example, you would use:

```sh
$ bazel run @nodejs//:bin/yarn -- add <package>
```

Note: the arguments passed to `bazel run` after `--` are forwarded to the executable being run.
```

docs/Built-ins.md
alexeagle: No, this is okay since there is no hyphen on the arg.
note this was just moved from lower in the file

docs/Built-ins.md
alexeagle: hmm yeah this is the one problem of doc'ing the rule instead of the macro. the macro sets the name... what should we do?

alexeagle:
I think that doc for "self-managed dependencies" is actually missing a bigger topic: do you install/use `bazel` and it fetches yarn/npm (that doc belongs in such a section) or do you install node/npm and use that to fetch bazel.

Seems like an orthogonal topic to whether you use self-managed/bazel-managed coarse/bazel-managed fine deps

Needs a bit of an overhaul really. I'll merge this one just to refactor this one docset.

alexeagle:
maybe we should have a new doc page just about npm dependencies, separate from install or built-in rules

PR #1148, opened 2019-09-18
Nomerge - WIP for comment - Angular example generator

kylecordes:
For comment and nudges, probably from @alexeagle .

(hack alert) I used a a bit of Bash to generate the list. Seems like a bad idea.

Still TODO

* Push the resulting files to another git repo - easy.
* Understand whether to do that push from outside or inside Bazel.
* Understand whether to do that from CircleCI or BazelCI/Buildkite.

I'm at the limit of my Bazel knowledge on a key bit here: Need to generate code which includes build config and dependencies. Should that happen in a repository rule?

### DIFF
--- a/examples/angular/generate.sh
+++ b/examples/angular/generate.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+set -e
+
+yarn
+yarn generate:clean
+yarn generate
+
+FN=src/app/modules.bzl
+echo "APP_MODULES = [" >$FN
+find src/app -mindepth 2 -maxdepth 2 -name BUILD.bazel -printf '    "//%h",\n' >>$FN
+echo "]" >>$FN
+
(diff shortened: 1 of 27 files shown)


alexeagle:
Why can't the generator JS code update that app_modules.bzl file? why does it need to happen outside the generator in shell?

The simple way to wire this up is for it to have nothing to do with Bazel, right? Just at the end of a green build we'll update the CI to call `yarn generate && git push`.

If we wanted to do it inside bazel I'd love to see a rule that produces a directory identical to what we publish. then it would be `bazel build :example_with_generation && cd there && git push`. But I don't think it's worth spending time on that, we don't really care if the generation step can be incremental.

alexeagle:
Hey Kyle, any interest on picking this up still? My guess is no core contributors have time for reviving this :(

PR #1150, opened 2019-09-19
cleanups

alexeagle:


gregmagolan:
Any reason to keep `/packages/less` & `/packages/stylus` around at this point?

alexeagle:
No I'll remove those packages soon too

Issue #1156, opened 2019-09-19
Remove /internal/http-server & /internal/history-server


gregmagolan:
User's can now use the new npm_package_bin auto-generated targets for http-server & history-server so these wrappers can be removed

alexeagle:
fixed in #1158

PR #1158, opened 2019-09-19
refactor: remove http_server and history_server rules

alexeagle:
BREAKING CHANGE:
history_server and http_server rules are no longer built-in.

To use them, first install the http-server and/or history-server packages
Then load("@npm//http-server:index.bzl", "http_server")
(or replace with history-server, noting that the rule has underscore where the package has hyphen)
and change your templated_args attribute to args


internal/node/node_repositories.bzl
```diff
@@ -706,28 +705,6 @@ def node_repositories(package_json = [], **kwargs):
         symlink_node_modules = False,
     )
 
-    _maybe(
```
gregmagolan: 🌮 🌮 🌮 🌮 🌮 🌮

internal/npm_install/generate_build_file.js
```diff
@@ -1046,19 +1050,32 @@ function printIndexBzl(pkg) {
...
+    for (const [name, path] of executables.entries()) {
+      result = `${result}
 
 # Generated helper macro to call ${name}
-def ${name}(**kwargs):
-    npm_package_bin(tool = "@${WORKSPACE}//${pkg._dir}/bin:${name}", **kwargs)
-`;
+def ${name.replace(/-/g, '_')}(**kwargs):
+    if "outs" in kwargs:
+        npm_package_bin(tool = "@${WORKSPACE}//${pkg._dir}/bin:${name}", **kwargs)
```
gregmagolan: a comment here perhaps?. why `outs` is the determining attribute?

also, do additionalAttributes belong in npm_package_bin? if there is a `"configuration_env_vars"` additional that would likely be needed in both.

gregmagolan:
Resolves https://github.com/bazelbuild/rules_nodejs/issues/1156

internal/npm_install/generate_build_file.js
alexeagle: yes, additionalAttributes are here,line 1073

alexeagle: the doc for this has to go somewhere new, since it's a generated macro. I'll make a new doc for all the stuff we generate into your `@npm` and how to use those things

PR #1159, opened 2019-09-20
refactor: move ts_proto_library to @bazel/labs

### LINKED ISSUE #1151: Move ts_proto_library to @bazel/labs

### LINKED ISSUE #1152: Remove extra yarn_installs
* build_bazel_rules_typescript_devserver_deps should go 
* build_bazel_rules_typescript_protobufs_compiletime_deps is going into labs along to ts_proto_library

alexeagle:
We don't intend to fully support this for 1.0
Also remove a nested install, by vendoring the one file we need for ts_devserver

Fixes #1152
Fixes #1151


packages/labs/src/package.bzl
```diff
@@ -0,0 +1,14 @@
+"Install toolchain dependencies"
+
+load("@build_bazel_rules_nodejs//:defs.bzl", "yarn_install")
+
+def npm_bazel_labs_dependencies():
+    yarn_install(
+        name = "build_bazel_rules_typescript_protobufs_compiletime_deps",
```
gregmagolan: nit: poorly named as it is now in `build_bazel_rules_nodejs`

packages/labs/src/protobufjs/BUILD.bazel
```diff
@@ -69,8 +70,24 @@ nodejs_binary(
     install_source_map_support = False,
 )
 
+# Runtime libraries needed by the protobufjs library.
+# Any JS code produced by the ts_proto_library rule has a runtime dependency on these scripts.
+js_library(
```
gregmagolan: hmmmmmm.. looks like js_library has the additional amd_names attribute which is being used. will note that in https://github.com/bazelbuild/rules_nodejs/issues/149

packages/typescript/src/internal/ts_repositories.bzl
```diff
@@ -28,23 +28,3 @@ def ts_setup_workspace():
 
     # 0.16.8: ng_package fix for packaging binary files
     check_rules_nodejs_version("0.16.8")
-
-    yarn_install(
-        name = "build_bazel_rules_typescript_devserver_deps",
```
gregmagolan: 🌮 🌮 🌮 🌮 🌮 🌮

packages/labs/src/package.bzl
gregmagolan: also, we could probably remove this extra yarn_install entirely by moving its deps to the @bazel/labs package.json but not a big deal as its in labs which is experimental code anyway

alexeagle: yeah also I seem to recall this workspace name appeared in a few places which is why we didn't rename it last time we moved it. Going for a minimal edit here to clear out the 1.0 space and yeah, don't mind a mess in labs

packages/labs/src/protobufjs/BUILD.bazel
alexeagle: sigh, yes that needs a new home

Issue #1162, opened 2019-09-20
Handle peerDependencies a bit better.


bweston92:
# 🚀 feature request

### Relevant Rules

- `yarn_install`
- `npm_install`

### Description

Currently creating a React without using TypeScript will require adding many peer dependencies that React has for TypeScript even though it is not used at all however Bazel will fail the install.

### Describe the solution you'd like

Either have some sort of `ignoreDepedencies` in `package.json` which are read and will no longer fail, this will mean you've acknowledged and are ok with these dependencies being ignored.

Or we can just not hard fail at all on peer dependencies assuming the developer knows best.

### Describe alternatives you've considered

You have to add peer dependencies of all packages and sub packages that have them as dependencies in your own `package.json`

alexeagle:
Latest release no longer errors when peerDeps are missing

PR #1163, opened 2019-09-20
Add ability to ignore peer depdenencies from package.json

bweston92:
## PR Checklist

## PR Type

Issue Number: #1162

## What is the new behavior?

Ability to pass a `canIgnore` list of strings containing peer dependency names that can be ignored.

- [ ] Yes
- [x] No

## Other information

This is one implementation, we could make it so we can pass it via the `*_install` rule instead if it might be cleaner.

### DIFF
--- a/internal/npm_install/generate_build_file.js
+++ b/internal/npm_install/generate_build_file.js
@@ -696,7 +696,7 @@ def _maybe(repo_rule, name, **kwargs):
             });
         }
         findDeps(dep.dependencies, true, 'dependency');
-        findDeps(dep.peerDependencies, true, 'peer dependency');
+        findDeps(dep.peerDependencies, false, 'peer dependency');
         // `optionalDependencies` that are missing should be silently
         // ignored since the npm/yarn will not fail if these dependencies
         // fail to install. Packages should handle the cases where these
--- a/internal/npm_install/generate_build_file.ts
+++ b/internal/npm_install/generate_build_file.ts
@@ -779,7 +779,7 @@ function flattenPkgDependencies(pkg: Dep, dep: Dep, pkgsMap: Map<string, Dep>) {
   }
 
   findDeps(dep.dependencies, true, 'dependency');
-  findDeps(dep.peerDependencies, true, 'peer dependency');
+  findDeps(dep.peerDependencies, false, 'peer dependency');
   // `optionalDependencies` that are missing should be silently
   // ignored since the npm/yarn will not fail if these dependencies
   // fail to install. Packages should handle the cases where these


internal/npm_install/generate_build_file.js
```diff
@@ -745,8 +748,11 @@ function flattenPkgDependencies(pkg, dep, pkgsMap) {
     });
   }
 
-  findDeps(dep.dependencies, true, 'dependency');
-  findDeps(dep.peerDependencies, true, 'peer dependency');
+  findDeps(dep.dependencies, () => true, 'dependency');
+  findDeps(dep.peerDependencies, (targetDep) => {
+    // if it is not in the can ignore then is is required
+    return canIgnore.indexOf(targetDep) === -1;
```
alexeagle: so the field `canIgnore` is a list in the user's package.json? a new field you're adding?

Would be good to have some doc change so users can figure out how to use this, along with some test (can just be that you find a way to reproduce the failure in this repo by not installing a peer dep, then use your new feature to fix that breakage)

bweston92: Yes, `canIgnore` is a string list in the package.json, I was going to add documentation but I don't know whether or not to do it this way, we also have an alternative of putting in a rule attribute, which would you prefer?

Toxicable: I'd prefer an attribute over a new key added to the package.json

bweston92: @Toxicable I personally think I am with you, it was something I thought about after I wrote this but will amend asap.

alexeagle: yeah it's a behavior of `npm_install`/`yarn_install` so seems like it should go there. I think the default should be to act like `npm` and `yarn` which means we should print peerDep warnings but proceed. Offer users a `strictPeerDependencies` boolean option. (I guess you could have a list of specific peerDep warnings to whitelist in strict mode)

bweston92: I have changed the way it is done, will check when at work again tomorrow to make sure it still works if so I will add documentation if it is ok with you?

internal/npm_install/npm_install.bzl
```diff
@@ -89,6 +89,10 @@ which are installed, but failures on installation are ignored.
     "exclude_packages": attr.string_list(
         doc = """DEPRECATED. This attribute is no longer used.""",
     ),
+    "ignore_peer_dependencies": attr.string_list(
+        default = [],
+        doc = """A list of peer depedencies that can be ignored.""",
```
Toxicable: Could you add to this doc that `strict_peer_dependencies ` must be set for you to use this?  
Or maybe change the logic a little bit so that you can set this without needing `strict_peer_dependencies `

bweston92: How would you suggest a change the logic which can have the following 3 use cases:

`a = strict_peer_dependencies `
`b = ignore_peer_dependencies `

1. You want to ignore peer dependencies. (`a = False`, `b = []`)
1. You want to have strict dependencies and ignore a few of them. (`a = True`, `b = ['dep1', 'dep2']`)
1. You want to have strict dependencies and ignore none of them.  (`a = True`, `b = []`)

alexeagle: Can you find any other popular tooling that gives a strict mode for peer deps? I would like to follow some precedent rather than invent something new.

If there is no precedent at all, maybe we shouldn't even have a strict mode - that should be someone else's job, orthogonal to whether the package manager is run by bazel

bweston92: I don't really know of any, Yarn and NPM both give warning but nothing more.

Toxicable: Yeah tha's a good point @alexeagle I don't know of any that have strict deps.
Maybe we should just remove the checks, if anything it should be yarn checking things like this right, not us?

bweston92: Ok so should I revert it all and put the “true” to “false” for required?

bweston92: I have reverted and set the peer dependencies as a non requirement altogether. Is that ok?

Toxicable: Yeah sounds good to me, sorry for all the back and forward.
We appciate you taking the time here!

alexeagle:
Could you rebase past the conflict, squash commits and update commit message to pass the .commitlint check (seems like your precommit hooks are not working?)
If that's hard for you I can do it

bweston92:
@alexeagle is that ok?

alexeagle:
I think you've done something wrong, did you look at the resulting diff? This PR now proposes editing only two lines which can't be right.
https://github.com/bazelbuild/rules_nodejs/pull/1163/files

I guess I'll grab your earlier snapshot and just fix this for you...

bweston92:
That is right, before we wanted to allow us to ignore some of the peer dependencies, now we're just turning off the flag which makes peer dependencies required in the first place.

httpete:
For anyone struggling, it seems that optionalDepencies in your package.json works to bypass it.

`  "optionalDependencies": {
    "popper.js": "latest",
    "jquery": "latest",
    "@types/prismjs": "latest",
    "karma-firefox-launcher": "latest"
  }`

bweston92:
@httpete yes it does since this PR, prior to this PR peerDepedencies where strictly required by rules_nodejs.

PR #1164, opened 2019-09-20
feat(builtin): npm_package_bin can produce directory output

alexeagle:


internal/node/npm_package_bin.bzl
```diff
@@ -8,26 +8,45 @@ _ATTRS = {
     "outs": attr.output_list(),
     "args": attr.string_list(mandatory = True),
     "data": attr.label_list(allow_files = True, aspects = [module_mappings_aspect]),
+    "out_dir": attr.string(),
```
gregmagolan: should be added to doc string for rule and comment on `$@` usage in args

gregmagolan: could out_dir be a label to give the user more freedom as to where it will be?

alexeagle: I don't think that's possible since https://docs.bazel.build/versions/master/skylark/lib/actions.html#declare_directory is the only API to produce TreeArtifact and it takes string - and you can't use a label to reference something that doesn't exist yet
anyway it's probably good that the output directory has to be inside the package that owns it?

gregmagolan: 👍

alexeagle: added missing doc

Also handle the case of `args=["--out-dir=$@"]` (it's a substring of the arg)

alexeagle:
thanks @c-parsons for advising on this

internal/node/npm_package_bin.bzl
```diff
@@ -45,17 +63,22 @@ def npm_package_bin(tool = None, package = None, package_bin = None, **kwargs):
     https://docs.bazel.build/versions/master/skylark/macros.html#full-example
 
     Args:
-        data: identical to [genrule.srcs](https://docs.bazel.build/versions/master/be/general.html#genrule.srcs)
+        data: similar to [genrule.srcs](https://docs.bazel.build/versions/master/be/general.html#genrule.srcs)
               may also include targets that produce or reference npm packages which are needed by the tool
-        outs: identical to [genrule.outs](https://docs.bazel.build/versions/master/be/general.html#genrule.outs)
+        outs: similar to [genrule.outs](https://docs.bazel.build/versions/master/be/general.html#genrule.outs)
+        out_dir: use this instead of `outs` if you want the output to be a directory
```
gregmagolan: nit: mention this is a string a relative to the current package?

PR #1168, opened 2019-09-20
feat(examples): replace examples/webapp with new rollup_bundle

alexeagle:
This demonstrates how to stitch together a complete app.
Note that injecting script tags no longer works because rollup_bundle produces only a directory output.
We'll try to figure out how to restore that feature. In the meantime we just hard-code script tag paths

### DIFF
--- a/internal/node/npm_package_bin.bzl
+++ b/internal/node/npm_package_bin.bzl
@@ -8,7 +8,7 @@ _ATTRS = {
     "outs": attr.output_list(),
     "args": attr.string_list(mandatory = True),
     "data": attr.label_list(allow_files = True, aspects = [module_mappings_aspect]),
-    "out_dir": attr.string(),
+    "output_dir": attr.bool(),
     "tool": attr.label(
         executable = True,
         cfg = "host",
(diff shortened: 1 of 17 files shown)


internal/node/npm_package_bin.bzl
```diff
@@ -66,14 +66,15 @@ def npm_package_bin(tool = None, package = None, package_bin = None, **kwargs):
         data: similar to [genrule.srcs](https://docs.bazel.build/versions/master/be/general.html#genrule.srcs)
               may also include targets that produce or reference npm packages which are needed by the tool
         outs: similar to [genrule.outs](https://docs.bazel.build/versions/master/be/general.html#genrule.outs)
-        out_dir: use this instead of `outs` if you want the output to be a directory
+        output_dir: set to True if you want the output to be a directory
                  Exactly one of `outs`, `out_dir` may be used.
```
jkrems:
```suggestion
                  Exactly one of `outs`, `output_dir` may be used.
```

alexeagle: nice catch

alexeagle:
merged as 7231aaa1

PR #1174, opened 2019-09-22
Fix repository references in CHANGELOG.

bweston92:
Links in the changelog currently point to Alex's fork which doesn't have the issues making them 404.


alexeagle:
Thanks! I don't know how to fix this for the future though. I guess there is something in my `.git` folder that the changelog script is picking up. I poked around in the changelog generator once and didn't find why. Can you see if there's some setting I'm missing so it uses the `upstream` remote to make these URLs?

alexeagle:
found the answer: just need to add `"repository": "https://github.com/bazelbuild/rules_nodejs",` to the root package.json to make `conventional-changelog` make good links

bweston92:
Sweet I did look but didn’t see the tool you was using to know where to look

PR #1180, opened 2019-09-23
feat(builtin): linker should resolve workspace-absolute paths

alexeagle:


internal/rollup/test/rollup/rollup.spec.js
```diff
@@ -1,28 +1,27 @@
 check = require('check');
-const path = __dirname;
 
 describe('bundling', () => {
   it('should work', () => {
-    check(path, 'bundle.min.js', 'bundle-min_golden.js_');
+    check('bundle.min.js', 'bundle-min_golden.js_');
```
gregmagolan: this is nice 🌮

PR #1181, opened 2019-09-23
feat(rollup): ensure that sourcemaps work end-to-end

alexeagle:


packages/rollup/src/rollup_bundle.bzl
```diff
@@ -218,7 +219,7 @@ def _rollup_outs(sourcemap, name, entry_point, entry_points, output_dir):
             fail("Multiple entry points require that output_dir be set")
         out = entry_point_outs[0]
         result[out] = out + ".js"
-        if sourcemap:
+        if sourcemap == "true":
```
gregmagolan: I think case insensitive here as with the current bool value users will have `True` or `False`. This will keep compat as you are just adding the `inline` option then.

alexeagle:
we never released this rule yet so it's not an observable breakage :)

Issue #1182, opened 2019-09-23
terser_minified should attach sourcemap comment


alexeagle:
Currently our terser rule outputs a .js.map but no comment in the .js file pointing there. This means you have to manually connect the sourcemap in Chrome devtools to get debugging of a minified production build.

Terser CLI says
> `--source-map "url='<URL>'"` to specify the URL where the source map can be found. Otherwise Terser assumes HTTP X-SourceMap is being used and will omit the `//# sourceMappingURL= ` directive.

Since we run terser on directories, we can't just use their CLI naively to get this behavior. Our `@bazel/terser/index.js` will have to adapt the command line for each subprocess to get the right sourcemap link in the written file.

alexeagle:
Note that it seems you could configure your server to include a header `SourceMap: path` on the .js file requests but from some quick research it seems like this is a lot of work and not idiomatic - the comment at the end of the .js file is the way to go.

alexeagle:
fixed by 1186

PR #1186, opened 2019-09-24
feat(terser): add source map links

Toxicable:
## PR Checklist

## PR Type

Issue Number: #1182
minified files do not link back to their source maps

## What is the new behavior?
minified files no link back to the source maps

- [x] No

### DIFF
--- a/packages/terser/src/index.js
+++ b/packages/terser/src/index.js
@@ -27,7 +27,7 @@ function isDirectory(input) {
 }
 
 /**
- * Replaces <OUTPUT_MAP_FILE> with the outputFile name in the source-map options argument
+ * Replaces directory url with the outputFile name in the url option of source-map argument
  */
 function directoryArgs(residualArgs, outputFile) {
   const sourceMapIndex = residualArgs.indexOf('--source-map');
@@ -35,18 +35,15 @@ function directoryArgs(residualArgs, outputFile) {
     return residualArgs;
   }
 
-  // copy args so we don't accidently mutate and process the same set of args twice
-  const argsCopy = [...residualArgs];
+  const sourceMapOptions = residualArgs[sourceMapIndex + 1].split(',');
+  const newSourceMapOptions = sourceMapOptions.map(
+      o => o.startsWith('url=') ? `url='${path.basename(outputFile)}.map'` : o);
 
-  // the options for the source map arg is the next one
-  // if it changes in terser_minified.bzl this needs to be updated
-  const sourceMapOptsIndex = sourceMapIndex + 1
-  const sourceMapOptsStr = argsCopy[sourceMapOptsIndex];
-
-  argsCopy[sourceMapOptsIndex] =
-      sourceMapOptsStr.replace('<OUTPUT_MAP_FILE>', `${path.basename(outputFile)}.map`);
-
-  return argsCopy;
+  return [
+    ...residualArgs.slice(0, sourceMapIndex + 1),
+    newSourceMapOptions.join(','),
+    ...residualArgs.slice(sourceMapIndex + 2),
+  ];
 }
 
 function terserDirectory(input, output, residual, terserBinary) {
--- a/packages/terser/src/index.js
+++ b/packages/terser/src/index.js
@@ -22,40 +22,39 @@ function log_error(...m) {
   console.error('[terser/index.js]', ...m);
 }
 
-// Peek at the arguments to find any directories declared as inputs
-let argv = process.argv.slice(2);
-// terser_minified.bzl always passes the inputs first,
-// then --output [out], then remaining args
-// We want to keep those remaining ones to pass to terser
-// Avoid a dependency on a library like minimist; keep it simple.
-const outputArgIndex = argv.findIndex((arg) => arg.startsWith('--'));
-
-// We don't want to implement a command-line parser for terser
-// so we invoke its CLI as child processes when a directory is provided, just altering the
-// input/output arguments. See discussion: https://github.com/bazelbuild/rules_nodejs/issues/822
-
-const inputs = argv.slice(0, outputArgIndex);
-const output = argv[outputArgIndex + 1];
-const residual = argv.slice(outputArgIndex + 2);
-
-// user override for the terser binary location. used in testing.
-const TERSER_BINARY = process.env.TERSER_BINARY || require.resolve('terser/bin/uglifyjs')
-// choose a default concurrency of the number of cores -1 but at least 1.
-const TERSER_CONCURENCY = (process.env.TERSER_CONCURRENCY || os.cpus().length - 1) || 1
-
-log_verbose(`Running terser/index.js
-  inputs: ${inputs}
-  output: ${output}
-  residual: ${residual}`);
-
 function isDirectory(input) {
   return fs.lstatSync(path.join(process.cwd(), input)).isDirectory();
 }
 
-function terserDirectory(input) {
+/**
+ * Replaces <OUTPUT_MAP_FILE> with the outputFile name in the source-map options argument
+ */
+function directoryArgs(residualArgs, outputFile) {
+  const sourceMapIndex = residualArgs.indexOf('--source-map');
+  if (sourceMapIndex === -1) {
+    return residualArgs;
+  }
+
+  // copy args so we don't accidently mutate and process the same set of args twice
+  const argsCopy = [...residualArgs];
+
+  // the options for the source map arg is the next one
+  // if it changes in terser_minified.bzl this needs to be updated
+  const sourceMapOptsIndex = sourceMapIndex + 1
+  const sourceMapOptsStr = argsCopy[sourceMapOptsIndex];
+
+  argsCopy[sourceMapOptsIndex] =
+      sourceMapOptsStr.replace('<OUTPUT_MAP_FILE>', `${path.basename(outputFile)}.map`);
+
+  return argsCopy;
+}
+
+function terserDirectory(input, output, residual, terserBinary) {
   if (!fs.existsSync(output)) {
     fs.mkdirSync(output);
   }
+
+  const TERSER_CONCURENCY = (process.env.TERSER_CONCURRENCY || os.cpus().length - 1) || 1
 
   let work = [];
   let active = 0;
@@ -63,7 +62,8 @@ function terserDirectory(input) {
 
   function exec([inputFile, outputFile]) {
     active++;
-    let args = [TERSER_BINARY, inputFile, '--output', outputFile, ...residual];
+    let args =
+        [terserBinary, inputFile, '--output', outputFile, ...directoryArgs(residual, outputFile)];
 
     spawn(process.execPath, args)
         .then(
@@ -116,19 +116,6 @@ function terserDirectory(input) {
   });
 }
 
-if (!inputs.find(isDirectory) && inputs.length) {
-  // Inputs were only files
-  // Just use terser CLI exactly as it works outside bazel
-  require(TERSER_BINARY || 'terser/bin/uglifyjs');
-
-} else if (inputs.length > 1) {
-  // We don't know how to merge multiple input dirs to one output dir
-  throw new Error('terser_minified only allows a single input when minifying a directory');
-
-} else if (inputs[0]) {
-  terserDirectory(inputs[0]);
-}
-
 function spawn(cmd, args) {
   return new Promise((resolve, reject) => {
     const err = [];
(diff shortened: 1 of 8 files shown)


packages/terser/src/index.js
```diff
@@ -48,6 +48,35 @@ log_verbose(`Running terser/index.js
   output: ${output}
   residual: ${residual}`);
 
+/**
+ * @param {string[]} residuleArgs
```
alexeagle: spelling

packages/terser/src/index.js
```diff
@@ -48,6 +48,35 @@ log_verbose(`Running terser/index.js
...
+ * @param {string} output
+ */
+function processSourceMapArg(residuleArgs, output) {
```
alexeagle: this function seems like it would also benefit from some unit testing?

packages/terser/src/index.js
```diff
@@ -117,6 +148,7 @@ function terserDirectory(input) {
 }
 
 if (!inputs.find(isDirectory) && inputs.length) {
+  process.argv = processSourceMapArg(process.argv, output);
```
alexeagle: why mess with the argv if we are just a pass-through to terser? it should be a no-op

packages/terser/test/inline_sourcemap/spec.js
```diff
@@ -14,4 +14,14 @@ describe('terser sourcemap handling', () => {
       expect(pos.column).toBe(14);
     });
   });
+
+  it('should link the source to the source map', () => {
+    const file = require.resolve(__dirname + '/out.min.js');
+    const expectedSourceMapUrl = require.resolve(__dirname + '/out.min.js.map');
+    const content = fs.readFileSync(file, 'utf8');
+    const sourceMapLine = content.split('\n').find(l => l.startsWith('//#'));
```
alexeagle: maybe need to split on `/\r?\n/` for Windows?

packages/terser/src/index.js
```diff
@@ -48,6 +48,35 @@ log_verbose(`Running terser/index.js
...
+  const sourceMapIndex = argsCopy.findIndex(a => a === '--source-map');
+  if (sourceMapIndex != -1) {
+    const sourceMapOptsIndex = sourceMapIndex + 1
+    const sourceMapOpts = argsCopy[sourceMapOptsIndex];
+
+    const baseKeyValue =
+        sourceMapOpts.split(',').map(g => g.split('=')).find(([key, value]) => key === 'base');
+    if (baseKeyValue) {
+      const [_, baseUrl] = baseKeyValue;
+      // TODO: maybe there's a better api to remove this part of the url
```
alexeagle: fix the TODO?

packages/terser/src/index.js
```diff
@@ -48,6 +48,35 @@ log_verbose(`Running terser/index.js
...
+      // the .slice(1) remove the starting `/`
+      const relativeUrl = output.replace(baseUrl, '').slice(1);
+      // if we use this relative one them terser throws
```
alexeagle: them -> then

packages/terser/src/index.js
Toxicable: terser reads the args back out of `process.argv` it's the only api we have for it

packages/terser/src/index.js
Toxicable: it's a little hard to do, should, we'd have to export it from this file to be able to unit test, do you think exporting is a good idea?

alexeagle: sure, exporting the function is fine I think, it's not like we need to document it as public API

packages/terser/src/index.js
alexeagle: yeah I know, but why do you need to do any processing on argv?

Toxicable: we need to process the source map argument, in processing we modify it to include the extra option, the same as the directory method

packages/terser/src/index.js
Toxicable: Aight, added some tests, required a bit of refactoring so needs a rereview

Toxicable:
@alexeagle Looks like we're good on CI, had a few failures, guess they were flakes?

packages/terser/test/inline_sourcemap/spec.js
```diff
@@ -14,4 +15,19 @@ describe('terser sourcemap handling', () => {
       expect(pos.column).toBe(14);
     });
   });
+
+  it('should link the source to the source map', () => {
+    const file = require.resolve(__dirname + '/out.min.js');
```
soldair: Because __dirname is an absolute path `require.resolve` can be replaced with `path.join`
This is helpful because when we look at require.resolve use to refactor out loader.js logic there will be fewer of them to understand.

packages/terser/src/terser_minified.js
```diff
@@ -0,0 +1,199 @@
...
+ * @param {string[]} residualArgs
+ * @param {string} output
+ */
+function processSourceMapArg(residualArgs, output) {
+  // copy args so we don't accidently mutate and process the same set of args twice
+  const argsCopy = [...residualArgs];
+  const sourceMapIndex = argsCopy.findIndex(a => a === '--source-map');
+  if (sourceMapIndex != -1) {
+    // the options for the source map arg is the next one
+    // if it changes in terser_minified.bzl this needs to be updated
```
soldair: we should consider reintroducing minimist for option parsing at the very least to organize which flags we read and why.

Toxicable: i'd appciate having a lib for handling args, doing it manually can get error prone.
Considering almost everthing we handle is CLIs maybe this is a good idea.
Or at very least implement our own standward for CLI args parsing if we dont want to pull in a full lib

packages/terser/src/terser_minified.js
```diff
@@ -0,0 +1,199 @@
+#!/usr/bin/env node
```
alexeagle: please rename to the old path, or at least put the rename in its own commit so I don't have a giant block of "newly added" code that I can't tell what you changed

Toxicable: Yeah that's fair, thought the naming of the files made more sense this way, but for the sake of actually able to code review it I'm happy flipping it around

Toxicable:
It's failing on windows with the brlow error, not really sure why
```
1) terser sourcemap handling should link the source to the source map
--
  | Message:
  | Error: ENOENT: no such file or directory, open 'D:\b\bk-windows-java8-kfjp\bazel\rules-nodejs-nodejs\packages\terser\test\inline_sourcemap\out.min.js'
```

packages/terser/src/terser_minified.js
Toxicable: put the file rename in a seprate commit

packages/terser/src/index.js
alexeagle: Sorry I still don't get it. If you're not doing a directory, you don't need our wrapper. The original terser arguments should work as-is

packages/terser/test/sourcemap-arg.spec.js
```diff
@@ -0,0 +1,19 @@
+const {processSourceMapArg} = require('npm_bazel_terser/index')
+
+describe('processSourceMapArg', () => {
+    it('return a new array ref', () => {
+  const args = [];
+  expect(processSourceMapArg(args)).not.toBe(args)
+    })
+
+it('should strip the base and leading / from the output url', () => {
```
alexeagle: why is the formatter not working for you? there should be a husky-installed pre-commit hook to check that you formatted your changes

packages/terser/src/index.js
```diff
@@ -147,3 +150,49 @@ function spawn(cmd, args) {
     });
   })
 }
+
+function main() {
```
alexeagle: please don't do unrelated refactorings. this just makes me search around to see what you changed.

Especially as you're on your first few contributions I would like to get minimal deltas where every edit is related to making the new test pass please

packages/terser/src/terser_minified.js
alexeagle: that didn't quite work because the first commit still references the new file, I wasn't able to just fetch and get the test failure off the first commit.

I don't even really agree with renaming the file. If there's only ever going to be one file in this package, why not name it index? it makes the rules shorter since there are no data deps

packages/terser/src/index.js
Toxicable: I needed to do this refactor so that this file wouldn't jsut execute the CLI work flow when required into the tests

packages/terser/src/terser_minified.js
Toxicable: > I wasn't able to just fetch and get the test failure off the first commit.

I have a couple of fixups that apply to the first commit, they'd need to be rebased in to allow for that.

I couldn't get `if(require.module === module) { main() }` to work here under the testing environment. It would still call the main method even when that file was being required and wasn't the entry point.
I'd prefer a single file aswell but I don't get how to make it work.

packages/terser/src/index.js
Toxicable: >The original terser arguments should work as-is

The original terser arguments don't have the `url` set in the sourcemap options, calling this methods sets it.

packages/terser/test/sourcemap-arg.spec.js
Toxicable: Im pretty sure thats what the formatter wanted to do.
I even ran it manually, as in `clang-format` instead of `git-clang-format`

Toxicable: ah! it was formatting correctly, but I forgot some semi colons - im use to prettier isnerting them for me

packages/terser/src/index.js
alexeagle: I see, so then what I'm looking for is that the original terser arguments should include `url` in the sourcemap options so that the original binary works in the single-file case.

packages/terser/src/terser_minified.js
alexeagle: you mean `if (require.main === module) {` ?
What didn't work about it? Let's please figure that out rather than introduce a refactoring purely as a workaround for a problem we don't understand

alexeagle:
sorry this is not getting fast enough back and forth - happy to discuss on Slack. This still doesn't look like what I was expecting.

packages/terser/src/index.js
Toxicable: hmm ok ill see what I can do, but I assume this means we'll be duplicating the logic in the function I added between here and in the `.bzl` that goes with this.

packages/terser/src/terser_minified.js
Toxicable: The refactor is still required, if we can get `if (require.main === module) {` to work then that just removed the need for two files.

Toxicable: ok tried it again and it's working now, not sure what happened before

packages/terser/src/index.js
```diff
@@ -22,48 +22,65 @@ function log_error(...m) {
...
-
-log_verbose(`Running terser/index.js
-  inputs: ${inputs}
-  output: ${output}
-  residual: ${residual}`);
+/**
+ * @param {string[]} residualArgs
+ * @param {string} output
+ */
+function processSourceMapArg(residualArgs, output) {
```
gregmagolan: a better name so that it is clear that this is for the directory case only would be better I think (maybe just `directoryArgs()`. I would also define this function within terserDirectory as well next to exec() so that its clear what the scope is.

packages/terser/src/index.js
```diff
@@ -22,48 +22,65 @@ function log_error(...m) {
...
+  // copy args so we don't accidently mutate and process the same set of args twice
+  const argsCopy = [...residualArgs];
+  const sourceMapIndex = argsCopy.findIndex(a => a === '--source-map');
```
gregmagolan: nit:

a little simpler to use indexOf: `const sourceMapIndex = residualArgs.indexOf('--source-map');`

You can also do this check on `residualArgs` and escape with
```
if (sourceMapIndex === -1)
    return residualArgs;
```
and then make a copy of the args after that.

packages/terser/src/index.js
```diff
@@ -22,48 +22,65 @@ function log_error(...m) {
...
+
   let work = [];
   let active = 0;
   let errors = [];
 
   function exec([inputFile, outputFile]) {
     active++;
-    let args = [TERSER_BINARY, inputFile, '--output', outputFile, ...residual];
+    let args =
+        [terserBinary, inputFile, '--output', outputFile, ...processSourceMapArg(residual, output)];
```
gregmagolan: Should be `...processSourceMapArg(residual, outputFile)` right?

packages/terser/src/index.js
gregmagolan: `directoryArgs()` will also be a good place to check if `${inputFile}.map` exists and if it does then `content=inline` needs to be replaced with `content='${inputFile}.map'`. That can be a follow up PR.

Toxicable: we cant define it in `terserDirectory ` unforunately since we need to export it to unit test it

packages/terser/src/index.js
Toxicable: yes! must have messed that up in a rebase or something

packages/terser/src/index.js
Toxicable: Turns out no logic is required in the .bzl, we know the exact name.

packages/terser/src/index.js
gregmagolan: ahhh. yes. that makes sense. the name change is clear enough then :)

packages/terser/src/BUILD.bazel
```diff
@@ -42,6 +42,8 @@ filegroup(
 
 nodejs_binary(
     name = "terser",
-    data = ["@npm//terser"],
+    data = [
```
gregmagolan: nit: this diff is not needed

packages/terser/src/index.js
```diff
@@ -22,48 +22,50 @@ function log_error(...m) {
...
-  inputs: ${inputs}
-  output: ${output}
-  residual: ${residual}`);
-
 function isDirectory(input) {
   return fs.lstatSync(path.join(process.cwd(), input)).isDirectory();
 }
 
-function terserDirectory(input) {
+/**
```
gregmagolan: nit: these param strings are not very useful; better to just add a short description of what this function does IMO

packages/terser/src/terser_minified.bzl
```diff
@@ -138,6 +138,11 @@ def _terser(ctx):
         else:
             fail("When sourcemap is True, there should only be one or none input sourcemaps")
 
+        if len(directory_srcs) == 0:
+            source_map_opts.append("url='%s.js.map'" % ctx.label.name)
+        else:
+            source_map_opts.append("url='<OUTPUT_MAP_FILE>'")
```
gregmagolan: a comment that this special key is replaced with the actual filename in the .js file would be good here

packages/terser/src/index.js
Toxicable: I find them helpful for types, but maybe we just convert this file to TS at some point.

alexeagle:
The magic token `<URL_GOES_HERE>` was both smelly, because you need a comment in one file saying to update along with the other file, and also was breaking some shells that interpret the `<` as a file indirection.

I think it's better just to comma-parse the `source-map` argument and replace the `url` option altogether. Simplifies the `.bzl` code in a way that's symmetric with the output option - in both cases the starlark can be naive about assembling the command-line assuming that the `ctx.label.name` is going to be used everywhere.

I added a commit to make that refactoring.

PR #1187, opened 2019-09-24
Typescriptify generate_build_files.js

soldair:
Direct conversion of generate_build_file.js to typescript.
no behavior should have changed. tests pass.

this prepares the code to refactor for async.


.vscode/settings.json
```diff
@@ -1,4 +0,0 @@
-{
```
Toxicable: im trialing using a `settings.json.default` with a few things like the ones in here in them for share around.
speaking of the code in here, i've submitted this: https://github.com/bazelbuild/vscode-bazel/issues/132

internal/npm_install/generate_build_file.ts
```diff
@@ -0,0 +1,1140 @@
...
+          additionalAttributes(pkg, name)}
+            **kwargs
+        )
+  `;
+    }
+  }
+  return result;
+}
+
+type Dep = {
```
Toxicable: should be an interface

internal/npm_install/generate_build_file.ts
```diff
@@ -0,0 +1,1140 @@
...
+  }
+
+  // none of the methods we tried resulted in a file
+  log_verbose(`could not find entry point for npm package ${pkg._name}`);
+
+  // at this point there's nothing left for us to try, so return nothing
+  return undefined;
+}
+
+type Bag<T> =
```
Toxicable: this type is usually called a Dictionary or Dict, typescript has one called a `Record<T,V>` so you could default it like `type Dictionary<T> = Record<string,T>`

internal/npm_install/generate_build_file.ts
```diff
@@ -0,0 +1,1140 @@
...
+    mkdirp(path.dirname(p));
+    fs.mkdirSync(p);
+  }
+}
+
+/**
+ * Writes a file, first ensuring that the directory to
+ * write to exists.
+ */
+function writeFileSync(p: string, content: any) {
```
Toxicable: content should be `string`, right?

internal/npm_install/generate_build_file.ts
```diff
@@ -0,0 +1,1140 @@
...
+}
+
+/**
+ * Reformat/pretty-print a json object as a skylark comment (each line
+ * starts with '# ').
+ */
+function printJson(pkg: Dep) {
+  // Clone and modify _dependencies to avoid circular issues when JSONifying
+  // & delete _files array
+  const cloned: any = {...pkg};
```
Toxicable: cloned looks like a `Dep`

.vscode/settings.json
soldair: i removed this because it seems like an error that you should be configuring my local vscode.
ill read the issue for background

internal/npm_install/generate_build_file.ts
soldair: Interfaces can be extended. Because this type is not exported there isn't a benefit that i know of.

i'm sure there is a style guide but somewhere about this but i'd like to only make style changes if they're backed up by documentation.

also this type is just popped in the middle of the file. it might make sense to move it to another file that just exports interfaces and clean things up.

internal/npm_install/generate_build_file.ts
soldair: the nodejs type expose it as any and i opted to propagate those though a case could have been made to only type how we use it.

internal/npm_install/generate_build_file.ts
soldair: yeah. i wish. it turns out its cloned._dependencies is a string array not a dep array.
because we only use it in debugging within this method i felt it was more maintainable to use any rather than extend Dep

internal/npm_install/generate_build_file.ts
soldair: I can empathize with that argument coming from python. I'm ambivalent on this because Dicts are not exactly javascript objects and Object is reserved in js for obvious reasons. I hadn't considered utility types. thanks

internal/npm_install/generate_build_file.ts
Toxicable: ah... Right, well that's unfortunate.
Might be worth refactoring when looking at this afterwards then

.vscode/settings.json
Toxicable: Yeah I think this should be ignored aswell.
It would be nice if we could check in some settings for the workspace, however vs code dosen't currently support having two settings files that get merged together.
That's why my work around is to make a default one that devs can use as a template

internal/npm_install/generate_build_file.ts
Toxicable: Fair point, Im pretty sure there's some docs for it somewhere but I cant find it from a quick search I just did.
Either way, for me it's just tidier, but you're right there's no functional difference in this situation aside from possible extension later on, but you could just refactor it then.

alexeagle: there's a google typescript style guide which we ought to nominally follow

PR #1189, opened 2019-09-25
docs: index.bzl -> defs.bzl

connorjclark:
It was wrong.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -42,7 +42,7 @@ See [Built-ins]
 If you have installed the [rollup] package, you could write this rule:
 
 ```python
-load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_binary")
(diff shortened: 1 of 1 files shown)


alexeagle:
Ah, sorry, this was correct but it refers to the next release.
You should use the docs at the tag you're using, e.g. 
https://github.com/bazelbuild/rules_nodejs/blob/0.37.1/docs/Built-ins.md

I think we should update some pointer, I guess the one from the readme?

alexeagle:
Actually this is fine, it will get overwritten by the next release.

We should just leave the docs/ folder alone until it gets updated during the release process, so it doesn't leak things that are only on HEAD

connorjclark:
> Ah, sorry, this was correct but it refers to the next release.

Makes sense to me, thanks for the explanation!

> We should just leave the docs/ folder alone until it gets updated during the release process, so it doesn't leak things that are only on HEAD

Sounds like a good solution.

PR #1192, opened 2019-09-25
refactor(builtin): remove git_repository_under_test as it is no longer needed

gregmagolan:


alexeagle:
Yay less code to maintain. You can still enjoy having written it :wine_glass:

PR #1193, opened 2019-09-25
Debug log in bazel integration test

gregmagolan:
This log makes the workflow of running `yarn run_integration_test --define=DEBUG=1 //examples:examples_angular` and then `cd /path/to/workspace/tmp` easier as when `DEBUG=1` test_runner.js will persist the `tmp` folders it creates so it should also output the path to the `tmp` folder for the workspace under test in the test logs.

Also fixes dup scripts in package.json


internal/bazel_integration_test/test_runner.js
```diff
@@ -271,6 +271,16 @@ const isWindows = process.platform === 'win32';
 const bazelBinary =
     require.resolve(`${config.bazelBinaryWorkspace}/bazel${isWindows ? '.exe' : ''}`);
 
+if (DEBUG) {
```
alexeagle: this should be VERBOSE_LOGS right? DEBUG should affect the build outputs, not affordances we need to see what the build system is doing

gregmagolan: At one point I had `DEBUG || VERBOSE_LOGS`. In this case `DEBUG` is what bazel_integration_test keys off of to keep the `tmp` folders around instead of having them auto-deleted so it would make sense to print out the `tmp` folder if it is being kept around.

gregmagolan: hmmm. actually `DEBUG || VERBOSE_LOGS` is more correct as you can CTRL-C the test halfway through with only VERBOSE_LOGS and the tmp folders are not deleted

PR #1194, opened 2019-09-25
test: cleanup e2e/typescript build file

gregmagolan:
These rollup_bundles snuck into e2e/typescript/BUILD.bazel but they are not testing anything


gregmagolan:
If I recall it was added with the node_repositories work to allow for downstream nodejs_binary support in the docker rules but I can't remember why. I think it was just a test the rollup_bundle built fine with ts_library output.

PR #1195, opened 2019-09-25
Terser sourcemap support

gregmagolan:


packages/terser/src/terser_minified.bzl
```diff
@@ -40,7 +40,7 @@ Can be a .js file, a rule producing .js files as its default output, or a rule p
 
 Note that you can pass multiple files to terser, which it will bundle together.
 If you want to do this, you can pass a filegroup here.""",
-        allow_files = [".js"],
+        allow_files = True,
```
alexeagle: why not continue to constrain it? do we no longer have a list of file extensions that should be legal?

packages/terser/src/terser_minified.bzl
```diff
@@ -95,6 +95,12 @@ so that it only affects the current build.
     ),
 }
 
+def _filter_js(files):
+    return [f for f in files if f.is_directory or f.extension == "js" or f.extension == "mjs"]
+
+def _filter_sourcemap(files):
```
alexeagle: nit: I don't feel like this function is carrying its weight, simple enough list comprehension to inline?

packages/terser/src/terser_minified.bzl
gregmagolan: It could be constrained to `.js` and `.mjs` and `.map` I  suppose and that would be fine. Should I change it to that?

packages/terser/src/terser_minified.bzl
gregmagolan: valid point 🌮

packages/terser/src/terser_minified.bzl
gregmagolan: I'll change it to that

gregmagolan: 🌮

PR #1197, opened 2019-09-26
refactor(builtin): determine correct root for linking packages

alexeagle:
This should work better than trial-and-error inside the process.


internal/linker/link_node_modules.bzl
```diff
@@ -95,7 +96,18 @@ def get_module_mappings(label, attrs, vars, srcs = [], workspace_name = None):
                     fail(("duplicate module mapping at %s: %s maps to both %s and %s" %
                           (label, k, mappings[k], v)), "deps")
                 _debug(vars, "target %s propagating module mapping %s: %s" % (dep, k, v))
-                mappings[k] = v
+
+                # A package which was reachable transitively via a _binary
+                # target is assumed to be in the runfiles of that binary,
```
gregmagolan: nit: "via a _binary **rule**"

I was thrown off by `target` as I first assumed it meant target name

internal/linker/link_node_modules.ts
```diff
@@ -260,30 +260,39 @@ export async function main(args: string[], runfiles: Runfiles) {
...
+    let target: string = '<package linking failed>';
+    if (kind === 'bin') {
+      // FIXME(#1196)
       target = path.join(workspaceRelative, bin, toWorkspaceDir(modulePath));
+      // Spend an extra FS lookup to give better error in this case
       if (!await exists(target)) {
-        // Try the execroot
-        target = path.join(workspaceRelative, toWorkspaceDir(modulePath));
+        return Promise.reject(`ERROR: no output directory found for package ${modulePath}
+        Did you mean to declare this as a from-source package?`);
```
gregmagolan: nit: add a // TODO: explain how to do this bit once there is a way

internal/linker/link_node_modules.ts
```diff
@@ -260,30 +260,39 @@ export async function main(args: string[], runfiles: Runfiles) {
...
-    links.push(linkModule(m, modules[m]));
+    const [kind, modulePath] = modules[m];
+    links.push(linkModule(m, kind, modulePath));
   }
 
-  await Promise.all(links);
+  let code = 0;
+  await Promise.all(links).catch(e => {
+    console.error(e);
+    code = 1;
```
gregmagolan: 🎉 nice catch

internal/linker/link_node_modules.ts
```diff
@@ -260,30 +260,45 @@ export async function main(args: string[], runfiles: Runfiles) {
...
 
   const linkModule =
-      async (name: string, modulePath: string) => {
-    let target = runfiles.resolve(modulePath);
-
-    // It sucks that we have to do a FS call here.
-    // TODO: could we know which packages are statically linked??
-    if (!target || !await exists(target)) {
-      // Try the bin dir
+      async (name: string, kind: 'bin'|'src'|'runfiles', modulePath: string) => {
```
gregmagolan: nit: a comment with a brief explanation of the 3 choices: `bin`,`src` and `runfiles` for when we come back to this file 6 months from now?

internal/linker/link_node_modules.ts
```diff
@@ -260,30 +260,45 @@ export async function main(args: string[], runfiles: Runfiles) {
...
+        // expected to declare ahead of time where the package is loaded,
+        // how that relates to npm link scenarios,
+        // and where the configuration can go.
+        return Promise.reject(`ERROR: no output directory found for package ${modulePath}
+        Did you mean to declare this as a from-source package?
+        See https://github.com/bazelbuild/rules_nodejs/pull/1197
+        until this feature is properly documented.`);
       }
     }
+    if (kind === 'src') {
```
gregmagolan: nit: `else if`

internal/linker/link_node_modules.ts
```diff
@@ -260,30 +260,45 @@ export async function main(args: string[], runfiles: Runfiles) {
...
+      target = path.join(workspaceRelative, toWorkspaceDir(modulePath));
+    }
+    if (kind === 'runfiles') {
```
gregmagolan: nit: `else if`

internal/linker/link_node_modules.ts
alexeagle: `switch/case/break`!

internal/linker/link_node_modules.ts
alexeagle: extracted a typedef and commented there

gregmagolan: 🐙

internal/linker/link_node_modules.ts
gregmagolan: is there a widely accepted threshold? 2 or less use if/else if/else; 3 or more use switch/case/break/default?

PR #1201, opened 2019-09-26
feat(typescript): add JSEcmaScriptModuleInfo provider to ts_library outputs

gregmagolan:


examples/app/BUILD.bazel
```diff
@@ -20,18 +22,47 @@ ts_devserver(
 
 rollup_bundle(
     name = "bundle",
-    enable_code_splitting = False,
     entry_point = ":app.ts",
     deps = [":app"],
 )
 
+babel(
```
alexeagle: seems like we shouldn't try to do differential loading in every example, if that's not a built-in feature anymore

examples/kotlin/BUILD.bazel
```diff
@@ -21,28 +23,66 @@ kt_js_library(
     deps = [":kotlinx-html-js"],
 )
 
+# Copy boostrap to the bin folder as entry_point.js
```
alexeagle: nit: boostrap

examples/kotlin/BUILD.bazel
```diff
@@ -21,28 +23,66 @@ kt_js_library(
...
+    # TODO: make this example work with format = "esm"
+    format = "cjs",
+    output_dir = True,
     deps = [
         "@npm//kotlin",
         "@npm//kotlinx-html-js",
     ],
 )
 
+babel(
```
alexeagle: no need to have diff. loading in this example

examples/protocol_buffers/rollup.config.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  // indicate which modules should be treated as external
+  external: [
```
alexeagle: actually globals is the one feature that's still present in the new rule.

It's more ergonomic than having to repeat the things twice here.

BUT maybe we should drop the feature. The attrs should only support parts of the rollup CLI that differ between compilation units. So we could drop unless you can make an argument that a single rollup.config.js would need to be used in different packages where the globals differ.

internal/common/collect_es6_sources.bzl
```diff
@@ -36,9 +40,10 @@ def collect_es6_sources(ctx):
     # be collected if the file is a js file.
     if hasattr(ctx.attr, "entry_point"):
         non_rerooted_files += [s for s in ctx.files.entry_point if s.extension == "js"]
+
     for dep in ctx.attr.deps:
-        if hasattr(dep, "typescript"):
-            non_rerooted_files += dep.typescript.transitive_es6_sources.to_list()
+        if JSEcmaScriptModuleInfo in dep:
```
alexeagle: is this change needed?

packages/rollup/src/rollup_bundle.bzl
```diff
@@ -98,7 +99,7 @@ rollup_bundle(
 )
 ```
 """,
-        allow_single_file = [".js"],
+        allow_single_file = True,
```
alexeagle: can it be constrained at all?

it's easier to make this looser over time but never tighter

packages/rollup/test/integration/BUILD.bazel
```diff
@@ -0,0 +1,44 @@
+load("@build_bazel_rules_nodejs//internal/golden_file_test:golden_file_test.bzl", "golden_file_test")
```
alexeagle: could you move this integration test to a separate PR? a lot of extra code and we should discuss why have both this and the existing integration test

packages/typescript/src/internal/build_defs.bzl
```diff
@@ -269,6 +270,15 @@ def _ts_library_impl(ctx):
         devmode_compile_action = _devmode_compile_action,
         tsc_wrapped_tsconfig = tsc_wrapped_tsconfig,
     )
+
+    # Add in new JS providers
```
alexeagle: link to issue or design about adding more and deleting the "typescript" one

packages/rollup/src/rollup_bundle.bzl
gregmagolan: rollup bundle can take pretty much anything as an input given that a user can add any plugins so I think the only way to support everything is to be unconstrained

packages/rollup/test/integration/BUILD.bazel
gregmagolan: existing integration test was using the goldens from the old rollup_bundle to make sure that the outputs were identical; I expanded on that but wanted to keep the existing one as an assertion until the old rollup_bundle is removed

internal/common/collect_es6_sources.bzl
gregmagolan: nope. I'll take that out

examples/protocol_buffers/rollup.config.js
gregmagolan: hmmmmm. that's a good question. having it as an attribute means you can take advantage of starlark to populate the attribute.

examples/app/BUILD.bazel
gregmagolan: hmmm. maybe add ts_library to the differential_loading macro in examples/webapp instead?

examples/kotlin/BUILD.bazel
gregmagolan: bazel will complain about bootstrap.js matching the input... maybe `_bootstrap.js` for the bin folder.

examples/kotlin/BUILD.bazel
gregmagolan: agreed

packages/rollup/test/integration/BUILD.bazel
gregmagolan: I'll factor it out to a separate PR

examples/app/BUILD.bazel
gregmagolan: agreed. I'll put ts_library into examples/webapp so it shows the full path from ts_library -> rollup_bundle -> babel -> terser -> serve

packages/typescript/src/internal/build_defs.bzl
gregmagolan: 👍

examples/kotlin/BUILD.bazel
```diff
@@ -21,28 +22,41 @@ kt_js_library(
...
-        # The injector will put two complimentary script tags in the index.html
-        ":bundle.min.js",
-        ":bundle.min.es2015.js",
-    ],
-    data = [":bundle"],
+    data = [":bundle.min"],
     index_html = "index.html",
 )
 
+# TODO: fix prodserver; require.js is needed if rollup_bundle output is cjs
```
alexeagle: can remove since we have only esm

examples/kotlin/yarn-error.log
```diff
@@ -0,0 +1,56 @@
+Arguments:
```
alexeagle: oops, delete

examples/protocol_buffers/rollup.config.js
alexeagle: resolved by removing it

packages/labs/src/protobufjs/ts_proto_library.bzl
```diff
@@ -104,16 +107,29 @@ def _ts_proto_library(ctx):
...
+            DefaultInfo(files = declarations),
+            DeclarationInfo(
+                declarations = declarations,
+                transitive_declarations = declarations,
+            ),
+            JSEcmaScriptModuleInfo(
+                sources = es6_sources,
+            ),
+        ],
         typescript = struct(
```
alexeagle: does this do anything?

gregmagolan: yes. needed to update examples/protocol_buffers which ties ts_proto_library -> ts_library -> rollup_bundle

examples/kotlin/yarn-error.log
gregmagolan: 👍

examples/kotlin/BUILD.bazel
gregmagolan: actually to get this example working I had to use ```    # TODO: make this example work with format = "esm"
    format = "cjs",
``` in the rollup_bundle to match what the old rollup_bundle produced with its incorrect `format = "es"` bug

gregmagolan: node takes issue with `import('./hello.js')` since it doesn't know that its an esm module and it imports it as CJS

gregmagolan: the prodserver never worked in the old example either

PR #1203, opened 2019-09-26
test(rollup): rollup integration test

gregmagolan:
New /packages/rollup integration test to replace the old one that relied on the /internal/rollup goldens.


alexeagle:
will review when it's green

internal/common/typescript_mock_lib.bzl
```diff
@@ -19,19 +19,42 @@ rules_typescript without introducing a circular dependency between
 rules_nodejs and rules_typescript repositories.
 """
 
+load("@build_bazel_rules_nodejs//:declaration_provider.bzl", "DeclarationInfo")
+load("@build_bazel_rules_nodejs//:providers.bzl", "JSEcmaScriptModuleInfo")
+
 def _mock_typescript_lib(ctx):
```
alexeagle: opt: we mocked typescript because the built-in couldn't depend on real ts_library but maybe now it should?

internal/common/typescript_mock_lib.bzl
```diff
@@ -19,19 +19,42 @@ rules_typescript without introducing a circular dependency between
...
-            transitive_declarations = transitive_decls,
+            es6_sources = es6_sources,
+            replay_params = None,
+            transitive_declarations = declarations,
+            transitive_es5_sources = es5_sources,
+            transitive_es6_sources = es6_sources,
+            tsickle_externs = [],
+            type_blacklisted_declarations = depset(),
         ),
+        providers = [
```
alexeagle: feels like it just wants to be a real thing

unless we want this to be "produces_jsprovider.bzl" and drop all references to typescript

packages/rollup/test/integration/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
+load("@build_bazel_rules_nodejs//internal/golden_file_test:golden_file_test.bzl", "golden_file_test")
+load("@npm_bazel_rollup//:index.from_src.bzl", "rollup_bundle")
+
+_BUNDLE_FORMATS = [
+    "amd",
```
alexeagle: do we really need coverage of all of them?

I can't imagine given our thin layering on top of rollup that we'd have a bug in only one format

gregmagolan: hmmmmm. I think the point was more to validate that all the possible values in the attribute constraints
```
values = ["amd", "cjs", "esm", "iife", "umd", "system"],
```
produced the expected bundle types. Verified that manually by looking at them and golden will validate that they continue to do so.

However, on that note, having the attribute constrained when rollup is a peerDependency seems wrong as well. If a newer version of rollup adds another bundle type in the future it won't be possible to use that without changing the constraints here. The constraints do have the benefit of validating that the user doesn't accidentally type `es` instead `ems` (for example) however it doesn't seem like the scope of the rule to prevent user typos and if they are expecting `ems` output and get `cjs` instead then they should notice on their end.

internal/common/typescript_mock_lib.bzl
gregmagolan: I think so too. I'll rip out mock_typescript_lib in a pre-factor PR

internal/common/typescript_mock_lib.bzl
gregmagolan: I'll rip out mock_typescript_lib in a pre-factor PR

packages/rollup/test/integration/BUILD.bazel
alexeagle: yeah maybe we shouldn't constrain them - but we should give a better error message cuz rollup will tell you to change the wrong thing

in theory we want to constrain when we add the JSModuleInfo provider so everyone is supposed to agree on the options

I think since we have the field at all, it's nice to duplicate the Rollup validation rules on it.

gregmagolan: ok. sounds like that decision is out of the scope of this PR so lets discuss elsewhere

gregmagolan:
https://github.com/bazelbuild/rules_nodejs/pull/1212 pre-factored out and should land first

internal/common/typescript_mock_lib.bzl
gregmagolan: Done

gregmagolan: in #1212

internal/common/typescript_mock_lib.bzl
gregmagolan: Done in #1212

packages/rollup/test/integration/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
...
+
+[
+    golden_file_test(
+        name = "test_%s" % format,
+        actual = "bundle.%s.js" % format,
+        golden = "//%s:golden.%s.js_" % (
+            package_name(),
+            format,
+        ),
+        # base64 encoded inline sourcemap data in golden files differs on Windows
```
alexeagle: let's use `sourcemap = "true"` so that the sourcemaps aren't inline

alexeagle: pushed a commit with that change, let's see if it's green on Windows

PR #1212, opened 2019-09-30
test: replace mock_typescript_lib with ts_library from index.from_src.bzl

gregmagolan:


examples/user_managed_deps/BUILD.bazel
```diff
@@ -19,13 +18,11 @@ filegroup(
     visibility = ["//visibility:public"],
 )
 
-mock_typescript_lib(
-    name = "mock_ts_lib",
+filegroup(
```
gregmagolan: ts_library from source does not work yet. see https://github.com/bazelbuild/rules_nodejs/issues/1213

PR #1224, opened 2019-10-01
test: add run_integration_test_debug package.json script to run integration test with DEBUG=1

gregmagolan:


package.json
```diff
@@ -77,6 +77,7 @@
         "test_e2e": "bazel --host_jvm_args=-Xms256m --host_jvm_args=-Xmx1280m test --test_tag_filters=e2e --local_resources=792,1.0,1.0 --test_arg=--local_resources=13288,1.0,1.0 ...",
         "test_examples": "bazel --host_jvm_args=-Xms256m --host_jvm_args=-Xmx1280m test --test_tag_filters=examples --local_resources=792,1.0,1.0 --test_arg=--local_resources=13288,1.0,1.0 ...",
         "run_integration_test": "bazel --host_jvm_args=-Xms256m --host_jvm_args=-Xmx1280m run --local_resources=792,1.0,1.0 --test_arg=--local_resources=13288,1.0,1.0",
+        "run_integration_test_debug": "bazel --host_jvm_args=-Xms256m --host_jvm_args=-Xmx1280m run --local_resources=792,1.0,1.0 --test_arg=--local_resources=13288,1.0,1.0 --define=DEBUG=1",
```
alexeagle: would be better chained I think, to reduce duplication?

"run_integration_test_debug": "npm run run_integration_test --define=DEBUG=1"

gregmagolan: `npm run run_integration_test --define=DEBUG=1` doesn't pass the `--define=DEBUG=1` forward for some reason but `yarn run_integration_test --define=DEBUG=1` does the trick.

PR #1226, opened 2019-10-02
feat(examples): updated to angular 8.2.8 in examples/angular

gregmagolan:


gregmagolan:
> Could you take a look at the renovate config so these stay more evergreen automatically?

👍

PR #1229, opened 2019-10-02
feat(examples): add closure compiler example

alexeagle:


Toxicable:
Is there no advantage to using `rules_closure` directly?

alexeagle:
rules_closure is super complicated and uses private closure APIs that keep breaking. I probably want to turn it down and port the minimal required feature set here. This is just starting to explore what's required

Toxicable:
@alexeagle that makes sense, thanks for whipping this up!

PR #1231, opened 2019-10-03
refactor: e2e and examples no longer nested workspaces

alexeagle:
Instead use .bazelignore so we can just glob() them up as a bunch of files.
This lets us run our tests on them without them exposing filegroups to the outer workspace, or having their dependencies installed in the root @npm workspace


package.json
```diff
@@ -68,7 +66,7 @@
         "typescript": "3.1.6",
         "unidiff": "1.0.1",
         "v8-coverage": "1.0.9",
-        "webpack": "~4.29.3",
+        "webpack": "~4.41.0",
```
gregmagolan: still needed in root package.json for tests under `//packages/labs/test...`

PR #1234, opened 2019-10-04
refactor: switch collect_es6_sources to use JSEcmaScriptModuleInfo

gregmagolan:
collect_es6_sources is going to be removed along with the legacy rollup_bundle but for now this change removes one of the few uses of the legacy “typescript” provider remaining.

### DIFF
--- a/internal/common/collect_es6_sources.bzl
+++ b/internal/common/collect_es6_sources.bzl
@@ -18,6 +18,10 @@ These are expected to be used by the production toolchain, such as bundlers.
 The tree will be flattened, such that all the es6 files are under a single tree.
 """
 
+load("@build_bazel_rules_nodejs//:providers.bzl", "JSEcmaScriptModuleInfo")
+
+# TODO(gregmagolan): Only the legacy rollup_bundle depends on this as it still uses
+# re-rooting. This can be removed along with the legacy rollup_bundle.
 def collect_es6_sources(ctx):
     """Returns a file tree containing only production files.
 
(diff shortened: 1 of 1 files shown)


internal/common/collect_es6_sources.bzl
```diff
@@ -18,6 +18,10 @@ These are expected to be used by the production toolchain, such as bundlers.
 The tree will be flattened, such that all the es6 files are under a single tree.
 """
 
+load("@build_bazel_rules_nodejs//:providers.bzl", "JSEcmaScriptModuleInfo")
+
+# TODO(gregmagolan): Only the legacy rollup_bundle depends on this as it still users
```
jbedard: "users" => "uses" ?

jbedard:
Worth moving this beside the rollup rule so they can all be deleted together? 🤷‍♂

internal/common/collect_es6_sources.bzl
gregmagolan: 👍

gregmagolan:
> Worth moving this beside the rollup rule so they can all be deleted together? 🤷‍♂

Probably not as the angular repo still uses it from this location (`load("@build_bazel_rules_nodejs//internal/common:collect_es6_sources.bzl", collect_es2015_sources = "collect_es6_sources")`) so that would be another thing to patch until angular is switched over to the new rollup_bundle (in progress).

gregmagolan:
Depends on https://github.com/bazelbuild/rules_nodejs/pull/1238 which should land first

PR #1238, opened 2019-10-04
feat(examples): update examples/angular to new rollup_bundle

gregmagolan:
Note: the new rollup_bundle will still have trouble with ng_module outputs that are non-Ivy as the generated `.ngfactory` & `.ngsummary` files have absolute imports that are not resolvable without the require patch. There is also a flaky `'default' is not exported` error that shows up while bundling much of the time but not always related to the bundling the `.ngfactory` & `.ngsummary` files. In short, there is some work to do to get the new rollup_bundle working with ViewEngine.


examples/angular/src/lib/shorten/BUILD.bazel
```diff
@@ -4,9 +4,8 @@ package(default_visibility = ["//:__subpackages__"])
 
 ts_library(
     name = "shorten",
-    srcs = ["shorten.ts"],
+    srcs = ["index.ts"],
```
gregmagolan: `module_root` not yet supported by linker

examples/angular/patches/@angular+bazel+9.0.0-next.8.patch
```diff
@@ -33,3 +54,22 @@ index 9b88fbb..d2eb8aa 100755
...
          declaration_files.append(bundle_index_typings)
          if is_legacy_ngc:
+@@ -603,7 +604,17 @@ def ng_module_impl(ctx, ts_compile_actions):
+     return providers
+ 
+ def _ng_module_impl(ctx):
+-    return ts_providers_dict_to_struct(ng_module_impl(ctx, compile_ts))
++    ts_providers = ng_module_impl(ctx, compile_ts)
++
++    # Add in new JS providers
```
alexeagle: will you send the Angular PR also if you haven't already?

examples/angular/src/BUILD.bazel
```diff
@@ -43,6 +45,13 @@ ts_library(
     ],
 )
 
+filegroup(
```
alexeagle: needs a comment about why to do this?
Also it's pretty gross that it ends up in user code. Maybe we should just use `tsc()` from `@npm//typescript:index.bzl` to transpile the initialize_testbed? but then we're further from the g3 approach

packages/karma/src/karma_web_test.bzl
```diff
@@ -129,14 +129,16 @@ def _write_karma_config(ctx, files, amd_names_shim):
...
+    for dep in ctx.attr.runtime_deps:
+        if hasattr(dep, "typescript"):
+            for src in dep.typescript.es5_sources.to_list():
+                runtime_files.append(_to_manifest_path(ctx, src))
+        elif hasattr(dep, "files"):
+            # These are javascript files provided by DefaultInfo from a direct
+            # dep that has no "typescript" provider. These files must be in named
+            # AMD or named UMD format.
+            for src in dep.files.to_list():
+                runtime_files.append(_to_manifest_path(ctx, src))
```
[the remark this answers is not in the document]
gregmagolan: I think we still want users to be able to pass in their own JS files if they know what they are doing (they need to be named-AMD or named-UMD). The JSNamedModuleInfo PR transforms this to:
```
    for dep in ctx.attr.runtime_deps:
        if JSNamedModuleInfo in dep:
            for src in dep[JSNamedModuleInfo].sources.to_list():
                runtime_files.append(_to_manifest_path(ctx, src))
        if not JSNamedModuleInfo in dep and not NodeModuleInfo in dep and hasattr(dep, "files"):
            # These are javascript files provided by DefaultInfo from a direct
            # dep that has no JSNamedModuleInfo provider or NodeModuleInfo
            # provider (not an npm dep). These files must be in named AMD or named
            # UMD format.
            for src in dep.files.to_list():
                runtime_files.append(_to_manifest_path(ctx, src))
```

examples/angular/src/lib/shorten/BUILD.bazel
alexeagle: yeah let's add a package.json here with a main field to fix this in a followup

gregmagolan: roger. I'll add a TODO

examples/BUILD.bazel
```diff
@@ -20,6 +20,8 @@ example_integration_test(
     npm_packages = {
         "//packages/karma:npm_package": "@bazel/karma",
         "//packages/protractor:npm_package": "@bazel/protractor",
+        "//packages/rollup:npm_package": "@bazel/rollup",
```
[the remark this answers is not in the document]
gregmagolan: I agree. follow up PR I think so we don't mix concerns.

examples/angular/patches/@angular+bazel+9.0.0-next.8.patch
gregmagolan: already added to https://github.com/angular/angular/pull/32889

examples/angular/src/BUILD.bazel
gregmagolan: hmmmm. I think this change was only required with the JSNamedModuleInfo as that bring in all the transitives. I can remove it from here and kick the ball a bit further on this.

packages/karma/src/karma_web_test.bzl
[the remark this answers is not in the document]
gregmagolan: This change was also only required with the JSNamedModuleInfo as that bring in all the transitives. I can remove it from here and kick the ball a bit further on this.

PR #1246, opened 2019-10-07
chore: remove sourcemap from checked in ts

alexeagle:
This causes merge conflicts on any changes to the TS files and makes our distro big


alexeagle:
The problem with the switch to `tsc` is that we have some tests written in TS that depend on this code, and their `ts_library` doesn't permit arbitrary DefaultInfo-only rules in the deps. I think this is fair - you need to wrap `tsc` to at least produce `DeclarationInfo` in order to participate with other TS rules.

PR #1247, opened 2019-10-07
feat: document the escape hatch from ts_library

alexeagle:
This will give us enough room to avoid major `ts_library` changes for 1.0

### DIFF
--- a/examples/app/BUILD.bazel
+++ b/examples/app/BUILD.bazel
@@ -1,9 +1,10 @@
 load("@build_bazel_rules_nodejs//internal/web_package:web_package.bzl", "web_package")
 load("@npm//http-server:index.bzl", "http_server")
+load("@npm//typescript:index.bzl", "tsc")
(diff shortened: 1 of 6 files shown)


examples/app/BUILD.bazel
```diff
@@ -46,20 +47,25 @@ http_server(
...
-    srcs = ["app.e2e-spec.ts"],
-    tsconfig = ":tsconfig-test",
-    deps = [
+    outs = [
+        "app.e2e-spec.js",
+    ],
+    args = [
+        "-p",
+        "$(location tsconfig-test.json)",
+        "--outDir",
```
gregmagolan: comment explaining what `--outDir $@` expands to and why it is needed may be good for the purposes of this example

alexeagle: yes, also added a sentence to the `npm_package_bin` docs

PR #1248, opened 2019-10-08
docs(examples): add Gatsby example

bweston92:
## PR Checklist

## PR Type

Add an example for Gatsby sites.

## Other information

Currently struggling to get it working.


examples/gatsby/.bazelrc
```diff
@@ -0,0 +1 @@
+common --experimental_allow_incremental_repository_updates
```
alexeagle: see other examples - this ought to include our common bazelrc

examples/gatsby/.gitignore
```diff
@@ -0,0 +1,69 @@
+# Logs
+logs
+*.log
+npm-debug.log*
+yarn-debug.log*
+yarn-error.log*
+
+# Runtime data
```
alexeagle: reduce this file down to the lines needed in the example

examples/gatsby/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+  "endOfLine": "lf",
```
alexeagle: I don't think it's illustrative to have a formatter in this example, it's orthogonal to what we want to show

examples/gatsby/BUILD.bazel
```diff
@@ -0,0 +1,27 @@
+load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary")
+
+nodejs_binary(
```
alexeagle: should be

```
load("@npm//gatsby-cli:index.bzl", "gatsby")
```
which is a nodejs_binary already set up to run that bin

bweston92:
It shouldn't be green, it doesn't run. I will update tomorrow.

examples/gatsby/.prettierrc
bweston92: Whoops that is from the `gatsby new` tool.

bweston92: deleted.

alexeagle:
to fix the CI you need to `yarn bazel:update-deleted-packages`
(see comment in /.bazelrc)

examples/gatsby/BUILD.bazel
```diff
@@ -0,0 +1,25 @@
...
+    name = "build",
+    args = ["build"],
+    data = [
+        ":src",
+        "@npm//:node_modules",
+    ],
+)
+
+filegroup(
+    name = "src",
```
alexeagle: this rule comes first in the depgraph so I'd usually have it come first in the file

examples/gatsby/BUILD.bazel
```diff
@@ -0,0 +1,25 @@
+load("@npm//gatsby-cli:index.bzl", "gatsby")
+
+gatsby(
```
alexeagle: this has no outs[] or output_dir which means it doesn't do anything unless you `bazel run :build` - is that what you intend? how would you access the built artifacts? I would expect this target to be a dependency of a devserver or something

examples/gatsby/BUILD.bazel
```diff
@@ -0,0 +1,25 @@
...
+            "config/*.js",
+            "plugins/**/*",
+            "src/**/*",
+            "gatsby-*.js",
+        ],
+        exclude = ["node_modules/**"],
+    ) + [
+        "package.json",
+    ],
+)
```
alexeagle: can you add some test that proves we did the right thing?

examples/gatsby/BUILD.bazel
bweston92: Once I got the build working I would progress to that yes, at the minute can't get the build working.

Issue #1249, opened 2019-10-08
New rollup_bundle not working with Angular ViewEngine


gregmagolan:
Two issues with the generated .ngfactory/.ngsummary ViewEngine files create problems with the new rollup bundle.

1) Some of the requires in those generated files have `external/` in their paths. I think I can work around this with some rollup.config code that fixes these import paths

2) I ran into a flake where rollup complained that `default is not exported` from the commonjs named-AMD generated `.ngfactory/.ngsummary` files. I'll need to debug this. One possible resolution is to generate esm files in the ngc postinstall step and then use babel to transform them to named-UMD which is required for ts_devserver.

alexeagle:
fixed by #1252

PR #1252, opened 2019-10-09
feat(examples): add Angular ViewEngine example that uses new rollup_bundle

gregmagolan:
Working out the last kinks with ViewEngine and the new rollup_bundle rule


gregmagolan:
Fixes https://github.com/bazelbuild/rules_nodejs/issues/1249

PR #1255, opened 2019-10-09
feat(examples): demonstrate using Webpack to build and serve a React app

alexeagle:
`yarn serve` works and runs a production-style build. The watch mode is too slow; we need to add a webpack-dev-server which should be ibazel-aware

/cc @lewish


alexeagle:
Code is from @lewish so I'm treating myself as the reviewer, and approved!

PR #1263, opened 2019-10-10
refactor: all providers exported from //:providers.bzl

gregmagolan:
Put all shared provider code into .bzl files in /providers folder and re-export all from //:providers.bzl so that users have a consistent load location. Legacy load locations kept around so dependency sandwich with angular/angular is simpler to resolve. Should be able to remove them by 1.0.

### DIFF
--- a/providers.bzl
+++ b/providers.bzl
@@ -1,90 +1,48 @@
-"""Providers for interop between JS rules.
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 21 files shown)


providers.bzl
```diff
@@ -1,90 +1,48 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Public providers, aspects and helper surface is re-exported here.
+
+Users should not load files under "/providers"
```
alexeagle: if we really don't want users loading them, I think they should be `/internal/providers/*.bzl`

alexeagle: I pushed the commit for this since you're on the road

PR #1264, opened 2019-10-11
fix(terser): terser_minified should support .mjs files when running on directory

gregmagolan:
Directory mode should match file mode which already allows .mjs files

```
def _filter_js(files):
    return [f for f in files if f.is_directory or f.extension == "js" or f.extension == "mjs"]
```


gregmagolan:
@alexeagle There are probably a few other rules that filter to allow only `.js` that probably need updating as well

PR #1268, opened 2019-10-11
chore(examples): upgrade ruels_docker

Toxicable:
this removes the need for
--host_force_python=PY2

### DIFF
--- a/examples/angular/WORKSPACE
+++ b/examples/angular/WORKSPACE
@@ -108,10 +108,14 @@ http_archive(
 
 http_archive(
     name = "io_bazel_rules_docker",
-    sha256 = "aed1c249d4ec8f703edddf35cbe9dfaca0b5f5ea6e4cd9e83e99f3b0d1136c3d",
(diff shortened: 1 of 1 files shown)


examples/angular/.bazelrc
```diff
@@ -1,4 +1,8 @@
 import %workspace%/../../common.bazelrc
 # Use the Angular IVy compiler
 # See https://github.com/angular/angular/blob/master/docs/BAZEL.md#various-flags-used-for-tests
-build --define=compile=aot
\ No newline at end of file
+build --define=compile=aot
+
+# Override the one in common
```
alexeagle: why override it instead of removing it?

Toxicable: There's a comment in the common saying it's needed for `pkg_tar`, is this no longer true?

Toxicable: ah wait, I was reading the wrong .bazelrc, not common

PR #1272, opened 2019-10-14
WIP: feat(rollup): Added ability to bundle ts_libraries that use module_root

Kolpa:
## PR Checklist

## PR Type
- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently a `rollup_bundle` task that references a `ts_library` wont be able to link if the `ts_library` depends on the `module_root` attribute being correctly parsed. A showcase can be found here: https://github.com/Kolpa/bazel_rollup_module_root

## What is the new behavior?
The `module_root` attribute is applied to the `module_mappings` in the same way as the `rules_typescript`'s approach allowing for compilation that follows directories.

- [x] Yes
- [ ] No

if `module_root` is set to a specific file linking will currently fail.

### DIFF
--- a/internal/linker/link_node_modules.bzl
+++ b/internal/linker/link_node_modules.bzl
@@ -147,8 +147,8 @@ def get_module_mappings(label, attrs, vars, rule_kind, srcs = [], workspace_name
                 if short_path.startswith("../"):
                     short_path = "external/" + short_path[3:]
                 if not short_path.startswith(mr):
-                    fail(("all sources must be under module root: %s, but found: %s" % 
-                        (mr, short_path)))
+                    fail(("all sources must be under module root: %s, but found: %s" %
+                          (mr, short_path)))
 
     # since our module mapping is currently based on attribute names,
     # allow a special one to instruct the linker that the package has no output
--- a/internal/linker/link_node_modules.bzl
+++ b/internal/linker/link_node_modules.bzl
@@ -122,6 +122,34 @@ def get_module_mappings(label, attrs, vars, rule_kind, srcs = [], workspace_name
     elif label.workspace_root:
         mr = "%s/%s" % (label.workspace_root, mr)
 
+    # Consider the module_root attribute when resolving mappings to allow for ts_libraries compiled with it set.
+
+    if attrs.module_root and attrs.module_root != ".":
+        mr = "%s/%s" % (mr, attrs.module_root)
+        if attrs.module_root.endswith(".ts"):
+            if workspace_name:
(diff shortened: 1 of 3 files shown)


Kolpa:
@googlebot I signed it!

internal/linker/link_node_modules.bzl
```diff
@@ -122,6 +122,34 @@ def get_module_mappings(label, attrs, vars, rule_kind, srcs = [], workspace_name
...
+
+                # Execroot paths for external repositories should start with external/
+                # But the short_path property of file gives the relative path from our workspace
+                # instead. We must correct this to compare with the module_root which is an
+                # execroot path.
+                if short_path.startswith("../"):
+                    short_path = "external/" + short_path[3:]
+                if not short_path.startswith(mr):
+                    fail(("all sources must be under module root: %s, but found: %s" %
+                          (mr, short_path)))
```
Kolpa: this is currently lifted straight from here: https://github.com/bazelbuild/rules_typescript/blob/0ab3a13ab610ad033a3c1ebd8a2080f7f251b624/internal/common/module_mappings.bzl#L79-L107

packages/rollup/test/integration/fum/BUILD.bazel
```diff
@@ -7,7 +7,4 @@ js_library(
     srcs = ["index.js"],
     module_from_src = True,
     module_name = "fumlib",
-    # Don't allow deep imports under here,
-    # and give it the AMD name "fumlib", not "fumlib/index"
-    module_root = "index.d.ts",
```
Kolpa: was this intended here? I could not find any use of the module_root attribute on any build path. Also d.ts seems out of place for a js library.

alexeagle:
It doesn't seem like this PR does anything to support module_root in rollup_bundle?

packages/rollup/test/integration/fum/BUILD.bazel
alexeagle: yes, it's intended, I think there's a comment somewhere

internal/linker/link_node_modules.bzl
alexeagle: I think we want to drop this on purpose - new rollup bundle just uses npm mechanisms to resolve things so it has to be a `package.json#main` instead

Kolpa:
> It doesn't seem like this PR does anything to support module_root in rollup_bundle?

The patch to the linker allows all rules that use the linker to resolve module_root attributes, which also applies to rollup_bundle.

But i gathered that this is not the intended approach to module resolution anymore?

PR #1273, opened 2019-10-14
feat(builtin): support multiple dynamic deps of a package

alexeagle:


alexeagle:
killing dynamic_deps instead

PR #1274, opened 2019-10-15
feat(examples): add Jest example

purkhusid:
This PR adds an example of how you might create a Jest rule

Unfortunately Jest does not handle symlinks very well so we have to add some patches so that it behaves correctly(See: https://github.com/ecosia/bazel_rules_nodejs_contrib/issues/4#issuecomment-475291612)

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes
- [ x] Example:


purkhusid:
Note that this is not working yet since I can't get the autogenerated rule to resolve `babel-loader`

alexeagle:
@Globegitter I'm following the threads of your work on this.

@purkhusid if the problem is really that `jest-cli` doesn't have a dependency declared in its package.json on something it wants to `require()`, this can be solved now with an attribute of `yarn_install` that adds missing dependency edges to Bazel's graph:
```
dynamic_deps = {
        "jest-cli": "@npm//@babel/preset-env",
    },
```
but this doesn't seem to change anything for me. Let's skip that for now and unroll the macro one step.

```
nodejs_test(
      entry_point = "@npm//:node_modules/jest-cli/bin/jest.js",
      data = ["@npm//jest-cli:jest-cli", "@npm//:node_modules"] + kwargs.pop("data", []),
      **kwargs
    )
```
note the `@npm//:node_modules` there which just says that the jest binary should include *all* of node_modules as deps.

Even then, the failure still seems like a syntax error, not a module resolution problem:

```
    /usr/local/google/home/alexeagle/.cache/bazel/_bazel_alexeagle/34f8c1c4c845d6e8057f92b20b01a493/sandbox/linux-sandbox/17/execroot/examples_jest/bazel-out/k8-fastbuild/bin/test.sh.runfiles/examples_jest/index2.test.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import indexfile from './index.js';
                                                                                                    ^^^^^^^^^

    SyntaxError: Unexpected identifier

      at ScriptTransformer._transformAndBuildScript (../npm/node_modules/jest-runner/node_modules/@jest/transform/build/ScriptTransformer.js:537:17)
```

purkhusid:
@alexeagle `babel.config.js` was missing from `deps` so now I'm getting the expected error:

```
Cannot find module '@babel/preset-env' from '/private/var/tmp/_bazel_danielpurkhus/039f5a9ccff2ae48767831e8b1a82438/sandbox/darwin-sandbox/73/execroot/examples_jest/bazel-out/darwin-fastbuild/bin/test.sh.runfiles/examples_jest'
```

I believe that it's acutally babel that is complaining about not finding the package so this might be more complicated than I had imagined. So we actually have a dependency chain that looks like this:

```
jest-cli  -> babel-jest(Loaded in the jest.config.js) -> @babel/preset-env(loaded in the babel.config.js)
```

The runfiles folder structure looks like this:
```
- examples_jest
    - <source files>
- npm
    - node_modules
        - <modules>   
```

Is there any reason for that we can not symlink `npm/node_modules` into the root of the `examples_jest` workspace? That way I think most tools should work since they usually traverse the folder structure upwards until they find a `package.json` or a `node_modules` directory.

purkhusid:
@alexeagle @Globegitter If there are any ideas on how we might be able to recreate the traditional Node folder structure within the runfiles then I could take a stab at it. Bazel is still kind of new to me so I don't know all the available tools.

alexeagle:
I fixed the node_modules layout in https://github.com/bazelbuild/rules_nodejs/commit/bc848f9d58175dad727b3285f36e7db88e5aeb48 - the Jest example is passing for me now.

We should take out some of the patches and put them into the new bazel_require_script.js whose job is to patch the nodejs environment before calling the entry script.

alexeagle:
@purkhusid could you sign the CLA so I can merge this if green?

purkhusid:
@googlebot I signed it!

purkhusid:
@alexeagle You want the missing `fs` functions before you merge or just add them afterwards?

marcus-sa:
Doesn't seem to work for me.
At first I thought it'd be a flaky test, but it keeps running into module resolution issues.
It can't really be the wrong configuration, since it happens in this repo as well.

PR #1276, opened 2019-10-16
refactor: remove dynamic_deps feature

alexeagle:
This isn't the right way to declare that a rule depends on a plugin.

- It is too hard for users to understand what this means, how to
configure it correctly, and to discover that their broken tool needs a
dynamic dep.
- It is not npm-idiomatic. The dependency graph at install time is not
meant to reflect the actual dependencies that will be loaded at runtime.

In the rollup example, the `rollup.config.js` has `import
'rollup-plugin-json'`
so the corresponding BUILD file should have that package in the deps.

BREAKING CHANGE:

The dynamic_deps attribute of yarn_install and npm_install is removed,
in favor of declaring needed packages in the deps/data of the rule that
invokes the tool.


alexeagle:
This is broken because of the way directories are layed out. When we run the tool we have

```
execroot
- build_bazel_rules_nodejs/bazel-out/k8-fastbuild/bin/packages/rollup/test/plugins/_plugins.rollup_config.js [1]
- external/npm/rollup_binary.runfiles/node_modules/rollup [2]
- external/npm/node_modules/rollup-plugin-json [3]
- node_modules -> external/npm/node_modules
```

The custom module loader only looks under the runfiles.

Issue #1261, opened 2019-10-10
rollup_bundle doesn't support module_name with more then 2 segments


jbedard:
While linking the modules any `ng_module(module_name)`/`ts_library(module_name)` gets linked in based on that name, but there is a [2-segment restriction](https://github.com/bazelbuild/rules_nodejs/commit/94abf68#diff-ba12e7c9523323f8d5f10aa8a99d35dbR312) on that name to follow the node/npm conventions.

In terms of *npm* module names this restriction is correct. But previously `ng_module(module_name)` and `ts_library(module_name)` could be used in various ways beyond npm (or at least that's how I thought of it), essentially as aliases or AMD/CJS names etc.

moczix:
yeah, i use module_nam as path alias from tsconfig. And nowe we got this error

alexeagle:
We want to scale back our semantics to match npm. You'll have to have more module names `@my_org/one`, `@my_org/two` or use deep imports 
`deps=["//packages/one"]` -> `import '@my_org/one/thing/two`

moczix:
the problem is, that bazel dont read paths from tsconfig, thats why we use module_name with name starting exactly like alias. I tried now, to remove module_name at all and it dont work. I tried diffrent combinations, like have module_name with one segment, two segments and dont work. The main problem is that bazel dont read paths from tsconfig

Toxicable:
@ashi009 You can have a high level module name with deep import paths.
Say you have a module `@my-wksp/lib1` and you want to make the import `import {SomeClass} from @my-wksp/lib1/lvl1/lvl2'`

You'd start by defining the module at it's root to be `@my-wksp/lib1`.
You'd make sure that this package includes and compiles all files down in `lvl1` and `lvl2`
or
You'd make sure it depends other targets in those lower levels, making it more fine grained.

That way it hold sematics with npm and is easy to do.
Does that fit your use case?
>say export * from '/absolute/path/to/the/resolved/entry/point'

This wouldn't work since if the user tried to make a deep import it would bypass that generated import and the fiels woudln't exist, also inter-node_module requires wouldn't work either.

jbedard:
> You'd make sure that this package includes and compiles all files down in lvl1 and lvl2
> or
> You'd make sure it depends other targets in those lower levels, making it more fine grained.

[...]

PR #1281, opened 2019-10-17
fix(builtin): allow more than 2 segments in linker module names

### LINKED ISSUE #1278: uncaught error in linker
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: 

`rollup_bundle`

### Is this a regression?

Not sure

### Description

 A clear and concise description of the problem...
In case the build depends on a module that has more than 2 segment name (See #1261), linker throws, but exit with zero. Hence, the build will succeed but generate bad code.

## 🔬 Minimal Reproduction

N/A

## 🔥 Exception or Error

<pre><code>
$ bazel build ecnova/www/frontend:bundle
INFO: Analyzed target //ecnova/www/frontend:bundle (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
INFO: From Bundling JavaScript ecnova/www/frontend/bundle.js [rollup]:
(node:49238) ExperimentalWarning: The fs.promises API is experimental
(node:49238) UnhandledPromiseRejectionWarning: Error: module xrc/ecnova/common/frontend/runtimeconfig/store has more than 2 segments which is not a valid node module name
    at /private/var/tmp/_bazel_xiaoyi/2258f168242f85dcc9f2c0d3e9d2258a/external/build_bazel_rules_nodejs/internal/linker/index.js:272:27
    at Generator.next (<anonymous>)
    at fulfilled (/private/var/tmp/_bazel_xiaoyi/2258f168242f85dcc9f2c0d3e9d2258a/external/build_bazel_rules_nodejs/internal/linker/index.js:3:58)
(node:49238) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:49238) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

bazel-out/darwin-fastbuild/bin/ecnova/www/frontend/index.mjs → bazel-out/darwin-fastbuild/bin/ecnova/www/frontend/bundle.js...
(!) Unresolved dependencies
https://rollupjs.org/guide/en/#warning-treating-module-as-external-dependency
xrc/ecnova/common/frontend/googleanalytics (imported by bazel-out/darwin-fastbuild/bin/ecnova/www/frontend/index.mjs)
xrc/ecnova/common/frontend/navigator (imported by bazel-out/darwin-fastbuild/bin/ecnova/www/frontend/index.mjs)
xrc/ecnova/www/frontend/app/loader (imported by bazel-out/darwin-fastbuild/bin/ecnova/www/frontend/index.mjs)
react (imported by bazel-out/darwin-fastbuild/bin/ecnova/www/frontend/index.mjs)
react-dom (imported by bazel-out/darwin-fastbuild/bin/ecnova/www/frontend/index.mjs)
created bazel-out/darwin-fastbuild/bin/ecnova/www/frontend/bundle.js in 29ms
Target //ecnova/www/frontend:bundle up-to-date:
  bazel-bin/ecnova/www/frontend/bundle.js
INFO: Elapsed time: 106.641s, Critical Path: 32.73s
INFO: 366 processes: 305 darwin-sandbox, 61 worker.
INFO: Build completed successfully, 440 total actions
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
macOS 10.15 (19A583)
  </code>
</pre>

<pre>
  <code>
$ bazel version
Build label: 0.27.0
Build time: Mon Jun 17 13:03:58 2019 (1560776638)
Build timestamp: 1560776638
Build timestamp as int: 1560776638
  </code>
</pre>

**Rules version (SHA):**

<pre>
  <code>
urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/0.38.3/rules_nodejs-0.38.3.tar.gz"],
        sha256 = "ad4be2c6f40f5af70c7edf294955f9d9a0222c8e2756109731b25f79ea2ccea0",
  </code>
</pre>

**Anything else relevant?**

gregmagolan:
The linking code is made more robust to handle more than two segments in the module names. There is no technical reason to not allow this although it does diverge from the npm package standard of either `foo` or `@scope/foo` names.

We've had a number of users (including Angular) already using this feature with the legacy rollup_bundle and the restriction to 2 segments is a breaking change for them (see discussion in #1261).

Module names are sorted and grouped to ensure that nested modules such as `c` are symlinked first before `c/c`, `c/c/c/c`, etc and that groups of nested modules are symlinked in order while maintaining the parallelized symlink code for module names that are not nested. Gaps in the nested paths are filled with mkdir as needed and common paths between groups are mkdir'd before handing off to the parallelized symlink code.

Fixes #1278


gregmagolan:
Also unblocks https://github.com/angular/angular/pull/33201. The angular repo relies on deep module_names quite a bit and upgrading to the new rollup_bundle that has a 2 segment restriction would require moving some things around which I'd rather not do just to update a rule.

@alexeagle Perhaps we should add a 'strict' mode for the linker where it restricts the module_name usage to `foo` or `@scope/foo` but allow more than 2 segments by default?

alexeagle:
I'm not that worried about strict mode because when we change the module name to come from a package chase on it will look weird to have many path segments in the name of an NPM module

internal/linker/link_node_modules.ts
```diff
@@ -276,36 +297,72 @@ export async function main(args: string[], runfiles: Runfiles) {
...
+          grouped[grouped.length - 1].push(module);
+        } else {
+          grouped.push([module]);
+        }
+        return grouped;
+      }, []);
+
+  log_verbose(`grouped modules ${JSON.stringify(groupedModules)}`);
+  for (const group of groupedModules) {
+    // reduce links such as `@foo/b/c => /path/to/a/b/c` to their
```
gregmagolan: added this after discussing with @alexeagle night. If there is a `@foo/a/b/c/d` module mapping in your deps and it maps to a path `/path/to/foo/a/b/c/d`, for example, then the mapping can and should be reduced to `@foo` => `/path/to/foo`.

internal/linker/index.js
```diff
@@ -47,16 +47,33 @@ Include as much of the build output as you can without disclosing anything confi
...
         if (!fs.existsSync(p)) {
             mkdirp(path.dirname(p));
-            fs.mkdirSync(p);
+            log_verbose(`mkdir( ${p} )`);
+            try {
+                fs.mkdirSync(p);
+            }
+            catch (e) {
+                if (e.code !== 'EEXIST') {
+                    // can happen if path being created exists via a symlink
```
alexeagle: because of a race condition creating the directory? or because fs.existsSync doesn't detect symlinks?

internal/linker/index.js
```diff
@@ -47,16 +47,33 @@ Include as much of the build output as you can without disclosing anything confi
...
+                    throw e;
+                }
+            }
         }
     }
     function symlink(target, path) {
         return __awaiter(this, void 0, void 0, function* () {
             log_verbose(`symlink( ${path} -> ${target} )`);
+            if (!fs.existsSync(target)) {
+                // This can happen if a module mapping is propogated from a dependency
```
alexeagle: can we detect this case earlier somehow? we want to keep the FS operations to a minimum since this is overhead for every node process spawned

internal/linker/index.js
```diff
@@ -250,33 +269,66 @@ Include as much of the build output as you can without disclosing anything confi
...
-                        // FIXME(#1196)
-                        target = path.join(workspaceAbs, bin, toWorkspaceDir(modulePath));
-                        break;
-                    case 'src':
-                        target = path.join(workspaceAbs, toWorkspaceDir(modulePath));
-                        break;
-                    case 'runfiles':
-                        target = runfiles.resolve(modulePath) || '<runfiles resolution failed>';
-                        break;
+            const linkModules = (modules) => __awaiter(this, void 0, void 0, function* () {
```
alexeagle: nit: prefer syntax
`function foo() {}` rather than `const foo = () => {}`
that's in the goog js/ts style guide

internal/linker/link_node_modules.ts
```diff
@@ -276,36 +297,72 @@ export async function main(args: string[], runfiles: Runfiles) {
...
-      mkdirp(segments[0]);
-    }
-    const [kind, modulePath] = modules[m];
-    links.push(linkModule(m, kind, modulePath));
+  // Group nested modules as these need to be symlinked in order.
+  // For example, given a list of module keys such as:
+  // ['a', '@foo/c/c/c/c', 'b/b', 'b', '@foo/c', '@foo/c/c']
+  // this reducer should output the groups list:
+  // [ [ '@foo/c', '@foo/c/c', '@foo/c/c/c/c' ], [ 'a' ], [ 'b', 'b/b' ] ]
+  const groupedModules = Object.keys(modules).sort().reduce(
```
alexeagle: seems like it would be better to add unit tests for this function rather than testing all the conditions via the medium test that uses the filesystem

internal/linker/link_node_modules.ts
```diff
@@ -276,36 +297,72 @@ export async function main(args: string[], runfiles: Runfiles) {
...
+          target = path.join(workspaceAbs, toWorkspaceDir(m.modulePath));
+          break;
+        case 'runfiles':
+          target = runfiles.resolve(m.modulePath) || '<runfiles resolution failed>';
+          break;
+      }
 
-    await symlink(target, name);
+      // ensure the subdirectories exist
+      mkdirp(path.dirname(m.name));
```
alexeagle: extra FS calls inside mkdirp right? how did we deal with scoped packages before (there must be a mkdir somewhere else already?)

internal/linker/link_node_modules.ts
```diff
@@ -276,36 +297,72 @@ export async function main(args: string[], runfiles: Runfiles) {
...
+          // strip off the last segment as it is common
+          name = path.dirname(name);
+          modulePath = path.dirname(modulePath);
+          log_verbose(`module mapping ( ${name}/${bn} => ${modulePath}/${bmp} ) reduced to ( ${
+              name} => ${modulePath} )`);
+        } else {
+          break;
+        }
+      }
+      return {name: name, root: kind, modulePath};
```
alexeagle: nit: `name: name` same as just `name`

internal/linker/index.js
gregmagolan: the latter; fs.exists doesn't seem to look into directory symlinks which was surprising; fs.existsSync would return false but then mkdir would fail as the target directory already exists

internal/linker/link_node_modules.ts
gregmagolan: yes, there was a mkdirp already for the 2 segment case. for one segment module mappings path.dirname() will return `` and mkdirp is a noop.

internal/linker/link_node_modules.ts
gregmagolan: 👍

internal/linker/index.js
gregmagolan: 👍

internal/linker/link_node_modules.ts
gregmagolan: that's a good idea; I'll refactor the function so its exported and add some unit tests for it

internal/linker/index.js
gregmagolan: Can't think of a way to determine this earlier; we should discuss further. At the least I think this should be `await fs.promises.stat` so that an async operation.

internal/linker/link_node_modules.ts
gregmagolan: broke the function out and added a unit test that makes it very clear what it is supposed to do :)

```
  it('should group and reduce modules appropriate', () => {
    const modules = {
      'a': ['src', `${workspace}/path/to/lib_a`],
      '@foo/c/c/c/c': ['bin', `${workspace}/path/to/foo_cccc`],
      'b/b': ['src', 'other_wksp/path/to/lib_bb'],
      'b': ['src', 'other_wksp/path/to/lib_b'],
      '@foo/c': ['bin', `${workspace}/path/to/foo_c`],
      '@foo/c/c': ['bin', `${workspace}/path/to/foo_cc`],
      '@foo/d/bar/fum/far': ['bin', `${workspace}/path/to/foo_d/bar/fum/far`],
      '@foo/d/bar': ['bin', `${workspace}/path/to/foo_d/bar`],
      // don't include `@foo/d` as the group & reduce function  should derive that
      // from the lowest common denominator of the module name & module path
    };

    const result = linker.groupAndReduceModules(modules);

    const expected = [
      [
        // `@foo/c`, `@foo/c/c` and `@foo/c/c/c/c` should
        // be grouped but not reduced as their arrangement
        // on disk does not match the modules names
        {'name': '@foo/c', 'root': 'bin', 'modulePath': `${workspace}/path/to/foo_c`},
        {'name': '@foo/c/c', 'root': 'bin', 'modulePath': `${workspace}/path/to/foo_cc`},
        {'name': '@foo/c/c/c/c', 'root': 'bin', 'modulePath': `${workspace}/path/to/foo_cccc`}
      ],
      [
        // `@foo/d/bar/fum/far` and `@foo/d/bar` should
        // be grouped and reduced to `@foo/d`
        {'name': '@foo/d', 'root': 'bin', 'modulePath': `${workspace}/path/to/foo_d`},
      ],
      [{'name': 'a', 'root': 'src', 'modulePath': `${workspace}/path/to/lib_a`}],
      [
        // `b`, `b/b` should be grouped but not reduced as
        // their arrangement on disk does not match the modules names
        {'name': 'b', 'root': 'src', 'modulePath': 'other_wksp/path/to/lib_b'},
        {'name': 'b/b', 'root': 'src', 'modulePath': 'other_wksp/path/to/lib_bb'}
      ]
    ];

    expect(JSON.stringify(result)).toEqual(JSON.stringify(expected));
  })
```

internal/linker/test/link_node_modules.spec.ts
```diff
@@ -32,6 +32,48 @@ describe('link_node_modules', () => {
...
+      ],
+      [
+        // `@foo/d/bar/fum/far` and `@foo/d/bar` should
+        // be grouped and reduced to `@foo/d`
+        {'name': '@foo/d', 'root': 'bin', 'modulePath': `${workspace}/path/to/foo_d`},
+      ],
+      [{'name': 'a', 'root': 'src', 'modulePath': `${workspace}/path/to/lib_a`}],
+      [
+        // `b`, `b/b` should be grouped but not reduced as
+        // their arrangement on disk does not match the modules names
```
alexeagle: do you know whether this case occurs today? it feels like maybe we should error for this until we know why it should be supported. could someone be relying on this behavior outside of Bazel?

gregmagolan: That's a good question. Outside of bazel someone could do something like this in theory. @jbedard may be relying on this with the current rollup_bundle.

gregmagolan: Discussed in person. Decided that it is better to allow this behavior as it is possible to do something similar outside of bazel with module mapping.

PR #1287, opened 2019-10-22
refactor: use deps declared on rollup_bundle

alexeagle:
This replaces the dynamic_deps mechanism for pulling user-specified plugins into the node_modules tree

### DIFF
--- a/internal/node/bazel_require_script.js
+++ b/internal/node/bazel_require_script.js
@@ -1,0 +1,38 @@
+// Adapt node programs to run under Bazel
+// Meant to be run in a --require hook
+
+const fs = require('fs');
+const path = require('path');
+const orig = {};
+// TODO: more functions need patched like
+// the async and native versions
+orig['realPathSync'] = fs.realpathSync;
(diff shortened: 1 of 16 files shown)


internal/node/bazel_require_script.js
```diff
@@ -0,0 +1,38 @@
...
+// TODO: more functions need patched like
+// the async and native versions
+orig['realPathSync'] = fs.realpathSync;
+orig['lstatSync'] = fs.lstatSync;
+
+// To fully resolve a symlink requires recursively
+// following symlinks until the target is a file
+// rather than a symlink, so we must make this look
+// like a file.
+function lstatSync(p) {
```
soldair: can we call fs.statSync here instead of lstatSync? 
this would allow it to work as expected for links to directories.

alexeagle: discussed in-person: this script is currently the minimum to make rollup work and will need expansion in later PRs as we add tests for things like Jest that exercise the APIs more. @soldair will own it generally.

PR #1292, opened 2019-10-23
feat(examples): use closure-templates aka soy in the closure example

alexeagle:
This seems like a nice idea but fails with

```
bazel-out/k8-fastbuild/bin/index.soy.js:13 (JSC_MISSING_NAMESPACE_IMPORT)
Imported Closure namespace "goog.soy.data.SanitizedContentKind" never defined.
var SanitizedContentKind = goog.require('goog.soy.data.SanitizedContentKind');
^

bazel-out/k8-fastbuild/bin/index.soy.js:14 (JSC_MISSING_NAMESPACE_IMPORT)
Imported Closure namespace "google3.javascript.template.soy.api_idom" never defined.
var incrementaldomlib = goog.requireType('google3.javascript.template.soy.api_idom');
^
```

(note the `google3` in there...)
/cc @iteriani

### DIFF
--- a/examples/closure/BUILD.bazel
+++ b/examples/closure/BUILD.bazel
@@ -1,6 +1,23 @@
 load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_test")
 load("@npm//google-closure-compiler:index.bzl", "google_closure_compiler")
 
+java_binary(
+    name = "google_closure_templates_compiler",
+    main_class = "com.google.template.soy.SoyToIncrementalDomSrcCompiler",
+    runtime_deps = ["@maven//:com_google_template_soy"],
+)
+
+genrule(
+    name = "compile_soy",
+    srcs = ["index.soy"],
+    outs = ["index.soy.js"],
+    cmd = """$(location google_closure_templates_compiler) \
+      --srcs $(SRCS) \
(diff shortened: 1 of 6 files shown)


alexeagle:
FYI @Yannic if we could get this working it would be a good model for how to start to simplify rules_closure

Yannic:
(haven't really looked at the PR yet, so apologies if you already know this)
The google3 in there is hard-coded in the Java sources :/ (https://github.com/google/closure-templates/blob/7132b6ac3001718b31891f61c584772e55d54f69/java/src/com/google/template/soy/incrementaldomsrc/IncrementalDomRuntime.java#L41 I'd love to have this configurable, but haven't gotten to file an issue for it). I believe there is an opensource shim providing the google3 namespace, so you'll need to include that file in the soy runtime (https://github.com/google/closure-templates/blob/master/javascript/shim.js).

I was originally thinking of moving the (new, see https://github.com/bazelbuild/rules_closure/issues/389) soy-specific rules to the soy repo. I'll have a closer look at this PR next week.

As said before, I think it might be a good idea to schedule a meeting so we can talk about the future of rules_closure. I have some uses of closure that probably won't work with what I've seen in rules_nodejs (they actually don't even work with OSS rules_closure, but that's a different story).

alexeagle:
replaced by #1692

Issue #968, opened 2019-06-28
Use the newly added dependency pruning


pauldraper:
Bazel 0.27 has finally allowed Starlark access to dependency pruning, a feature that has previously been reserved only for native rules.

https://github.com/bazelbuild/bazel/commit/57363813d3c72129c61c35a2b03c3033b62cb49d

Implement `unused_inputs_list` for TypeScript compilation, thereby improving cache hits.

pauldraper:
Correct. The requirement of `unused_inputs_list` is this:

> The content of this file (generally one of the outputs of the action) corresponds to the list of input files that were not used during the whole action execution. Any change in those files must not affect in any way the outputs of the action.

TypeScript mostly fits that requirement.

Ambient declarations present a problem however. Not sure if it's a fixable one or not.

DavidANeil:
Currently we pass all transitive dependencies into the typescript compiler, and it reads them all.
This is a performance issue, so I (and, I believe, Evan Martin) am looking into ways to get `ts_library` to not need everything passed in through the `files` config.

Once that is done, we can look more into getting this working.

PR #1293, opened 2019-10-23
WIP: feat(typescript): use unused inputs

DavidANeil:
Resolves bazelbuild/rules_nodejs#968
Depends on bazelbuild/rules_typescript#470


DavidANeil:
I reckon this could be extended to allow the compiler to error and print a buildozer command if no files from a dep are used.

I got this far, wanted to test it, learned I couldn't easily test, and decided to just throw it here.

alexeagle:
Thanks - at Bazelcon I talked to @JaredNeil about re-writing ts_library as a much simpler wrapper around `tsc` and using all the modern things. Let's try that in the new year? For now I'll close this to keep our PRs tidy

PR #1299, opened 2019-10-24
chore: make defs.bzl error

### LINKED ISSUE #1068: Consistent load() locations
We should use `index.bzl` as the public API for all packages

alexeagle:
BREAKING CHANGE:
Users must now switch to loading from index.bzl

Fixes #1068

### DIFF
--- a/defs.bzl
+++ b/defs.bzl
@@ -12,77 +12,53 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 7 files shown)


alexeagle:
it's pretty annoying to land this right now. Should probably wait for rules_sass and rules_docker to accept my PRs and do a release so the patches aren't going to be required for all our users?

OR maybe change this to just warn instead of error?

PR #1302, opened 2019-10-24
chore: update bazel-toolchains

alexeagle:


gregmagolan:
These are meant to go with Bazel 1.1.0. I think we should update @bazel/bazel to 1.1.0 in the repo as well.

Note: this version of bazel toolchains has trouble with RBE on CircleCI due to a docker access issue (new error since 1.0.0) but we're not using RBE here so doesn't affect us. See comment here for more details: https://github.com/angular/angular/pull/33367#issue-331848454.

alexeagle:
updated bazel too

gregmagolan:
> updated bazel too

You tripped my wire `package.json @bazel/bazel '^1.1.0' does not match BAZEL_VERSION in //:index.bzl '0.28.1'`. Checks so that bazel_integration_test is using the correct bazel verison.

PR #1320, opened 2019-10-30
fix: fix nodejs_binary cross-platform RBE issue #1305

### LINKED ISSUE #1305: nodejs_binary incorrectly chooses the target platform with RBE
It appears that `nodejs_binary` incorrectly chooses the platforms for execution of its outputs.

I believe this is currently controlled by the block found [here](https://github.com/bazelbuild/rules_nodejs/blob/10d136adbca8e4d3fef90c2068ac44bbda223a18/toolchains/node/BUILD.bazel#L74)

I created a demo/proof in this commit: https://github.com/josephperrott/angular/commit/8dbec8cfac8801a1dbbdb887434109dde8589859

Running `yarn bazel run //tools/rbe-test:bin` locally, everything builds and executes as expected and you get the expected console.log message.
```
$ yarn -s bazel run //tools/rbe-test:bin
INFO: Invocation ID: ce95d7c0-b0d2-4c30-a280-25d015ed331c
...
run!
✨  Done in 1.55s.
```
When running `yarn bazel run //tools/rbe-test:bin --config=remote` to force remote execution, everything builds but then on attempted execution it fails as it can't run the binary
```
$ yarn -s bazel run //tools/rbe-test:bin --config=remote
WARNING: Option 'tls_enabled' is deprecated: Deprecated. Please specify a valid protocol in the URL (https or grpcs).
WARNING: option '--jobs' was expanded to from both option '--config=remote' (source command line options) and option '--config=remote' (source command line options)
INFO: Invocation ID: 26cf4261-e1fb-4b2b-a68f-8f1dc762d7ac
...
INFO: Build completed successfully, 60 total actions
/private/var/tmp/_bazel_jperrott/104b07d8216361462e08a3d288166690/execroot/angular/bazel-out/darwin-fastbuild/bin/tools/rbe-test/bin.sh: line 151: /private/var/tmp/_bazel_jperrott/104b07d8216361462e08a3d288166690/external/nodejs_linux_amd64/bin/nodejs/bin/node: cannot execute binary file
/private/var/tmp/_bazel_jperrott/104b07d8216361462e08a3d288166690/execroot/angular/bazel-out/darwin-fastbuild/bin/tools/rbe-test/bin.sh: line 151: /private/var/tmp/_bazel_jperrott/104b07d8216361462e08a3d288166690/external/nodejs_linux_amd64/bin/nodejs/bin/node: Undefined error: 0
```

As you can see from the error message, it is attempting to run the node binary from nodejs_*linux*_amd64 rather than nodejs_*darwin*_amd64.  This is controlled by line 118 of the `bin.sh` script that is created for the build:
```
readonly node=$(rlocation "nodejs_linux_amd64/bin/nodejs/bin/node")
```

Interestingly the line below, 119, correctly identifies the platform:
```
readonly repository_args=$(rlocation "nodejs_darwin_amd64/bin/node_repo_args.sh")
```

gregmagolan:
First commit reproduces https://github.com/bazelbuild/rules_nodejs/issues/1305.

Repo here: https://buildkite.com/bazel/rules-nodejs-nodejs/builds/4727#5a880596-2b15-49a3-947c-a7b2894f6cca from https://github.com/bazelbuild/rules_nodejs/pull/1324.

A bit hacky but the issue can be reproduced in run_targets in bazelci. There doesn’t seem to be a run_flags to pass `—config=remote` so I used shell_commands to set the `--platform` to linux in a new `mac_fake_rbe` job on buildkite. This sets up the same scenario where the build with --platform linux selected (as it would be on RBE) but the resulting nodejs_binary runnable target is run on osx.

2nd commit fixes the issue.

* node_launcher.sh now picks the node binary at runtime instead of it being a templated arg evaluated at build time
* node_binary now adds the correct node tool_files to the nodejs_binary runfiles

3rd commit updates the nodejs toolchain test to no longer test for runfiles. What should be tested is that `ctx.toolchains["@build_bazel_rules_nodejs//toolchains/node:toolchain_type"].nodeinfo.tool_files` are for the correct selected `--platform`.

Fixes #1305.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

nodejs_binary does not work with cross-platform RBE. For example, developer machine is osx and RBE is a linux box

## What is the new behavior?

nodejs_binary works with cross-platform RBE

- [ ] Yes
- [x] No


gregmagolan:
Still have to update vendored_node support after this fix its broken

internal/node/node.bzl
```diff
@@ -139,6 +139,9 @@ def _to_execroot_path(ctx, file):
 
     return ("<ERROR> _to_execroot_path not yet implemented for " + file.path)
 
+def _strip_external(path):
+    return path[len("external/"):] if path.startswith("external/") or path.startswith("external\\") else path
```
alexeagle: nit: I think we have similar logic in other places, just curious if this matches?

internal/node/node_launcher.sh
```diff
@@ -114,7 +114,24 @@ TEMPLATED_env_vars
...
+    MINGW*)     machine=windows;;
+    MSYS_NT*)   machine=windows;;
+    *)          machine=linux;;
+esac
+
+case "${machine}" in
+  linux) readonly node=$(rlocation "TEMPLATED_node_linux");;
+  darwin) readonly node=$(rlocation "TEMPLATED_node_darwin");;
+  windows) readonly node=$(rlocation "TEMPLATED_node_windows");;
+  *) readonly node=$(rlocation "TEMPLATED_node_linux");;
```
alexeagle: should log something noisy in this case, maybe even tell user to file an issue and include the uname?

internal/node/node_launcher.sh
```diff
@@ -114,7 +114,24 @@ TEMPLATED_env_vars
...
 
-readonly node=$(rlocation "TEMPLATED_node")
+# Check environment
+unameOut="$(uname -s)"
+case "${unameOut}" in
+    Linux*)     machine=linux;;
+    Darwin*)    machine=darwin;;
+    CYGWIN*)    machine=windows;;
+    MINGW*)     machine=windows;;
+    MSYS_NT*)   machine=windows;;
```
alexeagle: sigh, when we introduce a batch/powershell version of the launcher we'll have to pick through this logic again

internal/node/node.bzl
gregmagolan: Sounds right regarding slash. There was a common case in npm_package.bzl  so I factored it out `strip_external` to a new `internal/common/path_utils.bzl`.

.bazelci/presubmit.yml
```diff
@@ -156,6 +158,20 @@ tasks:
     # TODO(gregmagolan): make node_repositories acccept different archives for different platforms
     - "//examples:examples_vendored_node"
     - "//examples:examples_vendored_node_and_yarn"
+  ubuntu1804_cross_compile_darwin:
```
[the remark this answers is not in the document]
gregmagolan: 👍

internal/node/node_launcher.sh
gregmagolan: added something noisy and error handling if the chosen node executable is not in runfiles

gregmagolan:
> Need more error handling in the case fallthrough

Done

.bazelci/presubmit.yml
```diff
@@ -156,6 +162,34 @@ tasks:
...
+    name: ubuntu1804_cross_compile_darwin
+    platform: ubuntu1804
+    # Build on linux with the node --platform set to darwin
+    build_flags:
+    - "--platforms=@build_bazel_rules_nodejs//toolchains/node:darwin_amd64"
+    build_targets:
+    - "//..."
```
alexeagle: feels really expensive to re-build everything in each configuration. I wonder if we are using more CI resources than any other rules

internal/node/node.bzl
```diff
@@ -211,6 +214,8 @@ def _nodejs_binary_impl(ctx):
         # also be sure to include the params file in the program inputs
         node_tool_files.append(ctx.outputs.templated_args_file)
 
+    is_vendored_node = (ctx.attr._node.label.workspace_name != "nodejs_darwin_amd64" and ctx.attr._node.label.workspace_name != "nodenodejs_linux_amd64js_darwin_amd64" and ctx.attr._node.label.workspace_name != "nodejs_windows_amd64")
```
alexeagle: nodenodejs_linux looks wrong

internal/node/node_launcher.sh
```diff
@@ -114,7 +114,51 @@ TEMPLATED_env_vars
...
 
-readonly node=$(rlocation "TEMPLATED_node")
+readonly vendored_node="TEMPLATED_vendored_node"
+
+if [ -n "${vendored_node}" ]; then
+  # Use the vendored node path
+  readonly node=$(rlocation "${vendored_node}")
+
+  if [ ! -f "${node}" ]; then
+      printf "\n>>>> FAIL: The vendored node binary '${vendored_node}' not found in runfiles. <<<<\n\n" >&2
```
alexeagle: will users know what to do with this error?

internal/node/node_launcher.sh
```diff
@@ -114,7 +114,51 @@ TEMPLATED_env_vars
...
+                  printf "you would like to add your platform to the supported rules_nodejs node platforms.\n\n" >&2
+                  ;;
+  esac
+
+  case "${machine}" in
+    # The following paths must match up with _download_node in node_repositories
+    linux) readonly node_toolchain="nodejs_linux_amd64/bin/nodejs/bin/node" ;;
+    darwin) readonly node_toolchain="nodejs_darwin_amd64/bin/nodejs/bin/node" ;;
+    windows) readonly node_toolchain="nodejs_windows_amd64/bin/nodejs/node.exe" ;;
+    *) readonly node_toolchain="nodejs_linux_amd64/bin/nodejs/bin/node" ;;
```
alexeagle: nit: can you collapse with the linux) case?

internal/node/test/BUILD.bazel
```diff
@@ -183,3 +181,24 @@ nodejs_test(
...
+nodejs_toolchain_test(
+    name = "nodejs_toolchain_darwin_test",
+    expected_node_path = "external/nodejs_darwin_amd64/bin/nodejs/bin/node",
+    tags = ["manual"],
+)
+
+# must be run with --platforms=@build_bazel_rules_nodejs//toolchains/node:windows_amd64
+nodejs_toolchain_test(
+    name = "nodejs_toolchain_windows_test",
+    expected_node_path = "external/nodejs_windows_amd64/bin/nodejs/node.exe",
```
alexeagle: the path seems a bit brittle as the assertion to make here. what about running `node --version` or something to assert things are working?

packages/protractor/src/protractor.conf.js
```diff
@@ -43,6 +44,49 @@ function setConf(conf, name, value, msg) {
   conf[name] = value;
 }
 
+/**
+ * Helper function to find a particular namedFile
+ * within the webTestMetadata webTestFiles
+ */
+function findNamedFile(webTestMetadata, key) {
```
alexeagle: is this protractor change related to cross-platform build?

internal/node/node_launcher.sh
gregmagolan: It shouldn't happen as the vendored_node binary should be added to runfiles by nodejs_binary.

packages/protractor/src/protractor.conf.js
gregmagolan: Initially yes as I thought the tests needed to be set to `extract = "run"` and protractor didn't handle the extract browser archive case; but it turned out I could get it working with `extract = "build"` in the angular repo so I can factor the protractor change out into a separate PR if you prefer. It already is a separate commit.

internal/node/node_launcher.sh
gregmagolan: 👍

.bazelci/presubmit.yml
gregmagolan: we do have quite a few jobs now. it's probably not strictly necessary to build in every permutation but I think at the least we should build everything or some subset of everything with each platform.

internal/node/test/BUILD.bazel
gregmagolan: what is being tested is that `ctx.toolchains["@build_bazel_rules_nodejs//toolchains/node:toolchain_type"].nodeinfo.tool_files[0].path` is set to the correct value for each platform choice. not possible to execute as the node binary will not necessarily match the host platform here.

It is a bit brittle with the path. There might be a way to derive the expected path from the platform itself if it is available in the ctx.

internal/node/node.bzl
gregmagolan: Thanks!

gregmagolan: Improved this with
```
BUILT_IN_NODE_REPOSITORIES = [
    "nodejs_darwin_amd64",
    "nodejs_linux_amd64",
    "nodejs_windows_amd64",
]
is_builtin = ctx.attr._node.label.workspace_name in BUILT_IN_NODE_REPOSITORIES
```

internal/node/test/BUILD.bazel
gregmagolan: Cleaned up the API for the test to make it less brittle:
```
[nodejs_toolchain_test(
    name = "nodejs_toolchain_%s_test" % platform,
    platform = platform,
    # must be run with --platforms=@build_bazel_rules_nodejs//toolchains/node:PLATFORM
    # where PLATFORM is one of BUILT_IN_NODE_PLATFORMS
    tags = ["manual"],
) for platform in BUILT_IN_NODE_PLATFORMS]
```

.bazelci/presubmit.yml
alexeagle: maybe just `//internal/...` is sufficient to get the coverage we need?

packages/protractor/src/protractor.conf.js
alexeagle: yeah I'd rather have it in a separate PR

gregmagolan: 👍

.bazelci/presubmit.yml
gregmagolan: 👍 that should do it for now as only node_binary/node_test vary the build based on platform. We can expand coverage in the future if some of the packages with build differences based on platform.

Issue #1323, opened 2019-10-31
Use `--compilation_mode=dbg` instead of `--define DEBUG=1`


ashi009:
# 🚀 feature request

### Relevant Rules

All rules with debug capabilities

### Description

`--compilation_mode=dbg` is the canonical way of creating debugging outputs.  Different compilation mode uses dedicated execroots, therefore unlike setting `--define DEBUG=1`, changing compilation mode won't flush caches. So rebuilding a target in debugging mode will be much faster.

### Describe the solution you'd like

Use `ctx.var.get("COMPILATION_MODE")` instead.

### Describe alternatives you've considered

N/A

gregmagolan:
Resolved in #1329 and fixes made in https://github.com/bazelbuild/rules_nodejs/pull/1521

PR #1329, opened 2019-11-04
builtin: support --compilation_mode flag

ashi009:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Deprecate `--define DEBUG=1` and use `--compilation_mode=dbg` instead. The value of compilation mode will be passed to any nodejs executables as  'COMPILATION_MODE' environment variable. All builtin rules/tests that depends on `DEBUG` are now using `COMPILATION_MODE` instead.

Issue Number: #1323 

## What is the new behavior?

Use `--compilation_mode=dbg` to trigger the debugging build.

- [x] Yes
- [ ] No

Defining `--define DEBUG` will no longer trigger a debugging build. To allow users to use to the new flag, a warning message will be printed if `--define DEBUG` is set while `--compilation_mode` is not `dbg`.

`DEBUG` is kept in `default_env_vars`, as many npm packages use it to trigger the debugging behavior. However, I do believe using `--define` should be more conservative, as it always triggers a rebuild, which could be painfully slow.

### DIFF
--- a/common.bazelrc
+++ b/common.bazelrc
@@ -42,16 +42,17 @@ test --test_output=errors
 # --define=VERBOSE_LOGS=1
 #     Rules will output verbose logs if the VERBOSE_LOGS environment variable is set. `VERBOSE_LOGS` will be passed to
 #     `nodejs_binary` and `nodejs_test` via the default value of the `default_env_vars` attribute of those rules.
-# --define=DEBUG=1
-#     Rules may change their build outputs if the DEBUG environment variable is set. For example,
-#     mininfiers such as terser may make their output more human readable when this is set. `DEBUG` will be passed to
+# --compilation_mode=dbg
+#     Rules may change their build outputs if the compilation mode is set to dbg. For example,
+#     mininfiers such as terser may make their output more human readable when this is set. `COMPILATION_MODE` will be passed to
 #     `nodejs_binary` and `nodejs_test` via the default value of the `default_env_vars` attribute of those rules.
+#     See https://docs.bazel.build/versions/master/user-manual.html#flag--compilation_mode for more details.
 test:debug --test_output=streamed --test_strategy=exclusive --test_timeout=9999 --nocache_test_results --define=VERBOSE_LOGS=1
 # Use bazel run with `--config=debug` to turn on the NodeJS inspector agent.
 # The node process will break before user code starts and wait for the debugger to connect.
 run:debug --define=VERBOSE_LOGS=1 -- --node_options=--inspect-brk
 # The following option will change the build output of certain rules such as terser and may not be desirable in all cases
-build:debug --define=DEBUG=1
+build:debug --compilation_mode=dbg
 
 # Turn off legacy external runfiles
 # This prevents accidentally depending on this feature, which Bazel will remove.
--- a/docs/Built-ins.md
+++ b/docs/Built-ins.md
@@ -226,9 +226,12 @@ without losing the defaults that should be set in most cases.
 
 The set of default  environment variables is:
 
-- `DEBUG`: rules use this environment variable to turn on debug information in their output artifacts
+- `COMPILATION_MODE`: rules use this environment variable to produce optimized (eg. mangled and minimized) or debugging output
 - `VERBOSE_LOGS`: rules use this environment variable to turn on debug output in their logs
+- `DEBUG`: used by some npm packages to print debugging logs
 - `NODE_DEBUG`: used by node.js itself to print more logs
+
+Note that, `DEBUG` is derived from bazel compilation mode if not present in --define.
 
 
 #### `entry_point`
@@ -438,9 +441,12 @@ without losing the defaults that should be set in most cases.
 
 The set of default  environment variables is:
 
-- `DEBUG`: rules use this environment variable to turn on debug information in their output artifacts
+- `COMPILATION_MODE`: rules use this environment variable to produce optimized (eg. mangled and minimized) or debugging output
 - `VERBOSE_LOGS`: rules use this environment variable to turn on debug output in their logs
+- `DEBUG`: used by some npm packages to print debugging logs
 - `NODE_DEBUG`: used by node.js itself to print more logs
+
+Note that, `DEBUG` is derived from bazel compilation mode if not present in --define.
 
 
 #### `entry_point`
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -100,8 +100,8 @@ tasks:
     name: ubuntu1804_debug
     platform: ubuntu1804
     test_flags:
+    - "--compilation_mode=dbg"
     - "--define=VERBOSE_LOGS=1"
-    - "--define=DEBUG=1"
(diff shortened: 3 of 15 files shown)


ashi009:
Can somebody please rerun the test? It's failing for some unrelated ruels_go issue:

> (13:29:13) ERROR: /private/var/tmp/_bazel_buildkite/a2f482ea6b0ee7030787e3581fad8ee8/external/go_sdk/BUILD.bazel:43:1: Couldn't build file external/go_sdk/builder: error executing shell command: '/bin/bash -c 'external/go_sdk/bin/go' tool compile -o 'bazel-out/host/bin/external/go_sdk/builder'.a -I 'external/go_sdk' -trimpath=$PWD $@ && 'external/go_sdk/bin/go' tool link -o 'bazel-out/host/...' failed (Exit 1) bash failed: error executing command
> --
>   | (cd /private/var/tmp/_bazel_buildkite/a2f482ea6b0ee7030787e3581fad8ee8/sandbox/darwin-sandbox/607/execroot/build_bazel_rules_nodejs && \
>   | exec env - \
>   | GOROOT=external/go_sdk \
>   | /bin/bash -c ''\''external/go_sdk/bin/go'\'' tool compile -o '\''bazel-out/host/bin/external/go_sdk/builder'\''.a -I '\''external/go_sdk'\'' -trimpath=$PWD $@ && '\''external/go_sdk/bin/go'\'' tool link -o '\''bazel-out/host/bin/external/go_sdk/builder'\'' -L '\''external/go_sdk'\'' '\''bazel-out/host/bin/external/go_sdk/builder'\''.a && rm '\''bazel-out/host/bin/external/go_sdk/builder'\''.a' '' external/io_bazel_rules_go/go/tools/builders/ar.go external/io_bazel_rules_go/go/tools/builders/asm.go external/io_bazel_rules_go/go/tools/builders/builder.go external/io_bazel_rules_go/go/tools/builders/cgo.go external/io_bazel_rules_go/go/tools/builders/cgo2.go external/io_bazel_rules_go/go/tools/builders/compile.go external/io_bazel_rules_go/go/tools/builders/compilepkg.go external/io_bazel_rules_go/go/tools/builders/cover.go external/io_bazel_rules_go/go/tools/builders/env.go external/io_bazel_rules_go/go/tools/builders/extract.go external/io_bazel_rules_go/go/tools/builders/filter.go external/io_bazel_rules_go/go/tools/builders/filter_buildid.go external/io_bazel_rules_go/go/tools/builders/flags.go external/io_bazel_rules_go/go/tools/builders/generate_nogo_main.go external/io_bazel_rules_go/go/tools/builders/generate_test_main.go external/io_bazel_rules_go/go/tools/builders/importcfg.go external/io_bazel_rules_go/go/tools/builders/link.go external/io_bazel_rules_go/go/tools/builders/pack.go external/io_bazel_rules_go/go/tools/builders/replicate.go external/io_bazel_rules_go/go/tools/builders/stdlib.go external/io_bazel_rules_go/go/tools/builders/path.go)
>   | Execution platform: @local_config_platform//:host

docs/Built-ins.md
```diff
@@ -226,10 +226,13 @@ without losing the defaults that should be set in most cases.
 
 The set of default  environment variables is:
 
-- `DEBUG`: rules use this environment variable to turn on debug information in their output artifacts
+- `COMPILATION_MODE`: rules use this environment variable to produce optimized (eg. mangled and minimized) or debugging output
```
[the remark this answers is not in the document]
alexeagle: no, compilation mode definitely is used to produce different outputs. verbose logging is the other flags below

common.bazelrc
```diff
@@ -42,16 +42,17 @@ test --test_output=errors
...
+#     mininfiers such as terser may make their output more human readable when this is set. `COMPILATION_MODE` will be passed to
 #     `nodejs_binary` and `nodejs_test` via the default value of the `default_env_vars` attribute of those rules.
+#     See https://docs.bazel.build/versions/master/user-manual.html#flag--compilation_mode for more details.
 test:debug --test_output=streamed --test_strategy=exclusive --test_timeout=9999 --nocache_test_results --define=VERBOSE_LOGS=1
 # Use bazel run with `--config=debug` to turn on the NodeJS inspector agent.
 # The node process will break before user code starts and wait for the debugger to connect.
 run:debug --define=VERBOSE_LOGS=1 -- --node_options=--inspect-brk
 # The following option will change the build output of certain rules such as terser and may not be desirable in all cases
-build:debug --define=DEBUG=1
+build:debug --compilation_mode=dbg
```
alexeagle: I think we should just remove this:
- it's confusing that `--config=debug` does something for test and a different thing for build
- `-c=debug` is short so the syntax sugar has no benefit

docs/Built-ins.md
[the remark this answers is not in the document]
ashi009: I assume this requires no action

> you say
> 
> > many npm packages use it to trigger the debugging behavior
> 
> does that mean maybe we should always set `DEBUG=1` in the environment of node processes when we find `-c debug`?

Those packages use DEBUG differently, so I think it should be preserved, but not necessarily set by the rule. For instance, npmjs.com/debug supports passing a pattern with `DEBUG` to filter debug messages, having DEBUG=1 set doesn't really help.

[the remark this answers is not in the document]
ashi009: weird, why the reply ends up here?

ashi009:
> you say
> 
> > many npm packages use it to trigger the debugging behavior
> 
> does that mean maybe we should always set `DEBUG=1` in the environment of node processes when we find `-c debug`?

Those packages use `DEBUG` differently, so I think it should be preserved, but not necessarily set by the rule. For instance, npmjs.com/debug supports passing a pattern with `DEBUG` to filter debug messages, having `DEBUG=1` set doesn't really help.

common.bazelrc
ashi009: SGTM. I can send another CL later, as I want this change to be a drop in replacement for `DEBUG` usages, without messing too much about tests and CI configs.

Also, I haven’t tried if `-c debug` works, as the mode seems always being referred to as `dbg` and `opt`, which also matches the value of `COMPILATION`_MODE context variable.

meteorcloudy: We should not use `--compilation_mode=dbg` flag (also it's alias `-c dbg`), this flag only affects C/C++ build.
See https://docs.bazel.build/versions/master/user-manual.html#flag--compilation_mode

meteorcloudy: Sorry for commenting without really looking into the PR. I believe this PR is fine.

`--compilation_mode` was designed for C++, but that doesn't mean we cannot use it for other languages. The value of `--compilation_mode` is exported as `COMPILATION_MODE` variable. Since this PR is actually using the information to configure the build, I think it's totally fine.

meteorcloudy: BTW, do we have test to make sure this actually work?

ashi009: Yes. It’s all covered. Tests are not changed to assure that the behavior is consistent.

PR #1338, opened 2019-11-06
feat: add ngJitMode to the terser config

alxhub:
rules_nodejs controls the terser configuration used to minify Angular code,
and as a result the terser configuration in this package has some default
values for Angular global variables. These defaults are used to tree-shake
away parts of the Angular application that are only used for development,
when bundling for production.

The ngJitMode flag is one such global variable. It controls whether parts
of AOT-compiled Angular code are retained to enable JIT code to link with
it at runtime. If these parts are left in the global bundle, they will break
tree-shaking of unusued components/directives/pipes/etc. Removing them is
thus an important optimization which allows the tree-shaker to effectively
process the rest of the application.

Unfortunately there are some cases where an application may wish to use
JIT in production, and in those cases ngJitMode should be left undefined
(or set to true). Because rules_nodejs does not allow user editing of the
terser configuration, this commit effectively makes the choice to support
optimization of AOT-only apps at the expense of supporting JIT production
builds. Such hybrid apps will not work with rules_nodejs' terser pass.

### DIFF
--- a/internal/rollup/terser-wrapped.js
+++ b/internal/rollup/terser-wrapped.js
@@ -66,7 +66,7 @@ function runterser(inputFile, outputFile, sourceMapFile) {
     'compress': {
       'pure_getters': true,
       'passes': 3,
-      'global_defs': {'ngDevMode': false, 'ngI18nClosureMode': false},
+      'global_defs': {'ngDevMode': false, 'ngI18nClosureMode': false, 'ngJitMode': false},
       'keep_fnames': !debug,
       'reduce_funcs': !debug,
       'reduce_vars': !debug,
--- a/packages/terser/src/terser_config.default.json
+++ b/packages/terser/src/terser_config.default.json
@@ -1,6 +1,6 @@
 {
     "compress": {
-        "global_defs": {"ngDevMode": false, "ngI18nClosureMode": false},
+        "global_defs": {"ngDevMode": false, "ngI18nClosureMode": false, "ngJitMode": false},
         "keep_fnames": "bazel_no_debug",
         "passes": 3,
         "pure_getters": true,


alexeagle:
Now that users control the rollup config, I wouldn't expect we need to keep making layering violations of having angular-specific constants appear in rules_nodejs. @gregmagolan is working on moving Angular repo to the new terser_minified rule which should remove the need for this.
Let's discuss if this change is needed?

gregmagolan:
Most of the migration away from the legacy rollup_bundle is done now in the angular repo. The only dep left is in /packages/zone.js which will get resolved with https://github.com/angular/angular/pull/33329.

ng_rollup_bundle is using terser_minified already: https://github.com/angular/angular/blob/master/tools/ng_rollup_bundle/ng_rollup_bundle.bzl#L398

and ng_packager was harder to break up to use terser_minified in a macro but it no longer depends on internal/rollup but it takes a `rollup_config_tmpl` (https://github.com/angular/angular/blob/master/packages/bazel/src/ng_package/ng_package.bzl#L746) which defaults to https://github.com/angular/angular/blob/master/packages/bazel/src/ng_package/terser_config.default.json.

My suggestion is that if this is needed for ng_packager only then the change should go into https://github.com/angular/angular/blob/master/packages/bazel/src/ng_package/terser_config.default.json.

If it is also needed for ng_rollup_bundle than the angular repo should supply a custom terser config in the ng_rollup_bundle macro where terser_minified is used (https://github.com/angular/angular/blob/master/tools/ng_rollup_bundle/ng_rollup_bundle.bzl#L398). Actually we should do this anyway so we can remove the layering violation of the default terser config in rules_nodejs which has `"global_defs": {"ngDevMode": false, "ngI18nClosureMode": false},`

alxhub:
@gregmagolan yes, that would be helpful.

gregmagolan:
kk. I'll make a PR in Angular. Closing this.

PR #1342, opened 2019-11-09
feat(karma): remove ts_web_test and ts_web_test_suite rules

gregmagolan:
BREAKING CHANGES:

The `ts_web_test` and `ts_web_test_suite` rules were duplicates of `karma_web_test` and `karma_web_test_suite` rules minus the `config_file` attribute. The `karma_web_test` and `karma_web_test_suite`, which have identical APIs, should now be used instead.
They can be loaded from `load("@npm_bazel_karma//:index.bzl", "karma_web_test")` and `load("@npm_bazel_karma//:index.bzl", "karma_web_test_suite")`.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [x] Yes
- [ ] No


examples/web_testing/BUILD.bazel
```diff
@@ -56,24 +56,7 @@ karma_web_test_suite(
...
 karma_web_test_suite(
-    name = "testing_karma_sauce",
-    browsers = [
-        "@io_bazel_rules_webtesting//browsers/sauce:chrome-win10",
-    ],
-    tags = [
-        "exclusive",
-        "sauce",
-        "native",
-        # TODO(alexeagle): enable on CI once we have set the SAUCE env variables
```
alexeagle: do we need to keep this TODO around somewhere? will we ever test with saucelabs?

examples/web_testing/defaults.bzl
```diff
@@ -12,40 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Overrides for karma_web_test_suite & ts_web_test_suite to support bazelci testing
+"""Overrides for karma_web_test_suite & karma_web_test_suite to support bazelci testing
```
alexeagle: needs some cleanup of places where it's just duplicating like this

examples/web_testing/BUILD.bazel
gregmagolan: I don't think so. I think saucelabs is going to be a user-space concern as ended up being in the angular repo and not one for the karma_web_test or karma_web_test_suite rule to worry about. The rules_webtesting saucelabs browser target is not very useful.

gregmagolan:
> can you use this karma rule without named amd modules? that's part of where we got the "ts_" name

No you can't. Maybe this needs to be a `karma_concatjs_web_test` as it forces the use of the concatjs plugin? We should probably refactor at some point and make the concatjs plugin option so user's can just use karma as they do in a regular nodejs workflow.

Toxicable:
>We should probably refactor at some point and make the concatjs plugin option so user's can just use karma as they do in a regular nodejs workflow.

That's a really good point, since we're looking at other devservers, such as the rollup and webpack ones we need to consider how karma will work at the same time since karma is effectively a devserver at the same time as a test runner.

I have no idea if it would work, but it would be cool if you could somehow pass a devserver target to the karama rule.

examples/web_testing/defaults.bzl
gregmagolan: nice catch. just the one that I could see.

gregmagolan:
That's more of a protractor concern to supply your own devserver target which is already takes. AFAIK that is not something that karma supports as it always serves the files itself. Still, the mandatory concatjs plugin with the current implementation leads to the same problem with karma_web_test that ts_devserver has in that it requires named-AMD or named-UMD modules.

PR #1349, opened 2019-11-12
fix: support for extracting rules_webtesting archives in protractor_web_test_suite

gregmagolan:
This is supported in karma_web_test_suite but not yet in protractor_web_test_suite.

TODO: add tests

### DIFF
--- a/packages/protractor/src/protractor.conf.js
+++ b/packages/protractor/src/protractor.conf.js
@@ -15,6 +15,7 @@
  * limitations under the License.
  */
 const path = require('path');
+const child_process = require('child_process');
 
 function log_verbose(...m) {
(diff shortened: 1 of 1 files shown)


packages/protractor/src/protractor.conf.js
```diff
@@ -43,6 +44,49 @@ function setConf(conf, name, value, msg) {
   conf[name] = value;
 }
 
+/**
```
alexeagle: how will this work in general, if we have special code to handle it for protractor and karma? does it make sense for cypress to be able to use rules_webtesting-provisioned browsers? maybe this code should be factored differently to serve all test runners, should we discuss (maybe with @DrMarcII)

PR #1352, opened 2019-11-14
feat(protractor): protractor npm package is now a peer deps

gregmagolan:
Also cleanup us arguments and doc string for protractor_web_test_suite.

BREAKING CHANGES:

npm protractor deps for protractor_web_test and protractor_web_test_suite is now a peer dep so that the versions used can be chosen by the user.


packages/protractor/src/package.json
```diff
@@ -15,9 +15,8 @@
       "protractor",
       "bazel"
   ],
-  "main": "index.js",
-  "dependencies": {
-      "protractor": "^5.4.2"
+  "peerDdependencies": {
```
alexeagle: spelling error

gregmagolan: have to work on my studder :)

PR #1355, opened 2019-11-15
chore: remove old rollup_bundle

### LINKED ISSUE #740: build_bazel_rules_nodejs should rarely require updates
The starlark-only package is harder to update because the version is in the WORKSPACE and not in package.json

This means that every time we cut a rules_nodejs release, we have to update WORKSPACE in the @angular/bazel builder package for users to pick up the change.

We want the build_bazel_rules_nodejs starlark package to be "evergreen" and it should rarely require bugfixes, which means it should have minimal bootstrapping code.

alexeagle:
BREAKING CHANGE:
internal/rollup_bundle rule is removed. see https://github.com/bazelbuild/rules_nodejs/wiki for migration instructions

Fixes #740


gregmagolan:
You could also remove collect_es6_sources.bzl at this point as angular/angular no longer depends on that either.

PR #1358, opened 2019-11-18
public web_package rule

alexeagle:


alexeagle:
fixed in a different PR

PR #1359, opened 2019-11-18
chore: remove unused code

gregmagolan:
Removes old code that was part of bazel_integration_test but is no longer used


alexeagle:
Rebase and try again, there was an infra issue on CI

PR #1361, opened 2019-11-19
feat(builtin): export npm package directories from generated npm_install & yarn_install BUILD file

gregmagolan:
This is useful for referencing node_module directories with labels such as `@npm//:node_modules/typescript`.

Working on integration test rule in angular repository and it will be helpful to use labels such as `@npm//:node_modules/typescript` in the rule's `npm_package`s attribute. For example,

```
angular_integration_test(
    name = "cli_hello_world_test",
    test_files = ":cli_hello_world_files",
    npm_packages = {
        "//packages/animations:npm_package": "@angular/animations",
        "//packages/common:npm_package": "@angular/common",
        "//packages/compiler:npm_package": "@angular/compiler",
        "//packages/compiler-cli:npm_package": "@angular/compiler-cli",
        "//packages/core:npm_package": "@angular/core",
        "//packages/forms:npm_package": "@angular/forms",
        "//packages/language-service:npm_package": "@angular/language-service",
        "//packages/platform-browser:npm_package": "@angular/platform-browser",
        "//packages/platform-browser-dynamic:npm_package": "@angular/platform-browser-dynamic",
        "//packages/router:npm_package": "@angular/router",
        "//packages/zone.js:npm_package": "zone.js",
        "@npm//:node_modules/typescript": "typescript",
        "@npm//:node_modules/rxjs": "rxjs",
        "@npm//:node_modules/tslib": "tslib",
        "@npm//:node_modules/protractor": "protractor",
        "@npm//:node_modules/@types/node": "@types/node",
    },
```


internal/npm_install/generate_build_file.ts
```diff
@@ -171,10 +171,23 @@ for installation instructions.`);
...
+    // by the `@wksp//:node_modules/pkg` label
+    exportsStarlark += `    "node_modules/${pkg._dir}",\n`;
+    const dirs = [];
+    for (const f of pkg._files) {
+      dirs.push(path.dirname(f))
+      exportsStarlark += `    "node_modules/${pkg._dir}/${f}",\n`;
+    }
+    // export the all dirs so that they can be reference
+    // by the `@wksp//:node_modules/pkg/path/to/dir` label
+    for (const d of [...new Set(dirs)]) {
```
alexeagle: I'm a little worried about the generated file getting "too big" whatever that means.

Do you actually use the subdirectories also? isn't it enough to get the top-level packages?

gregmagolan: Its already pretty big :) I only need the top-level package for angular. Besides the size I didn't see any reason to not include the nested directories incase someone wants to reference them as labels. Maybe that reason is enough to not include them however. Globbed exports_files would be nice.

gregmagolan:
Turns out not to be needed in angular/angular for npm_integration_test so this can be put on ice for a bit

PR #1365, opened 2019-11-19
Injector separate pkg

alexeagle:


packages/inject-html/inject_html.bzl
```diff
@@ -0,0 +1,87 @@
...
+    "src": attr.label(doc = "html file", allow_single_file = [".html"]),
+    "out": attr.output(
+        doc = "Path to write the output. If not provided, src will be copied to the same path in the output tree",
+    ),
+    "additional_root_paths": attr.string_list(
+        doc = """Path prefixes to strip off all assets, in addition to the current package. Longest wins.""",
+    ),
+    "assets": attr.label_list(
+        allow_files = True,
+        doc = """Files which should be referenced from the index_html""",
```
gregmagolan: "referenced in the output html"

alexeagle: fixed

examples/angular/src/BUILD.bazel
```diff
@@ -161,28 +163,33 @@ terser_minified(
     src = ":bundle-es5",
 )
 
+inject_html(
+    name = "inject_scripts_for_prod",
```
gregmagolan: 🌮

examples/angular/src/BUILD.bazel
```diff
@@ -85,23 +86,22 @@ _ASSETS = [
     "@npm//:node_modules/zone.js/dist/zone.min.js",
 ]
 
+inject_html(
```
gregmagolan: 🌮

examples/angular/src/BUILD.bazel
```diff
@@ -161,28 +163,33 @@ terser_minified(
     src = ":bundle-es5",
 )
 
-web_package(
+inject_html(
+    name = "inject_scripts_for_prod",
+    src = "//src:example/index.prod.html",
+    # we can't produce "example/index.html" since that collides with the devmode file
+    # pkg_web rule will re-root this path because it starts with _prodapp
```
gregmagolan: This is sneaky and I think needs a more comprehensive explanation for the purposes of an example. Maybe
```
# we can't produce "example/index.html" since that collides with the devmode file
# so we output "_prodapp/src/example/index.html" since pkg_web rule will re-root paths
# that start with _{name} by default which will put this file under `src/example/index.html`
# in the web package.
```
?

examples/angular_view_engine/src/BUILD.bazel
```diff
@@ -83,23 +84,22 @@ _ASSETS = [
     "@npm//:node_modules/zone.js/dist/zone.min.js",
 ]
 
+inject_html(
+    name = "inject_scripts_for_dev",
+    src = "//src:example/index.html",
```
gregmagolan: nit: make the input file`//src:example/index.dev.html` and output to `src/example/index.html` so the distinction between dev & prod is clearer in the sources?

internal/web_package/test2/index_golden.html_
```diff
@@ -1,2 +1 @@
...
+<html></html>
```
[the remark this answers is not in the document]
gregmagolan: already fixed

internal/web_package/test/index_golden.html_
```diff
@@ -1,2 +1 @@
-<html><head></head><body>
-<script src="/bundle.min.js?v=123"></script></body></html>
\ No newline at end of file
+<html></html>
```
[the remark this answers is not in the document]
gregmagolan: already fixed

examples/webapp/index.html
```diff
@@ -2,7 +2,6 @@
     <head>
         <title>rules_nodejs webapp example</title>
         <link rel="icon" href="/favicon.png">
-        <!-- CSS will be linked here from web_package#assets -->
```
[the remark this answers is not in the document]
alexeagle: deleted styles.css, good catch

examples/app/BUILD.bazel
```diff
@@ -31,14 +31,21 @@ terser_minified(
...
+
 web_package(
     name = "package",
-    assets = [
-        ":bundle.min",
-        "//styles:base.css",
-        "//styles:test.css",
-    ],
-    index_html = "index.html",
+    srcs = [":inject_js_css_tags"] + _ASSETS,
```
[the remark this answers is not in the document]
alexeagle: added a comment above `inject_html` 
# Copy index.html to the output folder, adding <script> and <link> tags

packages/inject-html/BUILD.bazel
```diff
@@ -0,0 +1,36 @@
...
+# For local testing without an npm install
+nodejs_binary(
+    name = "injector",
+    data = ["@npm//parse5"],
+    entry_point = ":index.js",
+    install_source_map_support = False,
+    visibility = ["//packages/inject-html:__subpackages__"],
+)
+
+jasmine_node_test(
```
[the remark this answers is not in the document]
alexeagle: done

packages/inject-html/spec.js
```diff
@@ -1,4 +1,4 @@
-const injector = require('./injector');
+const injector = require('./');
```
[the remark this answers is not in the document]
alexeagle: done

examples/angular/src/BUILD.bazel
alexeagle: sounds good

examples/angular_view_engine/src/BUILD.bazel
alexeagle: I think it's nice that at least one of the files has the same path in the srcs and the package.

Ideally I think we should have only `src/index.html` and devmode and prodmode both do the right thing - need to figure out diff. loading again, maybe using the manifest file with hinting, Jan can probably help us get it right

gregmagolan: That's a good point. It seems to me that any auto-detection for diff loading will always be broken for some use cases out there. Maybe the correct shape is to have inject_html have different srcs attributes for css, plain scripts, scripts with `nomodule` and scripts with `type="module"`.

However, probably not worth the effort or complexity at this point as the proper solution is to have an stand-alone npm package that can handle those distinctions and user's would just use the generated `npm_package_bin`.

gregmagolan: There is no way for an injector to know that you'd want `nomodule` on scripts such as
```    <script nomodule src="/core.min.js"></script>
    <script nomodule src="/system.js"></script>
```
for example

internal/pkg_web/test2/index_golden.html_
```diff
@@ -0,0 +1 @@
+<html></html>
```
gregmagolan: That doesn't seem right?

gregmagolan: Golden test not doing much anymore and should probably just get removed or its functionality moved to inject-html as it was testing the injector

PR #1375, opened 2019-11-21
feat(builtin): wire up patches

alexeagle:
.ts files from @soldair PR


alexeagle:
This wires up the production changes from #1332 

We will use that PR to land the corresponding tests.

alexeagle:
replaced with #1381

PR #1390, opened 2019-11-25
fix: @npm//foobar:foobar__files target no longer includes nested node_modules

gregmagolan:
The expected behavior would be that @npm//foobar:foobar__files would just be the files in the package and not the nested node_modules that may or may not be there depending on the behavior of the package installer.

Pre-req for https://github.com/angular/angular/pull/33927

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


gregmagolan:
I thought about that but I think the current `foo__files` filegroup is wrong and anyone using it (which is probably no one) would not expect it to contain the nested node_modules.

gregmagolan:
The `@npm//foo`  depends on `@npm//foo:foo_all_files`  so that rules gets the transitive deps if they are not hoisted. If `foo` has a dep on `bar` and bar is hoisted then `@npm//foo` will also depend on `@npm//bar:bar__contents`.

alexeagle:
do we have to expose both `@npm//foo` and `@npm//foo:foo_all_files`? seems like you're saying users should never reference the latter

gregmagolan:
`@npm//foo:foo__all_files` and `@npm//foo:foo__nested_node_modules` could be private visibility.

`@npm//foo:foo__files` is useful to access the raw filegroup that is the npm package without any nested node_modules.

PR #1397, opened 2019-11-26
Html insert assets

alexeagle:


jbedard:
@googlebot I consent

e2e/ts_devserver/subpackage/BUILD.bazel
```diff
@@ -12,14 +12,25 @@
...
+        "--html",
+        "subpackage/index.tmpl.html",
+        "--out",
+        "$@",
+        "--assets",
+        "$(location //:red-body-style.css)",
+    ],
+    data = [
+        "index.tmpl.html",
+        "//:red-body-style.css",
```
gregmagolan: location not needed for `--html $(location :index.tmpl.html)`?

examples/kotlin/BUILD.bazel
```diff
@@ -23,7 +22,7 @@ kt_js_library(
     deps = [":kotlinx-html-js"],
 )
 
-# Copy boostrap.js to the bin folder as _bootstrap.js
```
gregmagolan: 👻

internal/node/npm_package_bin.bzl
```diff
@@ -26,8 +26,13 @@ def _expand_location(ctx, s):
         # We'll write into a newly created directory named after the rule
         outdir_segments.append(ctx.attr.name)
 
+    if not ctx.attr.output_dir:
+        if s.find("$@") != -1 and len(ctx.outputs.outs) > 1:
+            fail("$@ substitution may only be used with a single out")
```
gregmagolan: suggest to use `$(RULEDIR)` instead to make migration easier?

gregmagolan: Oh nevermind. Now I noticed the `if not ctx.attr.output_dir:`

internal/node/npm_package_bin.bzl
```diff
@@ -26,8 +26,13 @@ def _expand_location(ctx, s):
         # We'll write into a newly created directory named after the rule
```
gregmagolan: Here would be a good place to help out users:
```
        if s.find("$@") != -1:
            fail("$@ substitution may not be used with output_dir set to True, use $(RULEDIR) instead")
```

e2e/ts_devserver/subpackage/BUILD.bazel
alexeagle: for inputs, `$(location foo)` is always the same as `package_name() + "/foo"` so I've been using interchangeably. I guess more consistency is fine

internal/node/npm_package_bin.bzl
alexeagle: sure but there's a common case where you had only one out and won't get this message even though you need to change. May as well add it though, could help someone

PR #1398, opened 2019-11-26
ci: stop running examples and e2e on circleci

alexeagle:
BuildKite gives us the same coverage on more platforms


alexeagle:
@gregmagolan says we need to keep doing this because the browser tests don't run on buildkite due to missing shared libraries

PR #1400, opened 2019-11-26
fix(builtin): $(RULEDIR) npm_package_bin expansion should always be the root output directory

gregmagolan:
To match genrule behavior described here: https://docs.bazel.build/versions/master/be/make-variables.html#predefined_genrule_variables:
```
RULEDIR: The output directory of the rule, that is, the directory corresponding to the name of the package containing the rule under the genfiles or the bin tree.
```

Also added $(@D) to be equal to the path of the directory artifact when output_dir=True.

### DIFF
--- a/internal/node/npm_package_bin.bzl
+++ b/internal/node/npm_package_bin.bzl
@@ -21,20 +21,31 @@ _ATTRS = {
 # because the output_dir is a tree artifact
 # so we weren't able to give it a label
 def _expand_location(ctx, s):
-    outdir_segments = [ctx.bin_dir.path, ctx.label.package]
-    if ctx.attr.output_dir:
-        # We'll write into a newly created directory named after the rule
-        outdir_segments.append(ctx.attr.name)
+    rule_dir = [ctx.bin_dir.path, ctx.label.package]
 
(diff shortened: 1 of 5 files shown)


internal/node/npm_package_bin.bzl
```diff
@@ -21,20 +21,31 @@ _ATTRS = {
...
+    else:
         if s.find("$@") != -1 and len(ctx.outputs.outs) > 1:
             fail("""$@ substitution may only be used with a single out
             Upgrading rules_nodejs? Maybe you need to switch from $@ to $(RULEDIR)
             See https://github.com/bazelbuild/rules_nodejs/releases/tag/0.42.0""")
         s = s.replace("$@", ctx.outputs.outs[0].path)
+        if len(ctx.outputs.outs) == 1:
+            output_dir = ctx.outputs.outs[0].dirname.split("/")
+        else:
+            output_dir = rule_dir[:]
```
alexeagle: nit: not sure why you need to make a copy here

gregmagolan: No good reason looks like 🤷‍♂   I think i had an `output_dir.append(something)` at one point

PR #1408, opened 2019-12-01
fix(examples): jest module resolution

marcus-sa:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: Fix for Jest example

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/examples/jest/tools/patches/jest-resolve+24.7.1.patch
+++ b/examples/jest/tools/patches/jest-resolve+24.7.1.patch
@@ -9,21 +9,30 @@ index 64e7387..b5b74d3 100644
 -      result = _fs().default.realpathSync(result);
 +      // result = _fs().default.realpathSync(result);
      }
-
+ 
      return result;
 diff --git a/node_modules/jest-resolve/build/index.js b/node_modules/jest-resolve/build/index.js
-index 8e6bfe0..c3a9f27 100644
+index 8e6bfe0..decdac7 100644
 --- a/node_modules/jest-resolve/build/index.js
 +++ b/node_modules/jest-resolve/build/index.js
 @@ -57,7 +57,7 @@ function _defineProperty(obj, key, value) {
  const NATIVE_PLATFORM = 'native'; // We might be inside a symlink.
-
+ 
  const cwd = process.cwd();
 -const resolvedCwd = (0, _realpathNative().sync)(cwd) || cwd;
 +const resolvedCwd = cwd;
  const nodePaths = process.env.NODE_PATH
    ? process.env.NODE_PATH.split(_path().default.delimiter)
        .filter(Boolean) // The resolver expects absolute paths.
+@@ -215,6 +215,8 @@ class Resolver {
+   }
+ 
+   resolveModule(from, moduleName, options) {
++    return moduleName;
++
+     const dirname = _path().default.dirname(from);
+ 
+     const module =
 diff --git a/node_modules/jest-resolve/build/nodeModulesPaths.js b/node_modules/jest-resolve/build/nodeModulesPaths.js
 index 8d9d10f..5d55456 100644
 --- a/node_modules/jest-resolve/build/nodeModulesPaths.js
@@ -31,7 +40,7 @@ index 8d9d10f..5d55456 100644
 @@ -55,14 +55,7 @@ function nodeModulesPaths(basedir, options) {
    } // The node resolution algorithm (as implemented by NodeJS and TypeScript)
    // traverses parents of the physical path, not the symlinked path
-
+ 
 -  let physicalBasedir;
 -
 -  try {
@@ -41,6 +50,6 @@ index 8d9d10f..5d55456 100644
 -    physicalBasedir = basedirAbs;
 -  }
 +  let physicalBasedir = basedirAbs;
-
+ 
    const paths = [physicalBasedir];
-
+ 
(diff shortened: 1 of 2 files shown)


alexeagle:
I'm a bit confused. You labelled the pr "fix" but is there an issue associated that says what is wrong? Can you add a test case that fails without the change?

Also now that we have the node-patches package, we expect that jest should run correctly under Bazel with none of these patches required. I'm testing that in #1385 (it's blocked because it's passing unexpectedly when it should fail)

marcus-sa:
@alexeagle my bad.
When running the Jest example locally I get `Cannot find module '@babel/preset-env' ...` because the Jest module resolution wasn't being patched, which this fixes.

Aigeruth:
I cloned this repository and tried to run the `jest` example locally and I noticed that if I run `bazel test //examples:examples_jest` from the root, it passes, but if I change to the `examples/jest` directory and run `bazel test //:test`, it fails with:

```
==================== Test output for //:test:
FAIL ./index.test.js
  ● Test suite failed to run

    Cannot find module '@babel/preset-env' from '/private/var/tmp/_bazel_user/6e2dd2cd3b6540e0aea13f097ff4d11e/sandbox/darwin-sandbox/13/execroot/examples_jest/bazel-out/darwin-fastbuild/bin/test.sh.runfiles/examples_jest'

      at Function.module.exports [as sync] (../../../../../../../../../../../../../../../Users/user/dev/rules_nodejs/examples/jest/node_modules/resolve/lib/sync.js:74:15)
          at Array.map (<anonymous>)

FAIL ./index2.test.js
  ● Test suite failed to run

    Cannot find module '@babel/preset-env' from '/private/var/tmp/_bazel_user/6e2dd2cd3b6540e0aea13f097ff4d11e/sandbox/darwin-sandbox/13/execroot/examples_jest/bazel-out/darwin-fastbuild/bin/test.sh.runfiles/examples_jest'

      at Function.module.exports [as sync] (../../../../../../../../../../../../../../../Users/user/dev/rules_nodejs/examples/jest/node_modules/resolve/lib/sync.js:74:15)
          at Array.map (<anonymous>)

Test Suites: 2 failed, 2 total
Tests:       0 total
Snapshots:   0 total
Time:        3.689s
Ran all test suites within paths "./index.test.js", "./index2.test.js".
```

If I understand the recent changes correctly, the `node-patches` package should have an impact here, but I'm not sure why it might be being applied only when the test is run from the root directory. I would expect both versions to pass. What have I missed here?

I'm on macOS 10.15.1 and I use Bazelisk.

alexeagle:
Actually, the node-patches were not installed into the `jest_test` rule at all before. I'm cutting a release right now which fixes that, could you try again with 0.42.3?

Aigeruth:
@alexeagle, thank you for releasing 0.42.3! I tested it locally and both `bazel test //examples:examples_jest` and `examples/jest && bazel test //:test` are passing on `master` for me. I was also able to make `ts-jest` work in my project by upgrading to this version. Is was failing to load the preset configuration before.

marcus-sa:
@Aigeruth strange, `examples/jest` dIdn't work for me yesterday.

Aigeruth:
@marcus-sa, I ran the following command after pulling the latest `master`:

```sh
bazel test //examples:examples_jest > /dev/null; echo $?; cd examples/jest; bazel test //:test; echo $?
```

Both returned 0 as exit code and passed successfully.

I'm not sure what do you mean in your question, the latest `master` does not have the patch files anymore and the `WORKSPACE` file has stale references. I opened #1445 to remove those. If understand correctly, the new `@bazel/node-patches` package patches the `fs` module in nodejs to handle symlinks 'correctly' (as expected for Bazel).

I hope this helps.

marcus-sa:
@Aigeruth yeah it works now.
But the issue is still not resolved completely.
Jest cannot find npm modules imported inside tests.

Aigeruth:
@marcus-sa, I'm not sure what you mean by including dependencies twice. What I can see in the Stack Overflow example is that the ES5 output group is selected explicitly from the outputs of `ts_library()`, but I cannot see any duplication there.

I'm happy to share my approach in case it helps. I have a small React project with TypeScript source files. I don't use `ts_library()` as an input for `jest()`, I use the [`ts-jest` npm package](https://github.com/kulshekhar/ts-jest) to work with tests written in TypeScript. 

```python
TEST_PATTERNS = [
    "*.test.ts",
    "*.test.tsx",
]
APP_PATTERNS = [
    "*.ts",
    "*.tsx",
]
APP_FILES = glob(APP_PATTERNS, exclude = TEST_PATTERNS)
APP_TESTS = glob(TEST_PATTERNS)
APP_WITH_TESTS = glob(APP_PATTERNS)
APP_DEPENDENCIES = [
    "@npm//react",
    "@npm//react-dom",
    "@npm//@material-ui/core",
]
APP_TYPES = [
    "@npm//@types/react",
    "@npm//@types/react-dom",
    "@npm//@types/material-ui",
]


jest_test(
    name = "tests",
    args = [
        "--no-cache",
        "--no-watchman",
        "--ci",
        "--config=$(location :jest.config.js)",
    ] + [
        "--runTestsByPath $(location %s)" % src for src in APP_TESTS
    ],
    data = [
        ":jest.config.js",
	"//:tsconfig.json",
        "@npm//@types/jest",
        "@npm//ts-jest",
	"@npm//@testing-library/react",
    ] + APP_WITH_TESTS + APP_TYPES + APP_DEPENDENCIES,
)
```

marcus-sa:
@Aigeruth I did get it to work though, but thanks anyways.
Yeah I'm using `ts_library` under the hood, so for me I had to
1) include the same runtime dependencies in both `ts_library` and in `jest_test`
2) include the workspace name in `moduleNameWrapper` along with the module mapped paths since they don't get flattened somehow.

alexeagle:
This is a great discussion but I think we're in the wrong spot - the PR as originally proposed doesn't seem to be needed so I'll close this. Let's continue on the related linked threads where there's still a problem in rules_nodejs.

PR #1428, opened 2019-12-06
test: cleanup user_managed_deps e2e test

gregmagolan:
* e2e/user_managed_deps now tested in its own circleci job from within the working directory; I tried using working_directory in bazelci but for some reason the node_modules filegroup glob was always empty
* remove the old dormant test_legacy_runfiles circleci job
* remove @e2e_packages workspace as it is not being used anymore


gregmagolan:
Pre-req for https://github.com/bazelbuild/rules_nodejs/pull/1426 as user_managed_deps is causing problems with that PR

gregmagolan:
> Why does the user_managed_deps example have a new typescript dependency? It doesn't seem to have any typescript code.

I shall address that question in the leaky transitive deps section of the `Why Not To Have Bazel Dependencies` lightning talk at BazelCon next week :)

Issue #1382, opened 2019-11-22
enable linker with nodejs_binary


Toxicable:
# 🚀 feature request

Tracking enabling the linker with `nodejs_binary`

mrmeku:
# Making nodejs_binary use the linker
## Problem
Currently nodejs_binary makes use of a node_modules within `external/npm`. This has caused `rules_nodejs` to have to maintain a custom require script such that source files in a user's workspace can require files as if they existed within a `node_modules` directory in the root of their repo.

Unfortunately this strategy does not always work as expected in practice. Many libraries are written with the expectation of a directory layout that mirrors the node module resolution algorithm. Even Angular's own `karma` testing library has that expectation written in multiple places.


## Solution

Rather than making custom patches or PRs for every library which has such an expection, we instead would rather just provide the directory structure that programs expect to ensure maximum compatibility with execution under bazel

## Hooking up the linker

Taking [`run_node`](https://github.com/bazelbuild/rules_nodejs/blob/bc848f9d58175dad727b3285f36e7db88e5aeb48/internal/providers/node_runtime_deps_info.bzl#L41) as inspiration (since it already hooked up to the linker), we need to do a couple of things to successfully hook up `nodejs_binary`.

First and foremost we need to do something similar to what [`register_node_modules_linker`](https://github.com/bazelbuild/rules_nodejs/blob/bc848f9d58175dad727b3285f36e7db88e5aeb48/internal/linker/link_node_modules.bzl#L37) does under the hood.
1) Write a module manifest to a known location (we can use the existing implementation if we extract it to a helper function)
2) Inform our binary about the module manifest using the `--bazel_node_modules_manifest` flag. (NOTE: `register_node_modules_linker` can add this argument directly but we cannot since the build step of `nodejs_binary` does not invoke an action, but rather makes an `.sh` script which can be run using `bazel run`)

So assuming we extract the module manifest creation logic to a helper function and call it directly, we then need to add its location to our generated `.sh` script.

The current strategy for adding file locations to our runner script is via the use of `TEMPLATED` variables that are expanded using `ctx.actions.expand_tempalate` [here](https://github.com/bazelbuild/rules_nodejs/blob/bc848f9d58175dad727b3285f36e7db88e5aeb48/internal/node/node.bzl#L110). We will need to add one more templated variable for this purpose.

[...]

PR #1430, opened 2019-12-06
Hook up linker to nodejs_binary/nodejs_test

alexeagle:
For #1382 
Pairing with @mrmeku on it today

### DIFF
--- a/internal/linker/link_node_modules.bzl
+++ b/internal/linker/link_node_modules.bzl
@@ -34,14 +34,12 @@ def add_arg(args, arg):
     else:
         args.add(arg)
 
-def register_node_modules_linker(ctx, args, inputs, data = []):
-    """Helps an action to run node by setting up the node_modules linker as a pre-process
+def write_node_modules_manifest(ctx, extra_data = []):
+    """Writes a manifest file read by the linker, containing info about resolving runtime dependencies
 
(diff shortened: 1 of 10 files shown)


examples/jest/package.json
```diff
@@ -18,6 +18,6 @@
     },
     "scripts": {
         "test": "bazel test ...",
-        "postinstall": "patch-package --patch-dir=./tools/patches"
+        "// postinstall": "patch-package --patch-dir=./tools/patches"
```
gregmagolan: remove entirely?

mrmeku:
@alexeagle Looks like the only failing test was already failing in prior builds: https://buildkite.com/bazel/rules-nodejs-nodejs/builds/5046

gregmagolan:
Ahh. Thanks for pointing that out. It's a strange heisenbug that shows up only on bazelci osx. I haven't been able to reproduce it locally either.

mrmeku:
@gregmagolan The package the error references is `jest-haste-map` so maybe try to lock down the dependency for that dependency even more by using the `resolutions` section of package.json: https://yarnpkg.com/lang/en/docs/selective-version-resolutions/

And if that doesn't work just pray to your chosen deity and ask for forgiveness

PR #1432, opened 2019-12-07
fix(builtin): link module_name to directories recursively to avoid directory clashes

### LINKED ISSUE #1411: module_name linking creates invalid links in some cases
Will try to collect and provide more info, but I think at minimum there's issues when. A lot of these cases overlap though:

Inconsistent sibling mappings (one name/dir different):
```
@foo/bar/a => /path/x
@foo/bar/b => /path/b
@foo/bar/c => /path/c
```
Inconsistent siblings mappings (one path having an extra entry):
```
@foo/bar/a => /path/x/a
@foo/bar/b => /path/b
@foo/bar/c => /path/c
```
Parent having an extra mapping:
```
@foo/bar => /path/other/lib
@foo/bar/a => /path/lib/a
@foo/bar/b => /path/lib/b
@foo/bar/c => /path/lib/c
```
Siblings with different link types:
```
@foo/bar/a => /path/lib/a (bin)
@foo/bar/b => /path/lib/b (src)
@foo/bar/c => /path/lib/c (bin)
```
Parent with different link type:
```
@foo/bar => /path/lib (src)
@foo/bar/a => /path/lib/a (bin)
@foo/bar/b => /path/lib/b (bin)
@foo/bar/c => /path/lib/c (bin)
```

jbedard:
Fixes #1411

The overall change is to build a recursive hierarchy instead of just an array of arrays (one level of the hierarchy?). That hierarchy is then the order changes must be processed in, siblings in the hierarchy can be processed concurrently as long as the parent was already processed.

IDK if such a drastic change was really necessary, but it's the path I started down. Maybe we can just adjust the existing implementation but keep the tests?

This also does a bit less `mkdirp` calls, the `@foo/c/c/c/c` test now has I think 17 mkdirp/symlink calls instead of 21 🤷‍♂ 

~~2 of the tests fail before the change, I might try finding more cases though...~~
5 of the tests fail before the change, I think all of which are ~unique situations. I've outlined these more in the issue.

### DIFF
--- a/internal/linker/test/link_node_modules.spec.ts
+++ b/internal/linker/test/link_node_modules.spec.ts
@@ -414,7 +414,7 @@ describe('link_node_modules', () => {
         .toEqual('/*@foo/d/bar/fum/far*/exports = {}');
   });
 
-  it('should handle first-party packages with sibling directories', async () => {
+  it('should handle first-party packages with single parent link', async () => {
     // Set the cwd() like Bazel would in the execroot
     process.chdir(workspace);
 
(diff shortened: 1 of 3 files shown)


gregmagolan:
Nice! Thanks for taking a look at this @jbedard . I'll TAL on Friday after BazelCon.

internal/linker/test/link_node_modules.spec.ts
```diff
@@ -150,55 +310,361 @@ describe('link_node_modules', () => {
...
-    expect(fs.readFileSync(
-               path.join(
-                   process.env['TEST_TMPDIR']!, workspace, 'node_modules', '@foo', 'd', 'bar',
-                   'fum', 'far', 'index.js'),
-               'utf-8'))
+    expect(readWorkspaceNodeModules('@foo', 'd', 'bar', 'fum', 'far', 'index.js'))
         .toEqual('/*@foo/d/bar/fum/far*/exports = {}');
   });
 
+  it('should handle first-party packages with sibling directories', async () => {
```
alexeagle: These tests talking to the real FS are slower. It seems like you add all the same coverage below as in the IN/OUT unit tests above. Do we really need both? Do these new tests below add coverage that we don't get from the ones above?

jbedard: They go one step further and test the use of that OUT tree, concurrently traversing it and applying it to the FS.

They could also be run with both the old and new code so I could assure I'm actually fixing things, the test failed before and passes now etc.

We could potentially remove a few now that we're (I hope) more confident everything is working. Only keep the ones that concentrate more on applying the tree to the FS...

jbedard:
I've updated this with a couple changes:
* rebased
* [updated some comments](https://github.com/bazelbuild/rules_nodejs/pull/1432/commits/4782d9faf0c0238f7c6f4e6caa97ed1ca0ae5649)
* [added some `reduceLink` tests](https://github.com/bazelbuild/rules_nodejs/pull/1432/commits/16b3d555be163b3cd6ea7c47dd9f3e266a1d7682) for some cases previously only the fs tests covered
* [adjustd the fs tests](https://github.com/bazelbuild/rules_nodejs/pull/1432/commits/e868fa044cf8b99c7b04b0fa3154b5527b79ffe4) to test the unique fs/link layouts, not the unique module_name => link conversions (that's what the IN/OUT tests are for). This includes deleting some cases which I think are unique module_name=> link cases, not unique fs/link cases. @alexeagle I think this addresses your point and reduces the number of these tests.

I've left those as fixup commits for now so you can see the changes. If we're good I or whoever can squash+merge.

PR #1433, opened 2019-12-09
feat(karma): add pre_test attribute to karma

gregmagolan:
Runs optional arbitrary binaries before and after running karma test

### DIFF
--- a/packages/karma/src/karma_web_test.bzl
+++ b/packages/karma/src/karma_web_test.bzl
@@ -66,6 +66,18 @@ KARMA_WEB_TEST_ATTRS = {
         cfg = "target",
         allow_files = True,
     ),
+    "post_test": attr.label(
+        doc = "Optional binary to run after karma.",
+        allow_files = True,
+        cfg = "target",
+        executable = True,
+    ),
+    "pre_test": attr.label(
+        doc = "Optional binary to run before karma.",
+        allow_files = True,
+        cfg = "target",
+        executable = True,
+    ),
     "static_files": attr.label_list(
         doc = """Arbitrary files which are available to be served on request.
         Files are served at:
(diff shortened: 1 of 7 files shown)


packages/karma/src/karma_web_test.bzl
```diff
@@ -66,6 +66,12 @@ KARMA_WEB_TEST_ATTRS = {
         cfg = "target",
         allow_files = True,
     ),
+    "pre_test": attr.label(
+        doc = "Optional binary to run before karma.",
```
alexeagle: describe a scenario where this would be used? probably useful for users to understand what Angular plans to do with this

gregmagolan:
This is better accomplished with a custom karma js wrapper that calls a pre-test before executing karma:

```
// Execute the pretest & then launch karma
const shell = require('shelljs');
const karmaBin = require.resolve('karma/bin/karma');
const preTest = require.resolve(process.argv[2]);
process.argv = [
  process.argv[0],
  karmaBin,
  ...process.argv.splice(3),
]
try {
  console.log(`Executing pre-test (${preTest})...`)
  const result = shell.exec(preTest).code;
  if (result !== 0) {
    throw new Error(`Pre-test failed with code ${result}`);
  }
  console.log(`Launching karma ${karmaBin}...`)
  module.constructor._load(karmaBin, this, /*isMain=*/true);
} catch (e) {
  console.error(e.stack || e);
  process.exit(1);
}
```

See https://github.com/angular/angular/pull/34769

gregmagolan:
> why can't we just modify the karma binary to do this?
> 
> if there's a reason the test process can't just do the before/after behavior, then it feels like this isn't a karma-specific feature. Then maybe it makes sense for nodejs_test to have a way to give a binary to run before and after the primary one inside the action.

Maybe it would make sense for nodejs_test and nodejs_binary to have `pre` and `post` attributes which would allow you to chain these binaries together 🤔  Future problem.

PR #1435, opened 2019-12-10
perf: avoid unnecessary nested depset()

jbedard:
As described in [the bazel docs](https://docs.bazel.build/versions/master/skylark/performance.html#use-depsets) unnecessarily nesting `depset`s is a performance anti-pattern.

I think all these cases are pretty basic changes?

### DIFF
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -142,26 +142,31 @@ def _to_execroot_path(ctx, file):
 
 def _nodejs_binary_impl(ctx):
     node_modules_manifest = write_node_modules_manifest(ctx)
-    node_modules = depset(ctx.files.node_modules)
+    node_modules_depsets = []
+    node_modules_depsets.append(depset(ctx.files.node_modules))
 
     # Also include files from npm fine grained deps as inputs.
     # These deps are identified by the NpmPackageInfo provider.
     for d in ctx.attr.data:
         if NpmPackageInfo in d:
-            node_modules = depset(transitive = [node_modules, d[NpmPackageInfo].sources])
+            node_modules_depsets.append(d[NpmPackageInfo].sources)
 
-    # Using a depset will allow us to avoid flattening files and sources
+    node_modules = depset(transitive = node_modules_depsets)
+
+    # Using an array of depsets will allow us to avoid flattening files and sources
     # inside this loop. This should reduce the performances hits,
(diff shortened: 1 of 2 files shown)


internal/node/node.bzl
```diff
@@ -247,23 +254,27 @@ def _nodejs_binary_impl(ctx):
...
     if ctx.file.entry_point.extension == "js":
-        runfiles = depset([ctx.file.entry_point], transitive = [runfiles])
+        runfiles_depsets.append(ctx.file.entry_point)
 
     return [
         DefaultInfo(
             executable = executable,
             runfiles = ctx.runfiles(
-                transitive_files = runfiles,
+                transitive_files = depset(runfiles_depsets),
```
alexeagle: the fact that you don't use `transitive = runfiles_depsets` makes me think the variable is misnamed, and runfiles_depsets is actually just runfiles list

jbedard: Yeah I guess it is just plain files, previously those plain files were converted into transitive `depset`s though. I think this change is ok though, but the var should be renamed...

jbedard: Updated to just `runfiles`

internal/node/node.bzl
```diff
@@ -142,26 +142,31 @@ def _to_execroot_path(ctx, file):
...
-    # Using a depset will allow us to avoid flattening files and sources
+    node_modules = depset(transitive = node_modules_depsets)
+
+    # Using an array of depsets will allow us to avoid flattening files and sources
     # inside this loop. This should reduce the performances hits,
     # since we don't need to call .to_list()
-    sources_sets = []
+    # Also avoid deap transitive depset()s by creating single array of
+    # transitive depset()s
+    sources_depsets = []
```
jbedard: This is just a rename after rebasing. Revert?

jbedard:
Updated. I think the buildkite failure is random?

jbedard:
Or maybe it isn't random? I don't see how that could be related though...

jbedard:
@alexeagle the build failure seems to be the same thing as PRs such as https://github.com/bazelbuild/rules_nodejs/pull/1445 - so I guess it is random and unrelated to this change?

alexeagle:
Yes that test is terribly flaky right now, in that it sometimes but rarely passes. We need to jump on that this week.

PR #1439, opened 2019-12-10
fix(bazel): no longer always include the version_file in inputs

josephperrott:
While bazel properly handles not causing rebuilds based on version_file locally,
if the version_file is included in the inputs it prevents caching from ever
working when caching using remote cache or disk cache.

Instead of determining if stamping should occur based on the presence of the
version_file, we should choose to stamp based on an attribute stamp.


alexeagle:
Replaced by https://github.com/bazelbuild/rules_nodejs/pull/1441 which hooks it up using an internal target that gathers the config setting

PR #1440, opened 2019-12-10
refactor: always link

alexeagle:
discussed in meeting with @soldair - we can decouple enabling the linker/node-patches from disabling the custom monkey-patched _loader.js

We should be able to always enable the linker and node-patches, while keeping the loader.js for existing code that depends on it. require()s that aren't handled by loader.js will fall through to what the linker lays out on disk.


alexeagle:
Will file an issue to investigate the 8-char truncated windows paths to see where it comes from.

PR #1441, opened 2019-12-10
fix(builtin): only stamp artifacts when --stamp is passed to bazel

alexeagle:
Fixes remote cache misses of rollup_bundle and npm_package targets observed by Angular team


docs/index.md
```diff
@@ -130,12 +130,19 @@ Bazel is generally only a build tool, and is unaware of your version control sys
 However, when publishing releases, you typically want to embed version information in the resulting distribution.
 Bazel supports this natively, using the following approach:
 
-First, pass the `workspace_status_command` argument to `bazel build`. We prefer to do this with an entry in `.bazelrc`:
+To stamp a build, you must pass the `--stamp` argument to Bazel.
+
+> Previous releases of rules_nodejs stamped builds always.
+> However this caused stamp-aware actions to never be remotely cached, since the volatile
+> status file is passed as an input and its checksum cannot be known.
```
josephperrott: nit: technically its that its checksum always changes

internal/npm_package/npm_package.bzl
```diff
@@ -129,6 +130,11 @@ NPM_PACKAGE_ATTRS = {
         doc = """Files inside this directory which are simply copied into the package.""",
         allow_files = True,
     ),
+    "node_context_data": attr.label(
```
josephperrott: If this is internal should it be `_` prefixed?

alexeagle: in the rollup case, I override from a test. I guess I have no such test right now for npm_package doing the stamp thing but I like them being the same, and we ought to have such a test

alexeagle: ah, the test for npm_package stamping was just busted. when I fixed the test it required passing this data to override --stamp. Fixed

PR #1450, opened 2019-12-13
feat(builtin): introduce copy_to_bin rule

alexeagle:
Based on meteorcloudy@ awesome work in https://github.com/bazelbuild/bazel-skylib/pull/217
That PR isn't getting approved for upstream so we vendor it into our own ruleset


gregmagolan:
Beauty. This is really nice to have.

PR #1455, opened 2019-12-17
Hook up linker to all generated /bin:foo targets

alexeagle:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/node-patches/src/subprocess.ts
+++ b/packages/node-patches/src/subprocess.ts
@@ -47,7 +47,12 @@ fi
   }
 
   // fix execPath so folks use the proxy node
-  process.argv[0] = process.execPath = path.join(nodeDir, 'node');
+  if (process.platform == 'win32') {
+    // FIXME: need to make an exe, or run in a shell so we can use .bat
+  } else {
+    process.argv[0] = process.execPath = path.join(nodeDir, 'node');
+  }
+
 
   // replace any instances of require script in execArgv with the absolute path to the script.
   // example: bazel-require-script.js
--- a/examples/BUILD.bazel
+++ b/examples/BUILD.bazel
@@ -91,6 +91,10 @@ example_integration_test(
     npm_packages = {
         "//packages/jasmine:npm_package": "@bazel/jasmine",
     },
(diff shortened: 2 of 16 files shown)


alexeagle:
This was peeled out of #1430 to reduce scope there making it easier to land

alexeagle:
finally getting back to this now (I installed a windows boot camp so I can really debug, once I get things like git in cmd.exe sorted...)

@alan-agius4 you had some recommendation for fixing this (.exe or non-interactive shell IIRC) but I think it's lost to the bazel freebie slack dropping messages

packages/node-patches/src/subprocess.ts
```diff
@@ -47,7 +47,8 @@ fi
   }
 
   // fix execPath so folks use the proxy node
-  process.argv[0] = process.execPath = path.join(nodeDir, 'node');
+  process.argv[0] = process.execPath =
```
alan-agius4, line 50: In Windows, this will work only if `shell` is enabled when third parties `fork`or `spawn` a process.

alan-agius4:
@alexeagle I did have the same issue this week when I tried to look up an old convo.

So in Windows you cannot `spawn` or `fork` a `.bat` or `.cmd` files unless the shell option is enabled. (Which is not by default). The problem is that `fork` doesn't have a `shell` option and when provided it is ignored. See https://nodejs.org/api/child_process.html#child_process_child_process_fork_modulepath_args_options

My idea to get around this is to leverage the fact that when a child process is spawned/forked both `env` and `execArgv` are inherited from the parent process. This is important because this means that we shouldn't be needed by run the patch for all process creation but rather just for the parent one, because the patch updates the `execArgv` and environment variables.

So if, we run the patch and set after set `process.execPath` to a `node.exe`, we should be good. Alternatively we can have an `exe` with the patch which I think it was your initial idea.

gregmagolan:
@googlebot I consent.

alexeagle:
@googlebot I consent

PR #1463, opened 2019-12-17
refactor(labs): webpack_bundle

rayman1104:
## PR Checklist

Please check if your PR fulfils the following requirements:

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Current webpack_bundle api looks outdated and it's not clear how to use it for some cases (i.e. in the angular example).

## What is the new behavior?

I tried to make webpack_bundle api almost the same as in rollup_bundle. Also, after this update, rollup_bundle step can be easily substituted by webpack_bundle in rare situations when it's needed.
Here is the example of using it in the Angular app: https://github.com/rayman1104/bazel_angular_webpack/blob/master/src/BUILD.bazel#L58

Maybe. But breaking changes in Labs are expected.

## Other information
In rare cases, some external dependencies are hard to be built with rollup_bundle. Giving the example, I didn't find workaround how to build an Angular app via rollup_bundle, when having protobufjs dependency and code-splitting simultaneously. And looks like it's not the problem of Bazel but of protobufjs and rollup (with not-iife format) incompatibility. But it can be solved if we use webpack. Here is the repro for rollup issue:
https://github.com/rayman1104/bazel_angular_webpack/tree/protobufjs

For such cases, it may be convenient to have an option to pack the bundle via webpack. Using webpack-cli directly as in react_webpack example is a good option too, but it looks not so convenient in Angular app, because we can't just pass ts_library outputs to deps and pass entry_points via labels, etc.


marcus-sa:
@rayman1104 I'm currently loading up the https://github.com/marcus-sa/bazel-tools with a bunch of custom wrapper rules around NPM packages and whole lot of examples that doesn't require specific rules but rather macros.
Feel free to do a PR there if you want.

rayman1104:
@alexeagle Is it possible to pack the angular example app by just calling the webpack-cli? I guess we will need to extract all sources for each ts_library, ng_module and also for npm dependencies? I just didn't find a good way and implementing webpack_bundle looked like a simpler solution for me.
Also, I mentioned below in the "Other Information" section why I needed webpack in the angular app.

alexeagle:
@rayman1104 ah! Actually I think you want to go a step further - don't configure webpack yourself, if you want to build an Angular app you should probably let Angular CLI code (via their "architect" subsystem) do this for you. See https://github.com/alexeagle/prototype_bazel_architect/blob/master/BUILD.bazel for the direction we are going for out-of-the-box Bazel + Angular working in a fully compatible way.

Customizing the webpack setup is generally only in the realm of experts, maybe you are one of them.

rayman1104:
@alexeagle That looks interesting, thanks, I'll have a look.
The only concern there is that we don't separate tsc and sass compilation build steps.

I'm definitely not a webpack expert :) I just made angular bundling correctly with a simple webpack config in the prototype. But maybe it won't work at scale and I see only the tip of the iceberg.

Thanks for the feedback! Want to try now a macro wrapper for the webpack-cli rule.

alexeagle:
A macro wrapper is a great way to keep it simple. Note that you won't be able to access Bazel "providers" which is how the rollup_bundle rule ensures that it gets ESnext javascript from its transitive dependencies, so there are still trade-offs for writing a rule as you've done here

kylecordes:
I worked with webpack recently, and hit a spot where a more sophisticated rule rather than simple use of the default CLI would be nice to have. The explanation is quite long-winded, and is included in a video I posted a couple of days ago. But this presumably no one reading along here wants to go watch an hour video :-)

To make a long story short, you end up need to know quite a bit about how to webpack interacts with paths, exemplified by this line of code:

https://github.com/OasisDigital/mobx-bazel-build/blob/master/webpack-test/webpack_4_cjs/BUILD.bazel#L14

…in which you have to "just know" where and why to prefix a path location provided by Bazel ($(location …)) to make it do the right thing in the webpack.

I can imagine a comprehensive rule dealing this bit and many others.

However, I'm just mentioning this here for the sake of completeness of anyone reading later. I generally much agree with the idea that it is best to be able to immediately use any NPM-shipped command. There are literally thousands of these that someone might want to use, and there is a **not** lot of interest out there in creating or maintaining thousands of Bazel rules to keep up with the evolution of each!

PR #1464, opened 2019-12-18
refactor: remove install_source_map_support from nodejs_binary since it is vendored in

gregmagolan:
BREAKING CHANGE:
`install_source_map_support` attribute removed from `nodejs_binary`. `source-map-support` is vendored in at `/third_party/github.com/source-map-support` so it can always be installed.

### DIFF
--- a/internal/bazel_integration_test/BUILD.bazel
+++ b/internal/bazel_integration_test/BUILD.bazel
@@ -22,8 +22,6 @@ nodejs_binary(
     configuration_env_vars = ["BAZEL_INTEGRATION_TEST_DEBUG"],
     data = ["@npm//tmp"],
     entry_point = ":test_runner.js",
-    # reduce memory usage by disabling source_map_support
-    install_source_map_support = False,
     templated_args = ["--node_options=--max-old-space-size=1024"],
 )
 
--- a/examples/angular/tools/angular_prerender.bzl
+++ b/examples/angular/tools/angular_prerender.bzl
@@ -45,7 +45,6 @@ def ng_prerender(name, index, prerender_roots = [], **kwargs):
             "@npm//domino",
             "@npm//reflect-metadata",
         ],
-        install_source_map_support = False,
         entry_point = "//src:prerender.ts",
     )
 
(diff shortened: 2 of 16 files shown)


internal/node/node_loader.js
```diff
@@ -493,17 +492,15 @@ module.constructor._resolveFilename = function(request, parent, isMain, options)
...
-    const sourcemap_support_package = path.resolve(process.cwd(),
-          '../build_bazel_rules_nodejs/third_party/github.com/source-map-support');
-    require(sourcemap_support_package).install();
-  } catch (_) {
-    log_verbose(`WARNING: source-map-support module not installed.
-      Stack traces from languages like TypeScript will point to generated .js files.
-      Set install_source_map_support = False in ${TARGET} to turn off this warning.`);
-  }
+try {
+  const sourcemap_support_package = path.resolve(
```
alexeagle: @soldair has pointed out we can remove this from the loader (which we want to delete) by passing the script to the `--require` flag in the launcher.

internal/bazel_integration_test/BUILD.bazel
```diff
@@ -21,8 +21,6 @@ nodejs_binary(
     name = "test_runner",
     data = ["@npm//tmp"],
     entry_point = ":test_runner.js",
```
alexeagle: the comment here was about reducing memory - I still think it may be useful to leave the flag so users can get out of trouble if source_map_support is really the wrong thing for their program

gregmagolan: I was reducing memory at the time to get bazel-in-bazel working but I don't think it was material in the effort as its not a large library. The bazel and JVM memory flags are far more significant so I should just remove that comment.

As far as leaving it in. It doesn't hurt anything to do so. In the API review we decided to remove it which is where this PR came from.

internal/node/node_loader.js
Toxicable: another advantage of the `--require` method is that it'll be passed down to workers and child processes

gregmagolan: 👍

alexeagle:
I think we should wait for a rules_sass release so users don't have to patch it themselves

gregmagolan:
waiting on a rule_sass release

gregmagolan:
Rules sass 1.25.0 is release. This is unblocked but should wait to land until we're ready to release 2.0 since it is a BREAKING CHANGE.

PR #1465, opened 2019-12-18
refactor: renamed nodejs_binary `templated_args` to `args`

gregmagolan:
refactor: renamed nodejs_binary & nodejs_test `templated_args` to `args`

* with args, $(location) and $(locations) are automatically expanded but the expansions no longer include the workspace name; instead they are of the standard short path form `./path/in/local/repo` and `../external_repo/path/in/external/repo`
* also removed related `templated_args_file` and added a trivial `params_file` rule to public API to replace this functionality (see /internal/common/test/BUILD.bazel for usage example)
* to match standard $(location) and $(locations) expansion, `params_file` args location expansions are also in the standard short_path form (this differs from the old templated_args behavior which was not Bazel idiomatic)
* bug fix in linker where bin folder was being included in runfiles path for tests when link type was 'bin'

BREAKING CHANGE:

* nodejs_binary & nodejs_test attribute `templated_args` renamed to `args`; the name should not be tied to its implementation details

* bootstrap attribute in nodejs_binary, nodejs_test & jasmine_node_test removed

This can be replaced with the `--node_options=--require=$(location label)` argument such as,
```
nodejs_test(
    name = "bootstrap_test",
    args = ["--node_options=--require=./$(location :bootstrap.js)"],
    entry_point = ":bootstrap.spec.js",
    data = ["bootstrap.js"],
)
```
or
```
jasmine_node_test(
    name = "bootstrap_test",
    srcs = ["bootstrap.spec.js"],
    args = ["--node_options=--require=./$(location :bootstrap.js)"],
    data = ["bootstrap.js"],
)
```

* now that args in nodejs_binary are not templated, protractor_test_suite on-prepare may need to pass arguments to its server executable if this server executable target has required args


gregmagolan:
Will break this one up into smaller PRs

gregmagolan:
Replaced by https://github.com/bazelbuild/rules_nodejs/pull/1488 & https://github.com/bazelbuild/rules_nodejs/pull/1490

PR #1469, opened 2019-12-18
refactor(builtin): make bootstrap attribute in nodejs_binary/test a label

gregmagolan:
BREAKING CHANGE:
bootstrap attribute in nodejs_binary & nodejs_test is now a label


gregmagolan:
> Can we kill this feature instead? this controls behavior in the node_loader.js - you ought to instead pass a `--node_options=--require=myscript` argument

Good call. I'll update the tests and make sure that works and then kill it.

alexeagle:
#1488 removes the bootstrap attribute instead

PR #1488, opened 2019-12-20
refactor: remove `bootstrap` attribute & fix $(location) expansions in nodejs_binary templated_args

gregmagolan:
* fix(builtin): bug fix in linker where bin folder was being included in runfiles path for tests when link type was 'bin'

BREAKING CHANGE:
bootstrap attribute in nodejs_binary, nodejs_test & jasmine_node_test removed

This can be replaced with the `--node_options=--require=$(location label)` argument such as,

```
nodejs_test(
name = "bootstrap_test",
templated_args = ["--node_options=--require=$(rlocation $(location :bootstrap.js))"],
entry_point = ":bootstrap.spec.js",
data = ["bootstrap.js"],
)
```
or
```
jasmine_node_test(
name = "bootstrap_test",
srcs = ["bootstrap.spec.js"],
templated_args = ["--node_options=--require=$(rlocation $(location :bootstrap.js))"],
data = ["bootstrap.js"],
)
```

`templated_args` `$(location)` and `$(locations)` are now correctly expanded when there is no space before ` $(location`
such as `templated_args = ["--node_options=--require=$(rlocation $(location :bootstrap.js))"]`.

Path is returned in runfiles manifest path format such as `repo/path/to/file`. This differs from how $(location)
and $(locations) expansion behaves in expansion the `args` attribute of a *_binary or *_test which returns
the runfiles short path of the format `./path/to/file` for user repo and `../external_repo/path/to/file` for external
repositories. We may change this behavior in the future with $(mlocation) and $(mlocations) used to expand
to the runfiles manifest path.
See https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes-binaries.

PR #1490, opened 2019-12-20
refactor: remove templated_args_file from nodejs_binary & nodejs_test

gregmagolan:
*Branched from https://github.com/bazelbuild/rules_nodejs/pull/1488 which should land first*

@mattem This should still work for your use case and it provides a nice example of how to generate and pass in your own custom params file. See usage example below.

BREAKING CHANGE:
`templated_args_file` removed from nodejs_binary, nodejs_test & jasmine_node_test. This was a separation of concerns and complicated node.bzl more than necessary while also being rigid in how the params file is formatted. It is more flexible to expose this functionality as another simple rule named params_file.

To match standard $(location) and $(locations) expansion, `params_file` args location expansions are also in the standard short_path form (this differs from the old templated_args behavior which was not Bazel idiomatic)

Usage example:

```
load("@build_bazel_rules_nodejs//:index.bzl", "params_file", "nodejs_binary")

params_file(
    name = "params_file",
    args = [
        "--some_param",
        "$(location //path/to/some:file)",
        "--some_other_param",
        "$(location //path/to/some/other:file)",
    ],
    data = [
        "//path/to/some:file",
        "//path/to/some/other:file",
    ],
)

nodejs_binary(
    name = "my_binary",
    data = [":params_file"],
    entry_point = ":my_binary.js",
    templated_args = ["$(location :params_file)"],
)
```

### DIFF
--- a/index.bzl
+++ b/index.bzl
@@ -20,6 +20,7 @@ Users should not load files under "/internal"
 load("//internal/common:check_bazel_version.bzl", _check_bazel_version = "check_bazel_version")
 load("//internal/common:check_version.bzl", "check_version")
 load("//internal/common:copy_to_bin.bzl", _copy_to_bin = "copy_to_bin")
+load("//internal/common:params_file.bzl", _params_file = "params_file")
 load(
     "//internal/node:node.bzl",
     _nodejs_binary = "nodejs_binary",
@@ -39,6 +40,7 @@ pkg_npm = _pkg_npm
 npm_package_bin = _npm_bin
 pkg_web = _pkg_web
 copy_to_bin = _copy_to_bin
+params_file = _params_file
 # ANY RULES ADDED HERE SHOULD BE DOCUMENTED, see index.for_docs.bzl
 
 # Allows us to avoid a transitive dependency on bazel_skylib from leaking to users
--- a/internal/common/params_file.bzl
+++ b/internal/common/params_file.bzl
@@ -1,0 +1,98 @@
+# Copyright 2019 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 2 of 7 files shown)


mattem:
👍 Perhaps `params_file` could be merged with skylibs `write_file` [1], they are very similar, with the addition of params_file handling expanding locations.

[1] https://github.com/bazelbuild/bazel-skylib/blob/master/rules/private/write_file_private.bzl

gregmagolan:
@mattem Great idea. Done.

index.bzl
```diff
@@ -39,6 +40,7 @@ pkg_npm = _pkg_npm
 npm_package_bin = _npm_bin
 pkg_web = _pkg_web
 copy_to_bin = _copy_to_bin
+params_file = _params_file
 # ANY RULES ADDED HERE SHOULD BE DOCUMENTED, see index.for_docs.bzl
```
alexeagle: add to index.for_docs.bzl so it gets stardoc

internal/node/node.bzl
```diff
@@ -458,13 +433,6 @@ jasmine_node_test(
         `--node_options=--preserve-symlinks`
         """,
     ),
```
alexeagle: is it still the case that you can write `--node_options` into the params file? I suppose not because we don't read it at all before it is read in the node program that's already running.

I don't think anyone needed to spill node_options into the file (though I suppose if you need to bootstrap a hundred files you'll run into argv limit... you could solve that some other way like bundle up your bootstraps)

index.bzl
gregmagolan: done. thanks

internal/node/node.bzl
gregmagolan: good point. I'd can't think of why someone would want to --require 100 files but stranger things have happened. bundling would work in that case. 👍

internal/common/params_file.bzl
```diff
@@ -0,0 +1,98 @@
...
+        newline = "auto",
+        **kwargs):
+    """Generates a UTF-8 encoded params file from a list of arguments.
+
+    Handles $(location) expansions for arguments.
+
+    Args:
+      name: Name of the rule.
+      out: Path of the output file, relative to this package.
+      args: AArguments to concatenate into a params file.
```
mattem: nit: typo 'AArguments'
```suggestion
      args: Arguments to concatenate into a params file.
```

alexeagle: will fix during merge

alexeagle:
merged as 799acb45

internal/common/params_file.bzl
gregmagolan: thanks @mattem & @alexeagle

PR #1494, opened 2019-12-20
fix(builtin): make .pack and .publish targets work again

### LINKED ISSUE #1493: pkg_npm .pack and .publish bins are broken [1.0.0 regression]
See https://github.com/bazelbuild/rules_nodejs/commit/b4782b8eb064d07535b110d2cb7b3e2bc82f3599#commitcomment-36523065 for the root cause.

As a workaround you can patch out this change. Put this content into `rules_nodejs.patch` in the root of your repo (next to WORKSPACE):

```
diff --git internal/pkg_npm/pkg_npm.bzl internal/pkg_npm/pkg_npm.bzl
index c6263521..76210bc0 100644
--- internal/pkg_npm/pkg_npm.bzl
+++ internal/pkg_npm/pkg_npm.bzl
@@ -86,6 +86,7 @@ def create_package(ctx, deps_sources, nested_packages):
         inputs = inputs,
         outputs = [package_dir, ctx.outputs.pack, ctx.outputs.publish],
         arguments = [args],
+        execution_requirements = {"local": "1"},
     )
     return package_dir
```

And then add to the `http_archive(name = "build_bazel_rules_nodejs")` in the WORKSPACE this attribute:

`patches = ["//:rules_nodejs.patch"]`

alexeagle:
fixes #1493

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -6,6 +6,8 @@ tasks:
     platform: ubuntu1604
     run_targets:
     - "@nodejs//:yarn_node_repositories"
+    # Regression test for #1493
+    - "//packages/create:npm_package.pack"
     - "//internal/node/test:no_deps"
     - "//internal/node/test:has_deps_legacy"
     - "//internal/node/test:has_deps"
@@ -198,6 +200,8 @@ tasks:
     platform: macos
     run_targets:
     - "@nodejs//:yarn_node_repositories"
+    # Regression test for #1493
+    - "//packages/create:npm_package.pack"
     - "//internal/node/test:no_deps"
     - "//internal/node/test:has_deps_legacy"
     - "//internal/node/test:has_deps"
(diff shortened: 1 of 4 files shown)


.bazelci/presubmit.yml
```diff
@@ -6,6 +6,8 @@ tasks:
     platform: ubuntu1604
     run_targets:
     - "@nodejs//:yarn_node_repositories"
+    # Regression test for #1493
```
gregmagolan: Only a 6s task so I think also add to mac and Windows to make sure we're good on all platforms. I'll make a note to add a more rigorous test on the .pack output.

alexeagle: done

I don't think there's much need to test the logic or the result of the pack operation because the code here is just to run `npm pack` - very small surface area for bugs where this run command succeeds but does the wrong thing.

PR #1502, opened 2019-12-26
Nodejs binary module mappings

gregmagolan:
* fix(builtin): nodejs_binary collects module_mappings for linker

Module name cannot be equal to the workspace name with the linker so this change removes the regression test for 834. This was a corner case and was already broken with the 1.0 for npm_package_bin so this fix makes nodejs_binary consistent with npm_package_bin and the new desired behavior for 1.0.

* fix(builtin): legacy module_mappings_runtime_aspect handles dep with module_name but no module_root

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -25,7 +25,7 @@ load("//internal/common:expand_into_runfiles.bzl", "expand_location_into_runfile
 load("//internal/common:module_mappings.bzl", "module_mappings_runtime_aspect")
 load("//internal/common:path_utils.bzl", "strip_external")
 load("//internal/common:windows_utils.bzl", "create_windows_native_launcher_script", "is_windows")
-load("//internal/linker:link_node_modules.bzl", "write_node_modules_manifest")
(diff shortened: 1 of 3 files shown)


internal/node/node.bzl
```diff
@@ -277,7 +277,7 @@ _NODEJS_EXECUTABLE_ATTRS = {
     "data": attr.label_list(
         doc = """Runtime dependencies which may be loaded during execution.""",
         allow_files = True,
-        aspects = [node_modules_aspect, module_mappings_runtime_aspect],
+        aspects = [node_modules_aspect, module_mappings_aspect, module_mappings_runtime_aspect],
```
alexeagle: I was hoping to remove the module_mappings_runtime_aspect at the same time. This is a perf regression since we'll walk the deps two separate times, gathering the same data and returning it in slightly different format.
I guess we can do that as follow-up but ideally we should do it now? Soon?

PR #1504, opened 2019-12-26
feat(builtin): add deps to npm_package_bin that are passed down transitively to consuming rules

gregmagolan:
This allows for the following:

```
load("//tools:tsc.bzl", "tsc")
load("//tools:jest_test.bzl", "jest_test")

filegroup(
    name = "decl",
    srcs = glob(["src/**/*.d.ts"]),
)

tsc(
    name = "lib",
    srcs = glob(["*.ts", "src/**/*.ts"], exclude = ["**/*.d.ts"]),
    ts_config = ":tsconfig.json",
    data = [
        "//:tsconfig.json",
    ],
    deps = [
        "//packages/bar:mod",
        ":decl",
    ],
)

tsc(
    name = "tests",
    srcs = glob(["test/**/*.test.ts"], exclude = ["**/*.d.ts"]),
    ts_config = ":tsconfig.json",
    data = [
        "@npm//@types/jasmine",
        "//:tsconfig.json",
    ],
    deps = [
        ":lib",
    ],
)

jest_test(
    name = "test",
    tests = [":tests"],
    jest_config = "//:jest.config.js",
    data = [
        ":lib",
    ],
)
```
(https://github.com/gregmagolan/bazel_nodejs_demo/blob/master/packages/bar/BUILD.bazel)

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/internal/node/npm_package_bin.bzl
+++ b/internal/node/npm_package_bin.bzl
@@ -17,6 +17,7 @@ _ATTRS = {
         cfg = "host",
         mandatory = True,
     ),
+    "deps": attr.label_list(allow_files = True, aspects = [module_mappings_aspect, node_modules_aspect]),
 }
 
 def _expand_locations(ctx, s):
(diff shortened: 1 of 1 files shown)


alexeagle:
This seems like a well-formed change and gives an interesting new capability.

OTOH it's not sufficient for the TS/NG use case because those need to distinguish .js and .d.ts outputs as well.

Also it diverges more from the notion that npm_package_bin is a fancy genrule. Seems hard for users to correctly differentiate data and deps.

Also you might argue that `deps` is the right name for `data` in the case that no bin is run. So this could be `transitive_deps` or something.

Since we don't have a good case for why we need this yet, let's think about it some more...

Issue #1447, opened 2019-12-12
NPM modules can be required from npm/node_modules


mrmeku:
Reproduction:
https://github.com/mrmeku/prototype-nx-bazel/commit/0b090725039fe597db1afeccf757d6e136576ccb

Checkout the repo at the above commit.
Run `bazel test //apps/cart:test`

Examine the stacktrace to see the following lines of interest

`at batchedUpdates$1 (../../../npm/node_modules/react-dom/cjs/react-dom.development.js:24328:12)
      at act (../../../npm/node_modules/react-dom/cjs/react-dom-test-utils.development.js:1120:14)
      at Object.render (../../node_modules/@testing-library/react/dist/pure.js:82:26)
      at Object.<anonymous> (src/app/app.spec.tsx:12:29)`

Notice that the following has occured:
  - A linked npm module `@testing-library/react` attempts to import a library called `react-dom` (here is the line of interest: https://github.com/testing-library/react-testing-library/blob/c48b501bf683faae6fcf9cadb7ee96fc0da319ba/src/pure.js#L2)
  - The resolved module comes from `npm/node_modules` rather than from the linked node_modules directory

This causes problems in many packages. For react, an invariant of a workspace is that you only have one version of react installed. Because we have one linked copy and one copy within `npm` we have two copies and break that invariant.

I'm still confused how modules can be required from `npm` since the node modules resolution algorithm would not allow for that.

evahowe:
@mrmeku - can this get closed since it was addressed in #1505?

Issue #1489, opened 2019-12-20
Performance: Link node_modules using Runfiles symlinks


mrmeku:
See https://docs.bazel.build/versions/master/skylark/rules.html#runfiles-symlinks

Normally, the relative path of a file in the runfiles tree is the same as the relative path of that file in the source tree or generated output tree. If these need to be different for some reason, you can specify the root_symlinks or symlinks arguments. The root_symlinks is a dictionary mapping paths to files, where the paths are relative to the root of the runfiles directory.

If we were to link the node_modules directory using this bazel semantic, we could skip running the linker entirely since Bazel will have essentially done the work for us. Bazel is already doing this work when creating the run files directory, it unfortunately just links those files into `runfiles/npm/external` (which requires us to run the linker program to move the files to where they need to be)

alexeagle:
We still need the linker for the npm_package_bin case (a program run as a tool in the execroot to transform inputs to outputs under bazel build - no runfiles is set up for that)

This does look like an interesting refactor for the bazel run / bazel test case, we might be able to skip the linker for nodejs_binary/nodejs_test if we can configure it here. OTOH it means having two different ways to do the same mapping, since npm_package_bin would still need the linker as a standalone program.

gregmagolan:
No longer in scope for rules_nodejs which only supplies the Node.js toolchain as of [v6.0.0](https://github.com/bazelbuild/rules_nodejs/releases/tag/v6.0.0-beta0).

Downstream canonical JavaScript + Node.js ruleset is now https://github.com/aspect-build/rules_js.

PR #1505, opened 2019-12-26
fix(builtin): don't allow symlinks to escape into duplicate node_modules tree in runfiles

gregmagolan:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -181,7 +181,10 @@ def _nodejs_binary_impl(ctx):
 
     env_vars = "export BAZEL_TARGET=%s\n" % ctx.label
     env_vars += "export BAZEL_WORKSPACE=%s\n" % ctx.workspace_name
-    env_vars += "export NODE_MODULES_ROOT=%s\n" % node_modules_root
+    env_vars += """if [[ -z "${NODE_MODULES_ROOT:-}" ]]; then
+  export NODE_MODULES_ROOT=%s
+fi
+""" % node_modules_root
     for k in ctx.attr.configuration_env_vars + ctx.attr.default_env_vars:
         # Check ctx.var first & if env var not in there then check
         # ctx.configuration.default_shell_env. The former will contain values from --define=FOO=BAR
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -78,7 +78,7 @@ def _compute_node_modules_root(ctx):
         ] if f])
     return node_modules_root
 
-def _write_require_patch_script(ctx):
+def _write_require_patch_script(ctx, node_modules_root):
     # Generates the JavaScript snippet of module roots mappings, with each entry
     # in the form:
     #   {module_name: /^mod_name\b/, module_root: 'path/to/mod_name'}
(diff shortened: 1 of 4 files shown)


mrmeku:
@gregmagolan Just wanted to say that I tested out this PR and it was the magic bullet for me to get jest tests working in a react project. Nice work Greg!

mrmeku:
Fixes https://github.com/bazelbuild/rules_nodejs/issues/1447

mrmeku:
@gregmagolan You should consider applying the same fix to NPM package bin. With this fix there is now a mismatch in behavior between `npm_package_bin` and `nodejs_binary`. In the former, node modules can still be resolved from within `npm/modules`. To rectify, I think you can do the following:

Augment [`run_node` ](https://github.com/bazelbuild/rules_nodejs/blob/a096a9e2d291502b83633d10cbe4fda2f414abb1/internal/providers/node_runtime_deps_info.bzl#L70)to include a dictionary of environment variables. 

`{NODE_MODULES_ROOT: _compute_node_modules_root(ctx)}`

gregmagolan:
Good call. Thanks @mrmeku .

gregmagolan:
It's all within the same root but we want to keep node from following symlinks to different node_modules trees as programs will think there are multiple versions of dependencies.

In short,

`/root/repo/node_modules` --- this is where we want resolve to
`/root/npm/node_modules` --- this is where node can end up if it follows symlinks from the above
`/root/repo/external/npm/node_modules` --- alternate legacy runfiles version (not present if --nolegacy_external_runfiles is set)

If we remove node_modules out of runfiles in the future then this is mitigated for `bazel test` although may still be a problem for `run` and `build` which don't have runfiles and just run out of the execroot.

alexeagle:
What happens if we set `/root/repo` as the root directory for the node-patches?

Or, let's explore removing these node_modules from runfiles, maybe there is a convenient way to make that conditional on the target being opted-in to using the node-patches

gregmagolan:
@alexeagle That may actually work for `run` and `test`; we'd still have the legacy external runfiles `/root/external/npm/node_modules` path but those may not be a problem as the `/root/node_modules` should not be symlinks into that tree. For the `build` case `execroot/repo/node_modules` is a symlink to `execroot/repo/external/npm/node_modules` so we may still have to prevent that escape. I'll test this out.

mrmeku:
@alexeagle If you want to remove these node_modules from run files I'd recommend using the strategy I outlined in https://github.com/bazelbuild/rules_nodejs/issues/1489

But for `npm_package_bin` Greg's fix would still be necessary since it doesn't make use of a runfiles directory. Somehow or another you'd still need to blacklist grabbing modules from `external/npm`

gregmagolan:
Tried out moving the root to `execroot/wksp` in #1509. I still think this solution is better even with the bazel specific bits. It could be made generic by adding the concept of blacklisted folders within the root which would be `npm/node_modules` and `repo/external/npm_node_modules`.

Commented here on #1509: https://github.com/bazelbuild/rules_nodejs/pull/1509#issuecomment-569349657.

soldair:
i would accept a generic array of directories as "roots" for the patcher instead of the environment variables. That way they can be passed in in a code layer that has a right to care about them. Thanks for trying in #1509

mrmeku:
Have either of you considered using a white list instead of a black list? Right now rules_nodejs is non-hermetic in cases in which a sandbox is reused since prior commands might add other source files or npm dependencies to the sandbox's file tree. Currently the node patches black list some directories we do not want entry into, but alternatively, the strategy could be to use a whitelist of the manifest of files associated with a particular target.

alexeagle:
Discussed in PR review. We still feel like Bazel is laying out a bad FS (the execroot having the "legacy" layout which was removed from runfiles tree) and the best solution is to make the `--nolegacy_external_runfiles` behavior apply always to node programs, under both the execroot and runfiles root, so that there is no duplicate node_modules tree to worry about. We could do this in the linker step, just fix the directory layout so that we never have code observe node_modules in the wrong place

gregmagolan:
Replaced by #1800

PR #1506, opened 2019-12-26
feat: enable test file crawling for jest example

mrmeku:
## PR Checklist
Before, the jest example macro used the `runTestsByPath` command line flag to specify test locations. Now, using a small patch to jest to enable symlinks, the more iodmatic method of using a test match glob pattern is used.

`testMatch: ['**/*.test.js']`

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,3 @@
 {
-    "bazel.buildifierExecutable": "${workspaceFolder}/node_modules/.bin/buildifier",
     "bazel.buildifierFixOnFormat": true
 }
--- a/examples/jest/jest.bzl
+++ b/examples/jest/jest.bzl
@@ -10,8 +10,6 @@ def jest_test(name, srcs, deps, jest_config, **kwargs):
         "--ci",
     ]
     args.extend(["--config", "$(location %s)" % jest_config])
-    for src in srcs:
-        args.extend(["--runTestsByPath", "$(location %s)" % src])
 
     _jest_test(
         name = name,
--- a/examples/jest/patches/jest-haste-map+24.9.0.patch
+++ b/examples/jest/patches/jest-haste-map+24.9.0.patch
@@ -1,0 +1,16 @@
+diff --git a/node_modules/jest-haste-map/build/crawlers/node.js b/node_modules/jest-haste-map/build/crawlers/node.js
+index 23985ae..a728a9a 100644
+--- a/node_modules/jest-haste-map/build/crawlers/node.js
++++ b/node_modules/jest-haste-map/build/crawlers/node.js
+@@ -166,7 +166,11 @@ function find(roots, extensions, ignore, callback) {
+ 
+ function findNative(roots, extensions, ignore, callback) {
+   const args = Array.from(roots);
++  args.push('(');
+   args.push('-type', 'f');
++  args.push('-o');
++  args.push('-type', 'l');
++  args.push(')');
+ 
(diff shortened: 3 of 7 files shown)


.vscode/settings.json
```diff
@@ -1,4 +1,3 @@
 {
```
alexeagle: do you mean to check in this change?

examples/jest/patches/jest-haste-map+24.9.0.patch
```diff
@@ -0,0 +1,16 @@
+diff --git a/node_modules/jest-haste-map/build/crawlers/node.js b/node_modules/jest-haste-map/build/crawlers/node.js
+index 23985ae..a728a9a 100644
+--- a/node_modules/jest-haste-map/build/crawlers/node.js
++++ b/node_modules/jest-haste-map/build/crawlers/node.js
+@@ -166,7 +166,11 @@ function find(roots, extensions, ignore, callback) {
+ 
+ function findNative(roots, extensions, ignore, callback) {
+   const args = Array.from(roots);
++  args.push('(');
```
alexeagle: is there an issue or PR open on the upstream repo so there's a chance of unwinding this patch later?
if so add a comment somewhere pointing to it, so we'll know when it's time to unpatch

examples/jest/jest.bzl
```diff
@@ -10,8 +10,6 @@ def jest_test(name, srcs, deps, jest_config, **kwargs):
         "--ci",
     ]
     args.extend(["--config", "$(location %s)" % jest_config])
```
alexeagle: I was expecting to see a glob pattern being added here or in the place jest_test is called

examples/jest/patches/jest-haste-map+24.9.0.patch
```diff
@@ -0,0 +1,16 @@
...
+   args.push('-type', 'f');
++  args.push('-o');
++  args.push('-type', 'l');
```
alexeagle: I don't quite understand where the symlinks are appearing. The node-patches should make it so the source files appear to be in the runfiles tree, not symlinks out of the runfiles. Are you sure the fix belongs in jest and not in the node-patches? Can you show where the symlink appears and where it points?

.vscode/settings.json
mrmeku: Maybe it shouldn't be part of this change, but in general I have to do this to rules_nodejs everytime I try to edit the repo in vscode. It breaks the bazel vscode extensions ability to format build files

examples/jest/jest.bzl
mrmeku: The glob pattern was already present within `jest.config.js` (that's the idiomatic way of specifying tests to jest). Do you have a different API you would prefer?

examples/jest/patches/jest-haste-map+24.9.0.patch
mrmeku: For my own reference, he's the PR I've started with Jest: https://github.com/facebook/jest/pull/9351

examples/jest/patches/jest-haste-map+24.9.0.patch
mrmeku: The symlinks appear within the runfiles tree of the testable target. `nodejs_binary` creates a runfiles directory which is filled with symlinks to the original files.

alexeagle:
In a followup could you add something to the examples/jest README so users understand they also have to copy and apply the patch if they want their globs to work?

PR #1509, opened 2019-12-27
fix(builtin): set root of node symlink patches to execroot/my_wksp

gregmagolan:
Alternative fix to #1505  for run & test.

This will prevent symlinks to escaping to `.../runfiles/npm/node_modules`  as it sets the symlink node patch root to `.../runfiles/repo`. `--nolegacy_external_runfiles` may still be needed to so that there is no `.../runfiles/repo/external/npm/node_modules` to escape to.

For the build case, this will not prevent escaping to `.../execroot/repo/external/npm/node_modules` as the root is set to `.../execroot/repo`.

### DIFF
--- a/internal/node/node_launcher.sh
+++ b/internal/node/node_launcher.sh
@@ -203,9 +203,10 @@ if [[ -n "$MODULES_MANIFEST" ]]; then
   "${node}" "${link_modules_script}" "${MODULES_MANIFEST}"
 fi
 
-# Tell the bazel_require_script that programs should not escape the execroot
-# Bazel always sets the PWD to execroot/my_wksp so we go up one directory.
-export BAZEL_PATCH_ROOT=$(dirname $PWD)
+# Tell the bazel_require_script that programs should not escape execroot/my_wksp.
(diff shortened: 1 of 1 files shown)


gregmagolan:
I still have jest escaping into `test.sh.runfiles/wm/external/npm/node_modules/graphql` in my reproduction if `--nolegacy_external_runfiles` is *not* set.

So while this solution is cleaner than #1505, it will require users to add `build --nolegacy_external_runfiles` to the `.bazelrc` for programs to work and if that is missing the error messages are not helpful.

It also, does not solve the `build` problem of escaping to `/execroot/repo/external/npm/node_modules` when there is no runfiles tree.

It also may have the unexpected negative consequences when running `.js` files in external repos such as those setup up by yarn install (`npm_bazel_foobar` for example). Actually, we may want to move the linker node_modules up one folder so node_modules resolution works from those scripts as well.

gregmagolan:
Closing in favor of specifying multiple generic root paths to the node patches. See https://github.com/bazelbuild/rules_nodejs/pull/1505#issuecomment-569736184.

PR #1515, opened 2019-12-30
feat(examples): add nodejs_binary cluster example

Toxicable:
Adds a basic example of how to use the clustering module in Nestjs.

It currently fails with 
```
Error: Cannot find module 'examples_nestjs/src/app.module'
Require stack:
- /SNIP/execroot/examples_nestjs/bazel-out/darwin-fastbuild/bin/src/server.sh.runfiles/examples_nestjs/src/main.js
```

So also doubles as a test for the linker applying correctly to `nodejs_binary`
When this passes then this should be resolved https://github.com/bazelbuild/rules_nodejs/issues/227

### DIFF
--- a/examples/nestjs/package.json
+++ b/examples/nestjs/package.json
@@ -8,6 +8,7 @@
     "@nestjs/core": "6.5.2",
     "@nestjs/platform-express": "6.5.2",
     "@types/jasmine": "^3.5.0",
+    "@types/superagent": "^4.1.4",
     "@types/supertest": "^2.0.8",
     "minimist": "1.2.0",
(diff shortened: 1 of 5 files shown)


alexeagle:
You say this currently fails, but the CI is green. What is left to do before merging?

Toxicable:
@alexeagle It fails when I run the program locally, I guess there's no test that actually starts the program up.
I was also working on a parallel PR to add some e2e test for this example.

Toxicable:
Added some testting here https://github.com/bazelbuild/rules_nodejs/pull/1528

alexeagle:
Could you rebase this please?

Issue #1471, opened 2019-12-19
NPM unmet dependency errors, after running npm_install, appear to confuse webpack


gmishkin:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: npm_install

### Is this a regression?

Not that I'm aware of

### Description

When executing the npm_install rule, it changes the node_modules directory, in a way that when you run `npm ls` it says there are unmet dependencies. This appears to confuse webpack, although I have not yet added that part to the repro. I want to understand why this rule can change your node_modules at all, even if you have a lock file, in a way that causes "unmet dependency" errors.

Once I understand whether this is a bug, or what, I'll feel better able to do a minimal repro of the webpack error I'm getting

## 🔬 Minimal Reproduction

https://github.com/gmishkin/rules_nodejs_unmet_dependencies

## 🔥 Exception or Error

<pre><code>
geoff@linux-dev:~/code/rules_nodejs_unmet_dependency$ npm ls Base64
rules_nodejs_unmet_dependency@1.0.0 /home/geoff/code/rules_nodejs_unmet_dependency
├─┬ UNMET DEPENDENCY @okta/okta-auth-js@2.11.0
│ └── UNMET DEPENDENCY Base64@0.3.0 
└─┬ UNMET DEPENDENCY mongodb-stitch@3.7.0
  └── UNMET DEPENDENCY Base64@1.1.0 

npm ERR! missing: @okta/okta-auth-js@2.11.0, required by rules_nodejs_unmet_dependency@1.0.0
npm ERR! missing: mongodb-stitch@3.7.0, required by rules_nodejs_unmet_dependency@1.0.0
npm ERR! missing: Base64@0.3.0, required by @okta/okta-auth-js@2.11.0
npm ERR! missing: Base64@1.1.0, required by mongodb-stitch@3.7.0
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Ubuntu Linux 19.10
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
1.2.1
  </code>
</pre>

**Rules_nodejs version:**

(Please check that you have matching versions between WORKSPACE file and `@bazel/*` npm packages.)

<pre>
  <code>
0.42.3
  </code>
</pre>

**Anything else relevant?**

The webpack error I get in the big repo is:
<pre><code>
[...]

PR #1516, opened 2019-12-30
feat(examples): add angular bazel architect

Toxicable:
Checking in this example https://github.com/alexeagle/prototype_bazel_architect
Which shows how you can build an angular app with `@angular-devkit/build-angular` but under bazel


Toxicable:
It dosen't actually run since something deletes random dirs in the `node_modules`, such as `node_modules/patch-package/dist/patch/apply.js`, so you get errors like:
```
Error: Cannot find module './patch/apply'
Require stack:
- /Users/fabianwiles/Documents/rules_nodejs/examples/angular_bazel_architect/node_modules/patch-package/dist/applyPatches.js
- /Users/fabianwiles/Documents/rules_nodejs/examples/angular_bazel_architect/node_modules/patch-package/dist/index.js
- /Users/fabianwiles/Documents/rules_nodejs/examples/angular_bazel_architect/node_modules/patch-package/index.js
```
So something happens during `Fetching @npm; Running yarn install on //:package.json` that causes this dir `node_modules/patch-package/dist/patch` to be deleted 

If I ran `yarn install` with my local version, it works fine, the dir remains there.

if you skip the postinstall step then the build still fails with:
```
Error: Cannot find module './util/canReportError'
Require stack:
- /private/var/tmp/_bazel_fabianwiles/221afc9f487a381a9ffe5cc91d597ad0/sandbox/darwin-sandbox/2/execroot/angular_cli_demo/node_modules/@angular-devkit/architect/node_modules/rxjs/internal/Observable.js
```

Toxicable:
This looks related to the above npm_install issues im facing https://github.com/bazelbuild/rules_nodejs/issues/1471

Toxicable:
Both build and test work fine locally but fail in CI due to the error below, but I thought that the patch  `patches/@ngtools+webpack+9.0.0-rc.0.patch` should handle that, is it not applied during `example_integration_test`?
```
FAILED: //examples:examples_angular_bazel_architect (Summary)
~~SNIP~~
ERROR: /tmp/tmp-6275IrF4677WKijh/BUILD.bazel:3:1: Action build failed (Exit 1) architect.sh failed: error executing command bazel-out/host/bin/external/npm/@angular-devkit/architect-cli/bin/architect.sh frontend:build '--outputPath=bazel-out/k8-fastbuild/bin/build' ... (remaining 2 argument(s) skipped)

frontend:build: 
ERROR in ./src/main.ts
Module build failed (from ./external/npm/node_modules/@ngtools/webpack/src/index.js):
Error: Angular Compiler was detected but it was an instance of the wrong class.
This likely means you have several @ngtools/webpack packages installed. You can check this with `npm ls @ngtools/webpack`, and then remove the extra copies.
    at Object.ngcLoader (/home/circleci/.cache/bazel/_bazel_circleci/02fc451c9ce3986cffaddce25971cb5b/execroot/build_bazel_rules_nodejs/_tmp/07c12d54970c8cddb1ddf80282a55151/_bazel_circleci/d694256e64afdb0bd419bc134525f790/sandbox/processwrapper-sandbox/1/execroot/angular_cli_demo/external/npm/node_modules/@ngtools/webpack/src/loader.js:33:15)

```

examples/angular_bazel_architect/patches/@ngtools+webpack+9.0.0-rc.0.patch
```diff
@@ -0,0 +1,50 @@
...
+     plugin.done
+diff --git a/node_modules/@ngtools/webpack/src/ngcc_processor.js b/node_modules/@ngtools/webpack/src/ngcc_processor.js
+index 77f218a..d270644 100644
+--- a/node_modules/@ngtools/webpack/src/ngcc_processor.js
++++ b/node_modules/@ngtools/webpack/src/ngcc_processor.js
+@@ -49,6 +49,16 @@ class NgccProcessor {
+             this._processedModules.add(moduleName);
+             return;
+         }
++        // If the package.json is read only we should skip calling NGCC.
```
Toxicable: can be removed https://github.com/angular/angular-cli/commit/c188a301d44bae8178d5c95bf8a8c9abe245b5cb

Toxicable:
This all runs fine locally.
Failure in CI is due to the angular cli trying to write to a cache - https://github.com/angular/angular-cli/issues/16649

gregmagolan:
Those patches should be applied in the integration test as yarn_install for that example should run the same in CI as it does locally.

Toxicable:
Current status
- `npx bazel test ...` works locally on Ubuntu 18.04 fine.
- `npx bazel build ...` works in CI for Ubuntu and Mac
- `:test` target dosen't run in bazelci since there's no way to install the chrome system deps.
- `NG_BUILD_CACHE` must be set to `false` and passed through to  `architect` since build-angular would try to cache to bazel's readonly fs.
- `npx bazel build ...` fails in CI due to some spawn and a missing or wrong path to `node`
```
Error: spawn D:\b\j3ai4z2n\execroot\build_bazel_rules_nodejs\_tmp\00ec2cae7910fb9a26b84ca52e1907d7\_bazel_b\nxx36vy6\external\build_bazel_rules_nodejs\internal\node\_node_bin\node ENOENT
```

alan-agius4:
@Toxicable the last error might be related to the fact that under Windows the spawn process is being invoked without the shell option.

See: https://github.com/angular/angular-cli/issues/16119 I gave a more detailed explanation.

Toxicable:
I've added a `tags = ["fix-windows"],` since at least the other platforms work for now

gregmagolan:
From slack: https://bazelbuild.slack.com/archives/CSACXKUBE/p1580173655005700

I'm trying it out locally and hit a node_modules runfiles issue on OSX with puppeteer since the file `node_modules/puppeteer/.local-chromium/mac-706915/chrome-mac/Chromium.app/Contents/MacOS/Chromium Framework.framework/Versions/79.0.3945.0/Chromium Framework` is not included in runfiles as it has spaces (node_modules files with spaces (\x20) or unicode characters (<\x20 && >\x7E) are deliberately filtered out of the file lists of @npm  targets because Bazel runfiles can't handle them).

Chrome errors out in the architect_test with:
```
28 01 2020 00:35:22.564:ERROR [launcher]: ChromeHeadless stdout: 
28 01 2020 00:35:22.564:ERROR [launcher]: ChromeHeadless stderr: dlopen /private/var/tmp/_bazel_greg/e8c34d42b565fe5ec630462de6b01ea3/execroot/angular_cli_demo/bazel-out/darwin-fastbuild/bin/test.sh.runfiles/angular_cli_demo/node_modules/puppeteer/.local-chromium/mac-706915/chrome-mac/Chromium.app/Contents/MacOS/../Frameworks/Chromium Framework.framework/Versions/79.0.3945.0/Chromium Framework: dlopen(/private/var/tmp/_bazel_greg/e8c34d42b565fe5ec630462de6b01ea3/execroot/angular_cli_demo/bazel-out/darwin-fastbuild/bin/test.sh.runfiles/angular_cli_demo/node_modules/puppeteer/.local-chromium/mac-706915/chrome-mac/Chromium.app/Contents/MacOS/../Frameworks/Chromium Framework.framework/Versions/79.0.3945.0/Chromium Framework, 261): image not found
```
Spaces in runfiles almost got fixed a few months ago with https://github.com/bazelbuild/bazel/pull/9351 but it didn't pass internal review.

Could push to get that in for a future Bazel version... but in the meantime I'll think on a work-around since the Angular team would also like puppateer managing their chrome version for karma tests instead of rules_webtesting.

Its also very flakey on buildkite macos CI due to some node_modules filesystem flake there looks like. I did see it pass once on buildkite macos but it is running the karma tests there right now as only `bazel build ...` is being run there so it won't hit the puppeteer issue on buildkite

```
example_integration_test(
    name = "examples_angular_bazel_architect",
    bazel_commands = [
        # should be test but since `ng test` uses a browser we exclude it in bazelci
        # meaning that this fails since there's no tests to run
        # so once there is this should be flipped to test
        "build ...",
    ],
    npm_packages = {},
    # windows fails CI with an error to do with a process fork not being able ot find node
    tags = ["fix-windows"],
)
```

gregmagolan:
I haven't seen the filesystem flake locally on OSX at all. This seems to happen only on buildkite. Maybe it is something to do with something they are doing with macos virtualization?

But this PR is still blocked on the puppeteer usage for OSX as it won't work there yet until Bazel supports spaces in runfiles paths. Alternately, we don't include node_modules in runfiles but that is not ready to go yet.

gregmagolan:
I have a path to land this:

1. We need to remove puppeteer from the example as it will never work on OSX right now as Bazel can't handle spaces in runfiles. Karma should just use the local chrome at that point.

2. Turn off symlinked_node_modules & managed_directories fixes the flake. I'll follow up with buildkite as to what they are doing on the macos machines. Still thinking it has to do with their virtualization layer.

3. Run `bazel test ...` for the integration test. On CircleCI the local browser should work. Ubuntu on buildkite doesn't have the shared libs so you'll need to add a dummy_test target (see `examples/angular/dummy_test.sh`) or Bazel will complain that there are no test targets. On macos it should pass as the local chrome should work there. Windows is still broken and will be filtered out by `fix-windows`

.circleci/config.yml
```diff
@@ -12,6 +12,8 @@
 var_1: &default_docker_image circleci/node:10.16
 var_3: &cache_key node-0.16-{{ checksum "yarn.lock" }}-v2
 
+var_2: &browsers_docker_image circleci/node:10.16-browsers
```
gregmagolan: nit: belongs between var_1 & var3 above

examples/angular_bazel_architect/.bazelrc
```diff
@@ -0,0 +1,86 @@
+# Common Bazel settings for JavaScript/NodeJS workspaces
```
gregmagolan: To avoid duplication in the repo, change this to 
```
import %workspace%/../../common.bazelrc
<settings specific to this example go HERE>
```

gregmagolan: Example: https://github.com/bazelbuild/rules_nodejs/blob/master/examples/angular/.bazelrc

examples/angular_bazel_architect/.editorconfig
```diff
@@ -0,0 +1,13 @@
+# Editor configuration, see https://editorconfig.org
```
gregmagolan: Don't have this in the other examples so remove it from here? Could use one in the root looks like.

examples/angular_bazel_architect/README.md
```diff
@@ -0,0 +1,5 @@
+# Frontend
```
gregmagolan: Add a brief description of the example here and mention that the patch is still needed at this point?

gregmagolan:
@googlebot I consent

examples/angular_bazel_architect/WORKSPACE
```diff
@@ -0,0 +1,39 @@
...
+    # How this workspace would be referenced with absolute labels from another workspace
+    name = "angular_cli_demo",
+)
+
+# Install the nodejs "bootstrap" package
+# This provides the basic tools for running and packaging nodejs programs in Bazel
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "build_bazel_rules_nodejs",
```
gregmagolan: Nice! This is a great example.
Last few changes and I can merge this:
1) workspace name above should match the folder name `angular_bazel_architect`
2) rebase to master head & that should wipe out the root yarn.lock file diff as it was out-of-date
3) update the build_bazel_rules_nodejs http_archive here to 1.2.1 release

Toxicable:
I didn't properly update the sha256 on build_bazel_rules_nodejs - fixed now

PR #1524, opened 2020-01-03
Expand location features

gregmagolan:
feat(builtin): add support for $(rootpath) $(execpath) & predefined & custom variables to templated_args

* this applies to nodejs_binary, nodejs_test & jasmine_node_test

* adds support for $(rootpath) $(execpath) to params_file args

* also fixes ambiguity of `$(rlocation $(location ...))` expansion as it is not clear that `$(rlocation ...)` is not meant to be expanded but rather passed down to the shell script for execution. This is now more clear with `$$(rlocation $(rootpath ...))` which follows the Bazel "Make" variable expansion convention. For example, `templated_args = ["--node_options=--require=$$(rlocation $(rootpath :bootstrap.js))",],`.

* templated_args is now...

Subject to 'Make variable' substitution. See https://docs.bazel.build/versions/master/be/make-variables.html.

1. Predefined source/output path substitions is applied first:

Expands all `$(execpath ...)`, `$(rootpath ...)` and legacy `$(location ...)` templates in the
given string by replacing with the expanded path. Expansion only works for labels that point to direct dependencies
of this rule or that are explicitly listed in the optional argument targets.

See https://docs.bazel.build/versions/master/be/make-variables.html#predefined_label_variables.

Use `$(rootpath)` and `$(rootpaths)` to expand labels to the runfiles path that a built binary can use
to find its dependencies. This path is of the format:
- `./file`
- `path/to/file`
- `../external_repo/path/to/file`

Use `$(execpath)` and `$(execpaths)` to expand labels to the execroot (where Bazel runs build actions).
This is of the format:
- `./file`
- `path/to/file`
- `external/external_repo/path/to/file`
- `<bin_dir>/path/to/file`
- `<bin_dir>/external/external_repo/path/to/file`

The legacy `$(location)` and `$(locations)` expansion is DEPRECATED as it returns the runfiles manifest path of the
format `repo/path/to/file` which behaves differently than the built-in `$(location)` expansion in args of *_binary
and *_test rules which returns the rootpath.
See https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes-binaries.

The legacy `$(location)` and `$(locations)` expansion also differs from how the builtin `ctx.expand_location()` expansions
of `$(location)` and `$(locations)` behave as that function returns either the execpath or rootpath depending on the context.
See https://docs.bazel.build/versions/master/be/make-variables.html#predefined_label_variables.

The behavior of `$(location)` and `$(locations)` expansion may change in the future with support either being removed
entirely or the expansion changed to return the same path as `ctx.expand_location()` returns for these.

The recommended approach is to now use `$(rootpath)` where you previously used $(location).

To get from a `$(rootpath)` to the absolute path that `$$(rlocation $(location))` returned you can either use
`$$(rlocation $(rootpath))` if you are in the `templated_args` of a `nodejs_binary` or `nodejs_test`:

BUILD.bazel:
```
nodejs_test(
    name = "my_test",
    data = [":bootstrap.js"],
    templated_args = ["--node_options=--require=$$(rlocation $(rootpath :bootstrap.js))"],
)
```

or if you're in the context of a .js script you can pass the $(rootpath) as an argument to the script
and use the javascript runfiles helper to resolve to the absolute path:

BUILD.bazel:
```
nodejs_test(
    name = "my_test",
    data = [":some_file"],
    entry_point = ":my_test.js",
    templated_args = ["$(rootpath :some_file)"],
)
```

my_test.js
```
const runfiles = require(process.env['BAZEL_NODE_RUNFILES_HELPER']);
const args = process.argv.slice(2);
const some_file = runfiles.resolveWorkspaceRelative(args[0]);
```

NB: Bazel will error if it sees the single dollar sign $(rlocation path) in `templated_args` as it will try to
example `$(rlocation)` since we now expand predefined & custom "make" variables such as `$(COMPILATION_MODE)`,
`$(BINDIR)` & `$(TARGET_CPU)` using `ctx.expand_make_variables`. See https://docs.bazel.build/versions/master/be/make-variables.html.

To prevent expansion of `$(rlocation)` write it as `$$(rlocation)`. Bazel understands `$$` to be
the string literal `$` and the expansion results in `$(rlocation)` being passed as an arg instead
of being expanded. `$(rlocation)` is then evaluated by the bash node launcher script and it calls
the `rlocation` function in the runfiles.bash helper. For example, the templated arg
`$$(rlocation $(rootpath //:some_file))` is expanded by Bazel to `$(rlocation ./some_file)` which
is then converted in bash to the absolute path of `//:some_file` in runfiles by the runfiles.bash helper
before being passed as an argument to the program

NB: nodejs_binary and nodejs_test will preserve the legacy behavior of `$(rlocation)` so users don't
need to update to `$$(rlocation)`. This may be changed in the future.

2. Predefined variables & Custom variables are expanded second:

Predefined "Make" variables such as $(COMPILATION_MODE) and $(TARGET_CPU) are expanded.
See https://docs.bazel.build/versions/master/be/make-variables.html#predefined_variables.

Custom variables are also expanded including variables set through the Bazel CLI with --define=SOME_VAR=SOME_VALUE.
See https://docs.bazel.build/versions/master/be/make-variables.html#custom_variables.

Predefined genrule variables are not supported in this context.


alexeagle:
Hmm I guess any runfiles library has this problem, it needs to choose a path format to understand in its API. Do any of them adapt rootpaths to be manifest lookups? Seems like a windows-related issue since others rely on runfiles symlink forest - maybe @meteorcloudy has an idea?

gregmagolan:
I was thinking that a PR to Bazel may be in order to add `manifestpath` and `manifestpaths` as siblings of `execpath` and `rootpath`. Seems like this is missing and it should be there.

The trouble with converting from rootpath to manifestpath at runtime in the runfiles library is rootpaths look a lot like relative imports:

```
./file
path/to/file
../external_repo/path/to/file
```

how does the runfiles library know that `./file` is meant to be `repo/file.js` and not relative `file.js`? or `path/to/file` => `repo/path/to/file`? Seems like its much less likely to be ambiguous with the manifestpath single format: `repo/path/to/file`.

gregmagolan:
Looking at this in some more detail:

`$(manifestpath` is understood by templated_args but can't be evaluated by args as bazel expands args internally. This is not ideal to have special construct in templated_args --- +1 argument to add it to Bazel.

To bootstrap with node_options require you need the absolute path so you use the `$$rlocation` bash runfiles helper. For example: `templated_args = ["--node_options=--require=$$(rlocation $(manifestpath :bootstrap.js))"],`. This means that if `$(rlocation)` was customized in the launcher to handle `rootpath` we could make `templated_args = ["--node_options=--require=$$(rlocation $(rootpath :bootstrap.js))"],` work. +1 argument against `manifestpath`.

However, if the linker is enabled it would be sufficient to write `templated_args = ["--node_options=--require=$(manifestpath :bootstrap.js)"],` which is simpler but that simplicity would necessitate a `$(manifestpath)` in order for node_module resolution to resolve at `/wksp/node_modules/wksp/path/to/package/bootstrap.js`. +1 argument for `manifestpath`.

Outside of the bootstrap case, the resolved path is passed to the node program itself (such as jasmine_runner.js`). For example, `templated_args.append("$(manifestpath :%s_devmode_srcs.MF)" % name)` in `jasmine_node_test.bzl`. This means that if it a user entry_point or a custom rule entry_point then it can use the js runfiles helpers `const runfiles = require(process.env['BAZEL_NODE_RUNFILES_HELPER']);` to resolve with `runfiles.resolve`.

We can avoid `manifestpath` in this case by using `runfiles.resolveWorkspaceRelative` with `templated_args.append("$(rootpath :%s_devmode_srcs.MF)" % name)`. This assumes that the entry_point is in our control which is not always the case if we are using an npm tool from node_modules directory.

We can, however, avoid the js runfiles helper altogether with the bash `$rlocation`: `templated_args.append("$$(rlocation $(manifestpath :%s_devmode_srcs.MF))" % name)`. As mentioned above, this function could be made to understand rootpath: `templated_args.append("$$(rlocation $(rootpath :%s_devmode_srcs.MF))" % name)`.

NB: if not on Windows (or with Windows symlinked files enabled), an argument such as `$(rootpath :file)` returns a path that is resolvable from the cwd but we need to support Windows MANIFEST file case so this is not a general solution but does work for non-Windows users.

----

The general solution of either
* `$$(rlocation $(manifest :file))`
* or `$$(rlocation $(rootpath :file))` if rlocation can handle rootpath
*  or $(manifest :file)` with linker node_module resolution
however, circles back to being able to do something in templated_args that does not work with args:

args does not understand `$(manifestpath)` or `$(rlocation)` and passing through `$$(rlocation` does not work as its not evaluated by bash. This would be +1 to add `$(manifestpath` to Bazel itself which would make it work with templated_args and with args and with the linker it would also work without `$$rlocation` even on Windows.

gregmagolan:
Alright. I've managed to vendor in and patch the runfiles.bash helper to handle `$(rootpath)` so we can now use `$$(rlocation $(rootpath :file))` instead of `$$(rlocation $(manifestpath :file))`.

`$(manifestpath)` removed and shelved for now.

alexeagle:
Why do you have to vendor it? Everyone has @bazel_tools

internal/common/expand_into_runfiles.bzl
```diff
@@ -15,47 +15,91 @@
...
+    """Expands all $(execpath ...), $(rootpath ...) and legacy $(location ...) templates in the
+    given string by replacing with the expanded path. Expansion only works for labels that point to direct dependencies
     of this rule or that are explicitly listed in the optional argument targets.
 
-    Path is returned in runfiles manifest path format such as `repo/path/to/file`. This differs from how $(location)
-    and $(locations) expansion behaves in expansion the `args` attribute of a *_binary or *_test which returns
-    the runfiles short path of the format `./path/to/file` for user repo and `../external_repo/path/to/file` for external
-    repositories. We may change this behavior in the future with $(mlocation) and $(mlocations) used to expand
-    to the runfiles manifest path.
+    See https://docs.bazel.build/versions/master/be/make-variables.html#predefined_label_variables.
```
alexeagle: this is a nice and thorough docstring but does it appear anywhere after stardoc runs? will users ever see it? Is it a copy of the one on params_file? if so maybe just point to there so we don't end up with the copies diverging

internal/common/expand_into_runfiles.bzl
```diff
@@ -15,47 +15,91 @@
...
     See https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes-binaries.
+    This also differs from how the builtin ctx.expand_location() expansions of $(location) and ($locations) behave
+    as that function returns either the execpath or rootpath depending on the context.
+    See https://docs.bazel.build/versions/master/be/make-variables.html#predefined_label_variables.
+
+    The behavior of $(location) and $(locations) expansion may change in the future with support either being removed
+    entirely or the expansion changed to return the same path as ctx.expand_location() returns for these.
+
+    The recommended approach is to now use $(rootpath) where you previously used $(location). To get from a $(rootpath)
+    to the absolute path that $$(rlocation $(location)) returned you can either use $$(rlocation $(rootpath)) if you
```
alexeagle: it would be nice to explain this one level deeper - what are the two dollar signs doing, and does one expansion happen before the action and the other happens inside it? Do we describe the rlocation helper itself somewhere?

internal/common/expand_into_runfiles.bzl
```diff
@@ -67,28 +111,32 @@ def expand_location_into_runfiles(ctx, input, targets = []):
...
     path = ""
     length = len(input)
     last = 0
     for i in range(length):
-        if (input[i:i + 12] == "$(mlocation ") or (input[i:i + 13] == "$(mlocations "):
+        # Support legacy $(location) and $(locations) expansions which return the runfiles manifest path
+        # in the format `repo/path/to/file`. This expansion is DEPRECATED. See docstring above.
+        # TODO: Change location to behave the same as the built-in $(location) expansion for args of *_binary
+        #       and *_test rules. This would be a BREAKING CHANGE.
+        if input[i:].startswith("$(location ") or input[i:].startswith("$(locations "):
```
alexeagle: opt: seems like there's enough conditional logic and string manipulation in this code that we ought to have a starlark unit test for it

internal/node/node.bzl
```diff
@@ -283,8 +291,9 @@ The set of default  environment variables is:
 
 - `VERBOSE_LOGS`: use by some rules & tools to turn on debug output in their logs
 - `NODE_DEBUG`: used by node.js itself to print more logs
+- `RUNFILES_LIB_DEBUG`: print diagnostic message from Bazel runfiles.bash helper
```
alexeagle: why not use `VERBOSE_LOGS` for that?

packages/node-patches/BUILD.bazel
```diff
@@ -55,14 +55,12 @@ tsc(
...
     name = "unit_test",
-    args = ["$(location %s)" % s for s in test_js],
     data = js + test_js + [
         "@npm_node_patches//:node_modules",
     ],
     entry_point = "@npm_node_patches//:node_modules/mocha/bin/mocha",
-    tags = [
-        "fix-windows",
-    ],
+    tags = ["fix-windows"],
```
alexeagle: shouldn't this work on windows now?

internal/common/expand_into_runfiles.bzl
gregmagolan: Yea. I was thinking about that but wasn't top of list. I'll add it today.

internal/node/node.bzl
gregmagolan: `RUNFILES_LIB_DEBUG` is what is in the @bazel_tools runfiles.bash helper

packages/node-patches/BUILD.bazel
gregmagolan: I think it ran into some other error. I'll enable it and comment on the error.

internal/common/expand_into_runfiles.bzl
gregmagolan: 👍 I'll add more details to the comment.

Since this adds expansion of predefined & custom "make" variables (https://docs.bazel.build/versions/master/be/make-variables.html) such as $(COMPILATION_MODE) $(BINDIR) & $(TARGET_CPU) using `ctx.expand_make_variables`:

```
    # First expand predefined source/output path variables:
    # $(execpath), $(rootpath) & legacy $(location)
    expanded_args = [expand_location_into_runfiles(ctx, a, ctx.attr.data) for a in ctx.attr.templated_args]

    # Next expand predefined variables & custom variables
    expanded_args = [ctx.expand_make_variables("templated_args", e, {}) for e in expanded_args]
```

Bazel will error out if it sees the single dollar sign `$(rlocation path)`. For expand_make_variables not to expand $(rlocation) which is meant be be left in the arg and expanded in the launcher.sh we need to now use `$$(rlocation)` which expands to `$(rlocation)` so that `$$(rlocation $(rootpath //:some_file))` expands to `$(rlocation ./some_file)` and bash runs the `rlocation` runfiles.bash helper function to convert`./some_file` to its absolute path.

internal/common/expand_into_runfiles.bzl
gregmagolan: Its is slightly different from the param_files one and a subset of the nodejs_binary templated_args docstring. Not super thrilled with the duplication so I could point to either of those since params_file and nodejs_binary templated_args both use this helper function. OTOH, it is strange to document the helper function at the point it is used at and this docstring was intended more for rules_nodejs maintainers.

internal/node/node.bzl
gregmagolan: However, since we now vendor it in we could change that as well but I wanted to keep the changes to that file to minimum. I'll go back and comment around the rules_nodejs custom parts which add support for $(rootpath)

internal/common/expand_into_runfiles.bzl
gregmagolan: Thinking on this a bit more, I'll make the helper function docstring briefer and point to the Bazel docs & to the nodejs_binary & params_file docstrings for more info.

internal/common/expand_into_runfiles.bzl
gregmagolan: Added some more details

internal/common/expand_into_runfiles.bzl
gregmagolan: Done.

internal/common/expand_into_runfiles.bzl
gregmagolan: Done.

packages/node-patches/BUILD.bazel
gregmagolan: Yep. There are still failures on Windows in the node-patches test once enabled:

/cc @soldair 

https://buildkite.com/bazel/rules-nodejs-nodejs/builds/5480#bda746b5-f98d-4fc0-a4f5-9bb04c0a254e:
```
==================== Test output for //packages/node-patches:unit_test:
--
  | ["D:\\b\\ozpmhnle\\external\\nodejs_windows_amd64\\bin\\nodejs\\node.exe",["--preserve-symlinks","--preserve-symlinks-main"],["D:\\b\\ozpmhnle\\external\\nodejs_windows_amd64\\bin\\nodejs\\node.exe","npm_node_patches/node_modules/mocha/bin/mocha","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/escape.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/lstat.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/opendir.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/readdir.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/readlink.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/realpath.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/subprocess/shell-script.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/subprocess/spawn-script.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/subprocess/test.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/subprocess/worker-threads-script.js"],"D:\\b\\ozpmhnle\\external\\nodejs_windows_amd64\\bin\\nodejs;C:\\tools\\msys64\\usr\\bin;C:\\tools\\msys64\\usr\\bin;C:\\tools\\msys64\\usr\\bin;C:\\Windows;C:\\Windows\\System32;C:\\Windows\\System32\\WindowsPowerShell\\v1.0"]
  | ["D:\\b\\ozpmhnle\\external\\nodejs_windows_amd64\\bin\\nodejs\\node.exe",["--preserve-symlinks","--preserve-symlinks-main"],["D:\\b\\ozpmhnle\\external\\nodejs_windows_amd64\\bin\\nodejs\\node.exe","npm_node_patches/node_modules/mocha/bin/mocha","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/escape.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/lstat.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/opendir.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/readdir.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/readlink.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/realpath.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/subprocess/shell-script.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/subprocess/spawn-script.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/subprocess/test.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/subprocess/worker-threads-script.js"],"D:\\b\\ozpmhnle\\external\\nodejs_windows_amd64\\bin\\nodejs;C:\\tools\\msys64\\usr\\bin;C:\\tools\\msys64\\usr\\bin;C:\\tools\\msys64\\usr\\bin;C:\\Windows;C:\\Windows\\System32;C:\\Windows\\System32\\WindowsPowerShell\\v1.0"]
  |  
  |  
  | escape function
  | 1) isOutPath is correct
  | 2) isEscape is correct
  | √ isEscape handles relative paths
  |  
  | testing lstat
  | {"mainThread":["D:\\b\\ozpmhnle\\external\\nodejs_windows_amd64\\bin\\nodejs\\node.exe",["--preserve-symlinks","--preserve-symlinks-main"],["D:\\b\\ozpmhnle\\external\\nodejs_windows_amd64\\bin\\nodejs\\node.exe","npm_node_patches/node_modules/mocha/bin/mocha","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/escape.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/lstat.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/opendir.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/readdir.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/readlink.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/fs/realpath.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/subprocess/shell-script.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/subprocess/spawn-script.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/subprocess/test.js","D:/b/ozpmhnle/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/node-patches/test/subprocess/worker-threads-script.js"],"D:\\b\\ozpmhnle\\external\\nodejs_windows_amd64\\bin\\nodejs;C:\\tools\\msys64\\usr\\bin;C:\\tools\\msys64\\usr\\bin;C:\\tools\\msys64\\usr\\bin;C:\\Windows;C:\\Windows\\System32;C:\\Windows\\System32\\WindowsPowerShell\\v1.0"],"worker":["D:\\b\\ozpmhnle\\external\\nodejs_windows_amd64\\bin\\nodejs\\node.exe",["--preserve-symlinks","--preserve-symlinks-main"],["D:\\b\\ozpmhnle\\external\\nodejs_windows_amd64\\bin\\nodejs\\node.exe","D:\\b\\ozpmhnle\\execroot\\build_bazel_rules_nodejs\\bazel-out\\x64_windows-fastbuild\\bin\\packages\\node-patches\\test\\subprocess\\worker-threads-script.js"],"D:\\b\\ozpmhnle\\external\\nodejs_windows_amd64\\bin\\nodejs;C:\\tools\\msys64\\usr\\bin;C:\\tools\\msys64\\usr\\bin;C:\\tools\\msys64\\usr\\bin;C:\\Windows;C:\\Windows\\System32;C:\\Windows\\System32\\WindowsPowerShell\\v1.0"]}
  | √ can lstat symlink in root
  | √ lstat of symlink out of root is file.
  | √ not in root, symlinks do what they would have before.
  |  
  | testing opendir
  | 1 opendir: found link?  D:\temp\tmp-10368Zd1wSD3V59fA\a\apples false
  | 2 opendir: found link?  D:\temp\tmp-10368Zd1wSD3V59fA\a\link true
  | 2 opendir: escapes? [target] D:\temp\tmp-10368Zd1wSD3V59fA\b\file [link] D:\temp\tmp-10368Zd1wSD3V59fA\a\link false D:\temp\tmp-10368Zd1wSD3V59fA
  | 3) can opendir dirent in root
  | 1 opendir: found link?  D:\temp\tmp-10368DO9WPRyRX4cQ\a\apples false
  | 2 opendir: found link?  D:\temp\tmp-10368DO9WPRyRX4cQ\a\link true
  | 2 opendir: escapes? [target] D:\temp\tmp-10368DO9WPRyRX4cQ\b\file [link] D:\temp\tmp-10368DO9WPRyRX4cQ\a\link true D:\temp\tmp-10368DO9WPRyRX4cQ\a
  | 2 opendir: patching dirent to look like it's target D:\temp\tmp-10368DO9WPRyRX4cQ\b\file
  | Dirent { name: 'apples', [Symbol(type)]: 1 } Dirent {
  | name: 'link',
  | isFile: [Function],
  | isDirectory: [Function],
  | isBlockDevice: [Function],
  | isCharacterDevice: [Function],
  | isSymbolicLink: [Function],
  | isFIFO: [Function],
  | isSocket: [Function],
  | [Symbol(type)]: 3
  | }
  | 4) can opendir dirent link out of root
  | 1 opendir: found link?  D:\temp\tmp-10368T0ZXzrRSlk0r\a\apples false
  | 2 opendir: found link?  D:\temp\tmp-10368T0ZXzrRSlk0r\a\link true
  | 2 opendir: escapes? [target] D:\temp\tmp-10368T0ZXzrRSlk0r\b\file [link] D:\temp\tmp-10368T0ZXzrRSlk0r\a\link false D:\temp\tmp-10368T0ZXzrRSlk0r
  | √ can async iterate opendir
  | 1 opendir: found link?  D:\temp\tmp-10368Qoz16IUl0FJ1\a\apples false
  | 2 opendir: found link?  D:\temp\tmp-10368Qoz16IUl0FJ1\a\link true
  | 2 opendir: escapes? [target] D:\temp\tmp-10368Qoz16IUl0FJ1\b\file [link] D:\temp\tmp-10368Qoz16IUl0FJ1\a\link true D:\temp\tmp-10368Qoz16IUl0FJ1\a
  | 2 opendir: patching dirent to look like it's target D:\temp\tmp-10368Qoz16IUl0FJ1\b\file
  | √ can async iterate opendir link out of root
  |  
  | testing readdir
  | √ can readdir dirent in root
  | √ can readdir link dirents as files out of root
  |  
  | testing readlink
  | √ can resolve symlink in root
  | √ doesn't resolve as symlink outside of root
  |  
  | testing realpath
  | √ can resolve symlink in root
  | √ doesn't resolve as symlink outside of root
  |  
  | spawning child processes
  | 'NP_SUBPROCESS_BIN_DIR' is not recognized as an internal or external command,
  | operable program or batch file.
  | 5) get patched if run as shell script.
  | 'NP_SUBPROCESS_BIN_DIR' is not recognized as an internal or external command,
  | operable program or batch file.
  | 6) overwrites spawn related variables correctly.
  | 'NP_SUBPROCESS_BIN_DIR' is not recognized as an internal or external command,
  | operable program or batch file.
  | 7) can spawn node from the shell
  | 'NP_SUBPROCESS_BIN_DIR' is not recognized as an internal or external command,
  | operable program or batch file.
  | 8) can spawn node from spawn
  |  
  |  
  | 12 passing (167ms)
  | 8 failing
  |  
  | 1) escape function
  | isOutPath is correct:
  |  
  | AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
  |  
  | assert.ok(!isOutPath('/a/b'))
  |  
  | + expected - actual
  |  
  | -false
  | +true
  |  
  | at Context.<anonymous> (D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\test\fs\escape.js:27:16)
  | at processImmediate (internal/timers.js:439:21)
  |  
  | 2) escape function
  | isEscape is correct:
  |  
  | AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:
  |  
  | assert.ok(isEscape('/a/c/boop', '/a/b/l'))
  |  
  | + expected - actual
  |  
  | -false
  | +true
  |  
  | at Context.<anonymous> (D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\test\fs\escape.js:33:16)
  | at processImmediate (internal/timers.js:439:21)
  |  
  | 3) testing opendir
  | can opendir dirent in root:
  | Error: EPERM: operation not permitted, lstat 'D:\temp\tmp-10368Zd1wSD3V59fA\a'
  | at Object.lstatSync (fs.js:906:3)
  | at _rmdirRecursiveSync (D:\b\ozpmhnle\external\npm_node_patches\node_modules\tmp\lib\tmp.js:300:19)
  | at Object._cleanupCallback [as removeCallback] (D:\b\ozpmhnle\external\npm_node_patches\node_modules\tmp\lib\tmp.js:446:7)
  | at Object.withFixtures (D:\b\ozpmhnle\external\npm_node_patches\node_modules\inline-fixtures\build\src\fixtures.js:72:17)
  | at async Context.<anonymous> (D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\test\fs\opendir.js:35:9)
  |  
  | 4) testing opendir
  | can opendir dirent link out of root:
  | Error: EPERM: operation not permitted, lstat 'D:\temp\tmp-10368DO9WPRyRX4cQ\a'
  | at Object.lstatSync (fs.js:906:3)
  | at _rmdirRecursiveSync (D:\b\ozpmhnle\external\npm_node_patches\node_modules\tmp\lib\tmp.js:300:19)
  | at Object._cleanupCallback [as removeCallback] (D:\b\ozpmhnle\external\npm_node_patches\node_modules\tmp\lib\tmp.js:446:7)
  | at Object.withFixtures (D:\b\ozpmhnle\external\npm_node_patches\node_modules\inline-fixtures\build\src\fixtures.js:72:17)
  | at async Context.<anonymous> (D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\test\fs\opendir.js:61:9)
  |  
  | 5) spawning child processes
  | get patched if run as shell script.:
  | Error: Command failed: NP_SUBPROCESS_BIN_DIR=D:\temp\node-patches-test-tmp node -r D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\register.js -e 'console.log(JSON.stringify([process.execPath,process.execArgv,process.argv,process.env.PATH]))'
  | 'NP_SUBPROCESS_BIN_DIR' is not recognized as an internal or external command,
  | operable program or batch file.
  |  
  | at checkExecSyncError (child_process.js:621:11)
  | at Object.execSync (child_process.js:657:15)
  | at Context.<anonymous> (D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\test\subprocess\test.js:34:24)
  | at processImmediate (internal/timers.js:439:21)
  |  
  | 6) spawning child processes
  | overwrites spawn related variables correctly.:
  | Error: Command failed: NP_SUBPROCESS_BIN_DIR=D:\temp\node-patches-test-tmp node -r D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\register.js D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\test\subprocess\worker-threads-script.js
  | 'NP_SUBPROCESS_BIN_DIR' is not recognized as an internal or external command,
  | operable program or batch file.
  |  
  | at checkExecSyncError (child_process.js:621:11)
  | at Object.execSync (child_process.js:657:15)
  | at Context.<anonymous> (D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\test\subprocess\test.js:38:24)
  | at processImmediate (internal/timers.js:439:21)
  |  
  | 7) spawning child processes
  | can spawn node from the shell:
  | Error: Command failed: NP_SUBPROCESS_BIN_DIR=D:\temp\node-patches-test-tmp node -r D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\register.js D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\test\subprocess\shell-script.js
  | 'NP_SUBPROCESS_BIN_DIR' is not recognized as an internal or external command,
  | operable program or batch file.
  |  
  | at checkExecSyncError (child_process.js:621:11)
  | at Object.execSync (child_process.js:657:15)
  | at Context.<anonymous> (D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\test\subprocess\test.js:44:24)
  | at processImmediate (internal/timers.js:439:21)
  |  
  | 8) spawning child processes
  | can spawn node from spawn:
  | Error: Command failed: NP_SUBPROCESS_BIN_DIR=D:\temp\node-patches-test-tmp node -r D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\register.js D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\test\subprocess\spawn-script.js
  | 'NP_SUBPROCESS_BIN_DIR' is not recognized as an internal or external command,
  | operable program or batch file.
  |  
  | at checkExecSyncError (child_process.js:621:11)
  | at Object.execSync (child_process.js:657:15)
  | at Context.<anonymous> (D:\b\ozpmhnle\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\packages\node-patches\test\subprocess\test.js:52:24)
  | at processImmediate (internal/timers.js:439:21)
  |  
  |  
  |  
  | ================================================================================


```

gregmagolan:
> Why do you have to vendor it? Everyone has @bazel_tools

I have custom patches to the runfiles.bash helper to support $(rootpath) expansion so we don't need to add a $(manifestpath). I marked the custom code in the file with --begin-- and --end-- markers.

Toxicable:
> have custom patches to the runfiles.bash helper to support $(rootpath) ...

Do you think this is going to be permanent? or are we able to upstream these changes?

gregmagolan:
> > have custom patches to the runfiles.bash helper to support $(rootpath) ...
> 
> Do you think this is going to be permanent? or are we able to upstream these changes?

Here to 1.0 at least and we'll see how it goes. I think Bazel should either adopt a `$(manifest)` expansion or have the runfiles.bash `rlocation` function handle rootpaths. Have to figure out who the right person on the Bazel team is to talk to about it.

alexeagle:
@laszlocsomor do you have any suggestion for how to resolve paths such that they can be used to lookup files in runfiles manifest?
expand_location and genrule#args understand rootpath and execpath, but neither of these returns a manifest location which is what rlocation in the runfiles helper expects.
In some other language there must be something adapting for `./` and `../` paths? Having our own vendored and locally modified runfiles bash helper feels wrong so I wonder if you know a way to avoid that.

laszlocsomor:
> @laszlocsomor do you have any suggestion for how to resolve paths such that they can be used to lookup files in runfiles manifest?

No. `$(location)` is inconsistent: if the file is in the main repo, the result is path relative to the main repo root, and if the file is an ext.repo, the result starts with `external/<reponame>/`. The `rlocation` function could try and fix these paths but it must know the main workspace name for it, which is unavailable for `sh_binary`.

This needs to be fixed in Bazel. I reckon we can't fix `$(location)` because that'd break bug-for-bug compatibility, so we need to add a fixed version of it that returns consistent paths for main-repo and ext-repo labels.

> ctx.expand_location and genrule#args understand rootpath and execpath, but neither of these returns a manifest location which is what rlocation in the runfiles helper expects.

Yes, I think they are buggy.

> In some other language there must be something adapting for `./` and `../` paths?

Not sure I understand, can you show an example?

PR #1528, opened 2020-01-04
Examples nestjs test

Toxicable:
Adds a test which boots up the Nestjs application and ensures that and api endpoint can be hit

### DIFF
--- a/examples/nestjs/src/main.spec.ts
+++ b/examples/nestjs/src/main.spec.ts
@@ -1,0 +1,21 @@
+import {INestApplication} from '@nestjs/common';
+import * as request from 'supertest';
+
+import {bootstrap} from './main';
+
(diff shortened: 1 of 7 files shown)


Toxicable:
Should be good for review now

examples/nestjs/src/main.spec.ts
```diff
@@ -0,0 +1,21 @@
+import {INestApplication} from '@nestjs/common';
+import * as request from 'supertest';
+
+import {bootstrap} from './main';
+
+describe('App', () => {
+  let server: INestApplication;
+
+  beforeAll(async () => {
+    server = await bootstrap(3000);
```
alexeagle: ideally we should find an open port - this test could flake if some other test or app has this port open

PR #1542, opened 2020-01-11
@bazel/runfiles module

gregmagolan:
Adds a `@build_bazel_rules_nodejs//:runfiles` target that when added as a dep to nodejs_binary or nodejs_test makes the `@bazel/runfiles` module available.

For example (from /examples/closure):
BUILD.bazel
```
nodejs_test(
    name = "test",
    data = [
        "bundle.js",
        "@build_bazel_rules_nodejs//:runfiles",
    ],
    entry_point = "test.js",
)
```

test.js
```
const runfiles = require('@bazel/runfiles');
const closureOutput = runfiles.resolve('examples_closure/bundle.js');
```

### DIFF
--- a/internal/node/test/npm_package_bin.spec.js
+++ b/internal/node/test/npm_package_bin.spec.js
@@ -1,6 +1,6 @@
 const fs = require('fs');
 const path = require('path');
-const runfiles = require(process.env['BAZEL_NODE_RUNFILES_HELPER']);
+const runfiles = require('@bazel/runfiles');
 
 const min_js = path.join(runfiles.resolvePackageRelative('minified.js'));
 const content = fs.readFileSync(min_js, 'utf-8');
--- a/internal/runfiles/BUILD.bazel
+++ b/internal/runfiles/BUILD.bazel
@@ -1,0 +1,23 @@
+load("//internal/js_library:js_library.bzl", "js_library")
+
+exports_files(
+    ["index.js"],
+)
+
+js_library(
+    name = "runfiles_helper",
+    srcs = ["index.js"],
+    module_name = "@bazel/runfiles",
+    visibility = ["//visibility:public"],
+)
+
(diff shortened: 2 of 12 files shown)


internal/runfiles/index.js
```diff
@@ -0,0 +1 @@
+module.exports = require('build_bazel_rules_nodejs/internal/linker').runfiles;
```
alexeagle: feels like this is wishing we had already found a way to split up the linker code and move the runfiles helper here.
If we tried to do that, do we still run into the bootstrap problem of needing runfiles available before we've linked it to a place we can load from?

internal/node/test/npm_package_bin.spec.js
```diff
@@ -1,6 +1,6 @@
 const fs = require('fs');
 const path = require('path');
-const runfiles = require(process.env['BAZEL_NODE_RUNFILES_HELPER']);
+const runfiles = require('@bazel/runfiles');
```
alexeagle: hmm, it's a bit confusing that there is no such package on npm. But I think this is probably the best name to use (could be rules_nodejs_runfiles maybe, as this is part of the builtin?)

but I do like the idea of linking it at `node_modules/@bazel/runfiles` rather than use an environment variable.

internal/runfiles/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
+load("//internal/js_library:js_library.bzl", "js_library")
+
+exports_files(
+    ["index.js"],
+)
+
+js_library(
+    name = "runfiles_helper",
+    srcs = ["index.js"],
+    module_name = "@bazel/runfiles",
```
alexeagle: this will cause the linker to lay it out in the node_modules tree, right? we don't rely on custom patched resolve to make this load?

PR #1547, opened 2020-01-14
build: add snapshot build support

gregmagolan:
Example of a generated and published snapshot build at https://github.com/gregmagolan/rules_nodejs-builds/tree/labs.

The first argument of the script is the GitHub org to publish to. In the above case it is my GitHub account `gregmagolan`. Ideally, we'd publish snapshot builds from CI to https://github.com/bazelbuild/rules_nodejs-builds but some discussion as to if this is the best solution is needed first and we need to ask bazelbuild for another repo `rules_nodejs-builds `.

A few points from offline discussion:

* publishing each npm package to its own repo as angular does (https://github.com/angular/core-builds, https://github.com/angular/http-builds, etc...) is not a great solution since we would be publishing to the bazelbuild org and we would both spam their repositories list and need to ask for a new repository for every new npm package

* the last point is why I chose to publish .tar.gz files to this one repo for all of the npm packages. This works well for fetching these in your package.json with: `"@bazel/terser": "https://github.com/gregmagolan/rules_nodejs-builds/raw/1.1.0+7e1ec700/@bazel_terser-labs-snapshot.tar.gz"`

* publishing snapshots to npm is not desirable as it would span the npm versions and we want to reserve `next` for rcs

* publishing to the main repo https://github.com/bazelbuild/rules_nodejs is not desirable as it would make that repository grow in size by adding binary files to it and if you pull all branches on clone you would have to download all of the snapshot releases

* one benefit of the angular approach is that it makes it easy to see diffs in the release artifacts with each commit which we don't get here with the .tar.gz. It would be fairly easy to add the contents of the .tar.gz along side the binary archive in the commit so that diffs in the release artifacts would be available under a single repo (best of both worlds)

* considered making releases for each snapshot built to https://github.com/bazelbuild/rules_nodejs-builds using the GitHub API to upload release artifacts but that would not be as easy to navigate between branches and snapshots between branches would be interleaved in the release logs. You would also not get release artifacts diffs from the last point.

### DIFF
--- a/packages/create/BUILD.bazel
+++ b/packages/create/BUILD.bazel
@@ -1,3 +1,4 @@
+load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
 load("@build_bazel_rules_nodejs//:tools/defaults.bzl", "nodejs_test", "pkg_npm")
 load("//third_party/github.com/bazelbuild/bazel-skylib:rules/copy_file.bzl", "copy_file")
 
@@ -25,6 +26,15 @@ pkg_npm(
         ":copy_bazelrc",
         ":copy_bazelversion",
     ],
+)
+
+# This artifact is used for snapshot builds
+pkg_tar(
+    name = "archive",
+    srcs = [":npm_package"],
+    extension = "tar.gz",
+    strip_prefix = "./npm_package",
+    tags = ["manual"],
 )
 
 nodejs_test(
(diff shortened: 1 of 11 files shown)


gregmagolan:
@alexeagle I'll add the source trees to the commit so we get the diffs. I think the next step here is to ask bazelbuild for a new repo named `rules_nodejs-builds`

gregmagolan:
Added the release artifacts sources: https://github.com/gregmagolan/rules_nodejs-builds/tree/labs

gregmagolan:
Oh would you look at that. @bazel/typescript includes in external npm deps. Whoops: https://unpkg.com/browse/@bazel/typescript@1.1.0/external/npm/node_modules/

Should fix that

Issue #1548, opened 2020-01-14
Flip 


gregmagolan:
Once https://github.com/bazelbuild/bazel/issues/5980 is resolved and 

See https://bazel.build/designs/2016/06/21/environment.html#new-flag---action_env for more context.

A possible (but less than ideal) workaround is to set https://docs.bazel.build/versions/master/command-line-reference.html#flag--distinct_host_configuration to False and don't use `env{}` to pass `COMPILATION_MODE` to build actions. But a distinct host configuration is desirable so we should keep using `env{}` and wait until the #5890 is resolved and `use_default_shell_env=True` can be used without overwritting `env{}`

gregmagolan:
Related issues:
https://github.com/bazelbuild/bazel/issues/4008
https://github.com/bazelbuild/bazel/issues/6473
https://github.com/bazelbuild/bazel/issues/4643

gregmagolan:
This unblocks https://github.com/bazelbuild/rules_nodejs/pull/1516 which is failing as the `NG_BUILD_CACHE` needs to be passed to the `arcitect()` rule and neither `--define=NG_BUILD_CACHE=false` or `--action_env=NG_BUILD_CACHE=false` work right now.

With this change either of those will work if `configuration_env_vars` is used with the target:

```
architect(
    name = "build",
    args = [
        "frontend:build",
        "--outputPath=$(@D)",
    ],
    data = glob(
        [
            "src/*",
            "src/**",
        ],
        exclude = [
            "src/**/*.spec.ts",
            "src/test.ts",
        ],
    ) + [
        "angular.json",
        "browserslist",
        "tsconfig.app.json",
        "tsconfig.json",
        "@npm//mime",
        "@npm//@angular/cli",
        "@npm//@angular/core",
        "@npm//@angular/router",
        "@npm//@angular/platform-browser-dynamic",
        "@npm//@angular-devkit/architect-cli",
        "@npm//@angular-devkit/build-angular",
    ],
    output_dir = True,
    configuration_env_vars = ["NG_BUILD_CACHE"]
)
```

thesayyn:
#1648 We have this problem there too.

nagelp-bosch:
Is this still being considered? IMO defaulting to true would make more sense.

[...]

PR #1549, opened 2020-01-14
feat(builtin): add configuration_env_vars to npm_package_bin

gregmagolan:
This behaves similarly to `configuration_env_vars` in `nodejs_binary` except it passes the env vars to the process via the `env` attribute of `ctx.actions.run`.

Also add support for picking up env vars from `ctx.configuration.default_shell_env` if they are not found in `ctx.vars`.

* `ctx.vars` will contain values from `--define=FOO=BAR`
* `ctx.configuration.default_shell_env` will `--action_env=FOO=BAR` (but not from `--action_env=FOO`).

-----------

test: add test coverage for npm_package_bin configuration_env_vars

And test coverage for using --action_env=FOO=BAR to pass env variables to npm_package_bin & nodejs_binary

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -24,6 +24,7 @@ build --noincompatible_no_support_tools_in_action_inputs
 # Define environment value used by some tests such as //internal/npm_install/test:bazel_bin_test
 build --define=SOME_TEST_ENV=some_value
 test --define=SOME_TEST_ENV=some_value
+build --action_env=SOME_OTHER_ENV=some_other_value
 
 ###############################
 # Remote Build Execution support
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@
 ## IMPORTANT
 # If you change the `default_docker_image` version, also change the `cache_key` version
 var_1: &default_docker_image circleci/node:10.16
-var_3: &cache_key node-0.16-{{ checksum "yarn.lock" }}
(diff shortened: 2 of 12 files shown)


gregmagolan:
This unblocks https://github.com/bazelbuild/rules_nodejs/pull/1516 which is failing as the `NG_BUILD_CACHE` needs to be passed to the `arcitect()` rule and neither `--define=NG_BUILD_CACHE=false` or `--action_env=NG_BUILD_CACHE=false` work right now.

With this change either of those will work if `configuration_env_vars` is used with the target:

```
architect(
    name = "build",
    args = [
        "frontend:build",
        "--outputPath=$(@D)",
    ],
    data = glob(
        [
            "src/*",
            "src/**",
        ],
        exclude = [
            "src/**/*.spec.ts",
            "src/test.ts",
        ],
    ) + [
        "angular.json",
        "browserslist",
        "tsconfig.app.json",
        "tsconfig.json",
        "@npm//mime",
        "@npm//@angular/cli",
        "@npm//@angular/core",
        "@npm//@angular/router",
        "@npm//@angular/platform-browser-dynamic",
        "@npm//@angular-devkit/architect-cli",
        "@npm//@angular-devkit/build-angular",
    ],
    output_dir = True,
    configuration_env_vars = ["NG_BUILD_CACHE"]
)
```

gregmagolan:
Related issue for reference: https://github.com/bazelbuild/rules_nodejs/issues/1548

PR #1558, opened 2020-01-18
Replace grep with bash's regex operator

NathanHowell:
## PR Checklist

## PR Type

- [x] Bugfix

If `grep` is not in the path the launcher scripts don't fail, they just spam the console.

## What is the new behavior?

Substring checks are done with bash operators instead of shelling out to grep.

- [x] No


NathanHowell:
@soldair do those test failures look related? if not can you kick the tests?

alexeagle:
Thank you! nice simplification

PR #1564, opened 2020-01-20
Do not error with unbound variable when running karma web test with `bazel run`

devversion:
Previously, running a web test target with `bazel run` worked, and didn't result in an unbound variable exception. Now with the new runfile bash helpers v2, unbound variables are reported, and in `bazel run` the `TEST_TMPDIR` variable is not set.

There are two ways to fix this: (1) allowing unbound variables by running `set +u` after the helpers, or (2) by just properly handling unbound variables. The latter seems more future-proof to me and best practice.


gregmagolan:
Thanks for the fix!

PR #1571, opened 2020-01-22
build: update angular examples to angular 9.0.0-rc.10 

gregmagolan:
We can remove the @angular/bazel patches with this update.


kylecordes:
Thrilled to see enough has landed to discard the patches.

PR #1593, opened 2020-01-31
change ts_proto_library to default es5 wrapping

### LINKED ISSUE #562: [ts_proto_library] - Does not support Nodejs
# 🐞 bug report

### Affected Rule
ts_proto_library

### Is this a regression?
Nope

### Description
The module format of the code generated is not usable in a Nodejs environment due to it being AMD module format, where as Nodejs expects CommonJS.

From looking at the docs: https://github.com/dcodeIO/ProtoBuf.js/#pbts-for-typescript
They don't appear to support different modules so either a preprocessing step to change the module format is required to changing the protobufjs lib itself to support other module code generation.

## 🔬 Minimal Reproduction
https://github.com/angular/universal/pull/1003

## 🔥 Exception or Error
Error: `ReferenceError: define is not defined`

This is the generated code: 
<pre><code>define('nguniversal/modules/grpc-engine/src/engine_proto_ts', ["protobufjs/minimal"], function($protobuf) {
    "use strict"'
    var $Reader = $protobuf.Reader, $Writer = $protobuf.Writer, $util = $protobuf.util;</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>Mac OS X</code>
</pre>

<pre><code>Build label: 0.20.0-homebrew
Build time: Wed Dec 19 01:10:31 2018 (1545181831)
Build timestamp: 1545181831
Build timestamp as int: 1545181831</code></pre>

**Rules version (SHA):**

<pre>
  <code>

  </code>
</pre>

**Anything else relevant?**

CoopeyB:
This change wraps es5 sources generated with `ts_proto_library` to be wrapped in "default" mode. The "default" mode behaves like UMD, supporting both AMD and CommonJS. This change is necessary to support Node, as Node does not use AMD. Fixes #562

## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Build related changes
- [ ] CI related changes

When using `ts_proto_library`, the application fails at runtime with `ReferenceError: define is not defined`
Issue Number: #562

## What is the new behavior?
Works in Node - `ts_proto_library` will now emit UMD modules instead of AMD.

- [ ] Yes
- [X] No

### DIFF
--- a/packages/labs/test/protobufjs/test.ts
+++ b/packages/labs/test/protobufjs/test.ts
@@ -1,7 +1,5 @@
 import {TestMessage} from 'build_bazel_rules_nodejs/packages/labs/test/protobufjs/test_ts_proto';
 
-// const TestMessage = build_bazel_rules_nodejs.packages.labs.test.protobufjs.TestMessage;
-
 describe('protobufjs', () => {
   it('should work in node', () => {
     expect(TestMessage.verify({
--- a/packages/labs/test/protobufjs/test.ts
+++ b/packages/labs/test/protobufjs/test.ts
@@ -1,0 +1,11 @@
+import {TestMessage} from 'build_bazel_rules_nodejs/packages/labs/test/protobufjs/test_ts_proto';
+
+// const TestMessage = build_bazel_rules_nodejs.packages.labs.test.protobufjs.TestMessage;
+
+describe('protobufjs', () => {
+  it('should work in node', () => {
+    expect(TestMessage.verify({
+      testField: 'Hello',
+    })).toBeFalsy();  // verify returns an error if failed
+  });
+});
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -276,3 +276,8 @@ load("@build_bazel_integration_testing//tools:repositories.bzl", "bazel_binaries
 
 # Depend on the Bazel binaries
 bazel_binaries(versions = [BAZEL_VERSION])
+
+# Install labs dependencies
+load("@npm_bazel_labs//:package.bzl", "npm_bazel_labs_dependencies")
+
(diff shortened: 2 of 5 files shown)


CoopeyB:
@googlebot I fixed it.

packages/labs/test/protobufjs/test.ts
```diff
@@ -0,0 +1,11 @@
+import {TestMessage} from 'build_bazel_rules_nodejs/packages/labs/test/protobufjs/test_ts_proto';
+
+// const TestMessage = build_bazel_rules_nodejs.packages.labs.test.protobufjs.TestMessage;
```
alexeagle: stray comment?

packages/labs/test/protobufjs/test.ts
```diff
@@ -0,0 +1,11 @@
...
+
+describe('protobufjs', () => {
+  it('should work in node', () => {
+    expect(TestMessage.verify({
+      testField: 'Hello',
+    })).toBeFalsy();  // verify returns an error if failed
```
alexeagle: should there be a success case too? or is it enough to assert that it doesn't throw an error?

packages/labs/test/protobufjs/test.ts
CoopeyB: Oops, removed!

packages/labs/test/protobufjs/test.ts
CoopeyB: This test case is really just to run some code - I could change it to `expect(TestMessage).toBeTruthy();` and get the verification needed for this PR, would you prefer the simpler test case?

alexeagle:
Thanks!

PR #1609, opened 2020-02-05
ci: bump Bazel version to 2.1

alexeagle:


gregmagolan:
Hmmm. Looks like the bazel_binaries is looking for the 2.1.0rc4 archive and it is not there: https://github.com/bazelbuild/bazel-integration-testing/blob/master/tools/repositories.bzl

alexeagle:
I'm not surprised. Seems like this should be using bazelisk instead of still having its own thing

alexeagle:
I patched up bazel-integration-testing borrowing some logic from bazelisk in https://github.com/bazelbuild/bazel-integration-testing/pull/168

patched that in here locally so we can update.

Also added a placeholder for where we ought to be testing against multiple bazels

index.bzl
```diff
@@ -89,13 +89,21 @@ def yarn_install(**kwargs):
...
-BAZEL_VERSION = "2.0.0"
+BAZEL_VERSION = "2.1.0rc4"
+
+# Versions of Bazel which users should be able to use.
+# Ensures we don't break backwards-compatibility,
+# accidentally forcing users to update their LTS-supported bazel.
+# These are the versions used when testing nested workspaces with
+# bazel_integration_test.
+SUPPORTED_BAZEL_VERSIONS = [
+    # TODO: add LTS versions of bazel like 1.0.0, 2.0.0
```
gregmagolan: 😎 Tho it will increase our CI time by quite a bit; master only would be nice however buildkite is rather limited in its configuration

bazel-integration-testing.patch
```diff
@@ -0,0 +1,54 @@
...
+-
+ def _get_platform_name(rctx):
+   os_name = rctx.os.name.lower()
+ 
+@@ -45,7 +40,36 @@ def _get_installer(rctx):
+     extension = "sh"
+     installer = "-installer"
+ 
+-  urls = [url.format(version=version, installer=installer, platform=platform, extension=extension) for url in _BAZEL_BINARY_PACKAGES]
++  filename = "bazel-{version}{installer}-{platform}.{extension}".format(
```
gregmagolan: 😎 Should make a note to upstream

gregmagolan:
Suppose for integration testing we could just use bazelisk now instead of the bazel integration test repository rule binary

gregmagolan:
At some point in the future we could remove the `hide_build_files` in pkg_npm (or at least flip to False and mark it for future removal)

bazel-integration-testing.patch
alexeagle: I sent the PR already, sorry should have included PR number in the patch filename but I did them in the wrong order

PR #1619, opened 2020-02-08
feat(examples): add libs to angular baze examples

Toxicable:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


alexeagle:
sorry that giant line is a sore spot for merge conflicts... :(

examples/angular_bazel_architect/libs/ng-lib/BUILD.bazel
```diff
@@ -0,0 +1,18 @@
+load("@build_bazel_rules_nodejs//:index.bzl", "pkg_npm")
+load("@npm_bazel_typescript//:index.bzl", "ts_project")
+
+package(default_visibility = ["//visibility:public"])
+
+ts_project(
```
alan-agius4, line 6: @alexeagle / @Toxicable my only comments would be there a this will produce a different output from `ng-packagr`.

This works great if the user is not concerned about the final application bundle sizes, when having real-world cases however the user would probably care about bundle sizes this will be cause a problems because such libraries will be able to be optimised or changing this library to use `ng-packagr` might cause libraries consumption to break.

This is because there are a couple of fundamental differences between the output of a `ts_project`/`ng-packagr` library.

- `ng-packagr` produces FESMs, the main two reasons for this is for faster webpack compilations and better code splitting.
- `ng-packagr` by default will add a `sideEffect` property `package.json` which makes a library being sideEffect free. By default unless this property is specified Webpack will threat a library as having side effects.
- Most likely Build Optimizer will also not run on the output of this library because of the absence of a `package.json` and `typings` field.

A workaround for the the last 2 points is to add a `package.json` in the npm_package. Which can be done by providing a package.json file with the below contents via `srcs` attr in `pkg_npm`.

```json
{
  "name": "my-lib",
  "sideEffects": false,
  "main": "index.js",
  "typings": "index.d.ts"
}
```

alan-agius4:
Good work @Toxicable

examples/angular_bazel_architect/libs/ng-lib/BUILD.bazel
alan-agius4: Also, a drawback of not having an `ng-packagr/cli` library is for major versions, you cannot use `ng update` to do code modification to that library. That is because the library itself is part of the Angular CLI project.

Toxicable: Yeah so I think what we should do here is merge in this and your one as well, then write up a little something in the README.md on the different approaches, advantages and disadvantages.
Then let whoever wants to use it decide if they're happy to trade bundle size for performance.

Toxicable:
@alexeagle added a bit of info to the readme, let me know what you think.
I think after we get this and @alan-agius4 's one merged ill spend a bit more time writing up docs about how / when you should use the different methods

alexeagle:
Alan's merged first so this needs another rebase, sorry you keep losing the continuous integration game 😉

Toxicable:
@alexeagle Should be good now

PR #1625, opened 2020-02-11
chore(examples): upgrade rules_nodejs and angular

jbedard:


gregmagolan:
Thanks Jason! I should have mentioned that the ViewEngine example needs to say at Angular 8 for this PR.

Should dig further into that error however there are some issues with Bazel + ViewEngine (like the AMD patch needed & the necessary ngc postinstall) so I'd recommend that anyone updating to v9 with Bazel use Ivy anyway.

examples/angular/package.json
```diff
@@ -4,31 +4,31 @@
     "description": "Demo of building Angular apps with Bazel",
     "license": "Apache-2.0",
     "engines": {
-        "node": ">=10.9.0 <13.0.0",
+        "node": ">=10.9.0 <=13.8.0",
```
jbedard: @gregmagolan I hope this was ok? I had 13 locally and was just testing this out 🤷‍♂

gregmagolan: Noticed that and LGTM. Its still the supported range for rules_nodejs & for this example.

PR #1630, opened 2020-02-12
feat(typescript): add plugin support for type-checking lit-element

alexeagle:
/cc @runem


gregmagolan:
RBE issue 

```

(23:21:11) ERROR: /workdir/packages/typescript/test/strict_deps/BUILD:17:1: Couldn't build file packages/typescript/test/strict_deps/grandparent.js: Compiling TypeScript (devmode) //packages/typescript/test/strict_deps:grandparent failed (Exit 1)
--
  | internal/modules/cjs/loader.js:797
  | throw err;
  | ^
  |  
  | Error: Cannot find module 'bazel-out/host/bin/external/build_bazel_rules_typescript/internal/tsc_wrapped_bin.sh.runfiles/build_bazel_rules_nodejs/internal/linker/runfiles_helper.js'

```

should be fixed on a rebase to master since https://github.com/bazelbuild/rules_nodejs/pull/1634 has landed.

WORKSPACE
```diff
@@ -101,8 +101,10 @@ install_bazel_dependencies()
 # With http_archive it only sees releases/download/*.tar.gz urls
 git_repository(
     name = "build_bazel_rules_typescript",
-    commit = "5d79e42953eb8614d961ccf0e3440884e974eeb3",
+    commit = "9f4e2da7ea5ac2eb992b897f2a63cea66732fb62",
+    patches = ["//:rules_typescript.patch"],
```
gregmagolan: patch may no longer be needed now that the underlying BAZEL_NODE_RUNFILES_HELPER issue is fiexed

package.json
```diff
@@ -70,10 +71,11 @@
         "terser": "^4.4.0",
         "testy": "file:./tools/npm_packages/testy",
         "tmp": "0.1.0",
-        "tsickle": "0.33.1",
+        "ts-lit-plugin": "^1.1.9",
+        "tsickle": "0.38.0",
         "tsutils": "2.27.2",
         "typeorm": "0.2.18",
-        "typescript": "3.1.6",
+        "typescript": "^3.7.5",
```
gregmagolan: 🎉  I tried updating typescript here a week or so ago and had issues. Were there changes in rules_typescript to enable it?

packages/typescript/test/lit_plugin/BUILD.bazel
```diff
@@ -0,0 +1,36 @@
...
+
+ts_library(
+    name = "lit_plugin",
+    srcs = glob(["*.ts"]),
+    expected_diagnostics = [
+        """TS2322: \\[lit\\] Unknown tag "unknown-element". Did you mean 'lit-element'?""",
+        "TS2322: \\[lit\\] Type '222' is not assignable to 'string",
+        "TS2322: \\[lit\\] Type '333' is not assignable to 'string",
+        """TS2322: \\[lit\\] Type '"aaa"' is not assignable to 'number'""",
+        #"""TS2322: \\[lit\\] Type '"bbb"' is not assignable to 'number'""",
```
gregmagolan: nit: why the commented out assertion?

packages/typescript/test/lit_plugin/BUILD.bazel
```diff
@@ -0,0 +1,36 @@
...
+        """TS2322: \\[lit\\] Type '"ccc"' is not assignable to 'number'""",
+        "TS2322: \\[lit\\] Type '444' is not assignable to 'string",
+        "TS2322: \\[lit\\] Type '{ field: number; }' is not assignable to '{ field: string; }'",
+    ],
+    tsconfig = ":tsconfig.json",
+    deps = [
+        "@npm//lit-element",
+        # FIXME: plugins["@npm//ts-lit-plugin"] instead?
```
gregmagolan: is it `plugins` is google3?

package.json
marcus-sa: I've been using 3.7.5 since the release with no issues.

WORKSPACE
alexeagle: nope, still needed, I introduced a logic bug only observable when using workers outside of the local Runfiles case (like in `@bazel/worker`)

package.json
alexeagle: I had to do tsickle too, and a strange jasmine typings problem needed a node types update

packages/typescript/test/lit_plugin/BUILD.bazel
alexeagle: @rictic might TAL at at - I couldn't get this one to trigger. Not really important though, hard to imagine there's a bug in how we invoke the plugin, since we just pass it the whole options block. Will remove

packages/typescript/test/lit_plugin/BUILD.bazel
alexeagle: no, in google3 you have `ts_library(use_lit_plugin=True)`
But that's because we are highly locked down for allowing customization. Under Bazel we should probably let people choose their plugins and it's their choice to maintain the build to deal with bad behavior the plugins might cause

PR #1639, opened 2020-02-14
fix(examples): add fixes to angular architect

alan-agius4:
This PR adds;
1)  `lint` as a target
2) Changes npm script to use bazelisk
3) Disables `forkTypeChecker` as under Windows this currently errors out with:
```
Error: spawn C:\users\alag\_bazel_alag\jeen2jsq\external\build_bazel_rules_nodejs\internal\node\_node_bin\node ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:264:19)
    at onErrorNT (internal/child_process.js:456:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21)
Emitted 'error' event on ChildProcess instance at:
    at Process.ChildProcess._handle.onexit (internal/child_process.js:270:12)
    at onErrorNT (internal/child_process.js:456:16)
    at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  errno: 'ENOENT',
  code: 'ENOENT',
  syscall: 'spawn C:\\users\\alag\\_bazel_alag\\jeen2jsq\\external\\build_bazel_rules_nodejs\\internal\\node\\_node_bin\\node',
  path: 'C:\\users\\alag\\_bazel_alag\\jeen2jsq\\external\\build_bazel_rules_nodejs\\internal\\node\\_node_bin\\node',
  spawnargs: [
    '--require',
    'C:/users/alag/_bazel_alag/jeen2jsq/external/build_bazel_rules_nodejs/internal/node/node_patches.js',
    '--preserve-symlinks',
    '--preserve-symlinks-main',
    'c:\\users\\alag\\_bazel~1\\jeen2jsq\\execroot\\angula~1\\node_modules\\@ngtools\\webpack\\src\\type_checker_worker.js',
    '9d93e901-158a-4cf9-ba1b-2f0582ffcfeb'
  ]
}
```

- [ ] Yes
- [ ] No

### DIFF
--- a/examples/angular_bazel_architect/BUILD.bazel
+++ b/examples/angular_bazel_architect/BUILD.bazel
@@ -113,6 +113,26 @@ architect_test(
 )
 
 architect(
+    name = "lint",
+    args = ["frontend:lint"],
+    data = glob([
+        "src/**/*",
+        "e2e/src/**/*",
+    ]) + [
+        "angular.json",
+        "tsconfig.json",
+        "tsconfig.spec.json",
+        "tsconfig.app.json",
+        "tslint.json",
+        "e2e/tsconfig.json",
+        "@npm//@angular/cli",
+        "@npm//@angular-devkit/build-angular",
+        "@npm//codelyzer",
+        "@npm//tslint",
+    ],
+)
+
+architect(
     name = "serve",
     args = [
         "frontend:serve",
--- a/examples/angular_bazel_architect/angular.json
+++ b/examples/angular_bazel_architect/angular.json
@@ -17,6 +17,7 @@
         "build": {
           "builder": "@angular-devkit/build-angular:browser",
           "options": {
+            "forkTypeChecker": false,
             "preserveSymlinks": true,
             "outputPath": "dist/frontend",
             "index": "src/index.html",
--- a/examples/angular_bazel_architect/package.json
+++ b/examples/angular_bazel_architect/package.json
@@ -3,11 +3,11 @@
   "version": "0.0.0",
   "scripts": {
     "ng": "ng",
-    "start": "ng serve",
-    "build": "bazel build :build",
-    "test": "bazel test :test",
-    "lint": "ng lint",
-    "e2e": "ng e2e",
+    "start": "bazelisk run :serve",
+    "build": "bazelisk build :build",
+    "test": "bazelisk test :test",
+    "lint": "bazelisk run :lint",
+    "e2e": "bazelisk test :e2e",
     "postinstall": "patch-package && ngcc --properties es2015 browser module main --first-only"
   },
   "private": true,


examples/angular_bazel_architect/package.json
```diff
@@ -3,11 +3,11 @@
...
     "ng": "ng",
-    "start": "ng serve",
-    "build": "bazel build :build",
-    "test": "bazel test :test",
-    "lint": "ng lint",
-    "e2e": "ng e2e",
+    "start": "bazelisk run :serve",
+    "build": "bazelisk build :build",
+    "test": "bazelisk test :test",
+    "lint": "bazelisk run :lint",
```
alan-agius4: Should actually this be a using test so that it is not run for unchanged sources?

examples/angular_bazel_architect/BUILD.bazel
```diff
@@ -112,6 +112,26 @@ architect_test(
     ],
 )
 
+architect(
+    name = "lint",
+    args = ["frontend:lint"],
+    data = glob([
+        "src/**/*",
```
alan-agius4: From https://docs.bazel.build/versions/master/be/functions.html#glob

> /bar//*.txt matches every .txt file in every subdirectory of this package, if at least one directory on the resulting path is called bar, such as xxx/bar/yyy/zzz/a.txt or bar/a.txt (remember that ** also matches zero segments) or bar/zzz/a.txt

examples/angular_bazel_architect/BUILD.bazel
```diff
@@ -112,6 +112,26 @@ architect_test(
     ],
 )
 
+architect(
+    name = "lint",
```
alan-agius4: From https://docs.bazel.build/versions/master/be/functions.html#glob

> /bar//*.txt matches every .txt file in every subdirectory of this package, if at least one directory on the resulting path is called bar, such as xxx/bar/yyy/zzz/a.txt or bar/a.txt (remember that ** also matches zero segments) or bar/zzz/a.txt

examples/angular_bazel_architect/angular.json
```diff
@@ -17,6 +17,7 @@
         "build": {
           "builder": "@angular-devkit/build-angular:browser",
           "options": {
+            "forkTypeChecker": false,
```
Toxicable: Ahhh nice one on finding the cause of this issue.  
Could you try enabling windows ci now that this works also?

alan-agius4: Enabled, let's wait and see :)

alan-agius4: There seems to be the same error for `jest-workers`

```
Error: spawn C:\b\ojvxx6nx\execroot\build_bazel_rules_nodejs\_tmp\00ec2cae7910fb9a26b84ca52e1907d7\_bazel_b\r453ukw2\external\build_bazel_rules_nodejs\internal\node\_node_bin\node ENOENT
--
  | at Process.ChildProcess._handle.onexit (internal/child_process.js:264:19)
  | at onErrorNT (internal/child_process.js:456:16)
  | at processTicksAndRejections (internal/process/task_queues.js:80:21)
  | Emitted 'error' event on ChildProcess instance at:
  | at Process.ChildProcess._handle.onexit (internal/child_process.js:270:12)
  | at onErrorNT (internal/child_process.js:456:16)
  | at processTicksAndRejections (internal/process/task_queues.js:80:21) {
  | errno: 'ENOENT',
  | code: 'ENOENT',
  | syscall: 'spawn C:\\b\\ojvxx6nx\\execroot\\build_bazel_rules_nodejs\\_tmp\\00ec2cae7910fb9a26b84ca52e1907d7\\_bazel_b\\r453ukw2\\external\\build_bazel_rules_nodejs\\internal\\node\\_node_bin\\node',
  | path: 'C:\\b\\ojvxx6nx\\execroot\\build_bazel_rules_nodejs\\_tmp\\00ec2cae7910fb9a26b84ca52e1907d7\\_bazel_b\\r453ukw2\\external\\build_bazel_rules_nodejs\\internal\\node\\_node_bin\\node',
  | spawnargs: [
  | '--require',
  | 'C:/b/ojvxx6nx/execroot/build_bazel_rules_nodejs/_tmp/00ec2cae7910fb9a26b84ca52e1907d7/_bazel_b/r453ukw2/external/build_bazel_rules_nodejs/internal/node/node_patches.js',
  | '--preserve-symlinks',
  | '--preserve-symlinks-main',
  | 'C:\\b\\ojvxx6nx\\execroot\\build_bazel_rules_nodejs\\_tmp\\00ec2cae7910fb9a26b84ca52e1907d7\\_bazel_b\\r453ukw2\\execroot\\angular_bazel_architect\\node_modules\\jest-worker\\build\\workers\\processChild.js'
  | ]
  | }
```

I am thinking this might be related to something how the actual spawn of the architect is happening from bazel. This will need some digging.

alan-agius4: JFYI: https://github.com/angular/angular-cli/issues/16119

Toxicable: This is the code that puts the bazel node into the PATH https://github.com/bazelbuild/rules_nodejs/blob/master/internal/node/require_patch.js#L30
but that should be fine

alan-agius4: Most likely the problem is in this line https://github.com/bazelbuild/rules_nodejs/blob/25600ea4c307c120d04a14c6d41644a3e8d3ed3a/packages/node-patches/src/subprocess.ts#L50

That should probably have an extension of `exe` or  `bat` for windows.

Unfortunately the `node-patches` tests are all disabled for windows. 

//cc @gregmagolan & @alexeagle

alan-agius4: Changing it to `bat`  for windows does work, but eventually it will fail due `bad file descriptor, uv_pipe_open` in `jest-workers` when trying to fork the process.

That is because on Windows `bat` and `cmd` are non executable, on their own without the `shell` option defined. 

So, changing the above line to ` process.argv[0] = process.execPath` yield the correct result in Windows, though of course it's not setting the patched env variables which are configured in the batch file.

Is there any reason why we don't set `process.env` instead since forked/spawned processes should inherit the env vars from the parent process.

alan-agius4: Jest-workers do offer the possibility to use pass additional option to the fork process which in this case we can tell you use `shell`. But this would be a "workaround", as other libs that uses forked processes might not offer this functionality.

examples/angular_bazel_architect/package.json
alexeagle: I have a whole design for running lints. I don't think it should be a test because `bazel test //...` shouldn't fail just because you have something to fix before merging your change - basically an opinion on the developer workflow. Instead you use an Aspect (or an --experimental_extra_action_listener) so that you have a command-line way to say "re-run a portion of the build, but when you visit a ts_library target, run this other thing on its inputs". This way you get the benefits of caching and incremental re-lint but it's still a sidecar workflow.
Let's discuss separately?

examples/angular_bazel_architect/BUILD.bazel
alexeagle: Hmm! In copybara, which also uses starlark for configuration, their glob implementation has different semantics!
https://github.com/google/copybara/blob/master/docs/reference.md#matching-multiple-strings-with-one-expression
so #TIL that "glob" isn't a starlark builtin and things are not consistent

examples/angular_bazel_architect/BUILD.bazel
```diff
@@ -112,6 +112,26 @@ architect_test(
...
+    args = ["frontend:lint"],
+    data = glob([
+        "src/**/*",
+        "e2e/src/**/*",
+    ]) + [
+        "angular.json",
```
alexeagle: i keep wondering if we should extract a constant to make these less repetitive across rules, or if it just doesn't matter because the file is so trivial to generate

examples/angular_bazel_architect/angular.json
alexeagle: @soldair could chime in about the node-patches issue

examples/angular_bazel_architect/package.json
alan-agius4: Yes, let’s take this offline.

examples/angular_bazel_architect/BUILD.bazel
alan-agius4: Yeah definitely, though currently I was focused more on getting things “working” and polish after.

Also probably there are some deps that can be removed such as Angular CLI.

alan-agius4:
@alexeagle, can we get this in and address the linker issues on Windows in a separate PR? Since these changes doesn’t introduce any new failures compared to current HEAD.

examples/angular_bazel_architect/angular.json
soldair: we either need a different executable that works without shell or we can force the shell option to true on windows if the executable is node. Seems easiest to do the last one

soldair: if someone can make a really smol .exe to use instead of the bat that should work also but i'm less clear on how to do that.

alan-agius4: @soldair how can we force the shell open on processes forked in 3rd party libraries?

soldair: Like node-patches modifies the global fs object, we could patch the child_process object such that when the executable  to spawn is the batch file we set shell:true and we make execFile do similar trick. Much like baking in this workaround for execFile. https://nodejs.org/api/child_process.html#child_process_spawning_bat_and_cmd_files_on_windows

The alternative is to build the shim into a proper `.exe` file which @alexeagle mentioned and we don't have to modify child_process at the cost of whatever we need to do to build the exe. Right now it's dynamically created when you run node the first time.

PR #1640, opened 2020-02-14
feat(rollup): add worker process

jbedard:
POC of a worker process for rollup...


internal/rollup/worker.js
```diff
@@ -0,0 +1,202 @@
...
+    await bundle.write(outputOptions);
+  } catch (e) {
+    worker.log(e);
+    return false;
+  }
+
+  return true;
+}
+
+// Load the config file.
```
jbedard: This part is essentially copied from the rollup CLI which isn't great...

internal/rollup/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
...
+
+filegroup(
+    name = "node_modules_none",
+)
+
+nodejs_binary(
+    name = "worker",
+    data = [
+        "worker.js",
+        "@npm//@bazel/worker",
```
jbedard: Do we need to make the `@npm` part configurable? I think previously the `rollup_bin` rule attribute allowed that?

internal/rollup/BUILD.bazel
```diff
@@ -0,0 +1,33 @@
...
+        "@npm//rollup",
```
jbedard: Do we need to make the @npm part configurable? I think previously the rollup_bin rule attribute allowed that?

joeljeske:
@jbedard I am very interested in this work. My structure has intermediate rollup builds for package groups as opposed to a single rollup build at the top level of an application. I expect that keeping rollup warm could be very beneficial for me.

Have you seen good performance gains with this PR so far? Would you be interested in any assistance in getting this finished?

joeljeske:
Would we have the possibility of sandbox violations with a rollup worker? It seems that with TS worker there is a clear place to override the loading of files such that only approved files can be loaded, but with rollup, any plugin or core can query the FS to find files. Am I missing something as far as worker sandboxing goes?

packages/rollup/src/worker.js
```diff
@@ -0,0 +1,217 @@
...
+        const [entryName, entryPath] = args[++i].split('=', 2);
+        entries[entryName] = entryPath;
+      } while (i < args.length && !args[i + 1].startsWith('--'));
+    } else {
+      throw new Error(`${MNEMONIC}: unknown worker param: ${arg}`);
+    }
+  }
+
+  // Additional options passed via config file
+  // TODO: only on first run or if bazel says it changed
```
jbedard: This is probably worth doing at some point. Doesn't have to be in this initial PR, but only running `loadConfigFile()` when the config (and its deps?) changes would avoid a rollup run per worker loop...

packages/rollup/src/package.json
```diff
@@ -22,5 +25,8 @@
         "npm_bazel_rollup": {
             "rootPath": "."
         }
+    },
+    "dependencies": {
+        "@bazel/worker": "0.0.0-PLACEHOLDER"
```
jbedard: Should we make this a `peerDependency` or `optionaldependencies` since it is opt-in atm?

packages/rollup/src/rollup_bundle.bzl
```diff
@@ -304,6 +318,16 @@ def _rollup_bundle(ctx):
     # See CLI documentation at https://rollupjs.org/guide/en/#command-line-reference
     args = ctx.actions.args()
 
+    if ctx.attr.supports_workers:
+        # Set to use a multiline param-file for worker mode
+        args.use_param_file("@%s", use_always = True)
+        args.set_param_file_format("multiline")
+
+    # TODO: enforce here? Currently in js to allow a subset of args...
```
jbedard: I decided to let the .js fail on unknown args instead of disallowing any args like here. WDYT?

This allows supporting "key parameters" which aren't actually required, but might be worth supporting? I mainly switched from enforcing here to in .js to support `--environment` which one of the unit tests uses...

packages/rollup/src/index.js
```diff
@@ -0,0 +1,258 @@
...
+
+    if (config.output) {
+      outputOptions = {...config.output, ...outputOptions};
+    }
+
+    inputOptions = {...config, ...inputOptions};
+
+    // Delete from our copied inputOptions, not the config which
+    // may be external and persisted across runs
+    delete inputOptions.output;
```
jbedard: @alexeagle this was the main problem with config files being lost... previously we deleted `config.output` in the  `if (config.output)`. That `config` object gets cached, so next time we run the test the worker has bad cache data.

Now we simply delete it from our per-run instance of the options.

packages/rollup/src/index.js
```diff
@@ -0,0 +1,262 @@
...
+  };
+
+  // outputOptions: https://github.com/rollup/rollup/blob/v1.31.0/cli/run/loadConfigFile.ts#L35-L38
+  const outputOptions = {
+    exports: 'named',
+    format: 'cjs',
+    file: cjsConfigFile,
+  };
+
+  await runRollup(configFile, inputOptions, outputOptions);
```
jbedard: The cache key of the rollup run of the config is the configFile path.

packages/rollup/src/index.js
```diff
@@ -0,0 +1,262 @@
...
+  worker.debug(`${MNEMONIC}[${PID}][${cacheKey}].write()`, (bundleEndTime - rollupEndTime) / 1000);
+
+  return true;
+}
+
+// Run rollup, will use + re-populate the cache
+async function runRollupBundler(args /*, inputs */) {
+  const {inputOptions, outputOptions} = await parseCLIArgs(args);
+
+  return runRollup(inputOptions.input, inputOptions, outputOptions);
```
jbedard: The cache key of the main bundle is the input files array. So if other input/outputOptions change it will still share the cache. I think that's ok? I think that's a good thing since those rarely change, and if they do can probably still cache a lot?

jbedard:
@alexeagle one (I hope just one) more issue that I haven't figured out... the `rollup/test/integration` tests still fail because inline sourcemaps are being inserted for each import, and I still have no idea why.

For example `golden.amd.js_` has:
```
var far_a_b_c = `@far/a/b/c`;
```
The generated file when using the work has:
```
var far_a_b_c = `@far/a/b/c`;
//# sourceMappingURL=data:application/json;base64, ....
```

But the bottom of the file still has the correct
```
//# sourceMappingURL=bundle.amd.js.map
```

I haven't found any API/CLI docs on this so far, so I'm a bit lost once again...

packages/rollup/src/index.js
alexeagle: that sounds scary and incorrect though - if I change the config file my change isn't reflected in the next rebuild? Is that what it means?

packages/rollup/src/index.js
alexeagle: cool nice find

packages/rollup/src/index.js
jbedard: A cache miss doesn't mean rollup won't run, rollup just won't have any data from a previous run so it must do everything from scratch.

I think the best case would be if the key was the bazel target being executed, so each rollup bazel target has its own cache (actually 2, an extra for the config file).

Although sharing across rules where only the output type changes might be a good thing... meaning what I have here might be best? The same input points should mean the same file dep graph. I think parsing to create that graph would be the expensive thing.

Either way I don't think it would lead to bugs. A cache miss just means rollup must start from scratch. Too many (unrelated?) things sharing a key just means the cache hit might be useless, and then caches something probably useless to the next person as well (like when I had two keys hardcoded).

packages/rollup/test/integration/BUILD.bazel
```diff
@@ -56,12 +58,46 @@ nodejs_binary(
     entry_point = ":sha256.js",
 )
 
+# Verify the hash of bundle .map files
```
jbedard: I added a commit to test the `.map` file don't change similar to the `.js` tests. Not sure how valuable that is? But right now I'm getting errors with sourcemaps being inlined instead of in their own files so I wanted to verify the `.map`s were being generated correctly...

jbedard:
@gregmagolan you might want to merge the first commit first? I don't really understand it and just put what @alexeagle did into a separate commit. Or I could just update the commit message if someone has an explanation to put in there...

packages/rollup/src/index.js
alexeagle: sure, a de-optimization is not so critical as long as the build is correct. I think we still need to iterate on this some more to get the best performance in all cases

joeljeske:
Wow great work on this @jbedard and @alexeagle thanks! Did yall ever run into any issues related to https://github.com/bazelbuild/rules_nodejs/pull/1640#issuecomment-593676631 regarding file sandboxing? I'm just curious for my own edification. 

Thanks again!

packages/rollup/src/index.js
jbedard: Yeah I'd like to think about this more. Although I *think* using the bundle-input-files + bundle-config-file (which is also the "input file" for the config) as cache keys for the two rollup invocations might be best. I think the other option would be the specific bazel target.

I'd like to investigate + iterate more on potential perf improvements though 👍

PR #1650, opened 2020-02-17
fix: remove empty arguments from launcher

purkhusid:
## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently the launcher script adds a empty argument to the end of all invocations.
Some programs do not like it and fail. An example is `grpc_tools_node_protoc_plugin` in the `grpc-tools` NPM package.

## What is the new behavior?
If the argument list is empty we don't add an empty argument when invoking programs

- [ ] Yes
- [x] No


internal/node/launcher.sh
```diff
@@ -224,13 +224,28 @@ if [ "${EXPECTED_EXIT_CODE}" -eq "0" ]; then
   # handled by the node process.
   # If we had merely forked a child process here, we'd be responsible
   # for forwarding those OS interactions.
-  exec "${node}" "${LAUNCHER_NODE_OPTIONS[@]:-}" "${USER_NODE_OPTIONS[@]:-}" "${MAIN}" "${ARGS[@]:-}"
+  if (( ${#ARGS[@]} )); then
```
gregmagolan: I believe `if (( ${#ARGS} )); then` is equivalent and sufficient here.

The duplication is not great. How about something like:

```
command="${node}" "${LAUNCHER_NODE_OPTIONS[@]:-}" "${USER_NODE_OPTIONS[@]:-}" "${MAIN}"
# If there are no arguments we skip the arguments because some programs
# such as grpc_tools_node_protoc_plugin in the grpc-tools do not handle empty arguments well
if (( ${#ARGS} )); then
  command="${command}" "${ARGS[@]:-}"
fi
exec "${command}"
```

purkhusid: There was a issue with using `set -u` and empty arrays with older versions of bash.
See more information here: https://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u

purkhusid:
@gregmagolan Added a testt

internal/node/launcher.sh
```diff
@@ -224,12 +224,12 @@ if [ "${EXPECTED_EXIT_CODE}" -eq "0" ]; then
   # handled by the node process.
   # If we had merely forked a child process here, we'd be responsible
   # for forwarding those OS interactions.
-  exec "${node}" "${LAUNCHER_NODE_OPTIONS[@]:-}" "${USER_NODE_OPTIONS[@]:-}" "${MAIN}" "${ARGS[@]:-}"
+  exec "${node}" "${LAUNCHER_NODE_OPTIONS[@]:-}" "${USER_NODE_OPTIONS[@]:-}" "${MAIN}" ${ARGS[@]+"${ARGS[@]}"}
```
gregmagolan: That's neat 👍 👍

internal/node/test/BUILD.bazel
```diff
@@ -335,3 +335,10 @@ nodejs_test(
     expected_exit_code = 55,
     templated_args = ["--node_options=--require=$(rlocation $(location :bootstrap_with_patch.js))"],
 )
+
+nodejs_test(
+    name = "binary_with_no_arguments",
+    data = ["empty_args_fail.js"],
+    entry_point = "empty_args_fail.js",
+    expected_exit_code = 0,
```
gregmagolan: nit: `expected_exit_code = 0` is redundant

internal/node/test/empty_args_fail.js
```diff
@@ -0,0 +1,10 @@
+// Used to test that there is no empty argument
+// sent to the program when the argument list is empty.
+// Some programs do not handle empty arguments gracefully
+
+var theArgs = process.argv.slice(2);
+
+if (theArgs.length != 0) {
+  // Non-zero exit code if the argument list is not empty
+  process.exit(42)
```
gregmagolan: I approve of the error code

PR #1659, opened 2020-02-20
chore: switch to rules_pkg

alexeagle:


alexeagle:
/cc @aiuto 
I hope other rules authors hide the rules_pkg dev dependency from their distribution too

aiuto:
On Wed, Feb 26, 2020 at 11:46 AM Alex Eagle <notifications@github.com>
wrote:

> /cc @aiuto <https://github.com/aiuto>
> I hope other rules authors hide the rules_pkg dev dependency from their
> distribution too
>
Yes. I am a fan of packaging products so that your users do not need the
tools needed to build and test the product. The only dependencies you
should show are the ones needed to use the product. Mostly those should be
explicit, except for the rare cases where you need to pin to a very
specific version of a dependency and have to vendor it in yourself.

Issue #1621, opened 2020-02-10
pkg_web: Ability to stamp outputs. 


thesayyn:
# 🚀 feature request

Stamping pkg_web outputs

### Relevant Rules

pkg_web

### Description

pkg_web  rule cannot stamp outputs.

### Describe the solution you'd like

pkg_web rules should have the ability to stamp outputs as pkg_npm does.

### Describe alternatives you've considered

I've tried to combine pkg_npm and pkg_web to stamp builds although I couldn't succeed because pkg_npm does not pick up the transitive files either in `src` nor `deps` attribute.

thesayyn:
I will work on this if it's okay? @alexeagle

alexeagle:
Fixed by #1664

PR #1664, opened 2020-02-23
feat(builtin): support for substitutions

thesayyn:
## PR Checklist

- [ X] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ X ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #1621 

- [ ] Yes
- [ X ] No


thesayyn:
//internal/pkg_web/test3:test is failing. I'm gonna fix it asap.

internal/pkg_web/pkg_web.bzl
```diff
@@ -25,14 +25,17 @@ _ATTRS = {
     "additional_root_paths": attr.string_list(
         doc = """Path prefixes to strip off all srcs, in addition to the current package. Longest wins.""",
     ),
+    "substitutions": attr.string_dict(
+        doc = """Key-value pairs which are replaced in all the files while building the package.""",
```
thesayyn: "Key-value pairs are replaced in all files while building the package."

internal/pkg_web/test3/BUILD.bazel
```diff
@@ -0,0 +1,26 @@
+load("@build_bazel_rules_nodejs//:index.bzl", "pkg_web")
```
alexeagle: "test3" isn't a descriptive name, can you make it "substitutions" or similar?

internal/pkg_web/test3/spec.js
```diff
@@ -0,0 +1,15 @@
+const fs = require('fs');
+const path = require('path');
+
+process.chdir(path.join(process.env['TEST_SRCDIR'], 'build_bazel_rules_nodejs'));
+console.error(fs.readdirSync('.'));
+describe('pkg_web', () => {
+  it('should match the golden file', () => {
+    const output = 'build_bazel_rules_nodejs/internal/pkg_web/test/pkg/index.html';
+    const golden = 'build_bazel_rules_nodejs/internal/pkg_web/test/index_golden.html_';
```
alexeagle: use the `golden_file_test` rule instead

internal/pkg_web/pkg_web.bzl
```diff
@@ -48,12 +51,15 @@ def move_files(output_name, files, action_factory, var, assembler, root_paths):
     www_dir = action_factory.declare_directory(output_name)
     args = action_factory.args()
     args.add(www_dir.path)
+    args.add(version_file.path)
```
alexeagle: I think this needs some conditional logic because the version_file isn't always present?

I think we need to detect use of `--stamp` like pkg_npm does `stamp = ctx.attr.node_context_data[NodeContextInfo].stamp`

internal/pkg_web/assembler_spec.js
```diff
@@ -14,21 +15,42 @@ describe('assembler', () => {
...
+      ]);
+      expect(fs.readFileSync('output/path/to/thing1.txt', {
+        encoding: 'utf-8'
+      })).toBe('some other content');
+    })
+
+
+    it('should replace contents with dynamic text from volatile file', () => {
+      assembler.main(
+          [outdir, volatilePath, '{"content":"{TEST_KEY}"}', '--assets', 'path/to/thing1.txt']);
```
alexeagle: we should make the API match with pkg_npm. Over there we don't let you grab arbitrary keys from the version file, instead we know the key that indicates the version and the `replace_with_version` attribute which says what placeholder is used.

I suspect you prefer this more general implementation? but I think it's more important to be consistent.

internal/pkg_web/pkg_web.bzl
thesayyn: as far as I know, the version file is always present but the content changes regarding to the `--workspace_status_command`. Since we have logic in bundler which checks for the keys, well be okay I guess.

internal/pkg_web/assembler_spec.js
thesayyn: Yes. I think we should permit the users to replace the contents with the arbitrary keys from the version file. It would be more useful this way. Since we have a key check in bundler, the user would get an error if the key doesn't exist in the version file.

thesayyn: In addition, this approach is more general and gives you the ability to replace "placeholder texts" with arbitrary keys from the version file. 

For instance, In one of our projects, we are replacing "BUILD_SCM_HASH" with the real hash along with the  "BUILD_SCM_VERSION" key and it is not possible to accomplish the same behavior with the pkg_npm rule.

internal/pkg_web/test3/spec.js
thesayyn: Done ✅

internal/pkg_web/test3/BUILD.bazel
thesayyn: Done ✅

internal/pkg_web/assembler_spec.js
thesayyn: As far as I understand the rule has to obey --stamp which is stored in the node_context helper rule. I should make this work that way.

internal/pkg_web/pkg_web.bzl
thesayyn: As far as I understand the rule has to obey --stamp which is stored in the node_context helper rule. I should make this work that way.

internal/pkg_web/assembler.js
```diff
@@ -29,9 +30,47 @@ function mkdirp(p) {
...
+  const stampMap = getMappingsFromVolatileFile(volatileFilePath);
+
+  const normalizedSubstitutions = {};
+
+  for (const occurrence in substitutions) {
+    let substituteWith = substitutions[occurrence];
+    if (substituteWith.match(/^{.*?}$/)) {
+      substituteWith = substituteWith.replace(/^{(.*?)}$/, '$1');
+      if (!stampMap[substituteWith]) {
+        throw new Error(`Could not find ${substituteWith} key in volatile-status file.`);
```
mattem, line 54: I'm not sure about throwing this error here - if the user is replacing `0.0.0-PLACEHOLDER` as the version string for regular builds with a key from `--workspace_status_command` which is tied to `--stamp`, this will throw when not running a stamped build

internal/node/_node_bin/node
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
mattem, line 1: Revert this file?

internal/pkg_web/assembler_spec.js
mattem: Perhaps provide both APIs? The `replace_with_version` attr on pkg_web would provide a consistent API surface for both rules, and pkg_npm would change to support substitutions from  workspace_status?

internal/pkg_web/golden_test/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
...
+
+package(default_visibility = ["//visibility:public"])
+
+pkg_web(
+    name = "pkg",
+    srcs = [
+        "index.js",
+    ],
+    substitutions = {
+        "THIS_SHOULD_BE_REPLACED": "THIS_SHOULD_BE_REPLACED_WITH_THIS",
```
mattem, line 12: nit: I know this is a substitutions, but at first I didn't see the different between the strings after the build as only the suffix changed 😄 
Perhaps provide a real world example of a replacement here, as this is mostly centered around version replacement.

internal/node/_node_bin/node
thesayyn: This happened automatically when I was trying to commit the changes.

thesayyn: Sure, I'll try to revert it.

internal/pkg_web/assembler.js
thesayyn: When I tie this with the version file from node context, which only appears when --stamp option is present, this would be long gone. So I thought it would be better if we explicitly check for the keys when the user trying to stamp builds with --stamp option present.

internal/pkg_web/golden_test/BUILD.bazel
thesayyn: I agree, a bit confusing. we can add an example that replaces with an arbitrary key from the volatile file therefore we would have tests and examples for both use cases.

alexeagle:
Hey I'm finally picking up stamping again, I'll get this merged shortly...

thesayyn:
I had no time to work on this and there is still some work to do. Maybe I could get them done by Sunday.

internal/pkg_web/assembler_spec.js
alexeagle: update, I'm deprecating `replace_with_version` so let's just do it your way :)

alexeagle:
no worries @thesayyn I have time today and want to make this consistent with other stamping changes I just landed. I'll push a fixup commit

alexeagle:
@googlebot I consent.

PR #1676, opened 2020-02-28
Configures detailed output to XML_OUTPUT_FILE

joeljeske:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Build related changes
- [ ] CI related changes

I think this is considered a bugfix since a test runner is expected to write results to the `XML_OUTPUT_FILE`, however there was no explicit error that this is fixing. 

Currently, no test details are written to the testlogs directory, rather the output is captured via bazel core and written into a default file.

## What is the new behavior?

The `$XML_OUTPUT_FILE` is populated with proper test results detailing times/statuses/results from specs/suites from jasmine.

- [ ] Yes
- [X] No

It could be considered a breaking change if a consumer has changed the param `jasmine` which defaults `@npm//@bazel/jasmine` to a different package altogether (as opposed to just changing the repository label from the default `@npm`. If a consumer has done this, they should be responsible for keeping up to date with changes here 

## Other information

I considered making this an opt-in reporter, but it seems like this is always the desired behavior. I'd be happy to change it or move it behind a flag if other uses cases are presented. 

I also did not know how to test this well. It seemed as though most of jasmine_node_test own tests used various functionality but did not necessarily assert on the results. This led me to thinking that this did not need testing rather by virtue of not failing it was considered working.

### DIFF
--- a/packages/jasmine/src/index.js
+++ b/packages/jasmine/src/index.js
@@ -10,3 +10,11 @@ exports.boot = boot;
 // re-export jasmine and its transitive dep jasmine-core
 exports.jasmine = require('jasmine');
 exports.jasmineCore = jasmineCore;
+
+let JUnitXmlReporter = null;
+try {
+  JUnitXmlReporter = require('jasmine-reporters').JUnitXmlReporter;
+} catch (err) {
+  // fail quietly if jasmine-reporters is not available
+}
+exports.JUnitXmlReporter = JUnitXmlReporter;
--- a/packages/jasmine/src/jasmine_node_test.bzl
+++ b/packages/jasmine/src/jasmine_node_test.bzl
@@ -35,6 +35,10 @@ def jasmine_node_test(
         **kwargs):
     """Runs tests in NodeJS using the Jasmine test runner.
 
+    Detailed XML test results are found in the standard `bazel-testlogs`
+    directory. This may be symlinked in your workspace.
+    See https://docs.bazel.build/versions/master/output_directories.html
+
     To debug the test, see debugging notes in `nodejs_test`.
 
     Args:
--- a/packages/jasmine/src/jasmine_runner.js
+++ b/packages/jasmine/src/jasmine_runner.js
@@ -3,6 +3,7 @@ const path = require('path');
 const bazelJasmine = require('@bazel/jasmine');
 
 const JasmineRunner = bazelJasmine.jasmine;
+const JUnitXmlReporter = bazelJasmine.JUnitXmlReporter;
 
 let jasmineCore = null
 if (global.jasmine) {
@@ -112,6 +113,21 @@ function main(args) {
       noSpecsFound = false
     },
   });
+
+  if (JUnitXmlReporter) {
+    const testOutputFile = process.env.XML_OUTPUT_FILE;
+    if (testOutputFile) {
+      jrunner.addReporter(new JUnitXmlReporter({
+        filePrefix: path.basename(testOutputFile),
+        savePath: path.dirname(testOutputFile),
+        consolidate: true,
+        consolidateAll: true
+      }));
+    } else {
+      console.warn('Skipping XML Test Result: $XML_OUTPUT_FILE not found.')
+    }
+  }
+
   // addReporter throws away the default console reporter
   // so we need to add it back
   jrunner.configureDefaultReporter({});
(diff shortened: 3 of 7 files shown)


packages/jasmine/src/index.js
```diff
@@ -10,3 +10,4 @@ exports.boot = boot;
 // re-export jasmine and its transitive dep jasmine-core
 exports.jasmine = require('jasmine');
 exports.jasmineCore = jasmineCore;
+exports.JUnitXmlReporter = require('jasmine-reporters').JUnitXmlReporter;;
```
alexeagle: can you just wrap this with a try/catch to make the reporter optional? Just to make the change less breaking, if we don't find the jasmine-reporters package we could gracefully degrade to the current behavior

examples/user_managed_deps/package.json
```diff
@@ -3,6 +3,7 @@
   "devDependencies": {
     "jasmine": "~3.4.0",
     "jasmine-core": "~3.4.0",
+    "jasmine-reporters": "~2.3.2",
```
alexeagle: the fact that you had to update this spot suggests it's a breaking change - I'd rather just silently not produce the xml output if we don't find jasmine-reporters

alexeagle:
note, we ought to see test-case-level data showing up on the circleci run for this, but when I asked circleci support about it, they said they don't support symlinks in their `store_*` config file args :(

examples/user_managed_deps/package.json
joeljeske: Yup, that sounds good. I had not considered this style of dep management when I initially made the PR and stated that it was not breaking. I updated to remove this change and to make the reporters fail silently

joeljeske:
That’s a bummer for circle ci. I guess they could be copied out from the symlink but not sure if that’d be worth it.

internal/npm_install/test/BUILD.bazel
```diff
@@ -104,6 +104,7 @@ sh_test(
     jasmine_deps = [
         "@fine_grained_deps_%s//jasmine" % pkgmgr,
         "@fine_grained_deps_%s//jasmine-core" % pkgmgr,
+        "@fine_grained_deps_%s//jasmine-reporters" % pkgmgr,
```
alexeagle: curious, is this load-bearing? I assume the test passes whether or not this is included?

packages/jasmine/src/jasmine_runner.js
```diff
@@ -112,6 +113,21 @@ function main(args) {
       noSpecsFound = false
     },
   });
+  
+  if (JUnitXmlReporter) {
```
alexeagle: feels like we should have a markdown edit in here somewhere to let users know that the jasmine_node_test rule participates in Bazel's standard test case recording mechanism

internal/npm_install/test/BUILD.bazel
joeljeske: No, this is not needed. I have removed it in my latest changes

packages/jasmine/src/jasmine_runner.js
joeljeske: I updated the docs to mention the reporter now so users might think to leverage them

joeljeske:
Sorry about that, I have fixed my hooks to work properly now. Let me know if there is anything else!

joeljeske:
@alexeagle I think the build failure was unrelated due to build asset download. Could you restart the failure?

packages/jasmine/src/jasmine_node_test.bzl
```diff
@@ -35,6 +35,14 @@ def jasmine_node_test(
         **kwargs):
     """Runs tests in NodeJS using the Jasmine test runner.
 
+    Detailed XML test results are found in the standard `bazel-testlogs`
+    directory. This may be symlinked in your workspace.
+    See https://docs.bazel.build/versions/master/output_directories.html
+
+    If using self-managed npm dependencies, ensure that the `jasmine-reporters`
```
gregmagolan: nit: this is likely a very rare use case so may not be even worth mentioning as it may lead to some confusion; if we do mention it then `v8-coverage` is also required dep to use `jasmine_node_test` from `index.from_src.bzl`.

joeljeske: Ok sounds good. Shall I remove this note then?

gregmagolan: I think so. Anyone doing self-managed npm deps is going down the less supported path for now anyway.

joeljeske:
Great! Thanks for the help getting this through!

Issue #1661, opened 2020-02-21
Support custom flags passed to rollup


joeljeske:
# 🚀 feature request

### Relevant Rules

`rollup_bundle`

### Description

 A clear and concise description of the problem or missing capability...

It would be useful to allow extra flags to be passed to the rollup binary. Specifically, I want to be able to pass "--silent" to allow bazel to keep a pretty CLI output, but I think it would be useful to allow arbitrary args passed through to support the rest of Rollups API. 

### Describe the solution you'd like

 If you have a solution in mind, please describe it.

Add rule attr 

`args: [ ]`

to the rule `rolllup_bundle`

### Describe alternatives you've considered

 Have you considered any alternative solutions or workarounds?

Overriding the rollup tool attr for rollup bundle to set the proc arg before passing on to rollup. Seems a bit overkill

alexeagle:
Seems reasonable to me. But we already have an attribute corresponding to a bunch of flags, like `sourcemap` - if we introduced `args` would we still keep the sourcemap attribute? Want to be consistent about it

joeljeske:
That’s a good point. I think it would be nice to support override flags but I agree it could be confusing. I think the best experience would be to support every rollup flag as an attr but it could be difficult to keep up to date with rollup. 

Comparing to ts_library, we support some options from tsconfig but others are not respected. Would it be so bad to included a set of override args in addition to the current attrs?

gregmagolan:
I'm leaning towards a generic `args` so that users can override config values from their config file if they want to and also adds support for any args that are not available in config files such as `--silent` and `--environment`.

@josephperrott  checked and rollup will error out if there are duplicate args so if a user would to add `--format` it would fail.

In general, however, that would be up to the underlying tool to either handle duplicate CLI args or error out. Should we in general allow users to pass whatever args they want to tools that Bazel invokes?

joeljeske:
Seems like this is fixed as part of https://github.com/bazelbuild/rules_nodejs/pull/1681

PR #1680, opened 2020-03-02
feat(rollup): add silent attr to rollup_bundle to support --silent flag

josephperrott:
As rollup_bundle already supports specific rollup flags with specific
attributes on the rollup_bundle rule, adding a boolean attribute for
silent makes the most sense.

This boolean determines if the --silent flag is added to the rollup command
when called.

I was unable to add tests for this as its change is to the stdout of the execution
of bazel, not the results of the rollup command.

### DIFF
--- a/packages/rollup/src/rollup_bundle.bzl
+++ b/packages/rollup/src/rollup_bundle.bzl
@@ -169,7 +169,12 @@ Otherwise, the outputs are assumed to be a single file.
         default = "@npm//rollup/bin:rollup",
     ),
     "silent": attr.bool(
-        doc = "Whether to execute the rollup binary with the --silent flag",
+        doc = """Whether to execute the rollup binary with the --silent flag, defaults to False.
+
+Using --silent can cause rollup to [ignore errors/warnings](https://github.com/rollup/rollup/blob/master/docs/999-big-list-of-options.md#onwarn) 
+which are only surfaced via logging.  Since bazel expects printing nother on success, setting silent to True
+is a more Bazel-idiomatic experience, however could cause rollup to drop important warnings.
+""",
     ),
     "sourcemap": attr.string(
         doc = """Whether to produce sourcemaps.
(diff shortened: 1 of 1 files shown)


alexeagle:
Note: related to #1661 which was also for `--silent`.

My take is: we wrap rollup in a custom rule, and attrs like `sourcemap` make sense for Bazel to know about since they affect the output files. So we should continue to mirror the Rollup API in attributes.

This has an unfortunate meaning in rollup - `--silent` also causes builds to pass that would have otherwise failed, because they treat many fatal conditions as a warning and you have to supply a custom handler to find out about these if you use `--silent` https://github.com/rollup/rollup/blob/master/docs/999-big-list-of-options.md#onwarn
so that's why we can't make the default here True, which would make it more Bazel-idiomatic.

packages/rollup/src/rollup_bundle.bzl
```diff
@@ -168,6 +168,9 @@ Otherwise, the outputs are assumed to be a single file.
         cfg = "host",
         default = "@npm//rollup/bin:rollup",
     ),
+    "silent": attr.bool(
```
alexeagle: Could you add a warning in this doc that using silent can cause rollup to ignore problems in your code, and link to https://github.com/rollup/rollup/blob/master/docs/999-big-list-of-options.md#onwarn and maybe https://github.com/rollup/rollup/issues/2686 ?

You could also mention that Bazel expects tools to print nothing on success, so enabling `--silent` gives a more Bazel-idiomatic experience, but the default here is False due to how it also drops warnings.

josephperrott: Done, I added this information to the `attr` docstring.

packages/rollup/src/rollup_bundle.bzl
```diff
@@ -168,6 +168,14 @@ Otherwise, the outputs are assumed to be a single file.
...
+        doc = """Whether to execute the rollup binary with the --silent flag, defaults to False.
+
+Using --silent can cause rollup to [ignore errors/warnings](https://github.com/rollup/rollup/blob/master/docs/999-big-list-of-options.md#onwarn) 
+which are only surfaced via logging.  Since bazel expects printing nother on success, setting silent to True
```
alexeagle: nit: nother->nothing

PR #1688, opened 2020-03-05
docs: add LogiOcean as an adaptor to README.md

qzmfranklin:
LogiOcean is a lead fin-tech company in China focusing on powering up analytics and business intelligence.

We are an early adoptor of Angular + Bazel. We are also mentioned as such by Brad Green on May 3rd 2019 at ng-conf.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,7 @@ Organizations:
 - [Lucidchart](https://www.lucidchart.com)
 - [Webdox](https://www.webdox.cl)
 - [WeMaintain](https://www.wemaintain.com)
+- [LogiOcean](https://www.logiocean.com)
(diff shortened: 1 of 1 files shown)


qzmfranklin:
Not sure what's the proper agenda for this.

Upon request, I can provide evidence to show that we are really using rules_nodejs, Angular, and Bazel very heavily. Not sure what specifics are interesting though.

qzmfranklin:
@gregmagolan I'd believe there is interest in the stuff we've been doing. But there is a lack of integrated Bazel support for those features. That's where we are mostly concerned about.

If there is some stronger interest, I'd be happy to present something.

PR #1695, opened 2020-03-06
feat(labs): introduce a new ts_proto_library with grpc support

mrmeku:
This PR is based off the amazing work from https://github.com/Dig-Doug/rules_typescript_proto

At a high level, @Dig-Doug 's approach to generated a typescript compatible proto_library works as follows:
1) Generate commonjs and .d.ts files using an existing protoc plugin. His repo uses the plugin from `@improbable-eng/grpc-web` while alterations use the official `grpc-web` protoc plugin.
2) Use an aspect to take the output of the plugin and generate es5 and es6 sources by using string manipulations to transform the generated files into AMD / ES6 modules respectively
3) Return the ES5, ES6 and .d.ts files using the same typescript provider as ts_library

In the future, it is likely that the grpc-web protoc plugin will have full typescript file generation support. When that happens we can remove the need for the string manipulations.


WORKSPACE
```diff
@@ -107,6 +107,24 @@ git_repository(
...
+    name = "rules_proto",
+    sha256 = "4d421d51f9ecfe9bf96ab23b55c6f2b809cbaf0eea24952683e397decfbd0dd0",
+    strip_prefix = "rules_proto-f6b8d89b90a7956f6782a4a3609b2f0eee3ce965",
+    urls = [
+        "https://github.com/bazelbuild/rules_proto/archive/f6b8d89b90a7956f6782a4a3609b2f0eee3ce965.tar.gz",
+    ],
+)
+
+http_archive(
+    name = "io_bazel_rules_closure",
```
alexeagle: this is an unsupported ruleset and a very unfortunate dependency. do you understand why it's required? does the runtime depend on closure library? does it depend on `goog.base` for the module loading?

mrmeku: Yeah, its required only so that I can load in the com_google_protobuf workspace. Which I only grab for the protoc plugin. If you feel strongly I can potentially remove it since its not actually a dependency of the protoc plugin

mrmeku:
@alexeagle Just got the prodserver example tests to pass and bundle properly :D

mrmeku:
> FYI @siggisim has been working on proto/grpc under bazel web also
> 
> I think we should have a much lower bar for reviewing things in labs. I appreciate that this is interoperable with `ts_library` providers and the concatjs bundler in karma/ts_devserver.
> OTOH it's more of Bazel being weird. What would it look like if we modeled what the readme for grpc-web tells you to do?

If we modeled what grpc-web tells you to do you would essentially be generating code in place and checking it into your repository. At that point, you would do something like using the generated `.d.ts` files in your `ts_libraries` and then also referencing the generated commonjs modules in your tsdevserver / rollup bundle targets. Certainly doable, but not as clean or automatic. 

Plus whenever you update any proto you would need to regenerate the code and potentially update your consuming targets.

WORKSPACE
```diff
@@ -107,6 +108,15 @@ git_repository(
     shallow_since = "1582757372 -0800",
 )
 
+http_archive(
```
alexeagle: could you move it to package.bzl along with the other devDependencies? (assuming anyone building this workspace from src downstream needs the dep, and for consistency)

examples/protocol_buffers/BUILD.bazel
```diff
@@ -75,35 +84,24 @@ rollup_bundle(
...
+        "@npm//rollup-plugin-node-resolve",
+    ],
 )
 
 terser_minified(
     name = "bundle.min",
     src = ":bundle",
 )
 
-# Needed because the prodserver only loads static files that appear under this
```
alexeagle: nice deletion :)

examples/protocol_buffers/rollup.config.js
```diff
@@ -1,8 +1,9 @@
+const commonjs = require('rollup-plugin-commonjs');
+const nodeRequire = require('rollup-plugin-node-resolve');
+
 module.exports = {
-  // indicate which modules should be treated as external
-  external: [
-    'long',
-    'protobufjs/minimal',
+  plugins: [
+    nodeRequire(),
```
alexeagle: nice simplifications!

internal/node/_node_bin/node
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
```
alexeagle: oops don't check in this file! (we should fix it so we either stop writing it in the srcs or gitignore it)

packages/labs/src/grpc_web/change_import_style.js
```diff
@@ -0,0 +1,148 @@
+/**
+ * Converts a list of generated protobuf-js files from commonjs modules into named AMD modules.
```
alexeagle: I think you got this code or approach from somewhere? could you include a link so we can find that reference later?
(or add a README for us to understand the design)

packages/labs/src/mock_io_bazel_rules_closure.bzl
```diff
@@ -0,0 +1,28 @@
+"Mock transitive toolchain dependencies that are uneeded"
```
alexeagle: nice, should we doc this for users too?

packages/labs/src/grpc_web/change_import_style.js
mrmeku: Done! Added a nice little readme

packages/labs/src/mock_io_bazel_rules_closure.bzl
mrmeku: What would you recommend as far as docs?

examples/protocol_buffers/rollup.config.js
mrmeku: I try :D

mrmeku:
> either in this PR or a separate one we should add you in the codeowners for these directories, as you're clearly the expert now and I hope you'll help maintain it

I'd love to own/maintain this! I'm not exactly sure how to integrate with the with your custom codeowners rule though

mrmeku:
Anything else I can do push this along @alexeagle ?

alexeagle:
Adding the codeowners in https://github.com/bazelbuild/rules_nodejs/pull/1718

mancini0:
Thanks for this rule. Are there any usage examples ? The .proto files in the protobuf example directory  do not define any rpc services, yet I see grpc-web is a dependency in that module's package.json.

Issue #583, opened 2019-02-26
yarn_install logs to stdout, breaks bazel query


JaredNeil:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `yarn_install`

### Is this a regression?

No

### Description

The `yarn_install` repository action outputs logging info to stdout when `quiet=False`.
This breaks using the output from a `bazel query` command if the workspace actions have not run already.

## 🔬 Minimal Reproduction

1. Checkout master (currently 5c5d52fd) on https://github.com/angular/angular-bazel-example
2. Run the following commands:
```
bazel clean --expunge
T1=$(bazel query 'kind(ng_module, //...)')
echo $T1
T2=$(bazel query 'kind(ng_module, //...)')
echo $T2
```
3. Note that the yarn install logging appears in the output of the first query (`T1`).

## Example
```
~/code/angular-bazel-example$ bazel clean --expunge
INFO: Invocation ID: a331d265-45ea-4cc1-9715-df1401ba35d6
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
~/code/angular-bazel-example$ T1=$(bazel query 'kind(ng_module, //...)')
Starting local Bazel server and connecting to it...
INFO: Invocation ID: 97bde3aa-cd64-487d-93dc-2c4c02b1b084
INFO: SHA256 (https://github.com/bazelbuild/bazel-skylib/archive/d7c5518fa061ae18a20d00b14082705d3d2d885d.zip) = bb6f842f7f9ac9d4e5ef9092ba44a8d76520978d0ff3f730f26d7f8aa9343e5d
INFO: SHA256 (https://github.com/bazelbuild/rules_go/releases/download/0.16.3/rules_go-0.16.3.tar.gz) = b7a62250a3a73277ade0ce306d22f122365b513f5402222403e507f2f997d421
Loading: 17 packages loaded
~/code/angular-bazel-example$ echo $T1
yarn install v1.12.1 [1/4] Resolving packages... [2/4] Fetching packages... [3/4] Linking dependencies... [4/4] Building fresh packages... Done in 0.37s. //src:src //src/todos:todos //src/hello-world:hello-world //src/material:material
~/code/angular-bazel-example$ T2=$(bazel query 'kind(ng_module, //...)')
INFO: Invocation ID: daeb26a7-5d68-4657-ada7-b3c94b5239fd
Loading: 0 packages loaded
~/code/angular-bazel-example$ echo $T2
[...]

PR #1700, opened 2020-03-09
Do not print npm install output to stdout

devversion:
When people explicitly enable log output for `yarn_install` or `npm_install`, they don't expect that the installation output interferes with actual `stdout` from Bazel commands. i.e. `bazel query` no longer only prints the query results to `stdout`, but instead also contains NPM/Yarn log output. This means that results of `bazel query` can no longer be easily used in scripts that rely on query results.

### DIFF
--- a/internal/npm_install/npm_install.bzl
+++ b/internal/npm_install/npm_install.bzl
@@ -233,7 +233,7 @@ def _npm_install_impl(repository_ctx):
             content = """#!/usr/bin/env bash
 # Immediately exit if any command fails.
 set -e
-(cd "{root}"; "{npm}" {npm_args})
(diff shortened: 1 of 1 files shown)


DavidANeil:
I believe this is already done at https://github.com/bazelbuild/bazel/pull/10611
I believe it will be included in Bazel 2.3.0
See also:
https://github.com/bazelbuild/rules_nodejs/issues/583
https://github.com/bazelbuild/bazel/issues/8077

devversion:
@DavidANeil Oh great. Thanks for sharing. Closing this one then!

If anyone comes across this: Until this will be fixed, the workaround seems to be adding `args = ["1>&2]` to `npm_install` or `yarn_install`.

Issue #1702, opened 2020-03-12
make @bazel/jasmine jasmine dep a peerDeps


Toxicable:
If a repository wants to change their jasmine version they should be able to do it without waiting for @bazel/jasmine to adjust the version.
Using peerDeps fixes this by forcing the user to install jasmine and @bazel/jasmine will use whatever the user has installed.

This is a breaking change so could be good for 2.0

Toxicable:
Putting it on the 2.0 milestone for consideration

joeljeske:
Couldn’t this also be solved by using a minor patch floating version? Then the user could install a specific version if desired or else default to the latest one specified by bazel/jasmine?

evahowe:
@Toxicable is what @joeljeske saying true?  Or no?

alexeagle:
@Toxicable ping!

alexeagle:
Fixed by #1703

PR #1703, opened 2020-03-12
feat(jasmine): make deps peerDeps

Toxicable:
addresses https://github.com/bazelbuild/rules_nodejs/issues/1702

### DIFF
--- a/examples/kotlin/package.json
+++ b/examples/kotlin/package.json
@@ -7,6 +7,7 @@
         "@bazel/terser": "^1.6.0",
         "domino": "^2.1.3",
         "http-server": "^0.11.1",
+        "jasmine": "^3.5.0",
         "kotlin": "^1.3.41",
         "kotlinx-html-js": "^0.6.4",
         "rollup": "1.20.3",
--- a/e2e/bazel_managed_deps/package.json
+++ b/e2e/bazel_managed_deps/package.json
@@ -2,6 +2,7 @@
   "description": "runtime dependencies for bazel_managed_deps example",
   "devDependencies": {
     "@bazel/jasmine": "^1.6.0",
+    "jasmine": "^3.5.0",
     "typescript": "^3.0.1"
   },
   "scripts": {
--- a/e2e/bazel_managed_deps/yarn.lock
+++ b/e2e/bazel_managed_deps/yarn.lock
@@ -291,7 +291,7 @@ jasmine-reporters@~2.3.2:
     mkdirp "^0.5.1"
     xmldom "^0.1.22"
 
-jasmine@~3.5.0:
+jasmine@^3.5.0, jasmine@~3.5.0:
   version "3.5.0"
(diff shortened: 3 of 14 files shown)


alexeagle:
As you see it's a breaking change - anyone using `@bazel/jasmine` (including our own examples) needs to install the peerDeps too.
Also do we really mean to warn users if they are missing any of these four deps? some of them look like dependencies that our code has, rather than shared dependencies with users code. I would guess that jasmine-reporters and v8-coverage should remain `deps`

Toxicable:
Made just jasmine and jasmine-core a peerDep
added a breaking change message:
```
feat(jasmine): make jasmine a peerDep

BREAKING CHANGE
jasmine and jasmine-core are now peerDeps of @bazel/jasmine
please add them to your repository with:
yarn add jasmine jasmine-core
```

examples/kotlin/package.json
```diff
@@ -7,6 +7,8 @@
         "@bazel/terser": "^1.5.0",
         "domino": "^2.1.3",
         "http-server": "^0.11.1",
+        "jasmine": "^3.5.0",
+        "jasmine-core": "^3.5.0",
```
alexeagle, line 11: doesn't make sense to me why end users need to know about the `jasmine-core` package. If it isn't in jasmine's docs, you shouldn't need under Bazel either.

packages/jasmine/src/package.json
```diff
@@ -17,11 +17,13 @@
...
     "main": "index.js",
     "dependencies": {
-        "jasmine": "~3.5.0",
-        "jasmine-core": "~3.5.0",
         "jasmine-reporters": "~2.3.2",
         "v8-coverage": "1.0.9"
     },
+    "peerDependencies": {
+        "jasmine": "~3.5.0",
+        "jasmine-core": "~3.5.0"
```
alexeagle, line 25: I don't think we need to list this one since `jasmine` depends on `jasmine-core`

Toxicable: unfortunately since in jasmine_runner we require `jasmine-core` it must be specified as a dep, even though it's a transitive dep of `jasmine`, if it's not specified has a `dependencies` and `peerDependencies` then yarn wont hoist it to the top.
You can see the result of this in the CI failures
```
==================== Test output for //:test:
Error: Cannot find module 'jasmine-core'. Please verify that the package.json has a valid "main" entry
    at Function.module.constructor._resolveFilename (test_require_patch.js:482:17)
    at Function.Module._load (internal/modules/cjs/loader.js:687:27)
    at Module.require (internal/modules/cjs/loader.js:849:19)
```

Toxicable: So either we figure out how to remove the require tto `jasmine-core` or we have to put `jasmine-core` in our `peerDependencies`

alexeagle: got it, thanks for the explanation. we should add that as a comment here.

PR #1705, opened 2020-03-12
fix(builtin): include package.json files with DeclarationInfo

### LINKED ISSUE #1704: generated npm targets don't produce needed declarations for typescript
With latest rules_nodejs, just
`npm i -D @types/jest`
and query for what files you get from the __typings target
```
% bazel query 'deps("@npm//@types/jest:jest__typings")'
@npm//@types/jest:jest__typings
@npm//:node_modules/@types/jest/index.d.ts
```

`@types/jest` has
```
"dependencies": {
    "jest-diff": "^25.1.0",
    "pretty-format": "^25.1.0"
  },
```
so we should have followed those.

As a result, you can run `tsc` with the right `@types/jest` dependency but get 
```
node_modules/@types/jest/index.d.ts(485,51): error TS2307: Cannot find module 'jest-diff'.
node_modules/@types/jest/index.d.ts(539,44): error TS2307: Cannot find module 'pretty-format'.
```

alexeagle:
Fixes #1704

FIXME: needs some regression test

### DIFF
--- a/internal/npm_install/node_module_library.bzl
+++ b/internal/npm_install/node_module_library.bzl
@@ -26,7 +26,7 @@ def _node_module_library_impl(ctx):
     declarations = depset([
         f
         for f in ctx.files.srcs
-        if f.path.endswith(".d.ts") and
+        if (f.path.endswith(".d.ts") or f.basename == "package.json") and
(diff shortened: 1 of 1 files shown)


alexeagle:
Not surprising that this is a breaking change.
I think the answer is that if you want to read a package.json file, you should use an NpmPackageInfo instead of a DeclarationInfo.

PR #1709, opened 2020-03-15
chore(examples): upgrade html-insert-assets to 0.6.0

jbedard:
@alexeagle we discussed this on slack a bit and I went with a change to make paths relative to the html file: https://github.com/jbedard/html-insert-assets/commit/26d7118eaa73f0a3ee2b9bbcbcdf3388b298b182

This seems to work well. The only real difference in all the example apps is the css/js files are relative (`./` instead of `/`). This way the app can be placed in a subdirectory and a `<base href>` or just plain relative paths will work correctly.

FYI @gregmagolan

### DIFF
--- a/examples/app/yarn.lock
+++ b/examples/app/yarn.lock
@@ -584,12 +584,12 @@ he@^1.1.1:
   resolved "https://registry.yarnpkg.com/he/-/he-1.2.0.tgz#84ae65fa7eafb165fddb61566ae14baf05664f0f"
   integrity sha512-F/1DnUGPopORZi0ni+CvrCgHQ5FyEAHRLSApuYWMmrbSwoN2Mn/7k+Gl38gJnR7yyDZk6WLXwiGod1JOWNDKGw==
 
-html-insert-assets@^0.5.0:
-  version "0.5.0"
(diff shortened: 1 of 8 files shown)


examples/app/yarn.lock
```diff
@@ -844,13 +844,18 @@ minimist@~0.0.1:
...
 
-mkdirp@0.5.x, mkdirp@^0.5.0, mkdirp@^0.5.1, mkdirp@~0.5.x:
+mkdirp@0.5.x, mkdirp@^0.5.0, mkdirp@~0.5.x:
   version "0.5.1"
   resolved "https://registry.yarnpkg.com/mkdirp/-/mkdirp-0.5.1.tgz#30057438eac6cf7f8c4767f38648d6697d75c903"
   integrity sha1-MAV0OOrGz3+MR2fzhkjWaX11yQM=
   dependencies:
     minimist "0.0.8"
 
+mkdirp@^1.0.3:
```
alexeagle: nit: you could probably drop this dependency, mkdirp is a five-line implementation and maybe not worth the mechanics of keeping dependencies up-to-date

jbedard: I can look into that in the future...

PR #1710, opened 2020-03-16
feat(typescript): add ts_project rule

alexeagle:
This is a very thin layer on top of vanilla tsc


examples/react_webpack/BUILD.bazel
```diff
@@ -13,22 +13,14 @@ sass(
...
-        "react",
-    ],
-    data = [
+ts_project(
+    name = "tsconfig",
+    srcs = [
         "index.tsx",
         "types.d.ts",
+    ],
+    declaration = False,
```
mrmeku: Can you a comment describing why declaration is set to false?

packages/typescript/src/internal/ts_project.bzl
```diff
@@ -0,0 +1,192 @@
...
+    # ts, tsx, js, jsx, json
+    "srcs": attr.label_list(allow_files = True),
+    "extends": attr.label_list(allow_files = [".json"]),
+    "tsc": attr.label(default = Label("@npm//typescript/bin:tsc"), executable = True, cfg = "host"),
+    "tsconfig": attr.label(mandatory = True, allow_single_file = [".json"]),
+    "deps": attr.label_list(providers = [DeclarationInfo]),
+}
+
+_OUTPUTS = {
+    "js_outs": attr.output_list(),
```
mrmeku: Could you maybe address the difference in outputs of ts_projecrt and ts_library in a comment or readme file somewhere in the PR?

packages/typescript/src/internal/ts_project.bzl
```diff
@@ -0,0 +1,192 @@
...
+    run_node(
+        ctx,
+        inputs = inputs,
+        arguments = [arguments],
+        outputs = outputs,
+        executable = "tsc",
+        progress_message = "Compiling TypeScript project %s" % ctx.file.tsconfig.short_path,
+    )
+
+    runtime_files = depset(ctx.outputs.js_outs + ctx.outputs.map_outs)
```
mrmeku: I'm not sure if this is important by if a src file is a `.js` dependency might it also need to be included as a runtime file?

Toxicable: If you have .js inputs, does tsc write them to the our dir? or does it only emit diagnostics?

packages/typescript/src/internal/ts_project.bzl
```diff
@@ -0,0 +1,192 @@
...
+            Instructs Bazel to expect a `.js.map` output for each `.ts` source.
+        declaration_map: if the `declarationMap` bit is set in the tsconfig.
+            Instructs Bazel to expect a `.d.ts.map` output for each `.ts` source.
+        emit_declaration_only: if the `emitDeclarationOnly` bit is set in the tsconfig.
+            Instructs Bazel *not* to expect `.js` outputs for `.ts` sources.
+    """
+
+    if srcs == None:
+        srcs = native.glob(["**/*.ts"])
+    tsconfig = name + ".json"
```
Toxicable: Not sure if im a fan of this.
Could we possibly change the logic to default to `tsconfig = ctx.testonly ? 'tsconfig-test.json' : 'tsconfig.json'`.
But override with the tsconfig speciffied by the user if exists.
Usually we'd have libs where their "primary target" is a `ts_library` and this will use the same name as the dir, making it easier to reference.
This here prevents using a name that's different from your tsconfig, right?
Is there another reason why this might be a good idea?

packages/typescript/src/internal/ts_project.bzl
```diff
@@ -0,0 +1,192 @@
...
+
+def ts_project_macro(
+        name = "tsconfig",
+        srcs = None,
+        deps = [],
+        extends = None,
+        declaration = True,
+        source_map = False,
+        declaration_map = False,
+        emit_declaration_only = False,
```
Toxicable: maybe instead have a flag that says `emit_js = True`? then this can be derived from the inverse of that flag?

packages/typescript/src/internal/ts_project.bzl
```diff
@@ -0,0 +1,192 @@
+"ts_project rule"
+
+load("@build_bazel_rules_nodejs//:providers.bzl", "DeclarationInfo", "NpmPackageInfo", "run_node")
+
+_ATTRS = {
+    # TODO: limit to the extensions we expect?
+    # ts, tsx, js, jsx, json
+    "srcs": attr.label_list(allow_files = True),
+    "extends": attr.label_list(allow_files = [".json"]),
+    "tsc": attr.label(default = Label("@npm//typescript/bin:tsc"), executable = True, cfg = "host"),
```
Toxicable: i'd love to see if this works by replacing this with `ngtsc`

examples/react_webpack/BUILD.bazel
Toxicable: Looks like it's a demo showing a 1:1 with the old `tsc`  which didn't do declarations

alexeagle: added a comment, yes this is exercising the codepath where tsc wasn't configured with `declaration` or `composite`

packages/typescript/src/internal/ts_project.bzl
alexeagle: Ultimately I think that needs to go in packages/typescript/docs/install.md where there's currently a discussion of `ts_library` vs `@npm//typescript:index.bzl#tsc`

I'll add a comment here about just mirroring what tsc can produce

packages/typescript/src/internal/ts_project.bzl
alexeagle: yeah this needs a test case for .js inputs...

packages/typescript/src/internal/ts_project.bzl
alexeagle: and a TODO about updating the markdown docs after this lands

packages/typescript/src/internal/ts_project.bzl
alexeagle: I'll add a comment in the code explaining this.

```
Bazel's `name` property is arbitrarily controlled by the user. But here we constrain it to match the tsconfig file.
The reason for doing this is so that BUILD file generation can be trivial. If you know the path of a referenced tsconfig file, you know what label to use to include it in the deps of another rule.
If we allowed users to control it, then BUILD file generation would need heuristic semantics to do bazel query and try to figure out which label gives a target that is usable as a dep.
```

And on the rule docs
```
You can use a Bazel alias() to give an additional label for this typescript project compilation
```

alexeagle: Also note you could import the raw `ts_project` rule from the `internal` path, which I think we should leave undocumented, but it gives you the escape valve to avoid this macro logic.

packages/typescript/src/internal/ts_project.bzl
alexeagle: I'm trying to keep these flags named identically to the tsconfig file so the mapping between the two is totally obvious.

packages/typescript/src/internal/ts_project.bzl
alexeagle: yeah I played with it briefly and something didn't work, please give that a try if you have time :)

packages/typescript/src/internal/ts_project.bzl
Toxicable: Ah right, I wasn't aware of a flag in the tsconfig of the same name, sounds good!

packages/typescript/src/internal/ts_project.bzl
Toxicable: Ah yeah I see the constraint here, sounds good to me with the extra docs and code gen considered

examples/react_webpack/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+    "compilerOptions": {
+        "jsx": "react",
+        "lib": ["ES2015", "DOM"],
+        "listEmittedFiles": true,
+        "traceResolution": true
```
Toxicable: Do these two options need to be enabled?

packages/typescript/src/internal/ts_project.bzl
```diff
@@ -0,0 +1,225 @@
...
+#    cd C:/b/uuxnwop3/execroot/build_bazel_rules_nodejs
+#    SET COMPILATION_MODE=fastbuild
+#    bazel-out/host/bin/external/npm/typescript/bin/tsc.bat -p packages/typescript/test/ts_project/b/tsconfig.json --outDir bazel-out/x64_windows-fastbuild/bin/packages/typescript/test/ts_project/b --declarationDir bazel-out/x64_windows-fastbuild/bin/packages/typescript/test/ts_project/b --bazel_node_modules_manifest=bazel-out/x64_windows-fastbuild/bin/packages/typescript/test/ts_project/b/_tsconfig.module_mappings.json --nobazel_patch_module_resolver
+#  Execution platform: @local_config_platform//:host
+#  packages/typescript/test/ts_project/b/b.ts(1,17): error TS6305: Output file 'C:/b/uuxnwop3/execroot/build_bazel_rules_nodejs/packages/typescript/test/ts_project/a/a.d.ts' has not been built from source file 'C:/b/uuxnwop3/execroot/build_bazel_rules_nodejs/packages/typescript/test/ts_project/a/a.ts'.
+TS_37378_DISABLED = ["fix-windows"]
+
+_ATTRS = {
+    # TODO: limit to the extensions we expect? ts, tsx, js, jsx, json
+    "srcs": attr.label_list(allow_files = True, mandatory = True),
```
Toxicable: I don't think we should limit the extensions here, especically if you can swap out the compiler for `ngtsc` for example, since the inputs would then include .html and .css

examples/react_webpack/tsconfig.json
alexeagle: no, these are for debugging.
I think maybe the rule implementation could enable these when running with VERBOSE_LOGS in the environment...

packages/typescript/src/internal/ts_project.bzl
alexeagle: yeah I think even tsc keeps adding ability to natively type-check against require('./funny.thing')

packages/typescript/src/internal/ts_project.bzl
```diff
@@ -0,0 +1,231 @@
...
+            "--listFiles",
+            # Did tsc write all outputs to the place we expect to find them?
+            "--listEmittedFiles",
+            # Why did module resolution fail?
+            "--traceResolution",
+            # Why was the build slow?
+            "--diagnostics",
+            "--extendedDiagnostics",
+            # Prefer to show gory details rather than save space in console
+            "--noErrorTruncation",
```
evmar: We recently discovered that tsc can OOM when you set this flag, because without error truncation error generation can infinite loop.

alexeagle: good tip, thanks. I only set this here from reading through the tsc docs and it sounded helpful, will remove

evmar: Oh wait they fixed it by actually limiting it with the flag, heh https://github.com/microsoft/TypeScript/pull/37461

packages/typescript/src/internal/ts_project.bzl
alexeagle: Discussed with @gregmagolan - even if users can control `tsconfig` and `name` attributes independently, it's still pretty easy for the generator to print all `ts_project` rules in the referenced package to find the right one. So `tsconfig` is now independent (though we recommend naming it the basename of the tsconfig.json and that default is convenient)

examples/react_webpack/BUILD.bazel
```diff
@@ -13,27 +13,29 @@ sass(
...
+    deps = [
         "@npm//@types",
         "@npm//csstype",
     ],
 )
 
+# If you don't want to write //:tsconfig as the label for the TypeScript compilation,
+# use an alias like this, so you can write //:compile_ts instead.
+alias(
+    name = "compile_ts",
```
gregmagolan: nit: use tsconfig attribute and use "compile" as ts_project name to avoid alias?

packages/typescript/src/internal/ts_project.bzl
```diff
@@ -0,0 +1,240 @@
+"ts_project rule"
+
+load("@build_bazel_rules_nodejs//:providers.bzl", "DeclarationInfo", "NpmPackageInfo", "run_node")
+
+_ATTRS = {
```
gregmagolan: I think we should add "args" in general to rules to allow users to pass their own arguments as
1) config files do not always support all arguments. I recently added "args" to rollup for this reason so 
Angular could pass `--silent` which is command line one.
2) users may want to override config file settings with command line arguments in some cases
General pattern is
```
    "args": attr.string_list(
        doc = """Command line arguments to pass to <tool name>. Can be used to override config file settings.
```
and 
```
```
    # See CLI documentation at https://rollupjs.org/guide/en/#command-line-reference
    args = ctx.actions.args()

    # Add user specified arguments *before* rule supplied arguments
    args.add_all(ctx.attr.args)
```
where `args` is passed to `run_node`

File: packages/typescript/src/internal/ts_project.bzl [hunk 12]
```diff
@@ -0,0 +1,240 @@
...
+    # NB: no restriction on extensions here, because tsc sometimes adds type-check support
+    # for more file kinds (like require('some.json')) and also
+    # if you swap out the `compiler` attribute (like with ngtsc)
+    # that compiler might allow more sources than tsc does.
+    "srcs": attr.label_list(allow_files = True, mandatory = True),
+    "extends": attr.label_list(allow_files = [".json"]),
+    "tsc": attr.label(default = Label("@npm//typescript/bin:tsc"), executable = True, cfg = "host"),
```
gregmagolan: nit: set DEFAULT_COMPILER = "@npm//typescript/bin:tsc" as the string is in two places

examples/react_webpack/BUILD.bazel
alexeagle: oops, this BUILD file should have been updated for the change of `declaration` default to False too. will boil it down to the minimum

packages/typescript/src/internal/ts_project.bzl
```diff
@@ -0,0 +1,273 @@
...
+
+            Must include any tsconfig files "chained" by extends clauses.
+
+        tsc: Label of the TypeScript compiler binary to run.
+
+            Override this if your npm_install or yarn_install isn't named "npm"
+            For example, `tsc = "@my_deps//typescript/bin:tsc"`
+            Or you can pass a custom compiler binary instead.
+
+        declaration: if the `declaration` bit is set in the tsconfig.
```
gregmagolan: nit: links to typescript docs for these 1:1 mappings for more info?

packages/typescript/src/internal/ts_project.bzl
```diff
@@ -0,0 +1,273 @@
...
+
+    return [
+        DeclarationInfo(
+            declarations = depset(typings_files),
+            transitive_declarations = depset(typings_files, transitive = [
+                dep[DeclarationInfo].transitive_declarations
+                for dep in ctx.attr.deps
+            ]),
+        ),
+        DefaultInfo(
```
gregmagolan: nit: short comment on rationale for only including runtime files in DefaultInfo?

packages/typescript/src/internal/ts_project.bzl
```diff
@@ -0,0 +1,273 @@
...
+
+    deps_depsets = []
+    for dep in ctx.attr.deps:
+        if _TsConfigInfo in dep:
+            deps_depsets.append(dep[_TsConfigInfo].tsconfigs)
+        if NpmPackageInfo in dep:
+            # TODO: we could maybe filter these to be tsconfig.json or *.d.ts only
+            # we don't expect tsc wants to read any other files from npm packages.
+            deps_depsets.append(dep[NpmPackageInfo].sources)
+        elif DeclarationInfo in dep:
```
gregmagolan: should this just be an `if`? They seem orthogonal and should overlap anyway.

packages/typescript/src/internal/ts_project.bzl
```diff
@@ -0,0 +1,273 @@
...
+
+    inputs = ctx.files.srcs + depset(transitive = deps_depsets).to_list() + [ctx.file.tsconfig]
+    if ctx.attr.extends:
+        inputs.extend(ctx.files.extends)
+    outputs = ctx.outputs.js_outs + ctx.outputs.map_outs + ctx.outputs.typings_outs + ctx.outputs.typing_maps_outs
+    if ctx.outputs.buildinfo_out:
+        outputs.append(ctx.outputs.buildinfo_out)
+
+    if len(outputs) == 0:
+        return []
```
gregmagolan: should still return transitive declarations & tsconfigs?

packages/typescript/src/internal/ts_project.bzl
alexeagle: added `args` pass-through with a delightful test coverage

packages/typescript/src/internal/ts_project.bzl
alexeagle: added above

packages/typescript/src/internal/ts_project.bzl
alexeagle: # DefaultInfo is what you see on the command-line for a built library,
        # and determines what files are used by a simple non-provider-aware
        # downstream library.
        # Only the JavaScript outputs are intended for use in non-TS-aware
        # dependents.

packages/typescript/src/internal/ts_project.bzl
alexeagle: hmm, I think this is true. Need to verify if we have coverage for it

packages/typescript/src/internal/ts_project.bzl
alexeagle: yes, took a while to add a test case to expose this...

packages/typescript/src/internal/ts_project.bzl
gregmagolan: 👍

PR #1715, opened 2020-03-22
feat(builtin): add new LinkablePackageInfo to pkg_npm, js_library & ts_library

gregmagolan:
Linker no longer uses module_name but a LinkablePackageInfo provider instead.

Some gymnastics necessary to support existing ts_library module_name behavior with the current linker.


gregmagolan:
Waiting on:
https://github.com/bazelbuild/rules_nodejs/pull/1736
https://github.com/bazelbuild/rules_nodejs/pull/1737
https://github.com/bazelbuild/rules_nodejs/pull/1738
https://github.com/bazelbuild/rules_nodejs/pull/1739

gregmagolan:
All pre-factors merged. Ready for final review.

Toxicable:
Coudl you briefly explain the `bin` to `execroot` changes?

internal/js_library/js_library.bzl
```diff
@@ -46,23 +48,66 @@ def write_amd_names_shim(actions, amd_names_shim, targets):
...
+_js_library = rule(
+    implementation = _impl,
     attrs = {
-        "srcs": attr.label_list(allow_files = [".js"]),
+        "package_name": attr.string(),
+        "srcs": attr.label_list(allow_files = True),
         "amd_names": attr.string_dict(doc = _AMD_NAMES_DOC),
-        "module_from_src": attr.bool(),
-        # Used to determine module mappings
+        # module_name for legagy ts_library module_mapping support
```
alexeagle: nit: spelling "legacy"

internal/js_library/js_library.bzl
```diff
@@ -46,23 +48,66 @@ def write_amd_names_shim(actions, amd_names_shim, targets):
                 amd_names_shim_content += "define(\"%s\", function() { return %s });\n" % n
     actions.write(amd_names_shim, amd_names_shim_content)
 
-def _js_library(ctx):
-    return [
+def _impl(ctx):
+    if not ctx.files.srcs:
+        fail("No srcs specified")
```
alexeagle: could it just be `mandatory = True` ?

gregmagolan:
> Coudl you briefly explain the `bin` to `execroot` changes?

Sure. Those are implementation details so didn't feel necessary to docstring it too much. But in short, the old linker code always had the path as a `manifest` path (ie. wksp/path/to/file) and the root type the told the linker whether it was in the source tree (`src`) or in the output tree (`bin`). While working on this I realized that these are actually both paths in the `execroot` so the difference can be expressed by just using the full execroot path which includes `bazel-out` if it is in the output tree.

So the only two types of linker roots are now `execroot` and `runfiles`... the latter being the special case of runtime linker deps from rules of type `*_test` or `*_binary`:
```
                if ctx.rule.kind.endswith("_binary") or ctx.rule.kind.endswith("_test"):
                    v = ["runfiles", v[1]]
```

`execroot` links roots will still win over `runfiles` link roots of the same path.

internal/js_library/js_library.bzl
```diff
@@ -46,23 +48,66 @@ def write_amd_names_shim(actions, amd_names_shim, targets):
...
+
+    source_files = ctx.files.srcs[0].is_source
```
alexeagle: nit: source_files is a surprising name for type bool - maybe `all_are_input_artifacts` or sth

internal/js_library/js_library.bzl
```diff
@@ -46,23 +48,66 @@ def write_amd_names_shim(actions, amd_names_shim, targets):
...
-        "module_from_src": attr.bool(),
-        # Used to determine module mappings
+        # module_name for legagy ts_library module_mapping support
+        # TODO: remove once legacy module_mapping is removed
         "module_name": attr.string(),
-        "module_root": attr.string(),
     },
 )
+
+def js_library(
```
alexeagle: TODO to remove the macro after breaking change rolls?

internal/linker/link_node_modules.bzl
```diff
@@ -36,16 +37,21 @@ def add_arg(args, arg):
 
 def _link_mapping(label, mappings, k, v):
     if k in mappings and mappings[k] != v:
+        # Allow all other mappings to win over legacy "_tslibrary"
```
alexeagle: why?

internal/linker/link_node_modules.bzl
```diff
@@ -61,7 +67,7 @@ def write_node_modules_manifest(ctx, extra_data = []):
         # We always map the workspace to itself to support absolute require like
         # import from 'my_wksp/path/to/file'
         # and it's always a bin dependency, TODO: let user control via package.json
```
alexeagle: "always a bin dependency" needs update

alexeagle:
Design doc https://hackmd.io/9gqAG5WdSCazzNWyn5OSRw
(maybe want that in the commit msg so we can find it later)

internal/providers/linkable_package_info.bzl
```diff
@@ -0,0 +1,26 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""A docstring
+"""
+
+LinkablePackageInfo = provider(
+    doc = "A doc string",
+    fields = {
+        "package_name": "The package name",
```
alexeagle: + "should be the same as name field in package.json"

internal/providers/linkable_package_info.bzl
```diff
@@ -0,0 +1,26 @@
...
+        "files": "Depset of files in this package (must all be contained within path)",
+        "path": "The path to link to",
```
alexeagle: a bit more about what format path - and/or examples of what path should look like

internal/js_library/js_library.bzl
gregmagolan: 👍

internal/linker/link_node_modules.bzl
gregmagolan: Oh. Such a long story. I can get rid of it as a breaking change by adding package_name to ts_library for linking and reserving module_name for the AMD name in the UMD/AMD output.

I left a paragraph comment in the code about tthis.
```
    # Special case for ts_library module_name for legacy behavior and for AMD name work-around
    # Tag the mapping as "_tslibrary" so it can be overridden by any other mapping if found.
    #
    # In short, ts_library module_name attribute results in both setting the AMD name (which
    # desired and necessary in devmode which outputs UMD) and in making a linkable mapping. Because
    # of this, you can get in the situation where a ts_library module_name and a downstream pkg_name
    # package_name conflict and result in duplicate mappings. This work-around will make this
    # situation work however it is not a recommended pattern since a ts_library can be a dep of a
    # pkg_npm but not vice-verse at the moment since ts_library cannot handle directory artifacts as
    # deps.
```

internal/linker/link_node_modules.ts
```diff
@@ -111,16 +111,15 @@ async function resolveRoot(root: string|undefined, runfiles: Runfiles) {
   const fromManifest = runfiles.lookupDirectory(root);
   if (fromManifest) return fromManifest;
 
-  // Account for Bazel --legacy_external_runfiles
-  // which look like 'my_wksp/external/npm/node_modules'
-  if (await exists(path.join('external', root))) {
-    log_verbose('found legacy_external_runfiles, switching root to', path.join('external', root));
+  if (runfiles.execroot) {
```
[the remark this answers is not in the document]
gregmagolan: No. I think the two are orthogonal. Even with `legacy_external_runfiles` you get both external workspaces accessible via `../external_wksp` and via `./external/external_wksp`. This code just makes a distinction between being in the execroot for build rules where `./external/external_wksp` is the only resolution available and being in runfiles where `../external_wksp` is the preferred resolution since `./external/external_wksp` is only available with `legacy_external_runfiles`

internal/node/test/lib1/BUILD.bazel
```diff
@@ -4,8 +4,10 @@ package(default_visibility = ["//internal:__subpackages__"])
 
 js_library(
     name = "lib1",
-    srcs = glob(["**/*.js"]),
-    module_name = "lib1",
-    # Don't allow deep imports under here
-    module_root = "src/index.d.ts",
```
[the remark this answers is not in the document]
gregmagolan: Yup. Not a desirable feature for the internal js_library. We can now throw a package.json in there instead.

internal/providers/linkable_package_info.bzl
gregmagolan: 👍

internal/providers/linkable_package_info.bzl
gregmagolan: 👍

internal/pkg_npm/pkg_npm.bzl
```diff
@@ -77,6 +77,9 @@ You can pass arguments to npm by escaping them from Bazel using a double-hyphen,
 """
 
 PKG_NPM_ATTRS = {
+    "package_name": attr.string(
+        doc = """Optional package_name that this npm package may be imported as.""",
```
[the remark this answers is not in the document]
gregmagolan: yes. I was already thinking of a validator on a follow up PR. one that can be turned off as there are use cases where you may want to make linkable pkg_npm's with different linkable names from the package.json for testing or some other corner cases

internal/linker/link_node_modules.bzl
gregmagolan: 👍

internal/js_library/js_library.bzl
gregmagolan: 👍

internal/js_library/js_library.bzl
gregmagolan: 👍

internal/js_library/js_library.bzl
gregmagolan: not quite as you can still specify `srcs = []` which satisfies mandatory. but I can simplify with
```
is_all_sources = ctx.files.srcs and ctx.files.srcs[0].is_source
```
instead

internal/pkg_npm/pkg_npm.bzl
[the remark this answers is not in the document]
gregmagolan: It looks like the npm package name restrictions are pretty tight: https://github.com/npm/validate-npm-package-name. we should be good. also, we'd have run into it already if there were packages that were not valid labels as `@npm//@scope/pkg_name` and `@npm//pkg_name` would be issues.

third_party/github.com/gjtorikian/isBinaryFile/BUILD.bazel
```diff
@@ -10,12 +7,12 @@ licenses(["notice"])
...
 
 js_library(
-    # ts_library(
     name = "isBinaryFile",
-    srcs = ["src/index.js"],  # src/index.ts
-    module_name = "isbinaryfile",
-    module_root = "src/index",
-    #     deps = ["@npm//@types/node"],
+    srcs = [
+        "src/index.js",
```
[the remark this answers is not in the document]
gregmagolan: I'll try ts_project and put the resulting js file into the release pkg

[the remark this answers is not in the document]
gregmagolan: done

third_party/github.com/gjtorikian/isBinaryFile/BUILD.bazel
```diff
@@ -1,21 +1,19 @@
...
+        "BUILD.bazel",
+        "LICENSE",
+        "//third_party/github.com/gjtorikian/isBinaryFile/src:package_contents",
+    ]
 )
 
-filegroup(
-    name = "package_contents",
-    srcs = glob(["*"]) + [":isBinaryFile"],
+alias(
```
alexeagle: now that the ts_project rule is fixed I guess you can remove that nested BUILD file?

gregmagolan: 👍

internal/linker/link_node_modules.bzl
```diff
@@ -59,9 +65,11 @@ def write_node_modules_manifest(ctx, extra_data = []):
 
     mappings = {
         # We always map the workspace to itself to support absolute require like
-        # import from 'my_wksp/path/to/file'
-        # and it's always a bin dependency, TODO: let user control via package.json
-        ctx.workspace_name: ["bin", ctx.workspace_name],
+        # import from 'my_wksp/path/to/file' and it's always links to the bin_dir.
+        # If source files need to be required then they can be copied to the bin_dir
+        # with copy_to_bin.
+        # TODO: let user control the link path via package.json.
```
alexeagle: what would this look like? seems like the path is always the directory containing package.json?

third_party/github.com/bazelbuild/bazel-skylib/rules/private/copy_file_private.bzl
```diff
@@ -19,13 +19,28 @@ cmd.exe (on Windows). '_copy_xfile' marks the resulting file executable,
 '_copy_file' does not.
 """
 
+def _hash_file(file):
+    # Strip the minus off of hash if it is negative.
```
alexeagle: why is it a problem to have an extra hyphen appear in the generated filename?

gregmagolan: its not. I'm just pedantic. I'll remove the condition it 🙄

internal/linker/link_node_modules.bzl
gregmagolan: I don't know. That TODO was there from the original linker impl. Perhaps it is referring to the the root package.json to control how to root workspace is linked. Seems out of date at this point so I'll remove the TODO.

internal/js_library/js_library.bzl
```diff
@@ -12,11 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""js_library allows defining a set of javascript sources and assigning a module_name and module_root.
+"""js_library allows defining a set of javascript sources and assigning a package_name.
 
 DO NOT USE - this is not fully designed, and exists only to enable testing within this repo.
 """
 
+load("//internal/providers:linkable_package_info.bzl", "LinkablePackageInfo")
```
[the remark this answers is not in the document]
gregmagolan: 👍

PR #1722, opened 2020-03-23
Validate ts project

alexeagle:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/typescript/src/internal/ts_project_options_validator.ts
+++ b/packages/typescript/src/internal/ts_project_options_validator.ts
@@ -1,0 +1,83 @@
+import * as ts from 'typescript';
+
+const diagnosticsHost: ts.FormatDiagnosticsHost = {
+  getCurrentDirectory: () => ts.sys.getCurrentDirectory(),
+  getNewLine: () => ts.sys.newLine,
+  // Print filenames including their relativeRoot, so they can be located on
+  // disk
+  getCanonicalFileName: (f: string) => f
+};
+
+function main([tsconfigPath, output, target, attrsStr]: string[]): 0|1 {
+  // The Bazel ts_project attributes were json-encoded
+  // (on Windows the quotes seem to be quoted wrong, so replace backslash with quotes :shrug:)
+  const attrs = JSON.parse(attrsStr.replace(/\\/g, '"'));
+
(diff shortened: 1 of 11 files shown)


packages/typescript/src/internal/ts_project_options_validator.ts
```diff
@@ -0,0 +1,83 @@
...
+  check('incremental');
+
+  if (failures.length > 0) {
+    console.error(`ERROR: ts_project rule ${
+        target} was configured with attributes that don't match the tsconfig`);
+    failures.forEach(f => console.error(' - ' + f));
+    console.error('You can automatically fix this by running:');
+    console.error(`    npx buildozer ${buildozerCmds.map(c => `'${c}'`).join(' ')} ${target}`);
+    console.error('Or to suppress this error, run:');
+    console.error(`    npx buildozer 'set validate False' ${target}`);
```
gregmagolan: nit: user may not have buildozer? also add simpler manual instructions?

gregmagolan:
Looking this PR. another feature to add to the list to make it this feature compatible to ts_library is the `expected_diagnostics` attribute.

packages/typescript/src/internal/ts_project_options_validator.ts
alexeagle: Even worse: https://www.npmjs.com/package/buildozer
I think I'd better make this `npx @bazel/buildozer`
(note, that works even if you don't have it installed, so I think it's fine to keep the error message short since it should work in all cases)

alexeagle:
expected_diagnostics is annoying since you have to regex anyway. plus the kind of failure we want to test here is the action before tsc. Could we have a generic feature for run_node or npm_package_bin for expected failure?

gregmagolan:
> expected_diagnostics is annoying since you have to regex anyway. plus the kind of failure we want to test here is the action before tsc. Could we have a generic feature for run_node or npm_package_bin for expected failure?

Yes. Something with that could integration with run_node would be ideal. nodejs_test has expected_exit code but we have nothing for build rule failure testing

PR #1741, opened 2020-03-26
feat(builtin): add support for predefined variables and custom variable to params_file

gregmagolan:
args docs for params_files is now:

```
args: Arguments to concatenate into a params file.

    1. Subject to $(location) substitutions.

    NB: This substition returns the manifest file path which differs from the *_binary & *_test
    args and genrule bazel substitions. This will be fixed in a future major release.
    See docs string of `expand_location_into_runfiles` macro in
    `internal/common/expand_into_runfiles.bzl` for more info.

    2. Subject to predefined variables & custom variable substitutions.

    See https://docs.bazel.build/versions/master/be/make-variables.html#predefined_variables
    and https://docs.bazel.build/versions/master/be/make-variables.html#custom_variables.

    Predefined genrule variables are not supported in this context.
```


internal/common/test/params_file.spec.js
```diff
@@ -0,0 +1,50 @@
...
+  'some_value',
+  // $location (expands to runfiles manifest path of format repo/path/to/file)
+  'build_bazel_rules_nodejs/package.json',
+  // $locations (expands to runfiles manifest path of format repo/path/to/file)
+  'build_bazel_rules_nodejs/package.json',
+  'build_bazel_rules_nodejs/internal/common/test/foo/bar/a.txt',
+  'build_bazel_rules_nodejs/internal/common/test/params_file.spec.js',
+];
+
+if (params.length !== expected.length) {
```
alexeagle: nit: don't we have some assertion library in the node_modules easily available that does assertEquals on arrays?

internal/common/params_file.bzl
```diff
@@ -70,25 +98,41 @@ def params_file(
...
-      args: Arguments to concatenate into a params file.
-          Subject to $(location) substitutions
-      newline: one of ["auto", "unix", "windows"]: line endings to use. "auto"
-          for platform-determined, "unix" for LF, and "windows" for CRLF.
-      **kwargs: further keyword arguments, e.g. <code>visibility</code>
+        name: Name of the rule.
+        out: Path of the output file, relative to this package.
+        args: Arguments to concatenate into a params file.
+
+        1. Subject to $(location) substitutions.
```
alexeagle: this is just copy-pasta from somewhere right? if we can't share it then maybe a comment pointing to the canonical version of it so we have a chance to prevent skew

gregmagolan: thanks. It's not copy-pasta but similar to another docstring so I'll just reference that. And I'll remove the duplicate docstrings from _ATTRS and so only the macro has the strings.

internal/common/test/params_file.spec.js
gregmagolan: thanks. switched to jasmine. good call.

PR #1747, opened 2020-03-27
chore: rely less on the index.from_src pattern

alexeagle:
Instead use substitutions when building packages to switch out the local bin label with the npm one


packages/typescript/src/internal/build_defs.bzl
```diff
@@ -22,7 +22,8 @@ load("@build_bazel_rules_typescript//internal:common/compilation.bzl", "COMMON_A
 load("@build_bazel_rules_typescript//internal:common/tsconfig.bzl", "create_tsconfig")
 load("//internal:ts_config.bzl", "TsConfigInfo")
 
-_DEFAULT_COMPILER = "@npm//@bazel/typescript/bin:tsc_wrapped"
+# NB: substituted with "@npm//@bazel/typescript/bin:tsc_wrapped" in the pkg_npm rule
+_DEFAULT_COMPILER = "@build_bazel_rules_typescript//internal:tsc_wrapped_bin"
```
jbedard: Does `@build_bazel_rules_typescript` work different then the other `@npm_bazel_*` targets? So we can't stick with the `@npm//@bazel/X` => `@npm_bazel_X` convention here?

jbedard:
If we do this in a limited and consistent fashion I think it might be better then all the `from_src` imports. Knowing it will be substituted while reading the code is important, but might be better/no-worse then needing to know the rule might be executed from `from_src.bzl`?

packages/typescript/src/internal/build_defs.bzl
alexeagle: yeah it's a weird snowflake because rules_typescript has a g3 source-of-truth

alexeagle:
@jbedard agreed, we need the repo to be more contributor friendly so your take on it has more weight than the rocket science Greg or I might try to apply to it.

I can spend some more time to kill the other .from_src usages but they're all less trivial than this pattern.

Also this will help us down the road if we have a BUILD file generator, we'll be able to use it ourselves in these cases since the load statement looks like what we tell users.

Note, we plan at some point to kill the `@npm_something` workspaces since these should be load from `@npm//something` just like the generated npm_package_bin rules for non-bazel-aware packages

PR #1751, opened 2020-03-28
test: fix rollup integration tests with worker mode by removing sourcemaps from inputs

gregmagolan:
There is a bug with the rollup cache (which is on in worker mode) & sourcemaps.

In short, rollup seems to drop the `//# sourceMappingURL=` comment from an input file when the file is not cached (worker mode or not) but when the file is cached and rollup is run with the cache in worker mode it fails to drop the comment (which is also present in the cache) and it ends up in the bundle which makes the golden tests fail.

The flakiness is likely due to the rollup action running twice on occasion. Actions running twice on occasion another unrelated issue which we've seen in CI for other rules as well and we don't know why that happens yet.

I happens both if the comment is an inline source map:
```
//# sourceMappingURL=data:application/json;base64,eyJ2ZXJzaW9uIjo...
```
or pointing to an external file
```
//# sourceMappingURL=index.js.map
```

So it seems that rollup + cache + `//# sourceMappingURL` comment in inputs is broken either in general or more likely in the way we are calling rollup in the worker with the cache. @jbedard to investigate but now that we understand the issue and the rollup worker is opt-in we're good.


packages/rollup/test/integration/golden.umd.js.map.sha256_
```diff
@@ -1 +1 @@
-58ccd4b63bbd5ff5e2eb93b2da4ea0ea34dcda1c82b37ae32dd77ed7867f4b9e
\ No newline at end of file
+cbec59b9c2ca7bde7011dbd2b14072ad3d51cbb5c9198cece6cf299071abb090
```
jbedard: The hashes changed but not the files themselves?

gregmagolan: Input files are now `.js` instead of `.mjs`

jbedard: But I guess they're still es6 since the actual files didn't change 👍

PR #1755, opened 2020-03-29
chore: update to Bazel 3.0 rc

alexeagle:


alexeagle:
actually bazelisk looks up the directory tree past the WORKSPACE IIUC so the .bazelversion in the rules_nodejs root should still work

PR #1757, opened 2020-03-29
chore: remove hide-build-files functionality

### LINKED ISSUE #1613: Deprecation for @bazel/hide-build-files package
Bazel 2.1 supports .bazelignore files in external repos (thanks @meteorcloudy !)
Our workaround for this can now be deprecated.

I just merged a change which removes our own dependency on it (https://github.com/bazelbuild/rules_nodejs/commit/82599e605e606b28f6ab58a7b82fe422baa21eb2)

Next steps:
- [ ] remove the https://github.com/bazelbuild/rules_nodejs/blob/82599e605e606b28f6ab58a7b82fe422baa21eb2/docs/Built-ins.md#hide_build_files attribute (first flip default to False, but keep using this flag for our own packages until most users are on 2.1)
- [x] mark the npm package as deprecated
- [ ] remove the https://github.com/bazelbuild/rules_nodejs/blob/82599e605e606b28f6ab58a7b82fe422baa21eb2/docs/Built-ins.md#always_hide_bazel_files and equivalent one for yarn 
- [ ] delete the hide-bazel-files package code
- [ ] delete the code in generate_build_files.ts that hides bazel files

alexeagle:
Since 1.3.0, we write a .bazelignore file into the generated workspaces, so Bazel does not see BUILD files.
The features to hide the BUILD files are no longer needed.

We also stop hiding the BUILD files in packages published by pkg_npm. This means that users who publish packages built by that rule will now require *their* users to have rules_nodejs 1.3.0 or greater, which adds the .bazelignore file.

BREAKING CHANGE:
rules_nodejs now requires Bazel 2.1 or greater.
Also the hide_build_files attribute was removed from pkg_npm, and always_hide_bazel_files was removed from yarn_install and npm_install.

Fixes #1613


Toxicable:
since this is a breaking change might be time to open up a 2.0 branch?

alexeagle:
We have 1.x branch, master is 2.0 now

Toxicable:
As part of your BReAKING CHANGE notice could you explain why those attributes were removed?

PR #1763, opened 2020-03-30
build: patch ng_module to provide LinkablePackageInfo

gregmagolan:
Interim patch until https://github.com/angular/angular/pull/36333 lands and makes it into a release.
For ViewEngine however, upgrades past `@angular/bazel` 9.0.0-rc.10 are broken and given that ViewEngine will soon be deprecated there is not much impetus to fix that atm so the patch will be long lived for the legacy compilation mode if users are using ViewEngine and "module_name" with ng_module.


gregmagolan:
@googlebot I consent.

jbedard:
@googlebot I consent

PR #1764, opened 2020-03-30
docs(examples): add use of ng_module(module_name) in angular example

jbedard:
@gregmagolan 

This should fail without the angular patch...


gregmagolan:
Thanks @jbedard . I moved this to angular_view_engine in https://github.com/bazelbuild/rules_nodejs/pull/1763 since Alex is moving examples/angular to use ts_library

PR #1765, opened 2020-03-30
feat(typescript): wire up use_angular_plugin attribute

alexeagle:
Add tests and switch away from `ng_module` in the angular example


examples/angular/package.json
```diff
@@ -29,8 +29,8 @@
     "devDependencies": {
         "@angular/bazel": "9.0.0",
```
mrmeku: Maybe remove `@angular/bazel`?

PR #1770, opened 2020-04-01
Simple repro of ts_project unable to resolve files not in bazel-out

mrmeku:
Not an actual PR, just demonstrating a bug


alexeagle:
I'll close this for housekeeping, the link and commit are still useful/valid as a demo of your problem of course 💚

PR #1772, opened 2020-04-01
fix(builtin): preserve lone $ in templated_args for legacy support

gregmagolan:
**This fixes an unintentional breaking change introduced in 1.5.0.**

This preservation may be removed in a future version so that templated_args behaves the same as the built-in *_binary and *_test args


gregmagolan:
@filipesilva Ran into this in angular-cli repo after upgrading to 1.5.0

gregmagolan:
> maybe was easier to land it for 2.0 after all ... 🤷‍♂️

This fixes a corner case that should be very rare so I'm still happy we the feature in for 1.5.

PR #1773, opened 2020-04-01
build: ts_project windows ci fix

gregmagolan:


packages/rollup/test/integration/far/a/tsconfig.json
```diff
@@ -5,5 +5,11 @@
     "target": "es2015",
     "module": "esnext",
     "moduleResolution": "node"
-  }
+  },
+  // For Windows, which is not sand-boxed, we need to limit
```
jbedard, line 9: Because the default is `**/*.ts` which includes subdirectories?

packages/rollup/test/integration/far/a/b/c/BUILD.bazel
```diff
@@ -4,7 +4,7 @@ load("@npm_bazel_typescript//:index.bzl", "ts_project")
 package(default_visibility = ["//packages/rollup:__subpackages__"])
 
 ts_project(
-    srcs = ["index.ts"],
+    srcs = glob(["*.ts"]),
```
jbedard, line 7: Was this necessary for the fix? Or just as a better example for common use cases?

packages/rollup/test/integration/far/a/tsconfig.json
gregmagolan: Yes

packages/rollup/test/integration/far/a/b/c/BUILD.bazel
gregmagolan: Just a better explanation so that the ts_project srcs lines up with the tsconfig include

PR #1774, opened 2020-04-01
ci: shard all integration tests

gregmagolan:


scripts/ci/run_integration_tests.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
+# -e: exits if a command fails
+# -u: errors if an variable is referenced before being set
+# -o pipefail: causes a pipeline to produce a failure return code if any command errors
+
+readonly TEST_TARGETS=$(bazel query 'kind("bazel_integration_test", //examples/... union //e2e/...)')
```
alexeagle, line 8: Switching in this PR from selecting based on tags to using the rule kind. Did you check that it gives the same list of targets?

alexeagle: also we rely on these coming in the same order every time, might be good to comment on that, or maybe even do an explicit sort to be safe?

gregmagolan: yes. double checked that everything is still tested.

I'm making the assumption that bazel query is deterministic in its order. I'll add an explicit sort to be safe.

gregmagolan:
Before:
Screen Shot 2020-04-01 at 1 43 56 PM

gregmagolan:
After:
Screen Shot 2020-04-01 at 2 03 30 PM

gregmagolan:
Sharded with parallelism 8:
Screen Shot 2020-04-01 at 2 07 02 PM

PR #1777, opened 2020-04-02
chore: create update_deleted_packages script that works on osx & linux

gregmagolan:
Current script wasn't compatible with osx


gregmagolan:
Replaced by #1778

PR #1780, opened 2020-04-02
feat(builtin): make npm_package_bin output linkable with package_name attribute

gregmagolan:
This will allow for a simplified pattern for npm_package_bin rules that make fully linkable packages such as `architect()`:

previously:
```
architect(
    name = "build",
    args = [
        "frontend-lib:build",
        "--",
        "$(@D)"
    ],
    data = ...,
    output_dir = True,
)

 pkg_npm(
     name = "npm_package",
     package_name = "frontend-lib",
     nested_packages = [":build"],
 )
```

with this feature:
```
architect(
    name = "build",
    args = [
        "frontend-lib:build",
        "--",
        "$(@D)"
    ],
    data = ...,
    package_name = "frontend-lib",
    output_dir = True,
)
```

Since the output of `pkg_npm` above is identical to that of `architect`, unless you want the pack & publish functionality of `pkg_npm` you can simplify and save copying the files around by using `package_name` on `architect` instead.

### DIFF
--- a/examples/angular_bazel_architect/BUILD.bazel
+++ b/examples/angular_bazel_architect/BUILD.bazel
@@ -35,7 +35,7 @@ architect(
         "@npm//@angular/platform-browser-dynamic",
         "@npm//@angular-devkit/architect-cli",
         "@npm//@angular-devkit/build-angular",
-        "//projects/frontend-lib:npm_package",
+        "//projects/frontend-lib",
     ],
     output_dir = True,
 )
@@ -64,7 +64,7 @@ architect_test(
         "@npm//karma-coverage-istanbul-reporter",
         "@npm//karma-jasmine",
         "@npm//karma-jasmine-html-reporter",
-        "//projects/frontend-lib:npm_package",
+        "//projects/frontend-lib",
     ],
     tags = [
         "browser:chromium-local",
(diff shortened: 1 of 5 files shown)


alexeagle:
This feels like a strange special case. Very few npm bin scripts are intended to produce new npm packages. Does architect produce a package.json in this case so it's really a publishable package?
It also seems like feature creep for that rule (though the implementation is nice and small)

It feels like what's really desired here is a little macro around `npm_package_bin` that puts the output dir into a package.
If the user doesn't want to have that macro live in their project, I could imagine generating another target (akin to the `architect_test` target we generate to behave like a test) like `pkg_architect` that is a macro invoking `architect` and nesting the result into a `pkg_npm` WDYT?

gregmagolan:
Yes. In this case, architect() does create a publishable output directory complete with package.json and everything.

gregmagolan:
I'm leaning towards having this in as ng-packager likely won't be the only js tool out there that makes a package. A macro that takes the architect() output and pipes it to a pkg_npm() would work as well, which is what Alan's PR currently does, but there is some ambiguity as to how pkg_npm should handle directory artifacts which I'm not sure how to resolve.

Currently pkg_npm `deps` will keep the output directory artifact intact which doesn't work with architect ng-packager where the output directory artifact to the the root of the package. So piping it to `nested_packages` is a temporary work-around but we need a better pattern there and a good explanation of how it is supposed to to work.

alexeagle:
offline discussion: the nice thing about the `pkg_architect` macro is it keeps the layering distinct. Any rule which produces a tree artifact can work with `pkg_npm` that way. it could have a special case for a single src which is a TreeArtifact.

gregmagolan:
Replaced with better solution: https://github.com/bazelbuild/rules_nodejs/pull/1786

PR #1496, opened 2019-12-21
feat(examples): upgrade rules_docker to 0.13.0

Toxicable:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

PR #1782, opened 2020-04-03
chore: drop rules_docker patch

alexeagle:
The latest release has the fix we need


Toxicable:
You'll need a few other changes https://github.com/bazelbuild/rules_nodejs/pull/1496

alexeagle:
#1496 should go instead of this

PR #1785, opened 2020-04-03
test(jasmine): add coverage for fix #1796

gregmagolan:
Pre-1796, test expected to fail on Windows because a relative require from
the output tree to the source tree will not work unless the spec
is run from the correct spot. But without the fix in
`external\npm\node_modules\build_bazel_rules_nodejs\packages\jasmine\test\generated.spec.js` (since jasmine was let to do standard node_module resolution and that is where it resolved to).
```
==================== Test output for //packages/jasmine/test:generated_require_spec_test:
Randomized with seed 77827
Started
C:\b\t2skj56o\external\npm\node_modules\build_bazel_rules_nodejs\packages\jasmine\test\generated.spec.js
Failures:
1) output tree spec resolves should resolve runfiles relative
Message:
Failed: './test.json' should be resolved
Stack:
Error: Failed: './test.json' should be resolved
at <Jasmine>
at UserContext.<anonymous>
(C:\b\t2skj56o\external\npm\node_modules\build_bazel_rules_nodejs\packages\jasmine\test\generated.spec.js:8:7)
at <Jasmine>
at processImmediate (internal/timers.js:439:21)
1 spec, 1 failure
Finished in 0.011 seconds
Randomized with seed 77827 (jasmine --random=true --seed=77827)
```

#1796 fixes jasmine_node_test to use runfiles.resolve on specs so absolute paths are passed to jasmine and the specs are running out of the correct locations where the require patches can handle relative imports.


gregmagolan:
> the change description made me think there would be a removed `fix-windows` tag but I guess I just misunderstood

Yes, a bit confusing since I added the breaking test after the fix already landed but kept it as two commits so it was clear what broke and what the fix fixed.

PR #1789, opened 2020-04-05
fix(builtin): fix runfiles linking issue on Windows

### LINKED ISSUE #1781: //internal/pkg_npm/test/linking:test fails on Windows with remote caching enabled
# 🐞 bug report

### Is this a regression?

No, the test is added recently

### Description

`//internal/pkg_npm/test/linking:test` will fail in a second run with remote caching enabled.

https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1446#b6409cc9-0a26-4ddd-8828-74d639131091

## 🔬 Minimal Reproduction

```
bazel test //internal/pkg_npm/test/linking:test
bazel clean --expunge & rm -rf node_modules
bazel test //internal/pkg_npm/test/linking:test
```

## 🔥 Exception or Error

<pre><code>
Executing tests from //internal/pkg_npm/test/linking:test
-----------------------------------------------------------------------------
internal/modules/cjs/loader.js:797
    throw err;
    ^

Error: Cannot find module '@scoped/bar'
Require stack:
- C:\b\s7rxdkmr\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\internal\pkg_npm\test\linking\main.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:794:15)
    at Function.Module._load (internal/modules/cjs/loader.js:687:27)
    at Module.require (internal/modules/cjs/loader.js:849:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (C:\b\s7rxdkmr\execroot\build_bazel_rules_nodejs\bazel-out\x64_windows-fastbuild\bin\internal\pkg_npm\test\linking\main.js:3:13)
    at Module._compile (internal/modules/cjs/loader.js:956:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:973:10)
    at Module.load (internal/modules/cjs/loader.js:812:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1025:10) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    'C:\\b\\s7rxdkmr\\execroot\\build_bazel_rules_nodejs\\bazel-out\\x64_windows-fastbuild\\bin\\internal\\pkg_npm\\test\\linking\\main.js'
  ]
}
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Windows
  </code>
</pre>

<pre>
  <code>
3.0.0rc2
  </code>
</pre>

**Rules_nodejs version:**

<pre>
  <code>
HEAD
  </code>
</pre>

gregmagolan:
Fix 1 of 2 for linker issue on Windows when all build actions are cached #1781. This fixes incorrectly symlinks but it does not actually address the underlying issue of programs (tests) running outside of the sandbox and directly out of the execroot on Windows (either out of execroot sources or execroot bazel-out).

If all build actions are cached then the linker only runs for the test that is executed. If there are no third-part npm deps then it makes a node_modules folder in the runfiles for that test and puts its symlinks in there. The problem is that when the test runs, the .js spec is in execroot bazel-out and it doesn't see the node_modules created in the runfiles folder. Since no build actions have run that may have created the symlinks needed for the test in the execroot node_modules, the symlinks that it needs do not exist.

The 2nd fix (still coming) seems to be to always symlink into the execroot which is below the runfiles folder. Some mental gymnastics needed to figure out if this is correct when in the sandbox as well and if there are third-party deps or not. I'll investigate.

### DIFF
--- a/internal/linker/index.js
+++ b/internal/linker/index.js
@@ -440,37 +440,24 @@ Include as much of the build output as you can without disclosing anything confi
                     yield mkdirp(path.dirname(m.name));
                     if (m.link) {
                         const [root, modulePath] = m.link;
-                        const externalPrefix = 'external/';
                         let target = '<package linking failed>';
                         switch (root) {
                             case 'execroot':
                                 if (runfiles.execroot) {
                                     target = path.posix.join(runfiles.workspaceDir, modulePath);
+                                    break;
                                 }
-                                else {
-                                    // If under runfiles, convert from execroot path to runfiles path.
-                                    // First strip the bin portion if it exists:
-                                    let runfilesPath = modulePath;
-                                    if (runfilesPath.startsWith(`${bin}/`)) {
-                                        runfilesPath = runfilesPath.slice(bin.length + 1);
-                                    }
-                                    else if (runfilesPath === bin) {
-                                        runfilesPath = '';
-                                    }
-                                    // Next replace `external/` with `../` if it exists:
-                                    if (runfilesPath.startsWith(externalPrefix)) {
-                                        runfilesPath = `../${runfilesPath.slice(externalPrefix.length)}`;
-                                    }
-                                    target = path.posix.join(runfiles.workspaceDir, runfilesPath);
-                                }
-                                break;
+                            // If under runfiles, the fall through to 'runfiles' case
+                            // so that we handle case where there is only a MANIFEST file
                             case 'runfiles':
                                 // Transform execroot path to the runfiles manifest path so that
-                                // it can be resolved with runfiles.resolve()
+                                // it can be resolved with runfiles.resolve().
                                 let runfilesPath = modulePath;
+                                // First strip the bin portion if it exists
                                 if (runfilesPath.startsWith(`${bin}/`)) {
                                     runfilesPath = runfilesPath.slice(bin.length + 1);
                                 }
+                                const externalPrefix = 'external/';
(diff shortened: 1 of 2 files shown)


gregmagolan:
This is a partial fix for linker issue on Windows when all build actions are cached. This fixes incorrectly symlinks but it does not actually address the underlying issue of programs (tests) running outside of the sandbox and directly out of the execroot on Windows (either out of execroot sources or execroot bazel-out).

If all build actions are cached then the linker only runs for the test that is executed. If there are no third-part npm deps then it makes a node_modules folder in the runfiles for that test and puts its symlinks in there. The problem is that when the test runs, the .js spec is in execroot bazel-out and it doesn't see the node_modules created in the runfiles folder. Since no build actions have run that may have created the symlinks needed for the test in the execroot node_modules, the symlinks that it needs do not exist.

The principled fix seems to be to always symlink into the execroot which is below the runfiles folder. Some mental gymnastics needed to figure out if this is correct when in the sandbox as well and if there are third-party deps or not. I'll investigate.

gregmagolan:
Replaced by https://github.com/bazelbuild/rules_nodejs/pull/1805 which includes this fix and adds the 2nd fix as well.

PR #1792, opened 2020-04-06
fix(rollup): fix worker not picking up config file changes

jbedard:
Node caches `require()` calls, so when we re-`require()` the potentially changed `rollup.config.cjs.js` we need to ensure it's not cached.

Basically copying: https://github.com/rollup/rollup/blob/v1.31.0/cli/run/loadConfigFile.ts#L52

### DIFF
--- a/packages/rollup/src/index.js
+++ b/packages/rollup/src/index.js
@@ -39,10 +39,10 @@ try {
 // Store the cache forever to re-use on each build
 let cacheMap = Object.create(null);
 
-// Generate a unique cache ID based on the input/outputOptions
-function computeCacheKey(inputOptions, outputOptions) {
+// Generate a unique cache ID based on the given json data
+function computeCacheKey(cacheKeyData) {
   const hash = crypto.createHash('sha256');
-  const hashContent = JSON.stringify(inputOptions) + JSON.stringify(outputOptions);
+  const hashContent = JSON.stringify(cacheKeyData);
   return hash.update(hashContent).digest('hex');
 }
 
--- a/packages/rollup/src/index.js
+++ b/packages/rollup/src/index.js
@@ -105,6 +105,10 @@ async function loadConfigFile(configFile) {
 
   await runRollup(configFile, inputOptions, outputOptions);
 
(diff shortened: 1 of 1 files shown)


packages/rollup/src/index.js
```diff
@@ -105,6 +105,10 @@ async function loadConfigFile(configFile) {
 
   await runRollup(configFile, inputOptions, outputOptions);
 
+  // Ensure node isn't caching a previous version of the config file
+  // https://github.com/rollup/rollup/blob/v1.31.0/cli/run/loadConfigFile.ts#L52
+  delete require.cache[cjsConfigFile];
```
gregmagolan, line 110: Should this be `require.cache[require.resolve(cjsConfigFile)]` to ensure it matches the cache key?

jbedard: Because node probably does the same thing before caching? I think that would make sense, I can add `require.resolve` where we define `cjsConfigFile`...

jbedard: Actually the file doesn't exist until this line (after `runRollup` creates it above) so I've updated to exactly what you suggested here.

packages/rollup/src/index.js
```diff
@@ -39,10 +39,10 @@ try {
 // Store the cache forever to re-use on each build
 let cacheMap = Object.create(null);
 
-// Generate a unique cache ID based on the input/outputOptions
-function computeCacheKey(inputOptions, outputOptions) {
+// Generate a unique cache ID based on the given json data
+function computeCacheKey(cacheKeyData) {
```
alexeagle, line 43: interesting, it took two args before but we only passed one?

jbedard: That was me playing around with how the cache key was calculated and not cleaning up after myself.

Originally the key was basically all input/outputOptions. I think that's overkill and basically just the entry-files is enough.

Note that two keys conflicting isn't the end of the world, it just means the cache data passed to rollup might be useless (then get overwritten and useless for the other rollup rule with the conflicting key).

PR #1795, opened 2020-04-06
fix(npm_umd_bundle): include all json files

mrmeku:
To make a UMD bundle of the popular react testing library enzyme, bundling json files from @npm//entities is required

### DIFF
--- a/internal/npm_install/npm_umd_bundle.bzl
+++ b/internal/npm_install/npm_umd_bundle.bzl
@@ -40,7 +40,7 @@ def _npm_umd_bundle(ctx):
     inputs = [
         f
         for f in sources
-        if f.path.endswith(".js") or f.basename == "package.json"
+        if f.path.endswith(".js") or f.path.endswith(".json")
     ]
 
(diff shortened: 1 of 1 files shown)


internal/npm_install/npm_umd_bundle.bzl
```diff
@@ -40,7 +40,7 @@ def _npm_umd_bundle(ctx):
     inputs = [
         f
         for f in sources
-        if f.path.endswith(".js") or f.basename == "package.json"
+        if f.path.endswith(".js") or f.path.endswith(".json")
```
alexeagle, line 43: update the comment above please, to match the code change and also to make sure we don't think it's safe to change it back.
If you want to add test coverage you can add a case to https://github.com/bazelbuild/rules_nodejs/blob/master/internal/npm_install/test/browserify.spec.js
(though it has the sad consequence of adding more top-level dependencies to our project)

alexeagle:
can address the review feedback in a followup

PR #1796, opened 2020-04-06
refactor(jasmine): cleanup jasmine_node_test to no longer depend on require of the manifest path

gregmagolan:
Use runfiles helper instead.

This is a pre-req to removing the `node_modules/my_workspace` symlink which allows manifest path requires to work with standard node_modules resolution. The principled approach should be to have tests use the runfiles helper and not rely on being able to resolve `my_workspace/path/to/output/file.js` with standand node_modules resolution.

The problem with that symlink is that it leads to infinite symlink recursion as my_workspace contains the same node_modules folder that it is symlinked into so you get `/path/to/my_workspace/node_modules/my_workspace/node_modules/my_workspace/node_modules/...` to infinity and bey0nd.

Removing that symlink would be a breaking change, however, so it will wait until 2.0.0-next.0.


gregmagolan:
Also refactor devmode_js_sources helper rule out of common as it only used by jasmine_node_test

PR #1797, opened 2020-04-06
fix(builtin): linker no longer makes node_modules symlink to the root of the workspace output tree

gregmagolan:
For 2.0.

Thanks @mrmeku for updating `ts_proto_library` for this PR.

BREAKING CHANGES:

1. Any nodejs_binary/nodejs_test processes with the linker enabled (--nobazel_patch_module_resolver is set) that were relying on standard node_module resolution to resolve manifest file paths such as `my_workspace/path/to/output/file.js` must now use the runfiles helper such as.

Previously:
```
const absPath = require.resolve('my_workspace/path/to/output/file.js');
```
With runfiles helper:
```
const runfiles = require(process.env['BAZEL_NODE_RUNFILES_HELPER']);
const absPath = runfiles.resolve('my_workspace/path/to/output/file.js');
```

2. Angular ViewEngine + Bazel is no longer supported in rules_nodejs 2.0 as the generated ngfactory & ngsummary files depend on absolute imports. Users upgrading to rules_nodejs 2.0 will also need to upgrade to Ivy if they are on ViewEngine.


gregmagolan:
Pretty much done except for issues with the ts_proto_library labs rule which has examples & tests that depends on absolute imports. @mrmeku PTAL.

alexeagle:
What do we do about the recommendation in a monorepo that you can avoid crazy `import {} from '../../../../../../path/to/thing'` and instead always `import {} from my_wksp/path/to/thing`
do we have any revised guidance? is there a README somewhere that needs to be updated?

examples/angular_view_engine/src/index.html
```diff
@@ -1,761 +0,0 @@
...
-                <a
-                  href="https://medium.com/@Jakeherringbone/automated-testing-of-each-commit-ci-6f718d93d0da"
-                >
-                  Automated testing of each commit != CI
-                </a>
-                (January 2019)
-              </li>
-              <li>
-                <a
-                  href="https://medium.com/@Jakeherringbone/running-tools-under-bazel-8aa416e7090c"
```
alexeagle, line 632: all this content we didn't really mean to delete...

internal/linker/test/integration/program.js
```diff
@@ -7,8 +7,20 @@ const t = require('transitive_static_linked');
 // they should get resolved from the execroot
 const b = require('dynamic_linked');
 const d = require('@linker_scoped/dynamic_linked');
-// We've always supported `require('my_workspace')` for absolute imports like Google does it
-const c = require('build_bazel_rules_nodejs/internal/linker/test/integration/absolute_import');
+
+let c;
+try {
+  // As of 2.0, we no longer support `require('my_workspace/path/to/output/file.js')` for absolute
```
alexeagle, line 13: would be nice to link to this spot in the BREAKING CHANGE

IgorMinar:
Let's coordinate. I have a meeting with Greg later this week so we can cover this as well.

alexeagle:
@gregmagolan if you don't have time to re-base and land, we could hand to someone else?

alexeagle:
Need to check that it's still possible for users to link their project root by putting a `pkg_npm` or something in the /BUILD.bazel

alexeagle:
Windows failure FYI
```

Test output for //packages/typescript/test/ts_library_esm_with_jest:test:
--
  | FAIL ../../lib.test.js
  | ● Test suite failed to run
  |  
  | Cannot find module 'build_bazel_rules_nodejs/packages/typescript/test/ts_library_esm_with_jest/lib' from 'lib.test.js'
  |  
  | > 1 \| import {doStuff} from './lib';
  | \| ^
  | 2 \|
  | 3 \| describe('doStuff', () => {
  | 4 \|   it('should do some stuff', () => {
  |  
  | at Resolver.resolveModule (../../../../../../../node_modules/jest-resolve/build/index.js:296:11)
  | at ../../../../../../../packages/typescript/test/ts_library_esm_with_jest/lib.test.ts:1:1
  | at lib.test.js:3:17
  | at Object.<anonymous> (lib.test.js:9:3)

```

examples/protocol_buffers/BUILD.bazel
```diff
@@ -5,6 +5,7 @@ load("@npm//@bazel/rollup:index.bzl", "rollup_bundle")
 load("@npm//@bazel/terser:index.bzl", "terser_minified")
 load("@npm//@bazel/typescript:index.bzl", "ts_config", "ts_devserver", "ts_library")
 load("@npm//http-server:index.bzl", "http_server")
+load("@rules_proto//proto:defs.bzl", "proto_library")
```
alexeagle, line 8: was this just a warning you're fixing or is it related to the change?

gregmagolan: 🤔  where did that symbol come from before I added that load?

alexeagle:
Note, #2175 added back the ability to link the workspace root into the node_modules tree

PR #1800, opened 2020-04-07
fix(builtin): don't allow symlinks to escape or enter bazel managed node_module folders

gregmagolan:


internal/node/launcher.sh
```diff
@@ -217,6 +217,22 @@ fi
 # Bazel always sets the PWD to execroot/my_wksp so we go up one directory.
 export BAZEL_PATCH_ROOT=$(dirname $PWD)
 
+# Set all bazel managed node_modules directories as guarded so no symlinks may
+# escape and no symlinks may enter
+if [ "$(basename ${BAZEL_PATCH_ROOT})" == "execroot" ]; then
+  # We are in execroot, linker node_modules is in the PWD
+  export BAZEL_PATCH_GUARDS="${PWD}/node_modules"
```
alexeagle, line 224: seems like this doesn't work if your node_modules is in a subdirectory, which we support - but I think that's what BAZEL_NODE_MODULES_ROOT is doing below? maybe this needs a comment that we do this in a catch-all case where we don't have a BAZEL_NODE_MODULES_ROOT - but then why does that happen?

internal/node/node_patches.js
```diff
@@ -483,23 +484,36 @@ exports.patcher = (fs = fs$1, root) => {
         }
     }
 };
-exports.escapeFunction = (root) => {
-    // ensure root is always absolute.
+exports.escapeFunction = (root, guards) => {
+    // ensure root & guards are always absolute.
     root = path.resolve(root);
+    guards = [...guards].map(g => path.resolve(g));
```
jbedard, line 490: Why clone `guards` just to invoke `.map`? 🤔

gregmagolan: good point :)

internal/node/launcher.sh
gregmagolan: BAZEL_NODE_MODULES_ROOT is the path to the external workspace such as `npm/node_modules`.
This path is actually the path to the linker symlink node_modules under execroot which should work regardless of where node_modules is.

gregmagolan: BAZEL_NODE_MODULES_ROOT is always set in templated_args above so below is just a sanity check:
```
    env_vars += """if [[ -z "${BAZEL_NODE_MODULES_ROOT:-}" ]]; then
  export BAZEL_NODE_MODULES_ROOT=%s
fi
```

soldair:
\o sorry for the delay. i somehow missed most of my communication last week

PR #1806, opened 2020-04-08
WIP fix: use puppeteer in examples/angular_bazel_architect so it doesn't depend on local chrome

gregmagolan:
Currently example may fail if it pulls down the wrong webdriver version for the locally installed chrome as seen here on master: https://app.circleci.com/pipelines/github/bazelbuild/rules_nodejs/670/workflows/c3e2b283-7e34-4833-8f99-74e628e22611


examples/angular_bazel_architect/projects/frontend-lib/karma.conf.js
```diff
@@ -28,7 +31,23 @@ module.exports = function (config) {
     customLaunchers: {
       ChromeHeadlessNoSandbox: {
         base: 'ChromeHeadless',
-        flags: ['--no-sandbox', '--disable-gpu', '--disable-dev-shm-usage']
+        // `--no-sandbox` flag disables the chrome sandbox because it causes Chrome to crash on some
```
alexeagle, line 34: should we do sth about the code duplication here? we could generate karma.conf but maybe that makes it too different from standard angular?

examples/angular_bazel_architect/scripts/puppeteer-chrome-versions.js
```diff
@@ -0,0 +1,16 @@
+// Mapping of puppeteer releases to their default Chrome version
+// derived from https://github.com/puppeteer/puppeteer/blob/master/docs/api.md.
+// The puppeteer package.json file contains the compatible Chrome revision such as
+// "chromium_revision": "722234" but this does not map easily to the Chrome version
+// so we use this mapping here instead.
+module.exports = {
```
alexeagle, line 6: ouch, feels like it belongs somewhere other than the users app, not sure where though

PR #1809, opened 2020-04-08
WIP ci: pin webdriver-manager version for examples/angular_bazel_architect

gregmagolan:
Alternate to https://github.com/bazelbuild/rules_nodejs/pull/1806 which is proving difficult to green up


Toxicable:
maybe also pin the docker image?

gregmagolan:
> maybe also pin the docker image?

That is pinned already in CircleCI config: `var_2: &browsers_docker_image circleci/node:12.14-browsers`

Toxicable:
I was thinking pinning to a hash, but probably not needed, I wouldn't think they'd push another version on the same label

PR #1811, opened 2020-04-09
chore: remove typescript index.from_src

alexeagle:
Fixes a bug in ts_project where it couldn't accept generated tsconfig.json files

This continues the cleanup of the from_src variants of the public entry points.
Some notes on this one since it's tricky:
- we now use ts_project to pre-compile the .ts files that we check in as .js. It uses a generated tsconfig for funsies
- because ts_project labels the .js outputs, we can't refer to the foo.js output as well as a foo.js input in the source tree - the golden_file_test gets confused about the labels
  so we have to rename one of the outputs to index.js


internal/linker/index.js
```diff
@@ -1,4 +1,5 @@
-/* THIS FILE GENERATED FROM .ts; see BUILD.bazel */ /* clang-format off */var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
+/* THIS FILE GENERATED FROM .ts; see BUILD.bazel */ /* clang-format off */"use strict";
+var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {
```
jbedard, line 2: This file changed quite a bit, but it seems like it really only a) dropped (all?) the comments  and b) dropped the UMD def which I guess was pointless?

is it the generated tsconfig.json vs previous ts_library config that changed this?

alexeagle: sorry, I should have explained in the commit message - yes it's just as you say, now that it uses `ts_project` we have more control so we don't bother with UMD, and as Greg is adding a lot of comments in the linker right now I figured we could save the 1% size on the builtin tgz.

https://github.com/bazelbuild/rules_nodejs/pull/1811/commits/ee3e065b763b7eb59777cff6217909f63977c31f#diff-28cc3d70512445827e84e7bd2f4adb49R31-R32

jbedard: Yeah that makes sense 👍 

I assume the other checked-in generated file is the same changes? That one was also renamed so you can't really see the diff...

alexeagle: right, identical changes. the rename is unfortunate but I couldn't find a less disruptive way to switch to ts_project, since it predeclares/labels the output js and that causes collision with input artifact

PR #1812, opened 2020-04-09
fix(typescript): don't mix worker mode and linker

### LINKED ISSUE #1803: Memory leak of ts_library
# 🐞 bug report

### Affected Rule

 ts_library

### Is this a regression?

 No.

### Description

 Build lot of `ts_library` target via command such as `bazel build src/...` would cause a memory leak problem.
    
The bazel would create a lot of node processes to compile the ts.
wrong
Then the computor memory is all ate by these node processes.
wrong_memory
I found that this commit caused this problem.
feat(typescript): use run_node helper to execute tsc 066a52c43052cfb5b32cc2effe52c4d31c133715
And  I change the code like below and the problem is solved, node processes number became normal.
correct_code
correct

And I make deep debuging, found the extra arg `--bazel_node_modules_manifest=%s` to the `tsc_wrapper` caused this probel, and I comment this, the problem is solved too.
image

## 🔬 Minimal Reproduction

It looks great when build a single ts_library target, but not great when build lots of ts targets via `bazel build ...`.

## 🔥 Exception or Error

<pre><code>
No exception, but the computor memory is overflow.
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Ubutu 18.04 LTS
  </code>
</pre>

<pre>
  <code>
2.1.0
  </code>
</pre>

**Rules_nodejs version:**

<pre>
  <code>
1.5.0
  </code>
</pre>

**Anything else relevant?**

alexeagle:
This causes hard-to-reproduce resource leaks as demonstrated in #1803
Also we didn't fully understand this before, for example if a worker process stayed running we'd never re-link under it when deps change
For now make the conservative fix to keep these two capabilities distinct

Fixes #1803

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -174,7 +174,6 @@ jobs:
           command: echo 'startup --host_jvm_args=-Xms256m --host_jvm_args=-Xmx1280m' >> .bazelrc
       - *init_bazel
       - *hide_node_and_yarn_local_binaries
-
       - run:
           command: ./scripts/ci/run_integration_tests.sh ${CIRCLE_NODE_INDEX} ${CIRCLE_NODE_TOTAL}
           no_output_timeout: 30m
--- a/e2e/webapp/BUILD.bazel
+++ b/e2e/webapp/BUILD.bazel
@@ -8,7 +8,8 @@ rollup_bundle(
     entry_point = "app.js",
     output_dir = True,
     # Enable experimental faster builds with Bazel workers
-    supports_workers = True,
+    # TODO: re-enable worker mode once fixed in bazel_integration_test
+    # supports_workers = True,
 )
 
 terser_minified(
--- a/examples/angular/package.json
+++ b/examples/angular/package.json
@@ -9,11 +9,11 @@
     },
     "dependencies": {
         "@angular/animations": "9.1.0",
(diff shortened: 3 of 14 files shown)


gregmagolan:
@googlebot I consent.

scripts/ci/run_integration_tests.sh
```diff
@@ -29,6 +29,10 @@ for target in ${TEST_TARGETS[@]:-}; do
   ((s=s+1))
 done
 
+# Stash system path in a special location that is picked up by test_runner.js
+# for inner bazel-in-bazel execution 
+echo "${PATH}" > /tmp/build_bazel_rules_nodejs__bazel_integration_test__orig_path
```
alexeagle, line 34: I guess it's too annoying to encode the whole value into a `--test_arg`?

gregmagolan: Actually. This may be incorrect as its not working on linux anyway. Still experimenting.

gregmagolan: Yes. It is not needed. Misdiagnosed that one.

alexeagle:
@googlebot thanks for all that you do!

Issue #1813, opened 2020-04-09
Jest's expect toBeInstanceOf fails in rules_nodejs


aherrmann:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `jest_test` from `@npm//jest-cli:index.bzl`

### Is this a regression?

Not that I'm aware of.

### Description

A `jest` `test` making use of `expect(...).toBeInstanceOf(...)` fails when it should succeed, the same test succeeds when run using `yarn test`.

## 🔬 Minimal Reproduction

* Checkout https://github.com/aherrmann/mock_class_mismatch (`master` `fec40bbe7159854dbcb6c6f9ab81e28aa624e5d8` at the time of writing).
* Execute
  ```
  $ bazel test //:repro --test_output=streamed
  ```
* Observe the following test failure:
  ```
    ...
    expect(received).toBeInstanceOf(expected)

    Expected constructor: Server
    Received constructor: Server

      10 |   console.log(ws.server.constructor);  // [Function: Server] { of: [Function: of] }
      11 |   console.log(Server);  // [Function: Server] { of: [Function: of] }
    > 12 |   expect(ws.server).toBeInstanceOf(Server);  // FAILS
         |                     ^
      13 | });
      14 |
    ...
  ```
* For comparison, run the test with `yarn` and see it succeed:
  ```
  $ yarn test
  ```

## 🔥 Exception or Error

```
  ...
  expect(received).toBeInstanceOf(expected)

  Expected constructor: Server
  Received constructor: Server

    10 |   console.log(ws.server.constructor);  // [Function: Server] { of: [Function: of] }
    11 |   console.log(Server);  // [Function: Server] { of: [Function: of] }
  > 12 |   expect(ws.server).toBeInstanceOf(Server);  // FAILS
       |                     ^
    13 | });
    14 |
  ...
```

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Ubuntu 19.10
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
2.0.0
  </code>
</pre>

**Rules_nodejs version:**

(Please check that you have matching versions between WORKSPACE file and `@bazel/*` npm packages.)

<pre>
  <code>
1.1.0
  </code>
</pre>

**Anything else relevant?**

* `yarn` version 1.21.11
* `node` version 13.7.0 under `yarn`, tried both 12.13.0 and 13.7.0 under `bazel`, issue observed in both cases.
[...]

PR #1805, opened 2020-04-08
fix(builtin): always symlink node_modules at `execroot/my_wksp/node_modules` even when running in runfiles

gregmagolan:
Fixes #1781.

Under runfiles, the linker should symlink node_modules at `execroot/my_wksp` so that when there are no runfiles (default on Windows) and scripts run out of `execroot/my_wksp` they can resolve node_modules with standard node_module resolution.

Also, restore BAZEL_WORKSPACE name environment variable. The optimization of deriving the workspace name from the path does not work on RBE. While we can derive the workspace from the pwd when running locally because it is in the execroot path `execroot/my_wksp`, on RBE the `execroot/my_wksp` path is reduced a path such as `/w/f/b` so the workspace name is obfuscated from the path. So we provide the workspace name here as an environment variable avaiable to all actions for the runfiles helpers to use.

PR #1814, opened 2020-04-09
test(builtin): add test for issue 1813

gregmagolan:
Issue #1813 was fixed by #1800 which is already landed. #1805 also fixes via a different mechanism and will also land soon.

### DIFF
--- a/internal/linker/test/issue_1813/BUILD.bazel
+++ b/internal/linker/test/issue_1813/BUILD.bazel
@@ -1,0 +1,21 @@
+load("@npm//jest-cli:index.bzl", "jest_test")
+
+jest_test(
+    name = "test",
+    data = [
+        "index.test.js",
+        "jest.config.js",
+        "@npm//jest",
+        "@npm//jest-websocket-mock",
+        "@npm//mock-socket",
+    ],
+    templated_args = [
+        "--no-cache",
+        "--no-watchman",
+        "--ci",
+        # On Windows, spawing workers is broken
+        "--runInBand",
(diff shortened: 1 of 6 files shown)


gregmagolan:
Also includes a windows fix commit:

 fix(builtin): fix node.bat proxy for Windows

Turns out this added new coverage under Windows

internal/linker/test/issue_1813/BUILD.bazel
```diff
@@ -0,0 +1,21 @@
...
+        "jest.config.js",
+        "@npm//jest",
+        "@npm//jest-websocket-mock",
+        "@npm//mock-socket",
+    ],
+    templated_args = [
+        "--no-cache",
+        "--no-watchman",
+        "--ci",
+        # On Windows, spawing workers is broken
```
alexeagle, line 16: oh! I wish I could use this to land the always-link-bins PR too

internal/linker/test/issue_1813/BUILD.bazel
```diff
@@ -0,0 +1,21 @@
...
+        "--runInBand",
+        "--config",
+        "$$(rlocation $(rootpath jest.config.js))",
```
alexeagle, line 19: random note: any time we write this under the `/examples` directory we should comment about it

alexeagle: It would be rad if this were `args=[".dist/bin/jest.config.js"])` and we fixed it up later in the toolchain

PR #1816, opened 2020-04-10
fix(builtin): look in the execroot for nodejs_binary source entry_points

### LINKED ISSUE #1787: Regression in 1.5.0 generated rules
# 🐞 bug report

### Affected Rule
The issue is related to the auto generated rules for `node_modules/.bin`

### Is this a regression?
Yes, this works in 1.4.1

### Description
I'm using the auto genrated rules to run GraphQL Code Generator(https://graphql-code-generator.com)

It works fine in version 1.4.1 but in 1.5.0 in fails. 
I've created a minimal reproduction of the issue here: https://github.com/purkhusid/glql-gen-repro

## 🔥 Exception or Error

<pre><code>
Cannot use GraphQLEnumType "__TypeKind" from another module or realm.

Ensure that there is only one instance of "graphql" in the node_modules
directory. If different versions of "graphql" are the dependencies of other
relied on modules, use "resolutions" to ensure only one version is installed.

https://yarnpkg.com/en/docs/selective-version-resolutions

Duplicate "graphql" modules cannot be used at the same time since different
versions may have different capabilities and behavior. The data from one
version used in the function from another could produce confusing and
spurious results.
[21:03:30] Generate types.ts [failed]
[21:03:30] → Cannot use GraphQLEnumType "__TypeKind" from another module or realm.

Ensure that there is only one instance of "graphql" in the node_modules
directory. If different versions of "graphql" are the dependencies of other
relied on modules, use "resolutions" to ensure only one version is installed.

https://yarnpkg.com/en/docs/selective-version-resolutions

Duplicate "graphql" modules cannot be used at the same time since different
versions may have different capabilities and behavior. The data from one
version used in the function from another could produce confusing and
spurious results.

</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
MacOS 1.14
  </code>
</pre>

<pre>
  <code>
Build label: 2.2.0
Build time: Tue Mar 3 09:28:15 2020 (1583227695)
Build timestamp: 1583227695
Build timestamp as int: 1583227695
  </code>
</pre>

**Rules_nodejs version:**
<pre>
  <code>
    1.5.0
  </code>
</pre>

alexeagle:
In a6e29c2add19dd487035b0951f5271053a0a11e2 we added support for generated entry_point by adding a secondary lookup.
In 863c7de48046c35dfe29f6a4f625ca3e7057b030 we reversed the order of the lookups to make rollup work in a certain use case.
Neither of these was principled, because we know ahead of time whether the entry_point is generated. We can use a single lookup.

This also makes sure that programs run in the location where the linker will put them (note that the logic in question runs before the linker has created the node_modules directory in the execroot.)
This is why #1787 observes that some node programs were broken - we were running the entry point as
bazel-out/host/bin/external/npm/@graphql-codegen/cli/bin/graphql-codegen.sh.runfiles/npm/node_modules/@graphql-codegen/cli/bin.js
when it should have been
node_modules/@graphql-codegen/cli/bin.js

Fixes #1787

### DIFF
--- a/internal/node/launcher.sh
+++ b/internal/node/launcher.sh
@@ -191,10 +191,7 @@ for ARG in "${ALL_ARGS[@]:-}"; do
   case "$ARG" in
     --bazel_node_modules_manifest=*) MODULES_MANIFEST="${ARG#--bazel_node_modules_manifest=}" ;;
     --nobazel_patch_module_resolver)
-      declare MAIN="TEMPLATED_entry_point_execroot_path"
-      if [[ ! -f "$MAIN" ]]; then
-        MAIN=$(rlocation "TEMPLATED_entry_point_manifest_path")
-      fi
+      MAIN=TEMPLATED_entry_point_execroot_path
       LAUNCHER_NODE_OPTIONS=( "--require" "$node_patches_script" )
 
       # In this case we should always run the linker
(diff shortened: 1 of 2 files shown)


alexeagle:
This is broken by targets which rely on a generated `bin` because those are *still* not getting linked.

We need https://github.com/bazelbuild/rules_nodejs/pull/1455 for this to pass

PR #1825, opened 2020-04-14
chore: add bazel-out and dist folders to bazelignore

alexeagle:
In the projects created with @bazel/create and in our examples.
Fixes issue reported on slack, where building examples/react_webpack fails on the second run
(because a glob in ts_project was picking up sources under bazel-out)

### DIFF
--- a/e2e/jasmine/.bazelignore
+++ b/e2e/jasmine/.bazelignore
@@ -1,1 +1,3 @@
 node_modules
+dist
+bazel-out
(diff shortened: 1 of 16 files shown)


gregmagolan:
@alexeagle Cherry pick onto 1.x branch?

alexeagle:
let's do lazy cherry-pick when/if we do another 1.x release. Will be a bit of archeology through the commit history

PR #1839, opened 2020-04-20
feat(examples): show how to use ts_library(use_angular_plugin) with w…

alexeagle:
…orker mode


marcus-sa:
I just tested it out.. and I must say that it's about several minutes slower than using `ng_module` per action.

PR #1842, opened 2020-04-21
feat(builtin): add DeclarationInfo to js_library

gregmagolan:
Adds a declaration_info provider factory which simplifies ts_project. Also add a ts_project test to validate that js_library be a dep to ts_project.

BREAKING CHANGE:

Removed provide_declarations() factory function for DeclarationInfo. Use declaration_info() factory function instead.

To discuss is whether or not to promote js_library to public API for 2.0?

### DIFF
--- a/internal/providers/declaration_info.bzl
+++ b/internal/providers/declaration_info.bzl
@@ -36,8 +36,14 @@ This prevents needing an aspect in rules that consume the typings, which improve
 def declaration_info(declarations, deps = []):
     """Constructs a DeclarationInfo including all transitive declarations from DeclarationInfo providers in a list of deps.
 
-Returns a single DeclarationInfo.
-"""
+    # TODO: add some checking actions to ensure the declarations are well-formed and don't have semantic diagnostics
+
+    Args:
+        declarations: list of .d.ts files
+        deps: list of labels of dependencies where we should collect their DeclarationInfo to pass transitively
+    Returns:
+        a single DeclarationInfo provider
+    """
     transitive_depsets = [declarations]
     for dep in deps:
         if DeclarationInfo in dep:
--- a/internal/providers/declaration_info.bzl
+++ b/internal/providers/declaration_info.bzl
@@ -14,20 +14,11 @@
 
 """This module contains a provider for TypeScript typings files (.d.ts)"""
 
-def provide_declarations(**kwargs):
-    """Factory function for creating checked declarations with externs.
-
-    Do not directly construct DeclarationInfo()
-    """
-
-    # TODO: add some checking actions to ensure the declarations are well-formed
-    return DeclarationInfo(**kwargs)
-
 DeclarationInfo = provider(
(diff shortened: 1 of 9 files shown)


internal/providers/declaration_info.bzl
```diff
@@ -41,3 +41,21 @@ This prevents needing an aspect in rules that consume the typings, which improve
         "type_blacklisted_declarations": """A depset of .d.ts files that we should not use to infer JSCompiler types (via tsickle)""",
     },
 )
+
+def declaration_info(declarations, deps = []):
```
alexeagle, line 45: I think this is meant to be `provide_declarations` above

gregmagolan: Was `provide_declarations` added as that is what it is called in g3? Not used at all in this repo & the pattern for the js provider factories is that the function is the same name as the provider but in lower case with underscores. js_module_info(), js_named_module_info(), etc...

Shall we just remove `provide_declarations` then for 2.0?

internal/providers/declaration_info.bzl
```diff
@@ -41,3 +32,21 @@ This prevents needing an aspect in rules that consume the typings, which improve
...
+    """Constructs a DeclarationInfo including all transitive declarations from DeclarationInfo providers in a list of deps.
+
+Returns a single DeclarationInfo.
+"""
+    transitive_depsets = [declarations]
+    for dep in deps:
```
alexeagle, line 42: keep the TODO here - we want to create an action to validate the .d.ts files at this point so a downstream compilation doesn't fail spooky-action-at-a-distance (and maybe use `--skipLibCheck` more safely)

packages/typescript/test/ts_project/js_library/BUILD.bazel
```diff
@@ -0,0 +1,16 @@
+load("@npm_bazel_typescript//:index.bzl", "ts_project")
+load("//internal/js_library:js_library.bzl", "js_library")
+
+js_library(
+    name = "lib_a",
+    srcs = [
+        "a.d.ts",
```
alexeagle, line 7: any thought about how these would typically be constructed? is this just for the case where you vendor a node_module? otherwise I'm not sure why you'd be authoring .js+.d.ts and if you didn't author it you'd get it through a node_module_lib

packages/typescript/test/ts_project/js_library/tsconfig.json
```diff
@@ -0,0 +1,15 @@
+{
+    "files": ["b.ts"],
+    "compilerOptions": {
```
alexeagle, line 3: this shouldn't be needed here since the .d.ts was in the source tree right?

alexeagle: oh, I see, js_library puts all its outputs into the out dir, so even though there's a .d.ts in the sources it's not the one the ts_project rule will see.
that's pretty confusing for users :(

alexeagle:
@googlebot I consent.

PR #1843, opened 2020-04-21
feat(builtin): add JSModuleInfo provider

gregmagolan:
**BREAKING CHANGE**

Adds JSModuleInfo provider as the common provider for passing & consuming javascript sources and related files such as .js.map, .json, etc.

For 1.0 we added JSNamedModuleInfo and JSEcmaScriptModuleInfo which were provided by ts_library and consumed by rules that needed to differentiate between the two default flavors of ts_library outputs (named-UMD & esm). We left out JSModuleInfo as its use case was unclear at the time.

For 2.0 we're adding JSModuleInfo as generic javascript provided for the rules_nodejs ecosystem. It is not currently opinionated about the module format of the sources or the language level. Consumers of JSModuleInfo should be aware of what module format & language level is being produced if necessary.

The following rules provide JSModuleInfo:

* ts_library (devmode named-UMD .js output flavor)
* ts_proto_library (devmode named-UMD .js output flavor)
* node_module_library (this is a behind the scenes rule used by yarn_install & npm_install)
* js_library (.js, .js.map & . json files)
* rollup_bundle
* terser_minfied
* ts_project

The following rules consume JSModuleInfo:

* nodejs_binary & nodejs_test (along with derivate macros such as jasmine_node_test); these rules no longer consume JSNamedModuleInfo
* npm_package_bin
* pkg_npm; no longer consumes JSNamedModuleInfo
* karma_web_test (for config file instead of JSNamedModuleInfo; JSNamedModuleInfo still used for test files)
* protractor_web_test (for config & on_prepare files instead of JSModuleInfo; JSNamedModuleInfo still used for test files)
* rollup_bundle (if JSEcmaScriptModuleInfo not provided)
* terser_minified

### DIFF
--- a/internal/js_library/js_library.bzl
+++ b/internal/js_library/js_library.bzl
@@ -17,7 +17,7 @@
 DO NOT USE - this is not fully designed, and exists only to enable testing within this repo.
 """
 
-load("//:providers.bzl", "LinkablePackageInfo", "declaration_info")
+load("//:providers.bzl", "LinkablePackageInfo", "declaration_info", "js_module_info")
 load("//third_party/github.com/bazelbuild/bazel-skylib:rules/private/copy_file_private.bzl", "copy_bash", "copy_cmd")
 
 _AMD_NAMES_DOC = """Mapping from require module names to global variables.
@@ -52,6 +52,7 @@ def write_amd_names_shim(actions, amd_names_shim, targets):
 def _impl(ctx):
     files = []
     typings = []
+    js_files = []
 
     for src in ctx.files.srcs:
         if src.is_source and not src.path.startswith("external/"):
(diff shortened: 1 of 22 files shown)


gregmagolan:
> could you update the commit description to explain what new use case is working now? I think from the terser tests you're doing something novel but it's hard to reverse-engineer

added description; terser tests suck in here somehow from an old version of this PR. I cleaned them up and they are just making basic assertions now that are easy to follow.

alexeagle:
It's a little fishy to assume we know all the js files by extension - no `.cjs`, `.mjs`, or assets that react code requires for example.
Also still would like to understand what this is good for, would we say anything in release notes about it? Makes it easier to write custom rules that interact with ours?

gregmagolan:
> It's a little fishy to assume we know all the js files by extension - no `.cjs`, `.mjs`, or assets that react code requires for example.
> Also still would like to understand what this is good for, would we say anything in release notes about it? Makes it easier to write custom rules that interact with ours?

Yes. For js_library that was the only fishy bit that I paused on and thought about improving but since that is still private API I let it go for now as can be fixed with breaking anyone.

Issue #1126, opened 2019-09-12
Add angular universal (SSR and PWA) to the example


alexeagle:
Probably for @Toxicable to do sometime.

Toxicable:
Ill take a look at writing a rule which calls the @angular/pwa cli manifest generation tool.
Since that should be low cost to put in.

Then(or in parallel) I can take a look at doing some universal stuff aswell

Toxicable:
I did a small start on running universal under bazel but ran into a few issues. 
Ill revisit when we support v9 since that resolve a few of them

marcus-sa:
Anything that I can help with regarding PWA support?

alexeagle:
hey @marcus-sa can you get something working under https://github.com/bazelbuild/rules_nodejs/tree/master/examples/angular ? you can just `cd` into that directory and hack. If you get something working we can merge it to the example

marcus-sa:
@alexeagle I've just fiddled around with it and IMO it would be better to create a Bazel definition for it called `ng_service_worker` in the `@angular/bazel` package as there would be a lot of redundant config/code needed for end users to be able to use it properly, since they'd need to copy the required files such as `ngsw-worker.js`, `safety-worker.js` and `worker-basic.min.js`.
This is done automatically by the Angular CLI as seen here:
https://github.com/angular/angular-cli/blob/da6d74a86e208cc5dc0a83718fb9122ca31addd4/packages/angular_devkit/build_angular/src/browser/index.ts#L682

flolu:
Are there any updates on Bazel with Angular Universal? This is the last step for my migration to Bazel, and I really need it.
If there is no progress, then how could it be implemented?

PR #1847, opened 2020-04-25
feat(example): pwa support

flolu:
# Web Manifest and Service Worker
manifest_sw

* The `service-worker.service.ts` will automatically update the Service Worker if there is a newer version available. This will cause the page to reload (maybe we should add a snackbar to ask the user to "update now" in the future)

# Lighthouse Audit
**Before**
audit_before

**After**
audit_after
- **Performance** might be improved by loading some content async?!
- **Accessibility** and **SEO** can easily be improved to 100% (just needs some changes in the html)
- **Best Practices** is at 93% because the app was served on localhost which has no HTTPs

_I've also added `compression` to the SSR server which decreased the content to load from 1.4 MB to just 321 kB!_

# Install App (Desktop and Mobile)
install

Huge thank to @marcus-sa! (He helped me with the implementation and is the author of `ngsw_config.bzl` which was the key to add the service worker.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

The Angular example application has no PWA support.

Issue Number: #1126

## What is the new behavior?

Full PWA support with web manifest, offline caching, service worker etc.

- [ ] Yes
- [x] N


examples/angular/src/server.ts
```diff
@@ -1,41 +1,28 @@
...
+const app = express();
+const port = process.env.PORT || 4000;
+const DIST_FOLDER = join(process.cwd(), 'src/pwa');
+
+/**
+ * text compression for smaller download sizes and thus faster load times
+ * without compression: 1.4 MB
+ * with compresion: 321 kB
+ */
+app.use(compression());
```
Toxicable, line 20: I wouldn't advise to use compression on your application server, you'd usually host this behind a reverse proxy which can handle more algorithms, freeing up resoureces here to run the SSR process

examples/angular/src/BUILD.bazel
```diff
@@ -193,30 +196,34 @@ pkg_web(
...
         ":inject_scripts_for_prod",
         "//src/assets",
+        ":robots.txt",
+        # Service worker
+        "@npm//:node_modules/@angular/service-worker/ngsw-worker.js",
         # Include polyfills that will be requested by old browsers
         "@npm//:node_modules/systemjs/dist/system.js",
         "@npm//:node_modules/core-js/client/core.min.js",
-        "index.html",
+        # NOW needed ?"index.html",
```
Toxicable, line 205: whats up with this comment?

examples/angular/tools/ngsw_config.bzl
```diff
@@ -0,0 +1,38 @@
+"Angular service worker support (credits: https://github.com/marcus-sa)"
+
+load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_binary")
+
+def ngsw_config(name, config, index_html, src, out = None, **kwargs):
```
Toxicable, line 5: are we able to use `@angular/service-worker/ngsw-config` from `npm_package_bin` instead?

examples/angular/src/server.ts
flolu: Alright, that's good to know! I will remove it :)

examples/angular/src/BUILD.bazel
flolu: I forgot to remove it

examples/angular/tools/ngsw_config.bzl
flolu: Do you mean solely using `@angular/service-worker/ngsw-config` instead of wrapping it inside `nodejs_binary`?

Maybe something for @marcus-sa to look at?

marcus-sa: There was an issue with `npm_package_bin` at the moment of coding it with resolving data (pre v1), so I'll have a look at it and test it out.

examples/angular/package.json
```diff
@@ -18,8 +18,10 @@
         "@angular/platform-browser-dynamic": "9.1.0",
         "@angular/platform-server": "^9.1.0",
         "@angular/router": "9.1.0",
+        "@angular/service-worker": "^9.1.3",
         "@ngrx/store": "9.0.0-beta.0",
         "@nguniversal/express-engine": "^9.0.0",
+        "compression": "^1.7.4",
```
Toxicable, line 24: you wont be needing this anymore

examples/angular/src/server.ts
```diff
@@ -1,41 +1,21 @@
...
+///<reference types='node'/>
+import 'zone.js/dist/zone-node';
 
-import { ngExpressEngine } from "@nguniversal/express-engine";
-import * as express from "express";
-import { join } from "path";
+import { ngExpressEngine } from '@nguniversal/express-engine';
+import * as express from 'express';
+import { join } from 'path';
+import * as compression from 'compression';
```
Toxicable, line 7: or this

examples/angular/src/BUILD.bazel
```diff
@@ -237,21 +243,20 @@ container_image(
...
     name = "universal_server_lib",
-    srcs = [
-        "server.ts",
-    ],
+    srcs = ["server.ts"],
     deps = [
         "//src/app:app_server",
         "@npm//@nguniversal/express-engine",
         "@npm//@types/node",
+        "@npm//compression",
```
Toxicable, line 251: also this

flolu: I think my brain was half-asleep this morning :man_facepalming:

examples/angular/tools/ngsw_config.bzl
Toxicable: I've prefer for downstream users to not have to do this if we can avoid it.  
@alexeagle any ideas what might be up here?

Toxicable:
Just that one comment, otherwise LGTM

examples/angular/tools/ngsw_config.bzl
marcus-sa: Ah yes I remember, since `genrule` is consuming the `nodejs_binary` as a tool, whatever `srcs` being provided to `genrule` will not be accessible at runtime for `npm_package_bin`, which means you'll need to provide the `srcs` to the actually binary (`nodejs_binary`)

flolu:
@Toxicable You're talking about the `npm_package_bin`, right?
If so, I couldn't make it work (probably because I don't know how to use `npm_package_bin` correctly)

Maybe we can do this as a follow-up issue?

examples/angular/tools/ngsw_config.bzl
alexeagle: Something feels wrong-shaped there - npm_package_bin should replace the need for genrule. do you have a branch that demonstrates that problem?

flolu:
@alexeagle Probably correct. Maybe we will just close this for now and it can be used for reference in the future? I won't have time to add tests.

matthewjh:
@flolu

I'm getting very close to needing this on our codebase. Before I try it, just wanted to ask, did it work correctly? Any rough edges or bugs worth knowing of? Or is the only reason this didn't get merge lack of test coverage?

flolu:
@matthewjh Honestly, I think it's just the lack of testing.

I used [this](https://github.com/flolu/centsideas) project to experiment with this exact PWA implementation. The angular client application is not yet on the master branch, but you can look at aa older working implementation [here](https://github.com/flolu/centsideas/tree/0ea60938b2c0826511090a88cfb96fc27178a91f/services/client).

So what can I say, it worked perfectly. I implemented PWA and server side rendering with user authentication and a backend API. It is fast and scores almost perfectly on the frontend.

f18bd9f9-cc06-40ed-8d6d-d973b88b3f51

The only issue I ran into was, that Socket.io doesn't work with the ts_devserver by default. But someone helped me to find an easy [solution](https://stackoverflow.com/questions/61423031/bazel-angular-socketio-causes-uncaught-typeerror-xmlhttprequest-is-not-a-c).

**Conclusion**
I don't see any reason to not use this.
We are currently developing a marketplace and we will be using this in production in couple of months.

Let me know if you want to know anything else.
If you want to, we can keep in contact, so that we can support each other on this journey :smile:

marcus-sa:
@matthewjh we've been using it in production for almost half a year, but as @flolu said it lacks testing.
As long as it works (which it does), then I don't see a reason as to why you shouldn't use it as well 😁

alexeagle:
@googlebot I consent.

alexeagle:
Thanks @flolu !!

Issue #1823, opened 2020-04-12
Regression in 1.6.0 autogenerated test rules


purkhusid:
### Affected Rule

Not so sure if it's only autogenerated rules, but the repro I have is failing in a auto generated test rule.

### Is this a regression?
Yes, this works in 1.5.0.

### Description
Auto generated rules fails to load resolve module from a relative file during a test run.

## 🔬 Minimal Reproduction
https://github.com/purkhusid/ts-jest-repro

## 🔥 Exception or Error

<pre><code>
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
Executing tests from //:test
-----------------------------------------------------------------------------
[link_node_modules.js] manifest file /private/var/tmp/_bazel_danielpurkhus/6514ebbf801be819708572b35f1a35af/sandbox/darwin-sandbox/5/execroot/jest_repro/bazel-out/darwin-fastbuild/bin/test.sh.runfiles/jest_repro/_test.module_mappings.json
[link_node_modules.js] manifest contents {
  "workspace": "jest_repro",
  "bin": "bazel-out/darwin-fastbuild/bin",
  "root": "npm/node_modules",
  "modules": {
    "jest_repro": [
      "execroot",
      "bazel-out/darwin-fastbuild/bin"
    ]
  }
}
[link_node_modules.js] startCwd /private/var/tmp/_bazel_danielpurkhus/6514ebbf801be819708572b35f1a35af/sandbox/darwin-sandbox/5/execroot/jest_repro/bazel-out/darwin-fastbuild/bin/test.sh.runfiles/jest_repro
[link_node_modules.js] isExecroot false
[link_node_modules.js] resolved node_modules root npm/node_modules to /private/var/tmp/_bazel_danielpurkhus/6514ebbf801be819708572b35f1a35af/sandbox/darwin-sandbox/5/execroot/jest_repro/bazel-out/darwin-fastbuild/bin/test.sh.runfiles/npm/node_modules
[link_node_modules.js] cwd /private/var/tmp/_bazel_danielpurkhus/6514ebbf801be819708572b35f1a35af/sandbox/darwin-sandbox/5/execroot/jest_repro
[...]

PR #1850, opened 2020-04-26
ci: import repro from #1823

alexeagle:


WesleyYue:
How come this never got merged? It looks like it regressed again in 2.0.0 https://github.com/bazelbuild/rules_nodejs/issues/2008

PR #1863, opened 2020-05-03
feat: support bazel+js packages that install into regular location

alexeagle:
…ge:index.bzl location

This PR shows how we don't need nested workspaces, which means
- easier to work in the repo, things are either `//packages/foo/...` labels when using from source or trivially map to `//@bazel/foo` when used in the npm-installed repository
- we don't have to hard-code `@npm` anywhere so if you have installs named something else, everything still works
- most users won't have to `load("@npm//...")` in their `WORKSPACE` anymore, which means if you're in a multi-lang monorepo, the non-frontend devs don't have to install from npm at all
- probably means you can grab rules_nodejs from github HEAD, and if you swap your load statements `@npm//@bazel/foo` to `@build_bazel_rules_nodejs//package/foo` things still work (seems true for a couple limited testing nested workspaces in this PR)
- simpler mapping of binaries, by having their source version in a `bin/` folder matching where the generated target will be called in the installed version
- more consistent for users - always load from `@npm//package` whether package is Bazel-authored or not. No more `@npm_bazel_foo` convention and strange inconsistency


packages/terser/terser_minified.bzl
```diff
@@ -98,8 +98,7 @@ so that it only affects the current build.
     ),
     "terser_bin": attr.label(
         doc = "An executable target that runs Terser",
-        # NB: will be substituted with "@npm//@bazel/terser/bin:terser" in the pkg_npm
```
jbedard, line 101: Is this still true?

jbedard: Looks like it is... keep/edit the existing comment? I guess other spots such as 20 lines up don't though 🤷

alexeagle: Yeah all of `//packages/foo` get swapped to `//@bazel/foo` which seems obvious enough, and we'd have to comment that in a lot of places... but maybe worth it

internal/npm_install/generate_build_file.ts
```diff
@@ -186,24 +186,63 @@ node_module_library(
  * Generates all BUILD & bzl files for a package.
  */
 function generatePackageBuildFiles(pkg: Dep) {
+  // If a BUILD file was shipped with the package, append its contents to the end of
+  // what we generate for the package.
+  let buildFilePath: string|undefined;
+  if (pkg._files.includes('BUILD')) buildFilePath = 'BUILD';
```
gregmagolan, line 192: Could this logic be simplified?
1) copy over all .bzl & BUILD files in the package's tree
2) write out the generated contents to BUILD[.bazel], bin/BUILD[.bazel], index.bzl --- if the file exists append otherwise create?

packages/typescript/internal/ts_project.bzl
```diff
@@ -206,7 +211,7 @@ def ts_project_macro(
         composite = False,
         incremental = False,
         emit_declaration_only = False,
-        tsc = _DEFAULT_TSC,
+        tsc = None,
```
jbedard, line 214: What's this change?

alexeagle: that's because the macro evaluates labels in the callers workspace
so `//typescript/bin` becomes `@my_wksp//typescript/bin`
but inside the rule it's evaluated in the context of the repo declaring the rule, so it becomes @npmwhatever//typescript/bin like we want
anyway it was always silly for that macro to bother with a default since the rule already had it

internal/npm_install/generate_build_file.ts
alexeagle: let's take a stab at that in a follow-up, but it's not that simple - I think it's naive to just merge them in some cases.

PR #1864, opened 2020-05-04
chore: cleanup jasmine from_src index

alexeagle:
Instead, use replacements so the distribution uses different labels than we use locally


gregmagolan:
Used https://github.com/bazelbuild/rules_nodejs/pull/1860 instead

PR #1874, opened 2020-05-06
refactor(jasmine): no longer a nested package

alexeagle:


packages/jasmine/jasmine_node_test.bzl
```diff
@@ -62,10 +62,10 @@ def jasmine_node_test(
         expected_exit_code = 0,
         tags = [],
         config_file = None,
-        # Replaced by pkg_npm with jasmine = "@npm//@bazel/jasmine",
-        jasmine = "@npm_bazel_jasmine//:jasmine__pkg",
+        # Replaced by pkg_npm with jasmine = "//@bazel/jasmine",
```
jbedard, line 65: Isn't it replaced with `@npm//@bazel/jasmine`?

alexeagle: nope, we don't want to hard-code that the user names their repository "@npm" because it only works in the simple case.

The rule will execute in the `@npm_whatever` repo so it shouldn't be fully-qualified

gregmagolan:
@googlebot I consent.

gregmagolan:
@jbedard There was a linker issue where `@bazel/jasmine` link path was reduced to `@bazel` but that path conflicted with the existing `node_modules/@bazel` folder so the symlink could not be created. I changed the logic to _only_ reduce if the parent is itself a link and updated all of the unit tests accordingly. Can you PTAL at the change & see if that will still meet your use case?

internal/linker/link_node_modules.ts
```diff
@@ -439,21 +439,6 @@ function liftElement(element: LinkerTreeElement): LinkerTreeElement {
     return {name, link};
   }
 
-  // No link but all child elements have aligning links
```
jbedard, line 442: I think this was just an "optimization" to avoid creating unnecessary folders just to symlink 1+ child elements within those folders. So if it was just an "optimization" then it would be ok to drop it?

You're saying this could be a problem when `a` is something special such as `@bazel`?

jbedard: I guess this made would convert real directories to links which was the issue? Where the logic above this would only convert links to simpler links, not real directories to links...

jbedard:
Are all the test changes necessary to get them to pass?

Maybe add a test explaining the issue, or change the description of an existing test?

PR #1876, opened 2020-05-06
Protractor nested workspace

alexeagle:


alexeagle:
Oh, that gross thing with `_relative` to make a label relative to the workspace containing the macro - that's what the `Label` constructor does! cleanup on the way

Issue #1775, opened 2020-04-01
Add prerendering with angular to examples


flolu:
# 🚀 feature request

### Description

Add an angular module to the [example angular app](https://github.com/bazelbuild/rules_nodejs/tree/master/examples/angular) which supports [prerendering](https://dev.to/michaeljota/how-to-prerender-your-angular-app-using-angular-universal-4g0b).

alexeagle:
That sounds cool, it looks like this is specifically build-time rendering (run angular as a nodejs program in a build step to produce a populated index.html file that goes into the app resources) as opposed to having any server at runtime doing it?

flolu:
@alexeagle Exactly! This is super useful for static pages like "About", "Terms" or even the Home page.

I created the issue because the implementation currently exceeds my own knowledge, otherwise I would have done a pull request. :laughing:

mattem:
I had an [example repo](https://github.com/Evertz/bazel-angular-prerender) for this some months ago. Things have probably changed since then, but should be able to bring it back up to date. I can take this one.

flolu:
@mattem Awesome! I'll check it out later. Thank you :)

mattem:
@flolu Added the prerendering example to examples/angular, let me know how you get on

closed by 8509d9c

flolu:
@mattem Awesome! Thank you so much!

PR #1879, opened 2020-05-10
example: add prerendering to example Angular app

mattem:
Adds example prerendering to the Angular app in examples, will prerender the root route, hello and todos. These new index files are then passed though `html_insert_assets` and served by the prodserver.

It might be worth adding a wrapping macro around `html_insert_assets` (and possibly the rest of `pkg_web`) to reduce the verbosity of each additional prerenderd route being added.

TODO: 
- [x] Update / add e2e tests

Related issue: #1775


mattem:
Some pointers how to run this against HEAD would be good - I based this off the 1.6.1 release commit then rebased, but HEAD has diverged quite a bit. The load sites are what's causing the CI failures.

mattem:
Ah, figured it out. Firefox fails locally for me but everything seems to pass on CI

mattem:
@alexeagle Are you good with the added comments and docs here?

PR #1898, opened 2020-05-17
try ts_project -> ts_project dep with named module

alexeagle:


longlho:
@alexeagle  any update on this? :)

alexeagle:
@gregmagolan and I discussed it yesterday. We aren't sure about adding package_name attributes to lots of rules. Feels like it will end up splattered in a lot of places. Probably better to figure out how to put a package rule in the middle so we solve it with composition. Might not make 2.0 but it's non breaking so maybe 2.1

longlho:
gotcha. So do u have any guidance for the solving this in the short term? (and for people porting from `lerna` basically)

alexeagle:
If you can name the workspace the same as a top-level package, then change
your imports to use that top-level package, then all your imports are deep
imports within the same package and that works. But of course you might not
have the freedom to do that if you publish them outside the workspace using
the different package names as references between them.
You can patch this one file
<https://github.com/bazelbuild/rules_nodejs/pull/1924/files#diff-ab3f36c8dfcccf8f7b945c8dc42d810f>
into
your node_modules (using patch-package maybe) and start putting module_name
on your ts_project's but you might need to change that in the future when
we have a better way. I do think ts_project is pretty stable so there's not
a lot of risk taking on that patch locally.

longlho:
circling back to this @alexeagle turns this is not enough. I think I'm actually looking for `pkg_npm` to be linkable. It contains the final shape of the ready-to-publish npm tarball that has non-code artifacts (e.g data, styles, config...) so `ts_project` is not enough :( Any suggestion here?

PR #1902, opened 2020-05-20
chore(examples): add example for ng packagr

lukasholzer:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

This example should demonstrate the usage of packaging an angular
library with the `ng_package` rule from [@angular/bazel](https://angular.io/guide/bazel) in combination with the `ts_library` rule in combination with the angular plugin.

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/examples/angular/package.json
+++ b/examples/angular/package.json
@@ -60,6 +60,7 @@
         "rollup": "^1.21.4",
         "rollup-plugin-commonjs": "^10.1.0",
         "rollup-plugin-node-resolve": "^5.2.0",
+        "rollup-plugin-sourcemaps": "^0.6.2",
         "terser": "4.3.1",
         "typescript": "3.8.3"
     },
--- a/examples/angular/src/lib/angular-lib/BUILD.bazel
+++ b/examples/angular/src/lib/angular-lib/BUILD.bazel
@@ -1,0 +1,14 @@
+load("@npm_angular_bazel//:index.bzl", "ng_package")
+
+package(default_visibility = ["//:__subpackages__"])
+
+ng_package(
+    name = "angular-lib",
+    srcs = [
+        "package.json",
+    ],
+    entry_point = ":index.ts",
+    deps = [
+        "//src/lib/angular-lib/button",
+    ],
+)
--- a/examples/angular/src/lib/angular-lib/button/BUILD.bazel
+++ b/examples/angular/src/lib/angular-lib/button/BUILD.bazel
@@ -1,0 +1,26 @@
+load("@io_bazel_rules_sass//:defs.bzl", "sass_binary")
+load("//tools:angular_ts_library.bzl", "ng_ts_library")
+
+package(default_visibility = ["//:__subpackages__"])
+
+ng_ts_library(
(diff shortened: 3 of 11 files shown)


alan-agius4:
@alexeagle, I got the same impression. 

I'd definitely recommend using `load("@npm//ng-packagr:index.bzl", "ng_packagr")` unless you really need more fine grain library builds.

lukasholzer:
@alan-agius4 I did The steps suggested by the Bazel Angular Docs https://angular.github.io/bazel-builds/src/ng_package/ng_package.html#ng_package

So I've used the installed `@angular/bazel` package where the rule `load("@npm_angular_bazel//:index.bzl", "ng_package")` is coming from.

alexeagle:
I think the Angular team is deprecating the `@angular/bazel` package (it was in Labs) so we'd rather not show users how to make new dependencies on it. The `ng_package` rule is likely to be removed from the public API. Instead we want everyone to use the supported mechanism which is the ng-packagr library.

alexeagle:
It looks like we have an ng-packagr example already here
https://github.com/bazelbuild/rules_nodejs/blob/master/examples/angular_bazel_architect/angular.json#L139
so I think this can be closed?

lukasholzer:
@alexeagle It is an example with the architect – I think there should be an example without using the architect as the architect only allows incremental rebuilds per library and not as granular as it should be with bazel. 

I'm not sure if the packagr is needed for the angular package format or if the same can be achieved with rollup? But I think there should be an example of creating an Angular publishable library (Angular Package Format) with the benefits of incremental builds.

alan-agius4:
@jbedard, was also looking to do something similar.

alan-agius4:
> @alexeagle It is an example with the architect – I think there should be an example without using the architect as the architect only allows incremental rebuilds per library and not as granular as it should be with bazel.

This is that same for this PR, there is no different in using my-packagr directly via using the architect-cli.

jbedard:
This seems to be `@angular/bazel`, not `ng-packagr`...?

alan-agius4:
Ohh, the title of this PR always confuses me 😓

jbedard:
> Ohh, the title of this PR always confuses me 😓

I agree :P

alexeagle:
updated the title to explain - the issue is when we drop `@angular/bazel` you lose the `ng_package` rule and it would be nice to have a replacement

jbedard:
I was trying `ng-packagr` but as a plain npm package it gets invoked quite differently then `ng_package`. I got `ng_package` working (thanks @gregmagolan !) by doing: https://github.com/angular/angular/pull/37727#issuecomment-656878664

alexeagle:
Yeah given time constraints I think you should stick with @angular/bazel for now.
Other angular ecosystem projects like components and ngrx use it so there's no risk it disappears.

alan-agius4:
@alexeagle, actually it seems that NGRX are moving away from Bazel in favor of NX with ng-packagr.

PR #1905, opened 2020-05-24
feat(builtin): create AssetInfo provider for JS targets

mrmeku:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Given that JS bundlers such as webpack and rollup both support
the loading of arbitrary files given the appropriate plugins,
it makes sense to associate asset files that will be needed
during bundling with a particular target so that downstream
bundling rules do not need to depend on those assets themselves.

- [ ] Yes
- [x] No

### DIFF
--- a/examples/react_webpack/BUILD.bazel
+++ b/examples/react_webpack/BUILD.bazel
@@ -14,6 +14,9 @@ sass(
 )
 
 ts_project(
+    assets = [
+        "styles.css",
+    ],
     deps = [
         "@npm//@types",
         "@npm//csstype",
@@ -32,8 +35,8 @@ webpack(
     ],
     data = [
         "index.js",
-        "styles.css",
         "webpack.config.js",
+        ":tsconfig",
         "@npm//:node_modules",
     ],
 )
--- a/internal/js_library/js_library.bzl
+++ b/internal/js_library/js_library.bzl
@@ -17,7 +17,7 @@
 DO NOT USE - this is not fully designed, and exists only to enable testing within this repo.
 """
 
-load("//:providers.bzl", "LinkablePackageInfo")
+load("//:providers.bzl", "AssetInfo", "LinkablePackageInfo")
 load("//third_party/github.com/bazelbuild/bazel-skylib:rules/private/copy_file_private.bzl", "copy_bash", "copy_cmd")
 
 _AMD_NAMES_DOC = """Mapping from require module names to global variables.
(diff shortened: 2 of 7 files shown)


jbedard:
Would it be possible to also add this to `ts_library`?

mrmeku:
I want to add it to ts_library but I'm a little unsure how to do so. Rules typescript is imported from rules nodejs so if I'll cause a cycle if I have it import this provider directly.

In general, I think I would like to see ts_library pulled into rules_nodejs so we can consolidate our effort in this monorepo.

@alexeagle @gregmagolan If you are both okay with me pulling in ts_library let me know. Or if you have a workaround that'd you'd prefer which doesn't cause the cycle I mentioned I'd love to know about it

jbedard:
If I understand correctly the `ts_library` only consumes functions, providers etc from rules_typescript, the [rule](https://github.com/bazelbuild/rules_nodejs/blob/1.6.1/packages/typescript/src/internal/build_defs.bzl#L336) itself is in rules_nodejs

Can the `assets` go right there alongside the [angular_assets](https://github.com/bazelbuild/rules_nodejs/blob/1.6.1/packages/typescript/src/internal/build_defs.bzl#L344-L348)?

alexeagle:
Yeah it's same story as DeclarationInfo

alexeagle:
@jbedard do you think this should interact with `html_insert_assets` in some way?

jbedard:
If the assets need to be inserted in the HTML file which is possible. I thought these assets would more likely be things which your js/ts or css reference (although for the css one the `sass_*` libraries should really support that :/).

alexeagle:
Hmm as a user it seems like it will be confusing that AssetInfo is only for TS -> bundlers and not related to assets getting injected into the page or served independently. How can we make this more clear?

jbedard:
> AssetInfo is only for TS -> bundlers

I was assuming anything that has runtime assets would output `AssetInfo`. Things like CSS reference images/fonts assets, HTML references css/html/image assets. For example `sass_binary` should have `assets` for assets referenced by the resulting css. Then rules like `pkg_web` would pull all those assets up into the package to be served at runtime...

Normally a small subset of those assets need to be inserted into an html file (favicon, initial css/js). How that subset of files are specified is another issue today... often you have a `filegroup` (or `AssetInfo`) of resources and only want a subset inserted into that html file. How can you reference that subset? One example today is the output of `rollup_bundle`... you want that .js inserted into the html but not the .map file, but both of those are outputted by `rollup_bundle` and I don't think can be referenced independently?

PR #1911, opened 2020-05-26
feat(angular): introduce an Angular CLI builder

alexeagle:
This replaces the one being removed in https://github.com/angular/angular/pull/37190

/cc @alan-agius4 @kyliau

### DIFF
--- a/packages/angular/README.md
+++ b/packages/angular/README.md
@@ -1,0 +1,29 @@
+# Angular support for Bazel
+
+This package is a replacement for parts of the deprecated @angular/bazel package previously maintained by the Angular team.
+
+Currently, this only provides an Angular CLI Builder, which can execute Bazel when triggered by `ng build`, `ng test`, etc.
+See https://angular.io/guide/cli-builder for more info about Builders.
+
+This builder assumes you have already created Bazel configurations (WORKSPACE and BUILD files).
+There is presently no tooling to generate these automatically that's supported by either Angular team or rules_nodejs maintainers.
+See the `@bazel/create` package for a quickstart to creating a Bazel workspace, or look at examples in rules_nodejs.
+
+To use it, you would just install this package (it doesn't hook into `ng add` because it has no schematics):
+
+```sh
+$ npm install --save-dev @bazel/angular
+```
+
+Then edit your `angular.json` to invoke Bazel. For example, to have `ng build` do `bazel build //:all` you would edit the `architect` block to have:
+
+```json
+"architect": {
+  "build": {
+    "builder": "@bazel/angular:build",
+    "options": {
+    "targetLabel": "//:all",
+    "bazelCommand": "build"
+  }
+}
+```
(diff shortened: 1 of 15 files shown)


packages/angular/tsconfig.json
```diff
@@ -0,0 +1,7 @@
+{
+    "compilerOptions": {
+        "strict": true,
+        "target": "ES2015",
```
alan-agius4, line 4: This can be ES2018 since Node supports it.

packages/angular/src/builders/index.ts
```diff
@@ -0,0 +1,47 @@
...
+        ps.kill('SIGTERM')
+      }
+
+      ps.on('SIGINT', shutdown);
+      ps.on('SIGTERM', shutdown);
+      return new Promise(resolve => {ps.on('close', e => resolve({success: e === 0}))})
+
+    } catch (err) {
+      context.logger.error(err.message);
+      return Promise.resolve({success: false});
```
alan-agius4, line 42: return Promise.resolve({success: false}); -> return{success: false};

packages/angular/src/builders/index.ts
```diff
@@ -0,0 +1,47 @@
...
+ */
+
+import {BuilderContext, BuilderOutput, createBuilder} from '@angular-devkit/architect';
+import {JsonObject} from '@angular-devkit/core';
+import {getNativeBinary as bazeliskBin} from '@bazel/bazelisk/bazelisk';
+import {getNativeBinary as ibazelBin} from '@bazel/ibazel';
+import {spawn} from 'child_process';
+import {Schema} from './schema';
+
+function _bazelBuilder(
```
alan-agius4, line 18: This can be an `async` function, and instead of using `return Promise.resolve({success: false});` you can simple `return {success: false}`.

packages/angular/src/builders/index.ts
```diff
@@ -0,0 +1,47 @@
...
+    options: JsonObject&Schema,
+    context: BuilderContext,
+    ): Promise<BuilderOutput> {
+  const {bazelCommand, targetLabel, watch} = options;
```
alan-agius4, line 22: Nice 👍

packages/angular/types.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module '@bazel/bazelisk/bazelisk' {
+  function getNativeBinary(): Promise<number>| string
+} declare module '@bazel/ibazel' {
```
alan-agius4, line 3: Nit: new line after `{`

packages/angular/src/builders/index.ts
alexeagle: Ah, I see, I can still return a Promise in the case where I need to nest the resolution of the promise inside the ps.on callback. See how often I do async programming lately...

packages/angular/tsconfig.json
alexeagle: cool

packages/angular/README.md
```diff
@@ -0,0 +1,19 @@
...
+See the `@bazel/create` package for a quickstart to creating a Bazel workspace, or look at examples in rules_nodejs.
+
+To use it, you would
+
+```sh
+$ npm install @bazel/angular
+# Edit your angular.json file to use it
+# See examples in this repo
+$ ng build
+```
```
gregmagolan, line 19: nit: add a bit of docs on what goes in the angular.json file

alexeagle: done

PR #1924, opened 2020-05-28
feat(typescript): ts_project is linkable

alexeagle:
This doesn't copy a package.json file, so it has to be resolvable with just index.ts


alexeagle:
we don't want more rules to grow a package_name attribute, instead we should use node_module_library

PR #1932, opened 2020-06-02
chore(examples): upgrade angular example to angular 10.0.0-rc

jbedard:
Current causes:
```
ERROR: /.../rules_nodejs/examples/angular/src/app/hello-world/BUILD.bazel:13:1: Compiling TypeScript (devmode) //src/app/hello-world:hello-world failed (Exit 1)
Compilation failed Error: ENOENT: no such file or directory, open '/private/var/tmp/.../execroot/examples_angular/bazel-out/darwin-fastbuild/bin/src/shared/material/material.module.ngtypecheck.d.ts'
```


alan-agius4:
This is the result that In version 10, incremental type-checking was introduced in Ivy, instead of a global type-checking file we now have a type-checking file for each file.

There are two problems here that I can see here.
1. In `tsc_wrapped`, the return of `setupCompilation` is ignored. https://github.com/bazelbuild/rules_typescript/blob/c4b325859b8d2c0579b094c67f449f1c3143271e/internal/tsc_wrapped/tsc_wrapped.ts#L389 This returns 2 pieces of information, files are shouldn't be emitted and files which shouldn't  be checked for diagnostics.  This will cause 2 extra files to be emitted for each original source file. `myfile.ngtypecheck.d.ts` and `myfile.ngtypecheck.js`.

2. In the TS source file, a triple-slash reference to the type-checking file is added, however this is not removed when emitting the DTS, this seems to be more of a bug in the plugin itself.

I am not sure if maybe the first issue should be handled directly by NG Compiler, maybe sometime similar to `ngtsc.wrapHost`,  because I am seeing that anyone who uses the plugin API will need to filter out those files from being emitted.

I did create a reproduction without Bazel https://github.com/alan-agius4/ngtsc-plugin-demo. 

//cc @alxhub

alexeagle:
@alan-agius4 thanks for digging into the cause, sounds like that is not something we can fix in rules_nodejs. Do you know if this issue is triaged to be fixed before Angular 10 final?

alan-agius4:
Hi @alexeagle, I don't think it's triaged. I'll ping @alxhub again.

alan-agius4:
Indeed the fix itself needs to happen at a NG plugin level, Unless you want to add temporary patch in `rules_typescript`.

That said, it's not clear to me who will address point 1 in `rules_typescript`.

alexeagle:
I guess it goes without saying that this plugin API is load-bearing for the
community who has opted-in to using Bazel already, as ng_module rule is
going unsupported in Angular 10, and our examples have stopped using
the @angular/bazel package. So there seems to be a mismatch in how we are
planning for this migration. The plugin is working today so I think it
needs to continue to work... I'll mention this to Igor.

marcus-sa:
I don't know whether or not to be sad that `ng_module` becomes unsupported in v10, because the new `ts_library(use_angular_plugin = True)` is about 10x slower than `ng_module`, but has support for strict transitive dependencies.. at least from what I've experienced in a project using 100+ components.

alexeagle:
@marcus-sa IIRC, a couple rules_nodejs contributors also saw ts_library being slower at first, but then in later testing it was a 20% improvement. I don't remember whether it was a user config change or a bug fix that made it faster. Could you try making this change again and do another measurement? If it is actually still slower, then we need to fix that.

marcus-sa:
@alexeagle I just benchmarked and it doesn't seem to be the case anymore.
We got strict transitive dependency diagnostics now, so hell yeah!

alexeagle:
Great to hear. What numbers did you get? Is ts_library faster?

Issue #231, opened 2018-06-14
Port to linux/arm64


vielmetti:
Support arm64 in these builds. There's a fully supported Node.JS for arm64, but Bazel doesn't know how to fetch the right images.

At minimum this is changes at https://github.com/bazelbuild/rules_nodejs/blob/master/internal/common/os_name.bzl#L28 plus pointers to the right Node.JS images in https://github.com/bazelbuild/rules_nodejs/blob/master/internal/node/node_repositories.bzl

alexeagle:
We also allow you to point to your own nodejs or even vendor it in your repo - does one of those options work for you?

vielmetti:
@alexeagle - the goal for arm64 support is full parity with other platforms to allow completely automated build enviornments. Patching and vendoring etc do not accomplish that goal.

alexeagle:
Sorry this has gone a while with no reply.

Why does vendoring arm64 version of node into the repo not solve the problem? You'd check in the copy of node you want to use and bazel in your build environment would call it. Is it because you also want to use the binaries for the other platforms in the same build?

photex:
Just skimming issues as I start to tread water with bazel here; but I thought I'd chime in to say it's my goal to be able to use bazel to cross-compile native node modules for aarch64 in addition to having modules built on the host architecture which is x86_64. 
We have embedded systems components and desktop components and while it's largely C, there are Python and Node tools and libraries that need to be built for one or both architecture.

enriched:
Chiming in here as well, it seems like support for this kind of use case is supposed to be coming in September based on the [Bazel Configurability Roadmap](https://bazel.build/roadmaps/configuration.html). Or maybe they are working on it this month based on the `User-Defined Configuration` section.  Does rules_nodejs already support these kinds of options?

My current situation is on a mac:
```
[typescript] -> [javascript] -> [Docker image] -> [Local k8s]
```
with some external npm modules sprinkled throughout. Unfortunately the `node_image` rule from [rules_docker](https://github.com/bazelbuild/rules_docker) doesn't work for me because it's adding in the `@nodejs//:node` binary and when you're running the build locally it will not run in the `gcr.io/google-appengine/debian9` image.

Toxicable:
https://github.com/bazelbuild/rules_nodejs/pull/645 would solve this

alexeagle:
Since we have toolchain support now maybe this is easy to solve? Could use someone trying it out

vielmetti:
[...]

PR #1936, opened 2020-06-04
Add linux arm64 support

kgreenek:
Connects to #231 

A start at adding support for linux/arm64. This unblocks me for my current use-case, but there is still work to do.

When I run `bazel test //...` on an aarch64 device, currently it fails with the following output:

```
ERROR: /root/.cache/bazel/_bazel_root/c6d1af4279bc20274c5613de1579a2dc/external/io_bazel_rules_webtesting/go/metadata/main/BUILD.bazel:19:1: Configurable attribute "actual" doesn't match this configuration (would a default condition help?).
Conditions checked:
 @io_bazel_rules_webtesting//common/conditions:linux
 @io_bazel_rules_webtesting//common/conditions:mac
 @io_bazel_rules_webtesting//common/conditions:windows
```

I think rules_webtesting needs to be updated in a similar way to add aarch64 support.

That being said, this PR shouldn't break any existing build, so it won't harm anything to merge.

I left a TODO to add arm64 to all the repositories until this PR is given high-level approval, but I am happy to fill those in in this PR.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No

### DIFF
--- a/docs/install.md
+++ b/docs/install.md
@@ -421,6 +421,7 @@ If you have an advanced use-case you can also register your own toolchains and c
 Toolchains allow us to support cross-compilation, e.g. building a linux binary from mac or windows. To tell Bazel to provide a toolchain for a different platform you have to pass in  the `--platforms` flag. Currently supported values are:
 
 - `@build_bazel_rules_nodejs//toolchains/node:linux_amd64`
+- `@build_bazel_rules_nodejs//toolchains/node:linux_arm64`
 - `@build_bazel_rules_nodejs//toolchains/node:darwin_amd64`
 - `@build_bazel_rules_nodejs//toolchains/node:windows_amd64`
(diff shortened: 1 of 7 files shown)


alexeagle:
As for rules_webtesting, that's maintained by a google-internal team who I don't think will care to make it work on more platforms. I think other mechanisms for getting a browser will work out better for use cases that need it - @gregmagolan has some thoughts about that.

As for the TODO in `node_repositories.bzl` that would be great in a follow-up PR, but there's already a mechanism for users to fill in their own values instead of relying on those defaults, so not much risk merging this as-is.

Thank you for helping!! We would never get around to the ARM platform without contributors :)

PR #1940, opened 2020-06-05
feat: add stdout capture to npm_package_bin

### LINKED ISSUE #1466: npm_package_bin - stdout as the output file
# 🚀 feature request

### Relevant Rules

`npm_package_bin()`

### Description

The documentation for this rule states “It must produce outputs. If you just want to run a program with bazel run, use the nodejs_binary rule.”

I'd like it to allow the stdout of the executed program to become the output, if no output file is specified. This could handle a variety of useful cases where the tool being executed does not produce a file.

This could be done with a `gen_rule()` as usual. Ugh.

mattem:
adds `stdout` and `stderr` attributes to `npm_package_bin` and `run_node` to capture stdout into the specified file

closes #1466

### DIFF
--- a/internal/node/launcher.sh
+++ b/internal/node/launcher.sh
@@ -194,9 +194,14 @@ ARGS=()
 LAUNCHER_NODE_OPTIONS=( "--require" "$require_patch_script" )
 USER_NODE_OPTIONS=()
 ALL_ARGS=(TEMPLATED_args $NODE_REPOSITORY_ARGS "$@")
+STDOUT_CAPTURE=""
+STDERR_CAPTURE=""
+
 for ARG in "${ALL_ARGS[@]:-}"; do
   case "$ARG" in
     --bazel_node_modules_manifest=*) MODULES_MANIFEST="${ARG#--bazel_node_modules_manifest=}" ;;
+    --bazel_capture_stdout=*) STDOUT_CAPTURE="${ARG#--bazel_capture_stdout=}" ;;
+    --bazel_capture_stderr=*) STDERR_CAPTURE="${ARG#--bazel_capture_stderr=}" ;;
     --nobazel_patch_module_resolver)
       MAIN=TEMPLATED_entry_point_execroot_path
       LAUNCHER_NODE_OPTIONS=( "--require" "$node_patches_script" )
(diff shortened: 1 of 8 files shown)


internal/node/npm_package_bin.bzl
```diff
@@ -38,12 +39,14 @@ def _inputs(ctx):
 def _impl(ctx):
     if ctx.attr.output_dir and ctx.outputs.outs:
         fail("Only one of output_dir and outs may be specified")
-    if not ctx.attr.output_dir and not ctx.outputs.outs:
-        fail("One of output_dir and outs must be specified")
+    if not ctx.attr.output_dir and not ctx.outputs.outs and not ctx.attr.stdout:
```
alexeagle, line 42: we don't have to restrict it in this way, you could get outputs or output_dir in addition to stdio outputs

internal/node/npm_package_bin.bzl
```diff
@@ -12,6 +12,7 @@ _ATTRS = {
     "configuration_env_vars": attr.string_list(default = []),
     "data": attr.label_list(allow_files = True, aspects = [module_mappings_aspect, node_modules_aspect]),
     "output_dir": attr.bool(),
+    "stdout": attr.string(),
```
alexeagle, line 15: this should be an `attr.output()`

internal/node/npm_package_bin.bzl
```diff
@@ -52,15 +55,22 @@ def _impl(ctx):
     for a in ctx.attr.args:
         args.add_all([expand_variables(ctx, e, outs = ctx.outputs.outs, output_dir = ctx.attr.output_dir) for e in _expand_locations(ctx, a)])
 
+    stdout_file = None
+    if ctx.attr.stdout:
+        stdout_file = ctx.actions.declare_file(ctx.attr.stdout)
```
alexeagle, line 60: you don't need to declare the file if you use `attr.output()`

internal/node/npm_package_bin.bzl
mattem: But wouldn't you want at least one of them? The extra condition is on the `and not ctx.attr.stdout`

internal/node/launcher.sh
```diff
@@ -188,9 +188,14 @@ ARGS=()
 LAUNCHER_NODE_OPTIONS=( "--require" "$require_patch_script" )
 USER_NODE_OPTIONS=()
 ALL_ARGS=(TEMPLATED_args $NODE_REPOSITORY_ARGS "$@")
+STDOUT_CAPTURE=/dev/stdout
+STDERR_CAPTURE=/dev/stderr
```
mattem, line 192: I'm unsure about this, as when a nodejs_binary is run, it will emit error to stderr, but when npm_package_bin is run and produces it currently swallows them, this would therefore force them all to stderr, I'm not sure if this is intentional

mattem:
`//examples:examples_angular_bazel_architect` is timing out on CircleCI, Windows is failing `/dev/stdout: No such file or directory` - fair enough, although there are other `/dev` references in that launcher

mattem:
Yeah, the failure seems to be just on Ubuntu, running locally on MacOS with `--compilation_mode=dbg` the test passes

mattem:
Running on Ubuntu with `--compilation_mode=dbg` also doesn't fail, however on buildkite, this is included in stderr 
```
bazel node patches enabled. root: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/420/execroot symlinks in this directory will not escape
```

Any ideas?

alexeagle:
You can see in `node_patches.js` where we do print that to stderr, so that seems like the expected output under `-c dbg`. I don't see why you'd get that output on one platform but not another.
Running it locally with `--subcommands`. I see that `--nobazel_patch_module_resolver` is passed to the program
```
(cd /private/var/tmp/_bazel_alex.eagle/6e198640a926b256bc76b9977cfe83fc/execroot/build_bazel_rules_nodejs && \
  exec env - \
    BAZEL_NODE_MODULES_ROOT='' \
    COMPILATION_MODE=fastbuild \
  bazel-out/host/bin/external/npm/terser/bin/terser.sh internal/node/test/terser_input_with_diagnostics.js --compress --verbose --warn '--bazel_node_modules_manifest=bazel-out/darwin-fastbuild/bin/internal/node/test/_diagnostics_producing_bin.module_mappings.json' '--bazel_capture_stdout=bazel-out/darwin-fastbuild/bin/internal/node/test/greeter.min.js' '--bazel_capture_stderr=bazel-out/darwin-fastbuild/bin/internal/node/test/diagnostics.out' --nobazel_patch_module_resolver)
```
which means we get `LAUNCHER_NODE_OPTIONS=( "--require" "$node_patches_script" )`
And I added some println debugging in the script
`echo ${LAUNCHER_NODE_OPTIONS[@]} >&2`
so we see
```
--require ./bazel-out/host/bin/external/npm/terser/bin/terser.sh.runfiles/build_bazel_rules_nodejs/internal/node/node_patches.js
```
So it's running that code... next I add some println debugging in internal/node/node_patches.js
and find that `VERBOSE_LOGS` is undefined even under `-c dbg`

So this finally explains the difference. in the .bazelci/presubmit.yml we have `"--define=VERBOSE_LOGS=1"` for ubuntu1804_debug
so to reproduce you need to run with that define flag.

Since I already have the bits locally I'll just do that and push to your PR

alexeagle:
@googlebot I consent.

mattem:
Thanks for the explanation Alex, appreciate it

PR #1941, opened 2020-06-05
feat: introduce generated_file_test

### LINKED ISSUE #1893: Add golden_file_test to public API
Our golden_file_test
https://github.com/bazelbuild/rules_nodejs/blob/master/internal/golden_file_test/golden_file_test.bzl
is pretty useful and I've pointed a bunch of users to it, as a way to check build outputs into their source tree (in cases where they really have to do that, it's a hacky workaround for dev workflows that can't rely on files in `bazel-out`)

It belongs in bazel-skylib https://github.com/bazelbuild/bazel-skylib/issues/90#issuecomment-453526001 but it feels it will be hard to add there. We already have the public api `copy_to_bin`
https://github.com/bazelbuild/rules_nodejs/blob/master/docs/Built-ins.md#copy_to_bin
which belonged in bazel-skylib but isn't getting merged there https://github.com/bazelbuild/bazel-skylib/pull/217

So we could probably commit to making it public

alexeagle:
This is a new public API for checking in golden or snapshot files.
If you used the private golden_file_test previously, you need to rename actual to generated, golden to src, and golden_debug to src_dbg

Fixes #1893


internal/generated_file_test/bin.js
```diff
@@ -0,0 +1,48 @@
...
+    return 0;
+  }
+  if (mode === '--out') {
+    // Write to golden file
+    fs.writeFileSync(runfiles.resolveWorkspaceRelative(golden), actualContents);
+    console.error(`Replaced ${path.join(process.cwd(), golden)}`);
+    return 0;
+  }
+  if (mode === '--verify') {
+    const unidiff = require('unidiff');
```
jbedard, line 24: Nothing needs to be done to ensure `unidiff` is available?

I know the `@bazel/*` npm packages use `peerDependencies`, but how does it work for rules built directly into rules_nodejs?

jbedard: I see `@npm//unidiff` is passed into the `nodejs_test|binary(data)`, but is that the consumers `@npm//` and we assume the package exists?

alexeagle: Immediate answer is this needs some e2e test to show how it's used outside our workspace.
Yes I think that will show it's broken like you say. This rule has to be in an npm package or else unidiff has to be vendored.

jbedard: That could always be a followup, wouldn't change what you've done.

alexeagle:
@jbedard I added a second commit where I run rollup on the bin.js so unidiff gets inlined into the code. Added a test by simplifying the parcel example with a golden file rather than a spec, to prove that it works.

PR #1953, opened 2020-06-11
docs: add Examples section to the root README.md

PendletonJones:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

It's not prominently stated in the README.md that the examples are only expected to be fully functional on tagged commits.

Issue Number: N/A

## What is the new behavior?

This PR adds a short section to the root README.md about how to get the examples running.

- [ ] Yes
- [x] No

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,9 @@ Do not expect the current HEAD of master to be in perfect working order, as that
 The latest release is [1.7.0](https://github.com/bazelbuild/rules_nodejs/releases/tag/1.7.0).
 
 `git checkout 1.7.0`
+
 `cd examples/angular`
+
 `bazel run //src:devserver`
 
 ## Adopters
--- a/README.md
+++ b/README.md
@@ -32,12 +32,12 @@ $ cd my_workspace
 ## Examples
 
 In order to run the examples checkout one of the tagged releases and then build your desired target.
-Do not expect the current head of master to be in perfect working order, as that branch reflects work currently still in progress and has not been fully validated.
+Do not expect the current HEAD of master to be in perfect working order, as that branch reflects work currently still in progress and has not been fully validated.
 
 The latest release is [1.7.0](https://github.com/bazelbuild/rules_nodejs/releases/tag/1.7.0).
 
 `git checkout 1.7.0`
-`git cd examples/angular`
+`cd examples/angular`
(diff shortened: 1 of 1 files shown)


PendletonJones:
@googlebot I signed it!

README.md
```diff
@@ -29,6 +29,19 @@ $ cd my_workspace
 > If you've used `@bazel/create` before, you may want to use `npx @bazel/create@latest` to get the most recent version.
 > Run without any arguments to see available command-line flags.
 
+## Examples
+
+In order to run the examples checkout one of the tagged releases and then build your desired target.
+Do not expect the current HEAD of master to be in perfect working order, as that branch reflects work currently still in progress and has not been fully validated.
+
+The latest release is [1.7.0](https://github.com/bazelbuild/rules_nodejs/releases/tag/1.7.0).
```
mattem, line 37: I think we don't want to list a single version here, it adds a bit of overhead for releasing to update this. The following steps are specific to the Angular example too, which does have it's [own readme](https://github.com/bazelbuild/rules_nodejs/tree/master/examples/angular). 
Perhaps this section should mention to the user to check the example specific readme? A spin off issue here is to the improve the example specific readme.

PendletonJones: I agree with that, I’ll make some updates tonight/this weekend and include updates to the other example Readmes. I’ll attempt to document the targets and what they in the other examples as well because I think it isn’t completely obvious in some cases.

alexeagle: Also we are only in this state because we're in the breaking changes window.

Maybe a more correct fix is to put the breaking changes on a different branch from the one users are browsing.
Either change the default branch in our GH settings (I asked Bazel team to do this) or use "master" to be 1.x and move 2.0 changes to another branch.

alexeagle:
1.x is now the default branch for web view and cloning the repo, so I think we've fixed this for now. We'll have to stay aware of the branching strategy in the future!

Thanks for the PR though! If you want to open a new one pointing to the examples index at https://github.com/bazelbuild/rules_nodejs/blob/master/docs/examples.md I'd love to merge that.

README.md
PendletonJones: Thanks for doing that, I'll work on making some other more helpful, docs contributions in the future

PR #1974, opened 2020-06-25
test: add repro case for run_node with ts_library built executable

mattem:
Repro of issue with `run_node` replacement for `ctx.actions.run` 
(contrived) custom rule using an executable via run_node
invoked via bazel `build //packages/typescript/test/some_module:test` as `--nobazel_patch_module_resolver` is added by default by run_node, the entry_point on the executable is resolved to the ts file.
This works for `bazel run //packages/typescript/test/some_module:writer_bin` as this will patch require and resolve the
ts to js

### DIFF
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -105,18 +105,26 @@ def _write_require_patch_script(ctx, node_modules_root):
         is_executable = True,
     )
 
+def _ts_to_js(entry_point_path):
+    """If the entry point specified is a typescript file then set it to .js.
+
+    Workaround for #1974
+    ts_library doesn't give labels for its .js outputs so users are forced to give .ts labels
+
+    Args:
+        entry_point_path: a file path
+    """
+    if entry_point_path.endswith(".ts"):
+        return entry_point_path[:-3] + ".js"
+    elif entry_point_path.endswith(".tsx"):
+        return entry_point_path[:-4] + ".js"
+    return entry_point_path
+
 def _write_loader_script(ctx):
     if len(ctx.attr.entry_point.files.to_list()) != 1:
         fail("labels in entry_point must contain exactly one file")
 
-    entry_point_path = _to_manifest_path(ctx, ctx.file.entry_point)
-
-    # If the entry point specified is a typescript file then set the entry
-    # point to the corresponding .js file
-    if entry_point_path.endswith(".ts"):
-        entry_point_path = entry_point_path[:-3] + ".js"
-    elif entry_point_path.endswith(".tsx"):
-        entry_point_path = entry_point_path[:-4] + ".js"
+    entry_point_path = _ts_to_js(_to_manifest_path(ctx, ctx.file.entry_point))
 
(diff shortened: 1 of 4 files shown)


alexeagle:
Good repro.
I thought adding `templated_args = ["--bazel_patch_module_resolver"]` to `writer_bin` would serve as a workaround but I get the same error.

This fixes it by selecting the .js file as the entry_point

```
ts_library(
    name = "writer",
    srcs = ["writer.ts"],
    deps = ["@npm//@types/node"],
)

filegroup(
    name = "writer_js",
    srcs = ["writer"],
    output_group = "es5_sources",
)

# nodejs_binary for generating an executable for the tool :writer
nodejs_binary(
    name = "writer_bin",
    data = [
        ":writer_js",
    ],
    entry_point = ":writer_js",
)
```

but it relies on having a single .js output from the ts_library and wouldn't work in general - you'd have to have an entry_point-only ts_library with the main, depending on another ts_library that builds the rest of the files.

The hack of swapping .ts -> .js on the fly is gross. Declaring .js outputs from ts_library might not be hard... maybe we could have another output_group from ts_library that's guaranteed to be a single file (like the first one in srcs maybe)

Needs a bit more thought.

alexeagle:
Another workaround is a genrule to copy out the one entry point JS file you need, then use that genrule default output as the entry_point.

Another workaround is `--bazel_patch_module_resolver` to turn on the require() patch

mattem:
Internal evertz patch used for workaround

```
diff --git a/internal/providers/node_runtime_deps_info.bzl b/internal/providers/node_runtime_deps_info.bzl
index 9a4e938..4ffa408 100755
--- internal/providers/node_runtime_deps_info.bzl
+++ internal/providers/node_runtime_deps_info.bzl
@@ -86,10 +87,12 @@ def run_node(ctx, inputs, arguments, executable, **kwargs):
         add_arg(arguments, "--bazel_capture_stderr=%s" % stderr_file.path)
         outputs = outputs + [stderr_file]

-    # By using the run_node helper, you suggest that your program
-    # doesn't implicitly use runfiles to require() things
-    # To access runfiles, you must use a runfiles helper in the program instead
-    add_arg(arguments, "--nobazel_patch_module_resolver")
+    disable_patch_module_resolver = kwargs.pop("disable_patch_module_resolver", True)
+    if disable_patch_module_resolver:
+        # By using the run_node helper, you suggest that your program
+        # doesn't implicitly use runfiles to require() things
+        # To access runfiles, you must use a runfiles helper in the program instead
+        add_arg(arguments, "--nobazel_patch_module_resolver")

     env = kwargs.pop("env", {})
```

alexeagle:
Greg and I remembered - this is the spot that's making it work today
https://github.com/bazelbuild/rules_nodejs/blob/1.x/internal/node/node.bzl#L117-L120
because the loader completely goes away under `--nobazel_patch_module_resolver`

alexeagle:
@googlebot I consent.

PR #1975, opened 2020-06-26
fix: allow multiple run_node calls to be made from the same rule context

mattem:
If two `run_node` calls are made from the same rule, the module mappings files generating actions will conflict. This adds the `mnemonic` into the file name (if provide) to separate the different actions generating the files.

### DIFF
--- a/internal/linker/link_node_modules.bzl
+++ b/internal/linker/link_node_modules.bzl
@@ -56,12 +56,13 @@ def _link_mapping(label, mappings, k, v):
     else:
         return True
 
-def write_node_modules_manifest(ctx, extra_data = []):
+def write_node_modules_manifest(ctx, extra_data = [], mnemonic = None):
     """Writes a manifest file read by the linker, containing info about resolving runtime dependencies
 
     Args:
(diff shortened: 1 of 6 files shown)


internal/linker/link_node_modules.bzl
```diff
@@ -56,12 +56,13 @@ def _link_mapping(label, mappings, k, v):
     else:
         return True
 
-def write_node_modules_manifest(ctx, extra_data = []):
+def write_node_modules_manifest(ctx, extra_data = [], mnemonic = ""):
```
alexeagle, line 59: nit: maybe it should default to `None`, then later we could avoid having the extra `_%s` appear if mnemonic is unset? my understanding is that it's called from run_node with kwargs.get("mnemonic") which means None is the value that will be commonly passed here anyway

mattem:
Added a more comprehensive test, couldn't find something really suitable to add it too, it also allows us to expand on this test case in the future

internal/linker/link_node_modules.bzl
mattem: Done

PR #1990, opened 2020-07-01
feat: add support for capturing and overriding the exit code within run_node

mattem:
## PR Checklist

## PR Type

## What is the new behavior?

Adds `exit_code_out` and `override_exit_code` to `npm_package_bin` (and handling in `run_node`) allowing the exit code of the binary to be captured to a file, and the subsequent exit code to then be overridden.


internal/node/npm_package_bin.bzl
```diff
@@ -10,8 +10,10 @@ _ATTRS = {
     "args": attr.string_list(mandatory = True),
     "configuration_env_vars": attr.string_list(default = []),
     "data": attr.label_list(allow_files = True, aspects = [module_mappings_aspect, node_modules_aspect]),
+    "exit_code_out": attr.output(),
```
mattem, line 13: `exit_code` on it's own seemed not descriptive enough, but can change these attr names to something better

internal/node/launcher.sh
```diff
@@ -179,8 +181,14 @@ for ARG in "${ALL_ARGS[@]:-}"; do
...
     # Supply custom linker arguments for first-party dependencies
     --bazel_node_modules_manifest=*) MODULES_MANIFEST="${ARG#--bazel_node_modules_manifest=}" ;;
+    # Captures stdout of the node process to the file specified
     --bazel_capture_stdout=*) STDOUT_CAPTURE="${ARG#--bazel_capture_stdout=}" ;;
+    # Captures stderr of the node process to the file specified
     --bazel_capture_stderr=*) STDERR_CAPTURE="${ARG#--bazel_capture_stderr=}" ;;
+    # Captures the exit code of the node process to the file specified
+    --bazel_capture_exit_code=*) EXIT_CODE_CAPTURE="${ARG#--bazel_capture_exit_code=}" ;;
+    # Overrides the exit code of the node process
+    --bazel_override_exit_code=*) OVERRIDE_EXIT_CODE="${ARG#--bazel_override_exit_code=}" ;;
```
alexeagle, line 191: is there a case where you'd want to capture the exit code, and not have it exit zero? If we can't think of one, I think I'd rather keep the feature smaller just so there's less surface area for us to support.

mattem: It seemed a natural additional option, but don't have a use case for it

alexeagle: okay do you mind simplifying then, and just have the one option to capture, and always exit zero when the option is used?
Also we should document here that you still have to produce all the expected outputs from the program even when expecting a non-zero exit code, as we discovered in your PR that consumes this feature. That restriction does make it a lot less useful... but I'm not sure what else we can do in this model where the program and the error translator are separate actions.

mattem: Will do 👍 Will add a note to the npm_package_bin docs too, I couldn't see where the run_node docs are surfaced, but I'll add something in there too I think.

PR #1992, opened 2020-07-01
chore: treat Bazel dev builds as 2.1+ for hiding BUILD files

adamyi:
I don't think anyone should be using a dev build for a really old bezel version

This makes the behavior consistent for new local dev builds and official releases.

### DIFF
--- a/internal/npm_install/generate_build_file.ts
+++ b/internal/npm_install/generate_build_file.ts
@@ -476,17 +476,20 @@ function findPackages(p = 'node_modules') {
                        .map(f => path.posix.join(p, f))
                        .filter(f => isDirectory(f));
 
+  // Starting in version 2.1, Bazel honors the .bazelignore file we wrote into the
+  // root of the external repository, and won't see BUILD files under node_modules
+  // This parsing of the version number isn't accurate in some cases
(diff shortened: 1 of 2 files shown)


alexeagle:
Note, this merged into 1.x which we don't plan to release anymore. Do you need a 1.x release to get this fix?

The master branch will be released as 2.0.0 (we already cut 2.0.0-rc.0 from there). I tried to cherry-pick the change, but we totally removed the hide BUILD files feature in https://github.com/bazelbuild/rules_nodejs/commit/5d1d0060159aaa90552daf0d1100ea714e503a9c#diff-7d9a49a369e423472adf00d06229a088 so I think there's nothing we need to do. Sound right?

adamyi:
> Note, this merged into 1.x which we don't plan to release anymore. Do you need a 1.x release to get this fix?

Yeah would be nice to have a release. Don't worry too much if another 1.x release is too much trouble - I can just depend on my fork for now.

> so I think there's nothing we need to do. Sound right?

Sounds about right. I'll take a closer look and try switching to 2.0 rc rules for my code. Will create another issue/pull request if I run into any other issue :)

PR #1993, opened 2020-07-01
fix(builtin): linker silently not generating expected links in windows

devversion:
Currently the builtin linker sometimes does not generate links as
defined in the module mappings on Windows. This results in unexpected
and confusing runtime issues, and differences with other platforms.

The linker fails to generate links if module mappings result in
a different symlink/module hierarchy (as computed by `reduceModules`).

For example, consider in a previous linker run, we linked the module
`@angular/cdk/overlay` into `node_modules/@angular/cdk/overlay`. In
a second run then, we actually link `@angular/cdk`. The linker will
fail to do that as the `node_modules/@angular/cdk` folder already
exists (due to missing sandbox/runfile symlinking in windows)

We fix this by clearing such leftover linker directories so that
the newly configured module mapping can be created. In order to
avoid race conditions in non-sandboxed environments, we need to
pay special attention to potential concurrent resource accesses,
and also need to preserve possible child links from previous or
concurrent linker runs.

### DIFF
--- a/internal/linker/index.js
+++ b/internal/linker/index.js
@@ -45,6 +45,56 @@ function mkdirp(p) {
         }
     });
 }
+function gracefulLstat(path) {
+    return __awaiter(this, void 0, void 0, function* () {
+        try {
+            return yield fs.promises.lstat(path);
+        }
+        catch (e) {
+            if (e.code === 'ENOENT') {
+                return null;
+            }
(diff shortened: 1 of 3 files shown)


devversion:
@alexeagle Thanks for reviewing and merging! 🎉

PR #1994, opened 2020-07-01
feat: augment TS2307 diagnostic message with missing deps info

mattem:
Augments TS2307 with additional info when a dependency is missing from the `deps` attr and compilation unit. 
This was branched from https://github.com/bazelbuild/rules_nodejs/pull/1990


packages/typescript/internal/build_defs.bzl
```diff
@@ -188,9 +192,26 @@ def _compile_action(ctx, inputs, outputs, tsconfig_file, node_opts, description
...
             env = {"COMPILATION_MODE": ctx.var["COMPILATION_MODE"]},
+            exit_code_out = exit_code_output,
+            override_exit_code = 0,
+        )
+
+        run_node(
+            ctx,
+            progress_message = "Checking TSC Diagnostics",
+            mnemonic = "tscdiag",
+            #            outputs = action_outputs,
```
mattem, line 207: @alexeagle This is the part that's tripping me up. On a failed compile, the tsc action is forced to exit with 0 so the following action can run and inspect stderr, but it still declares it's outputs that aren't created.

packages/typescript/internal/strict-deps-diagnostic-processor.js
```diff
@@ -0,0 +1,25 @@
...
+    const depsMessage = ` Ensure that the dependency for '${
+        moduleSpecifier}' is added to the targets deps attribute.`;
+    tsDiagnostics =
+        tsDiagnostics.slice(0, diagOffset) + depsMessage + tsDiagnostics.slice(diagOffset);
+  });
+}
+
+process.stderr.write(tsDiagnostics);
+
+process.exit(parseInt(exitcode, 10));
```
alexeagle, line 25: nit: Is this better than `Number(exitcode)` ?

packages/typescript/internal/build_defs.bzl
alexeagle: I don't quite understand, which action declares outputs that aren't created? tsc or the diagnostic processor? maybe an error message would help.

mattem: tsc, it produces the following when the exit code is forced 0 with missing deps (add `import * as fs from 'fs';` to `packages/typescript/test/strict_deps/grandparent.ts`)

```
bazel build //packages/typescript/test/strict_deps:grandparent
Starting local Bazel server and connecting to it...
INFO: Invocation ID: 7d2cbd7e-e9e7-459f-8c1c-fc02ca1d6eb1
INFO: Analyzed target //packages/typescript/test/strict_deps:grandparent (53 packages loaded, 878 targets configured).
INFO: Found 1 target...
INFO: Deleting stale sandbox base /private/var/tmp/_bazel_matt/4afd3aeed021b99395b4ccda14f35222/sandbox
ERROR: /Users/matt/Documents/workspace/rules_nodejs/packages/typescript/test/strict_deps/BUILD:17:1: output 'packages/typescript/test/strict_deps/grandparent.js' was not created
ERROR: /Users/matt/Documents/workspace/rules_nodejs/packages/typescript/test/strict_deps/BUILD:17:1: output 'packages/typescript/test/strict_deps/grandparent.d.ts' was not created
ERROR: /Users/matt/Documents/workspace/rules_nodejs/packages/typescript/test/strict_deps/BUILD:17:1: not all outputs were created or valid
Target //packages/typescript/test/strict_deps:grandparent failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 16.734s, Critical Path: 2.30s
INFO: 1 process: 1 darwin-sandbox.
```

alexeagle: Ah yes, right. I hadn't realized this flaw in the design. It needs to be run with `--emitOnError` if we want those outputs to be produced on failure, which I think is the only choice here, since we have to predict the outputs to Bazel before knowing whether the action will succeed.

mattem: Unfortunately even after forcing `noEmitOnError` to false (it's set true by [tsconfig.bzl](https://github.com/bazelbuild/rules_typescript/blob/master/internal/common/tsconfig.bzl#L269)), [tsc_wrapped exits early](https://github.com/bazelbuild/rules_typescript/blob/master/internal/tsc_wrapped/tsc_wrapped.ts#L446) if there are any diagnostics before emitting, so if we wanted to carry on here, I feel we'd need to work around that somehow, and it all starts feeling a bit messy?

This approach may still work with ts_project, I haven't tried that yet.

PR #2005, opened 2020-07-06
chore(examples): upgrade angular example to 10.0.2 and remove ngtsc p…

alexeagle:


jbedard:
@googlebot I consent.

Issue #1910, opened 2020-05-25
Can't forward the runfiles directory to nodejs_binary if its used as a subtool


a-ignatov-parc:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `nodejs_binary`

### Is this a regression?

Hard to say. The problem exist for a long time. Based on [this](https://github.com/bazelbuild/rules_nodejs/pull/587) PR, and [reported repo](https://github.com/Globegitter/rules_nodejs_sh_wrapper_repro) this issue exist since `0.27.8`.

### Description

Bazel has [a documented way](https://docs.bazel.build/versions/master/skylark/rules.html#tools-depending-on-tools) to create and use tools with runfiles that are based on other tools with their own runfiles.

The idea is simple:
1. The root tool collects and merge its subtools runfiles.
2. Each subtool should accept runfiles root as a parameter through environment or command line argument.

Unfortunately, there is no legit way to explicitly configure runfiles for `nodejs_binary`, which leads to non-deterministic execution, since `node_launcher.sh` is trying to resolve its runfiles based on heuristics described [here](https://github.com/bazelbuild/rules_nodejs/blob/8509d9c3b3a9d7558fea98f23f12da9743dc3288/internal/node/launcher.sh#L31).

`RUNFILES_DIR` seems to be the most suitable for solution, however `node_launcher.sh` has [a comment](https://github.com/bazelbuild/rules_nodejs/blob/8509d9c3b3a9d7558fea98f23f12da9743dc3288/internal/node/launcher.sh#L28) that searching for runfiles in this directory is not always correct. Sadly, I could not find any details about this.

If there are serious concerns about using `RUNFILES_DIR` we can use the [java binary approach](https://github.com/bazelbuild/bazel/blob/fd5d29aa4ef907a15ec6db0e359e86df1acf64fc/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt#L148) with a standalone env variable, which can be explicitly configured externally.

## 🔬 Minimal Reproduction

1. Checkout https://github.com/a-ignatov-parc/bazel-demo/
1. Switch to `runfiles-problem` branch.
1. Run `./bazel clean`.
1. Run `./bazel run //app:runner`. Works as expected.
1. Switch to `runfiles-problem-2` branch.
1. Run `./bazel run //app:runner` again. Works as expected.
1. Switch back to `runfiles-problem` branch.
1. Run `./bazel build //app/...`
[...]

PR #2015, opened 2020-07-08
[wip] Fix RUNFILES resolution for nested tools

samschlegel:
## PR Checklist

  - I can add a test case for the failing behavior if that's desired 

## PR Type

- [x] Bugfix
- [x] Build related changes
- [ ] CI related changes

Issue Number: #1910 

To clarify the cause of the issue, when a `node_binary` (X), is invoked from another tool (Y) the current RUNFILES resolution prefers X's runfiles directory when it should choose Y's. This is normally fine if X's runfiles directory doesn't exists, such as when it hasn't been built yet, but it breaks when it does. Y's runfiles directory should always be a superset of X's runfiles directory, so there's no reason to choose X's over Y's.

## What is the new behavior?

This PR switches to using the default `RUNFILES_DIR`, which will prefer the correct directory.

- [ ] Yes
- [x] No

## Other information

There were comments that `${PWD}/${RUNFILES_DIR}` is not always correct. I'm not sure what cases this is referring to, as that should always be correct. The only thing I can think of is that it's possible that X's runfiles are not contained in Y's runfiles, but that's an issue with the rule implementation of Y, not with RUNFILES_DIR.


samschlegel:
Ah right, `readlink -f` doesn't work on Mac, and Windows doesn't have the DIR. 😢 Will try and just fix the existing lookup instead.

samschlegel:
Added a failing test. The test succeeds in a clean env, but fails after building `//internal/node/test:nested_tool`. Not sure how to encode this in a Bazel test itself so will probably have to add some kind of ordering in another way.

Won't have time to tackle this the rest of the week but at least there's a fairly minimal repro 😄

alexeagle:
Ooh I was wondering if you were going to adopt Rome over there :)
Thanks for the repro, that's a good start. We might be able to test that ordering in a bazel-in-bazel test, each of our examples/e2e directories has one of those.
Not sure when we'll have time to dig in, but I'd like to do some bugfix sprint after 2.0 final

samschlegel:
> Ooh I was wondering if you were going to adopt Rome over there :)

Heh, probably at some point. I'm not super familiar with our front-end side of things. I mostly just want to use tools like `cdk8s` to build manifests for use with `rules_k8s` to replace Jsonnet 😅 

> Thanks for the repro, that's a good start. We might be able to test that ordering in a bazel-in-bazel test, each of our examples/e2e directories has one of those.
> Not sure when we'll have time to dig in, but I'd like to do some bugfix sprint after 2.0 final

🙏

PR #2040, opened 2020-07-15
chore: update examples/angular_view_engine to @angular/bazel 9.0.5

gregmagolan:
Updating @angular/bazel past 9.0.5 breaks ViewEngine which is only supported by ng_module due to https://github.com/angular/angular/pull/35841. It is recommended that Angular + Bazel users switch to Ivy and use ts_library with use_angular_plugin. See examples/angular.

### DIFF
--- a/examples/angular_view_engine/package.json
+++ b/examples/angular_view_engine/package.json
@@ -25,7 +25,7 @@
         "zone.js": "0.10.2"
     },
     "devDependencies": {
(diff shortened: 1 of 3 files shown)


gregmagolan:
@alexeagle After this lands we should link the patch file here in the 2.0 upgrade wiki for users of @angular/bazel.

PR #2048, opened 2020-07-17
feat(typescript): support for declarationdir on ts_project 

mistic:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

`declarationDir` option from [Typescript Compiler Options](https://www.typescriptlang.org/docs/handbook/compiler-options.html) is not supported on ts_project.

Issue Number: N/A

## What is the new behavior?
`declarationDir` option from [Typescript Compiler Options](https://www.typescriptlang.org/docs/handbook/compiler-options.html) is now supported on ts_project through a declarationdir param.

- [ ] Yes
- [x] No


packages/typescript/internal/ts_project.bzl
```diff
@@ -63,7 +64,9 @@ def _ts_project_impl(ctx):
     if len(ctx.outputs.typings_outs) > 0:
         arguments.add_all([
             "--declarationDir",
-            _join(ctx.bin_dir.path, ctx.label.package, ctx.attr.outdir),
+            (
+                _join(ctx.bin_dir.path, ctx.label.package, ctx.attr.declarationdir) if ctx.attr.declarationdir else _join(ctx.bin_dir.path, ctx.label.package, ctx.attr.outdir)
```
alexeagle, line 68: oh we made a naming mistake on the others - TS has "rootDir" so we should have named the attribute "root_dir". Same with "out_dir" and "declaration_dir"
I think maybe it was because other rules have "outdir" so we wanted that one to match

It's not too late to change, none of the attributes were released yet
https://github.com/bazelbuild/rules_nodejs/blob/1.x/packages/typescript/src/internal/ts_project.bzl

Would you change all three of them in a second commit here? If you don't have time today I'll merge this and do it.

alexeagle, line 68: this logic would be simpler if we make declarationDir default to outDir prior to this statement

packages/typescript/internal/ts_project.bzl
```diff
@@ -115,7 +118,7 @@ def _ts_project_impl(ctx):
     if ctx.outputs.buildinfo_out:
         outputs.append(ctx.outputs.buildinfo_out)
     runtime_outputs = depset(json_outs + ctx.outputs.js_outs + ctx.outputs.map_outs)
-    typings_outputs = ctx.outputs.typings_outs + [s for s in ctx.files.srcs if s.path.endswith(".d.ts")]
+    typings_outputs = ctx.outputs.typings_outs + ctx.outputs.typing_maps_outs + [s for s in ctx.files.srcs if s.path.endswith(".d.ts")]
```
alexeagle, line 121: why make this change? I don't think .map files belong in DeclarationInfo, they are meant to be consumed by editors not by other tsc actions that need to type-check

packages/typescript/internal/ts_project.bzl
```diff
@@ -357,6 +361,9 @@ def ts_project_macro(
 
         validate: boolean; whether to check that the tsconfig settings match the attributes.
 
+        declarationdir: a string specifying a subdirectory under the bazel-out folder where generated declaration
+            outputs are written.
```
alexeagle, line 365: Include what is the default value pls

alexeagle:
@googlebot I consent.

packages/typescript/internal/ts_project.bzl
alexeagle: checked with @gregmagolan - you are correct we transport .map files along with the thing they map (it's like that for .js.map files for example)

gregmagolan:
This is a BREAKING CHANGE and we should note it in release notes since the attributes renamed outdir => out_dir & rootdir => root_dir.

alexeagle:
Nope, non-breaking since none of these attributes were in 1.x

mistic:
Thanks both @alexeagle and @gregmagolan! Awesome work here 🎉

PR #2054, opened 2020-07-19
Fix2044

alexeagle:


packages/typescript/internal/build_defs.bzl
```diff
@@ -244,7 +244,9 @@ def tsc_wrapped_tsconfig(
     node_modules_root = _compute_node_modules_root(ctx)
     config = create_tsconfig(
         ctx,
-        files,
+        # Filter out package.json files that are included in DeclarationInfo
+        # tsconfig files=[] property should only be .ts/.d.ts
+        [f for f in files if f.path.endswith(".ts")],
```
gregmagolan, line 249: Hmmm. Should update `internal/providers/declaration_info.bzl` and note that package.json files may also be included. What about another field inside DeclarationInfo for package.json files? Feels off that consumers need to filter here. Not sure tho.

alexeagle: Updating the docs seems right to me.
I think at one point we discussed that file extensions are a nicer way to transmit some of this metadata than adding additional provider fields. Either way the producers/consumers of the provider may need to know about package.json, but I think this way fewer of them need to know.

gregmagolan: sgtm

Issue #1348, opened 2019-11-12
Allow custom rules to generate stardoc 


clintharrison:
Custom rules that `load()` providers/etc. from `rules_nodejs` cannot generate stardoc documentation.

`bzl_library` from skylib is mocked out to `filegroup`, which means your own rules have missing dependencies from Stardoc's perspective.
You can try creating your own `bzl_library` with those filegroups, but you end up needing to reference filegroups in `//internal` which don't have visibility outside of this workspace.

alexeagle:
Maybe one way to solve this is to stop mocking out the skylib dependency and make users add a WORKSPACE dependency so that bzl_library load statements work for everyone.

A bit more complex would be a "lite" distro with minimal deps and a heavier distro. A lot more complex would be for you to switch to getting rules_nodejs from sources rather than our tgz distro

alexeagle:
There's another possibility I didn't think of in the last comment - you could use a local patch to restore the original stardoc load statements.

mattem:
I made this patch for 1.6.1 that is used in my orgs repo to allow Stardoc generation where rules_nodejs is loaded, if you did want to go the patch route.

alexeagle:
I think we may already be in a state where you can use rules_nodejs from source rather than the distro artifacts. That gives another tidy answer to this, just switch to sources and then all our dev dependencies will be exposed.

mattem:
Added example in #2056 for using rules_nodejs from source and generating docs with stardoc

Issue #571, opened 2018-10-01
feature request: support using this rule in its source format


qzmfranklin:
What I want:

Git add this repo as a submodule in my repo, say, under `third_party/rules_typescript`.

Add a line in `WORKSPACE`:
```
local_repository(path='third_party/rules_typescript')
```

And it should just work.

But right now this is impossible. Something in this repo (or its dependencies) would break if I do that. I have been forced to use this rule via `yarn add @bazel/typescript`, which has a few undesirable consequences:

1. relying on running `yarn/npm` from the command line breaks the hermetcity of the build
2. `bazel` cannot run at all until `yarn` is finished, which prevents the use of `yarn_install/npm_install` (defined and recommended in rules_nodejs)
3. if I ever need to make a change to the source code of rules_typescript (for example, to make the ts_proto_library generate modules in commonjs format rather than amd format), I would have to patch the file under `node_modules/@bazel/typescript` *after* `yarn/npm` is run. This is just, ..., outrageous.

Given the above use cases, I'd like to request a feature to support using this bazel rule in its source format on top of as published artifact.

mattem:
Added example in #2056 for using rules_nodejs from source

PR #2056, opened 2020-07-20
docs: example of using the rules from source

mattem:
Adds an example of using rules_nodejs from source, including rules_typescript and generating stardoc from a custom macro.

The example outputs and test work as expected when run from inside the examples/from_source directory, but seems to fail when run as part of the `example_integration_test`,

```
example_integration_test(
    name = "examples_from_source",
    bazel_commands = [
        "build //:docs",
        "test ...",
    ],
    owners = ["@mattem"],
)
```

Results in:
```
ERROR: An error occurred during the fetch of repository 'build_bazel_rules_typescript':
        File "/private/var/tmp/_bazel_matt/0cbaade998ab768911d322862018291b/external/bazel_tools/tools/build_defs/repo/git.bzl", line 178
                patch(ctx)
        File "/private/var/tmp/_bazel_matt/0cbaade998ab768911d322862018291b/external/bazel_tools/tools/build_defs/repo/utils.bzl", line 135, in patch
                ctx.patch(patchfile, strip)
Not a regular file: /private/var/tmp/_bazel_matt/0cbaade998ab768911d322862018291b/external/build_bazel_rules_nodejs/rules_typescript_pr_494.patch
```

Also not sure the best way of updating the commit SHA in the example WORKSPACE, perhaps this is already handled.

PR to provide examples for #571 and #1348


examples/from_source/WORKSPACE
```diff
@@ -0,0 +1,57 @@
+workspace(
+    name = "from_source",
+    managed_directories = {"@npm": ["node_modules"]},
+)
+
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
+
+git_repository(
+    name = "build_bazel_rules_nodejs",
```
alexeagle, line 9: I think in this case it should just be `local_repository(path="../..")` maybe? That's what we do in a couple other cases of nested workspaces

examples/from_source/WORKSPACE
```diff
@@ -0,0 +1,57 @@
...
+    commit = "981e7c1ab431c79dd83bcf5d495fefc4a0e02e45",
+    remote = "https://github.com/bazelbuild/rules_nodejs.git",
+    shallow_since = "1595023729 -0700",
+)
+
+git_repository(
+    name = "build_bazel_rules_typescript",
+    commit = "10a5a86885f95ab788fd841ade47b6a16e0c13d6",
+    patches = [
+        "@build_bazel_rules_nodejs//:rules_typescript_pr_494.patch",
```
alexeagle, line 19: we should move this into a reusable function - I added another patch in a PR I put up yesterday

mattem: Added to package.bzl

PR #2060, opened 2020-07-20
propagate the tags to action

cristifalcas:
## PR Checklist

## PR Type

We are hitting this error on RBE with a react rule: `Note: Remote connection/protocol failed with: execution failed java.io.IOException: io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Input tree has 96692 files, above the maximum of 70000`

The solution is to build the affected target with "no-remote-exec", but this is not propagated to this action.

This PR fixes this issue

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/internal/node/npm_package_bin.bzl
+++ b/internal/node/npm_package_bin.bzl
@@ -78,6 +78,7 @@ def _impl(ctx):
         stdout = ctx.outputs.stdout,
         stderr = ctx.outputs.stderr,
         exit_code_out = ctx.outputs.exit_code_out,
+        execution_requirements = {tag: "" for tag in ctx.attr.tags},
     )
 
     return [DefaultInfo(files = depset(outputs + tool_outputs))]


cristifalcas:
@googlebot I fixed it.

bweston92:
Please run `yarn bazel:format --mode=fix`

internal/node/npm_package_bin.bzl
```diff
@@ -78,6 +78,7 @@ def _impl(ctx):
         stdout = ctx.outputs.stdout,
         stderr = ctx.outputs.stderr,
         exit_code_out = ctx.outputs.exit_code_out,
+        execution_requirements = {tag: "" for tag in ctx.attr.tags},
```
alexeagle, line 81: Hmm it would be nice to have some usage of the feature, as a regression test and also so others can see this workaround for `no-remote-exec` in action. I understand that's tricky, we don't have a 90k-input-count action to reproduce the same problem you had. Maybe we need a test that runs `bazel aquery` or something...

Also are there any docs about how remote execution honors certain tags and how they should be propagated via the execution requirements? I'd like a doc linked here so we can understand why it's like this.

cristifalcas: There were 2 issues on my case:
1. RBE was throwing that error when trying to build a react project. Bazel targets have been copied from [here](https://github.com/bazelbuild/rules_nodejs/tree/master/examples/create-react-app)
2. In order to not hit this RBE bug, I tried to force the target to always compile locally via "no-remote-exec" tag. This didn't worked because the tag was not propagated to this action

Here is the bazel issue about tags not being propagated: https://github.com/bazelbuild/rules_nodejs/tree/master/examples/create-react-app

alexeagle: That last link looks wrong, do you have a right one?
From https://github.com/bazelbuild/rules_go/issues/2534#issuecomment-642024716 it seems like `--experimental_allow_tags_propagation` is what you need, then it works everywhere not just this one spot

alexeagle:
Since I didn't hear back from you, I'll assume `--experimental_allow_tags_propagation` solves the problem, pls open a new PR if that's not right.

PR #2063, opened 2020-07-21
fix(typescript): ts_library should accept .tsx inputs

alexeagle:
These were accidentally filtered out of the generated tsconfig in 16cbc6fcf759e68482293ff674164064280a93f4


mrmeku:
One other change I think is worth including with this fix:

```
diff --git a/node_modules/@bazel/typescript/internal/common/module_mappings.bzl b/node_modules/@bazel/typescript/internal/common/module_mappings.bzl
index 286c3b3..f27c6ae 100755
--- a/node_modules/@bazel/typescript/internal/common/module_mappings.bzl
+++ b/node_modules/@bazel/typescript/internal/common/module_mappings.bzl
@@ -78,7 +78,7 @@ def get_module_mappings(label, attrs, srcs = [], workspace_name = None, mappings
         ] if p])
         if hasattr(attrs, "module_root") and attrs.module_root and attrs.module_root != ".":
             mr = "%s/%s" % (mr, attrs.module_root)
-            if attrs.module_root.endswith(".ts"):
+            if attrs.module_root.endswith(".ts") or attrs.module_root.endswith(".tsx"):
                 if workspace_name:
                     mr = mr.replace(".d.ts", "")
```

alexeagle:
@mrmeku do you have a repro to go along with that?

packages/typescript/test/ts_library_esm_with_jest/BUILD.bazel
```diff
@@ -5,7 +5,7 @@ load(":ts_jest_test.bzl", "ts_jest_test")
 ts_library(
     name = "lib",
     srcs = [
-        "lib.ts",
```
mattem, line 8: Nice 👍

PR #2068, opened 2020-07-22
Nextjs example

alexeagle:


mrtnzlml:
Hi @alexeagle, just wanted to check-in. Did you have any luck with Bazel vs. Next.js? :) Would you happen to know how to overcome the issues blocking this PR?

flolu:
Any updates on this? I think Next.js example would be highly valuable

alexeagle:
So far it just hasn't been either funded by a company that needs it, or a fun hobby for someone in their spare time. Would love to work on this if some company needs it

flolu:
Makes sense. What do you think is the estimated effort? (a couple days, weeks?)
We could add some funding, but our budget is small.

alexeagle:
Hard to say, I don't think we even identified all the issues. Depends what other assumptions are broken by next tooling. ts_project took me about two weeks I think.

flolu:
Btw the implementation with auto-generated `next` does work without `write_chdir_script` now.
But the [global styles issue](https://github.com/flolu/nextjs-bazel/issues/1) is still there. Other than that it works quite good, even with PWA support.

Do you think this is a bug in the `next` builder, or is there something "wrong" with Bazel?

PR #2072, opened 2020-07-23
chore: add config for https://github.com/apps/stale

alexeagle:
Keep our issues tidy without controversy !

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,23 @@
+# See https://github.com/apps/stale
+
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,22 @@
+# See https://github.com/apps/stale
+
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - "need: discussion"
```
mattem, line 9: Perhaps keep the `cleanup` label too? As that's an internal label collaborators may add to tickets

alexeagle: good call

alexeagle:
note, I expect we'll get an initial flurry of stale issues and will be forced to do a one-time triage to add preservation labels on the ones we want to keep. Maybe we'll need to devote next couple team meetings to that triage.

PR #2080, opened 2020-07-26
feat: augment TS2307 errors with missing dependency messages for ts_project

mattem:
Take 2, for `ts_project` as it presents an easier path to have tsc output files even when it fails, unlike ts_library which has a bunch of early exits depending on the diagnostics produced.

Last issue is the diagnostic processor must produce some output and something must request this, otherwise it will never be run, see comment on source

### DIFF
--- a/packages/typescript/internal/ts_project.bzl
+++ b/packages/typescript/internal/ts_project.bzl
@@ -23,6 +23,7 @@ _ATTRS = {
     "srcs": attr.label_list(allow_files = True, mandatory = True),
     "tsc": attr.label(default = Label(_DEFAULT_TSC), executable = True, cfg = "host"),
     "tsconfig": attr.label(mandatory = True, allow_single_file = [".json"]),
+    "_diagnostics_processor": attr.label(default = Label("//packages/typescript/bin:diagnostics_processor"), executable = True, cfg = "host"),
 }
 
 # tsc knows how to produce the following kinds of output files.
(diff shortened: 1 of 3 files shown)


packages/typescript/internal/ts_project.bzl
```diff
@@ -113,10 +114,15 @@ def _ts_project_impl(ctx):
...
+    # diagnostics outputs from tsc
+    exit_code_output = ctx.actions.declare_file("%s.exit.code" % ctx.label.name)
+    diagnostics_output = ctx.actions.declare_file("%s.diagnostics" % ctx.label.name)
+    processed_diagnostics_output = ctx.actions.declare_file("%s.diagnostics_processed" % ctx.label.name)
+
     outputs = json_outs + ctx.outputs.js_outs + ctx.outputs.map_outs + ctx.outputs.typings_outs + ctx.outputs.typing_maps_outs
     if ctx.outputs.buildinfo_out:
         outputs.append(ctx.outputs.buildinfo_out)
-    runtime_outputs = depset(json_outs + ctx.outputs.js_outs + ctx.outputs.map_outs)
+    runtime_outputs = depset(json_outs + ctx.outputs.js_outs + ctx.outputs.map_outs + [processed_diagnostics_output])
```
mattem, line 125: I'm not sure this is the right place to add this output, but it needed to cause the diagnostic processor to run as otherwise nothing is requesting the output of the processors action. However ts_project has its default outputs for JS and d.ts files etc

PR #2088, opened 2020-07-29
chore: upgrade rollup to ^2.3.0

jbedard:
... before attempting to use the rollup ^2.3.0 APIs


internal/node/node_patches.js
```diff
@@ -9,12 +9,18 @@ var fs$1 = _interopDefault(require('fs'));
 
 var commonjsGlobal = typeof globalThis !== 'undefined' ? globalThis : typeof window !== 'undefined' ? window : typeof global !== 'undefined' ? global : typeof self !== 'undefined' ? self : {};
 
-function unwrapExports (x) {
```
jbedard, line 12: This one was interesting. Seems to just be changes what rollup outputs, but interesting that the `unwrapExports` stuff is all gone, and I guess `createCommonjsModule` has a completely new implementation.

packages/rollup/test/code_splitting/spec.js
```diff
@@ -6,6 +6,6 @@ describe('rollup code splitting', () => {
     const chunks = runfiles.resolvePackageRelative('bundle');
 
     expect(fs.existsSync(chunks + '/bundle.js')).toBeTruthy();
-    expect(fs.existsSync(chunks + '/chunk.js')).toBeTruthy();
```
jbedard, line 9: Seems like rollup no longer renames to "chunk", I assume because `strings.js` doesn't import anything else. Might be worth adding a more complicated code-splitting test...

internal/node/node_patches.js
alexeagle: yeah I think we don't really care - in fact there's a mechanism to get GitHub to stop showing us this generated file which I'd like to look into sometime https://github.com/github/linguist#generated-code
(it excludes from the lines count but I dunno about code review and source browsing...)

packages/rollup/test/code_splitting/spec.js
alexeagle: yeah this is the simple "unit test" level assertion, I think your change here is fine since we still check that it outputs two files rather than one

Issue #2092, opened 2020-07-31
Documentation has load statements that produce errors


gonzojive:
# 🐞 bug report

### Affected Rule

ts_library, rollup_bundle, and probably many more

### Is this a regression?

Yes, previous cut and paste from the documentation worked.

### Description

`bazel query //...` fails with the following setup, generated by copy/paste from https://bazelbuild.github.io/rules_nodejs/Rollup.html and other documentation pages:

```skylark
# build.bazel
load("//packages/rollup:index.bzl", "rollup_bundle")
load("//packages/typescript:index.bzl", "ts_library")
```

```skylark
# WORKSPACE.bazel
# Bazel workspace created by @bazel/create 2.0.1

# Declares that this directory is the root of a Bazel workspace.
# See https://docs.bazel.build/versions/master/build-ref.html#workspace
workspace(
    # How this workspace would be referenced with absolute labels from another workspace
    name = "gtile",
    # Map the @npm bazel workspace to the node_modules directory.
    # This lets Bazel use the same node_modules as other local tooling.
    managed_directories = {"@npm": ["node_modules"]},
)

# Install the nodejs "bootstrap" package
# This provides the basic tools for running and packaging nodejs programs in Bazel
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "build_bazel_rules_nodejs",
    sha256 = "0f2de53628e848c1691e5729b515022f5a77369c76a09fbe55611e12731c90e3",
    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/2.0.1/rules_nodejs-2.0.1.tar.gz"],
)

# The yarn_install rule runs yarn anytime the package.json or yarn.lock file changes.
# It also extracts and installs any Bazel rules distributed in an npm package.
load("@build_bazel_rules_nodejs//:index.bzl", "yarn_install")
yarn_install(
    # Name this npm so that Bazel Label references look like @npm//package
    name = "npm",
    package_json = "//:package.json",
    yarn_lock = "//:yarn.lock",
)

```

```js
// package.json
{
    "name": "gtile",
    "version": "0.1.0",
    "private": true,
    "devDependencies": {
        "@bazel/bazelisk": "latest",
        "@bazel/buildifier": "latest",
        "@bazel/ibazel": "latest",
        "@bazel/rollup": "^2.0.1"
    },
    "scripts": {
        "build": "bazel build //...",
        "test": "bazel test //..."
[...]

Issue #2087, opened 2020-07-29
missing pieces in installation document


adb:
# 🐞 bug report

### Affected Rule

This is about some gaps and inconsistencies in the setup document for the typescript rules (e.g., ts_library):

https://github.com/bazelbuild/rules_nodejs/blob/stable/packages/typescript/install.md

### Is this a regression?

Maybe? The docs are out of sync with the code.

### Description

I'm walking through an install from scratch trying to debug an issue with another install I have, making notes of what's not quite right in the docs. So far:

- Installing the typescript package should be mentioned at the same time as @bazel/typescript. Or (I'm still relatively new to the node universe) should @bazel/typescript perhaps have an explicit dependency on typescript?

- The packages used in the example load() commands are inconsistent with each other and with the WORKSPACE "yarn create @bazel" generated for me: e.g.,

load("@npm//typescript:index.bzl", "tsc")
load("//packages/typescript:index.bzl", "ts_library")

vs what works on a freshly created project, which is

load("@npm//@bazel/typescript:index.bzl", "ts_devserver", "ts_library")

## 🔬 Minimal Reproduction

n/a

## 🔥 Exception or Error

n/a

## 🌍  Your Environment

cross-platform (I've been working on macOS Mojave and GCP's Debian buster)

**Output of `bazel version`:**

bazel 3.4.1 (Linux)
bazel 3.3.0 (Mac)

**Rules_nodejs version:**

(Please check that you have matching versions between WORKSPACE file and `@bazel/*` npm packages.)

2.0.1, still present in trunk code

**Anything else relevant?**

alexeagle:
Your yarn probably had an existing 1.x @bazel/create package and reused it.
Could you check 
yarn create @bazel@latest

evahowe:
@adb - did @alexeagle solve your problem?

PR #2101, opened 2020-08-08
docs: use @npm//@bazel for example load sites

mattem:
Have the docs for rules_nodejs use the `@npm//@bazel/foo` in the examples for the load sites for different packages.
The current docs show loading them from `//packages/foo`, which causes confusion. The common case seems to be to have a external workspace called `@npm`

closes: #2092, #2087

### DIFF
--- a/packages/labs/protobufjs/ts_proto_library.bzl
+++ b/packages/labs/protobufjs/ts_proto_library.bzl
@@ -175,7 +175,7 @@ a `ts_library` can appear, such as in the `deps[]` of another `ts_library`.
 Example:
 
 ```python
-load("//packages/typescript:index.bzl", "ts_library", "ts_proto_library")
+load("@npm//@bazel/typescript:index.bzl", "ts_library", "ts_proto_library")
 
 proto_library(
     name = "car_proto",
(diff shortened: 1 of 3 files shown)


packages/labs/protobufjs/ts_proto_library.bzl
```diff
@@ -175,7 +175,7 @@ a `ts_library` can appear, such as in the `deps[]` of another `ts_library`.
 Example:
 
 ```python
-load("//packages/typescript:index.bzl", "ts_library", "ts_proto_library")
+load("@npm//@bazel/typescript:index.bzl", "ts_library", "ts_proto_library")
```
alexeagle, line 178: Should we point out that this example assumes you installed dependencies under @npm ?

alexeagle:
Another solution could be to apply the same replacements in the docgen rule that we do for pkg_npm

packages/labs/protobufjs/ts_proto_library.bzl
mattem: I did wonder if we should do that, however the setup for [managed deps example](https://bazelbuild.github.io/rules_nodejs/install.html#using-bazel-managed-dependencies) shows using the `npm` workspace, so I thought it was reasonable to assume that if the user had followed along with the docs, they'd too have the `npm` workspace. 

I can add the note to those examples if we think it's need however

PR #2108, opened 2020-08-12
fix(rollup): allow config files to override default onwarn method

### LINKED ISSUE #2084: rollup_bundle overrides onwarn in worker mode
# 🐞 bug report

### Affected Rule

`rollup_bundle(supportWorker=True)`

### Description

The [default onwarn](https://github.com/bazelbuild/rules_nodejs/blob/2.0.1/packages/rollup/index.js#L152-L154) [overrides](https://github.com/bazelbuild/rules_nodejs/blob/2.0.1/packages/rollup/index.js#L224) any custom `onwarn` from the config file.

jbedard:
Fixes #2084

... will try adding a test still.

### DIFF
--- a/packages/rollup/index.js
+++ b/packages/rollup/index.js
@@ -103,11 +103,15 @@ function extractEnvironmentVariables(vars) {
 //  input:  https://rollupjs.org/guide/en/#inputoptions-object
 //  output: https://rollupjs.org/guide/en/#outputoptions-object
 async function parseCLIArgs(args) {
-  let inputOptions = {
+  // Options which the CLI args or config file can override
+  const defaultInputOptions = {
     onwarn(...warnArgs) {
       worker.log(...warnArgs);
     },
   };
+
+  // Options which can override the config file
+  let inputOptions = {};
 
   let outputOptions = {};
 
@@ -193,6 +197,9 @@ async function parseCLIArgs(args) {
     // may be external and persisted across runs
     delete inputOptions.output;
   }
+
+  // Provide default inputOptions which can be overwritten
+  inputOptions = {...defaultInputOptions, ...inputOptions};
 
   // The inputs are the rule entry_point[s]
   inputOptions.input = inputs;
--- a/packages/rollup/test/integration/foo/BUILD.bazel
+++ b/packages/rollup/test/integration/foo/BUILD.bazel
@@ -11,10 +11,12 @@ copy_file(
 )
 
 ts_project(
+    name = "foo_tslib",
     srcs = [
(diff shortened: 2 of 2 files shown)


jbedard:
Interesting this is now throwing:
```
(node:42) UnhandledPromiseRejectionWarning: Error: 'date-fns' is imported by node_modules/@foo/lib/index.js, but could not be resolved – treating it as an external dependency
    at Object.onwarn (/b/f/w/bazel-out/k8-fastbuild/bin/packages/rollup/test/integration/_bundle.umd.rollup_config.js.cjs.js:34:11)
    at Graph.onwarn (/b/f/w/node_modules/rollup/dist/shared/rollup.js:17999:25)
    at Graph.warn (/b/f/w/node_modules/rollup/dist/shared/rollup.js:17803:14)
    at ModuleLoader.handleResolveId (/b/f/w/node_modules/rollup/dist/shared/rollup.js:16823:24)
    at /b/f/w/node_modules/rollup/dist/shared/rollup.js:16709:26
    at async Promise.all (index 0)
    at async Promise.all (index 2)
    at async Promise.all (index 0)
    at async Promise.all (index 0)
    at async Promise.all (index 0)
```

However the `ts_project` using `date-fns` does have it in the `deps`. Is that because the `rollup_bundle(entry_point)` doesn't pull in the `deps` of that `entry_point`?

@alexeagle ?

jbedard:
@alexeagle @mistic note the change to the integration tests in the second commit.. should that be required? Is `js_library` not passing the `ts_project(deps)` along properly?

The rollup rule that test runs was not failing previously due to the bug this PR is fixing.

mistic:
@jbedard there is no way `js_library` could get the the transitive dependencies with the inputs you're using. You'll have to replace `"user.d.ts", ":index.js",` by `":tsconfig"`

mistic:
@jbedard I think you removed `":user.js",` by mistake

jbedard:
Thanks @mistic. Depending on the `ts_project` rule (instead of individual output files) is what I would have expected in the first place, idk why the test was written this way? But this rollup fix exposed that issue 🤷

jbedard:
Yet the error remains...? 🤔

mistic:
@jbedard I think the changes we discuss offline will now make the CI pass! My suggestion above was wrong and the output from the ts_project rule should be a dep for `js_library`

alexeagle (approved):

jbedard:
FWIW I never found an easy way to test this in my brief attempt. Thanks for merging.

Issue #1967, opened 2020-06-22
Support pkg_npm() targets as dependencies for ts_library()


nkeynes:
# 🚀 feature request

### Relevant Rules

ts_library() and pkg_npm()

### Description

Currently, attempting to use a pkg_npm() target as a ts_library dependency results in an error of the form "<merged target //Foo/src:pkg> doesn't contain declared provider 'DeclarationInfo', even if the pkg_npm was originally itself generated in part from another ts_library() target. For example,
... foo/BUILD:
ts_library(
   name = "foo_lib",
   module_name = "@Foo"
   srcs = glob(["**/*.ts"])
)

filegroup(
   name = "foo_static",
   srcs = glob(["**/*.json","**/*.css"])
)

pkg_npm(
  name = "foo",
  package_name = "@Foo",
  srcs = [":foo_static"],
  deps = [":foo_lib"]
)
... bar/BUILD:
ts_library(
   name = "bar",
   srcs = glob(["**/*.ts")
   deps = ["//Foo:foo"]
)
### Describe the solution you'd like

Ideally, pkg_npm() targets could be accepted as ts_library dependencies and treated the same as any imported node module. Our main use case for this is in libraries that bundle separate files (e.g. JSON, CSS, etc) along with TS code, where those files are resolved at runtime relative to the source files. Currently (pre-bazel) this is implemented by just copying all the additional files into the tsc build directory

### Describe alternatives you've considered

I've tried to work around this by putting a direct dependency between the ts_library() targets, which is sufficient for compilation, but then there doesn't seem to be an obvious way to get the correct, merged pkg_npm dependency at runtime - having both as dependencies leads to mapping errors ala:
attribute deps: conflicting mapping at '//Foo/src:pkg': '@Foo' maps to both ["execroot", "bazel-out/darwin-fastbuild/bin/lib/Foo/src".

alexeagle:
I tried to do this in https://github.com/bazelbuild/rules_nodejs/pull/1898/files
The problem is that `pkg_npm` produces a directory artifact, so it can't also expose individual files in that directory as labels. That means it also can't expose providers that let it serve as a dependency to other rules.
Our plan is to generalize `node_module_library` which is currently used as a rule to expose third-party packages from npm. It's basically the shape we need to depend on first-party libraries too.

alexeagle:
Decided that js_library is the right rule for this

alexeagle:
[...]

PR #2109, opened 2020-08-12
feat(builtin): new js_library rule

mistic:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently the `js_library` rule is very limited and is not built to be public. Additionally `node_module_library` was used only for internal purposes in order to represent the npm packages built upon yarn_install/npm_install. 

Issue Number: https://github.com/bazelbuild/rules_nodejs/issues/1967 , https://github.com/bazelbuild/rules_nodejs/issues/1896 , https://github.com/bazelbuild/rules_nodejs/issues/149

## What is the new behavior?
That PR introduces a new design for a new `js_library` that could replace both the legacy rules and that is supposed to act as the default/standard rule when you want to share first party libraries across the repository.

- [ ] Yes
- [x] No

## Other information
I've considered that PR doesn't include any breaking change as both `node_module_library` and `js_library` are not public rules available to be widely used out of this repo.

PS: For now I'm opening the PR as a draft one to collect feedback from @alexeagle and @mrmeku


e2e/nodejs_image/foolib/BUILD.bazel
```diff
@@ -1,8 +1,14 @@
+load("@build_bazel_rules_nodejs//:index.bzl", "copy_to_bin")
 load("@build_bazel_rules_nodejs//internal/js_library:js_library.bzl", "js_library")
 
+copy_to_bin(
```
alexeagle, line 4: FYI @gregmagolan our thought is that the copy feature can be orthogonal to js_library, and where composition is possible we should use it - thoughts?

internal/js_library/js_library.bzl
```diff
@@ -1,4 +1,4 @@
-# Copyright 2017 The Bazel Authors. All rights reserved.
+# Copyright 2020 The Bazel Authors. All rights reserved.
```
alexeagle, line 1: note for future - my understanding is that copyright dates doesn't need to be updated on edit

packages/terser/test/inline_sourcemap/input.js
```diff
@@ -1,11 +1,6 @@
-'use strict';
-exports.__esModule = true;
 // clang-format off
-var MyClass = /** @class */ (function () {
-    function MyClass(something) {
+export class MyClass {
```
alexeagle, line 2: can you explain why these changed?

internal/js_library/js_library.bzl
```diff
@@ -49,96 +53,124 @@ def write_amd_names_shim(actions, amd_names_shim, targets):
...
+        workspace_name = ctx.label.workspace_name if ctx.label.workspace_name else ctx.workspace_name
+        providers.append(NpmPackageInfo(
+            direct_sources = direct_sources,
+            sources = transitive_sources,
+            workspace = workspace_name,
+        ))
 
-    return providers
+    return struct(
+        typescript = struct(
```
alexeagle, line 138: what happens if we don't provide the legacy string-typed "typescript" provider? if something is broken we should maybe update that use site rather than carry this forward, as it will be a breaking change to remove it again

internal/js_library/js_library.bzl
```diff
@@ -49,96 +53,124 @@ def write_amd_names_shim(actions, amd_names_shim, targets):
...
+    transitive_declarations = depset(transitive = transitive_declarations_depsets)
+    transitive_sources = depset(transitive = sources_depsets + named_module_sources_depsets)
 
     providers = [
         DefaultInfo(
-            files = files_depset,
-            runfiles = ctx.runfiles(files = ctx.files.srcs),
+            files = direct_sources,
+        ),
+        DeclarationInfo(
```
alexeagle, line 102: does this produce DeclarationInfo when there are no .d.ts files? if so that's not nice as it takes away the error messaging when you put it in the deps of a ts_project

internal/js_library/js_library.bzl
```diff
@@ -49,96 +53,124 @@ def write_amd_names_shim(actions, amd_names_shim, targets):
...
+           # exclude eg. external/npm/node_modules/protobufjs/node_modules/@types/node/index.d.ts
+           # these would be duplicates of the typings provided directly in another dependency.
+           # also exclude all /node_modules/typescript/lib/lib.*.d.ts files as these are determined by
+           # the tsconfig "lib" attribute
+           len(f.path.split("/node_modules/")) < 3 and f.path.find("/node_modules/typescript/lib/lib.") == -1
+    ])
+
+    transitive_declarations_depsets = [declarations]
+
+    for dep in ctx.attr.deps:
```
alexeagle, line 85: in all these cases there are factory functions to construct the provider which bring in the transitive ones, just makes it a bit shorter

internal/js_library/js_library.bzl
```diff
@@ -49,96 +53,124 @@ def write_amd_names_shim(actions, amd_names_shim, targets):
...
-        else:
-            files.append(src)
-
-    for p in files:
-        if p.basename.endswith(".js") or p.basename.endswith(".js.map") or p.basename.endswith(".json"):
-            js_files.append(p)
-
-    files_depset = depset(files)
+        if src.is_source and src.path.startswith("external/"):
+            include_npm_package_info = True
```
alexeagle, line 65: this merits a comment about why it's done this way

packages/terser/test/inline_sourcemap/input.js
mistic: I've remembered at some point I've ran `yarn format` probably it was a result of it. I'll recover the original version

internal/js_library/js_library.bzl
mistic: I've got the year back to the original one

internal/js_library/js_library.bzl
mistic: included

internal/js_library/js_library.bzl
mistic: nice one 👍 fixed

internal/js_library/js_library.bzl
mistic: I've carried this from the old `node_module_library`. But now that I think better about it, I think it's safe to remove it

mistic:
@alexeagle yeah I believe an example should be added in a separate PR as soon as we finish the design of this rule and got the CI to pass!

> I like that you haven't exposed this as public API yet, so we can still iterate on it even after next release.

The idea is to make it public once we got around all the details and we feel confident about it.

internal/js_library/js_library.bzl
mistic: Do you mean for `JSModuleInfo` and `JSNamedModuleInfo` ?

internal/js_library/js_library.bzl
```diff
@@ -49,96 +53,117 @@ def write_amd_names_shim(actions, amd_names_shim, targets):
...
+            break
+
+    declarations = depset([
+        f
+        for f in ctx.files.srcs
+        if (
+               f.path.endswith(".d.ts") or
+               # package.json may be required to resolve "typings" key
+               f.path.endswith("/package.json")
```
mrmeku, line 79: Would a tsconfig.json also be necessary to resolve typings? I know under `compilerOptions` you can include an array of types to include

```"compilerOptions": {
    "types": ["node", "cypress", "cypress-plugin-snapshots"]
  }```

Comment (mistic) [reply internal/js_library/js_library.bzl:79]: I believe `tsconfig.json` is only need during the `tsc` compilations. If `js_library` represents a final state of a library, the only thing needed to map types when sharing it is the `package.json` 😸

File: internal/js_library/js_library.bzl (hunk 6 above)
Comment (mistic) [reply internal/js_library/js_library.bzl:90]: Alright, got it and simplified it!

File: packages/labs/grpc_web/BUILD.bazel [hunk 9]
```diff
@@ -42,10 +42,11 @@ well_known_protos = [
 
 js_library(
     name = "bootstrap_scripts",
-    srcs = [
+    named_module_srcs = [":google-protobuf-%s__umd" % p for p in well_known_protos],
```
alexeagle, line 45: this is where you get an error on CI
`ERROR: /home/circleci/rules_nodejs/packages/labs/test/grpc_web/BUILD.bazel:41:1: in srcs attribute of _karma_web_test rule //packages/labs/test/grpc_web:proto_with_deps_test_suite_wrapped_test: '//packages/labs/grpc_web:bootstrap_scripts' does not produce any _karma_web_test srcs files (expected .js). Since this rule was created by the macro 'karma_web_test_suite', the error might have been caused by the macro implementation`

it's suspicious that you moved some of the `deps` to `named_module_srcs` - seems like it would need to be all or none, since they're all __umd targets?

mistic: 👍

mistic:
@alexeagle @mrmeku @gregmagolan the CI is now green. I think the current state of the implementation represents a good point to start with. Do you have any other feedback or a design request to proceed with?

third_party/github.com/gjtorikian/isBinaryFile/BUILD.bazel
```diff
@@ -5,23 +5,31 @@ licenses(["notice"])
...
         "package.json",
-    ],
+    ]
+)
+
+js_library(
+    name = "isBinaryFile",
+    srcs = [":package_json_bin", "src/index.js"],
+    # BEGIN-INTERNAL
+    deps = [":tsconfig"],
```
alexeagle, line 31: why add this dep? seems unnecessary

third_party/github.com/gjtorikian/isBinaryFile/BUILD.bazel
```diff
@@ -5,23 +5,31 @@ licenses(["notice"])
 exports_files(["LICENSE"])
 
 load("@build_bazel_rules_nodejs//internal/js_library:js_library.bzl", "js_library")
+#load("@build_bazel_rules_nodejs//internal/js_library:js_library2.bzl", "js_library")
```
alexeagle, line 8: spurious comment line?

internal/js_library/js_library.bzl
```diff
@@ -50,95 +56,147 @@ def write_amd_names_shim(actions, amd_names_shim, targets):
...
 _js_library = rule(
     implementation = _impl,
     attrs = {
-        "amd_names": attr.string_dict(doc = _AMD_NAMES_DOC),
-        "is_windows": attr.bool(mandatory = True, doc = "Automatically set by macro"),
+        "amd_names": attr.string_dict(
+            doc = _AMD_NAMES_DOC,
+        ),
+        "deps": attr.label_list(
+            doc = "Transitive dependencies of the package",
```
alexeagle, line 159: we should explain a bit more here based on discussion today

e2e/nodejs_image/foolib/BUILD.bazel
gregmagolan: Since js_library is linkable with a LinkablePackageInfo I think the default behaviour should that everything is available via the link. Linker is linking to the output folder so it seems like a bad user experience for users to have to copy_to_bin all source files if they want them exposed.

internal/js_library/js_library.bzl
```diff
@@ -50,95 +56,147 @@ def write_amd_names_shim(actions, amd_names_shim, targets):
...
-        else:
-            files.append(src)
-
-    for p in files:
-        if p.basename.endswith(".js") or p.basename.endswith(".js.map") or p.basename.endswith(".json"):
-            js_files.append(p)
+    include_npm_package_info = False
+
+    for file in files:
+        if file.basename.endswith(".js") or file.basename.endswith(".js.map") or file.basename.endswith(".json"):
```
alexeagle, line 65: we always forget .jsx

e2e/nodejs_image/foolib/BUILD.bazel
alexeagle: @mistic the `js_library` symbol we import everywhere is already a macro, so it's pretty easy to fold the copy_to_bin as a separate rule (keeping our separation of concerns) but defer this design change, which is kinda orthogonal.

Also makes this PR a lot smaller since most of the use sites won't have to change.
Then we could tackle this design question separately, WDYT?

gregmagolan: Maybe not that simple. Embedded I was filtering on `is_source` which is not available in a macro:
```
        if src.is_source and not src.path.startswith("external/"):
            dst = ctx.actions.declare_file(src.basename, sibling = src)
            if ctx.attr.is_windows:
                copy_cmd(ctx, src, dst)
            else:
                copy_bash(ctx, src, dst)
```

gregmagolan: copy_to_bin may barf if you pass it a file already in the output tree

gregmagolan: Perhaps an option to disable to auto-copy to bin and then the rule will need to decide if it is linking to the output tree or the source tree again if that is set? Seems like a power-user feature optimization feature to me.

mistic:
I've promoted it to the state `ready for review` and I'll work and close all the feedback in the next couple of hours.

third_party/github.com/gjtorikian/isBinaryFile/BUILD.bazel
mistic: 👍

internal/js_library/js_library.bzl
mistic: @alexeagle I look at `.jsx` (and also `.tsx`) as source files that should be first processed before be able to ship in a node module and therefore probably should not be included in the js_library too. What are your thoughts around it?

third_party/github.com/gjtorikian/isBinaryFile/BUILD.bazel
mistic: 👍

internal/js_library/js_library.bzl
mistic: I've included a more complete description for deps

e2e/nodejs_image/foolib/BUILD.bazel
mistic: After reading the feedback both provided here, I believe I'll recover the old behaviour of auto-copy things into bin for now.

mistic: Alright, done

mistic:
@alexeagle I think I just need to remove the changes I've introduced on the files with `copy_to_bin` and we should be good to go. Will finish it tomorrow

internal/js_library/js_library.bzl
alexeagle: I'm not sure. I guess if packages on npm rarely include jsx then you're right, should be only internal within a package

mistic: yeah that is case. We can revisit later if we found the other way around!

packages/jasmine/BUILD.bazel
```diff
@@ -53,15 +53,26 @@ nodejs_test(
...
         "jasmine_runner.js",
     ],
 )
 
+# We need to use ":srcs" here instead of an
+# array for the source files otherwise our
+# logic on js_library will consider this a
+# node_module when used in examples/user_managed_deps
+# which will endup failing as we cannot use node_modules
+# from multiple workspaces.
```
mistic, line 69: /cc @alexeagle

PR #2112, opened 2020-08-13
feat(example): add targets in angular_bazel_architect for production serve and build

alan-agius4:


examples/angular_bazel_architect/BUILD.bazel
```diff
@@ -10,6 +10,21 @@ filegroup(
     visibility = ["//:__subpackages__"],
```
alan-agius4, line 10: Note: this can also be done with a macro, but I preferred to keep the example simple.

PR #2113, opened 2020-08-13
fix(examples): use ./ prefix on babel config file

alexeagle:
Otherwise if it's in a subfolder, babel will think it's an npm package


Aghassi:
I can't approve this, but looks good to me!

PR #2128, opened 2020-08-18
refactor: move node version list into its own file

alexeagle:
Add a package.json script for updating it


scripts/update-nodejs-versions.js
```diff
@@ -70,7 +71,11 @@ async function getNodeJsVersion(version) {
 async function main() {
   const versions = await getNodeJsVersions();
   const nodeRepositories = await Promise.all(versions.map(getNodeJsVersion));
-
+  console.log('""" Generated code; do not edit');
+  console.log('Update by running yarn update-nodejs-versions\n"""\n');
+  // Suppress buildifier
+  console.log('# @unsorted-dict-items');
+  console.log('NODE_VERSIONS = {');
```
mattem, line 78: nit: perhaps add to a single array and join, rather than logging multiple times? Kinda preference though :)

PR #2130, opened 2020-08-20
feat(typescript): generate tsconfig.json for ts_project

### LINKED ISSUE #2058: More ergonomic way to generate tsconfig.json
Some discussion on Slack points to a desire to use `ts_project` but not check in a `tsconfig.json` file for each compilation.

This is already possible in userspace and is demonstrated here:
https://github.com/bazelbuild/rules_nodejs/blob/d9ce5e8893a301238a7024899d850258e73c9ebf/packages/typescript/checked_in_ts_project.bzl#L18-L47

But this is a little unwieldy for users.

The core problem is that paths in the tsconfig file are relative to the location it was found on disk, so by generating it, you're making it a lot harder to point to the sources. Maybe one way to solve this is with a little syntax sugar around skylib's `write_file` like

```
def write_tsconfig(name, config, files, out):
    # workspace is up three dirs (bazel-out/arch/bin) plus number of segments in the package
    workspace_root = "/".join([".."] * (3 + len(native.package_name().split("/"))))

    amended_config = struct(
        files = ["/".join([workspace_root, native.package_name(), f]) for f in files],
        **config
    )
    write_file(
        name = name,
        content = [amended_config.to_json()],
        out = out,
    )
```

This would be used like
```
    tsconfig = "tsconfig_%s.json" % name

    write_tsconfig(
        name = "_gen_tsconfig_%s" % name,
        config = {
            "compilerOptions": struct(
                lib = ["es2017", "dom"],
                strict = True,
                target = "es2015",
                module = "commonjs",
                removeComments = True,
                declaration = True,
                skipLibCheck = True,
            ),
        },
        files = [src],
        out = tsconfig,
    )

    ts_project(
        name = name,
        srcs = [src],
        declaration = True,
        tsconfig = tsconfig,
        **kwargs
    )
```

It's pretty awkward to mix dict and struct in this way but I didn't find a better solution yet. You could have a `compiler_options` attribute containing the struct, but you still need a dict-type attribute for a catch-all of tsconfig top-level properties that we don't know about.

Another consideration here is keeping a good developer experience, if you have no tsconfig.json file in source control then your editor doesn't know what to do. Maybe a `generated_file_test` could be used to check in a tsconfig.json but assert it matches what Bazel would generate - but again the files can't actually be the same if they contain any tsconfig.json-relative paths. Another answer is to recommend having the base tsconfig.json (with the compilerOptions shared by editor and Bazel) and then only generate tsconfig that extends from it (like ts_library does)

#1931 suggests using the existing `ts_config` rule which makes sense too

alexeagle:
This is an experimental, opt-in feature where ts_project will generate a tsconfig.json file in place of
the user specifying one in the source directory.

I expect a long tail of compatibility bugs since any path appearing in this generated file needs to point from
bazel-out back to the source directory.

Fixes #2058


alexeagle:
TODO:
- you currently have to correctly specify things twice, like `declaration`
- we should look through deps and figure out how to add ts_project deps as project references
- we amend files to point back to srcs but I think we should handle includes/excludes, etc
- other path entries like typeRoots are going to be busted
- the write_tsconfig helper lives in the file with TsConfigInfo but has no relation, needs to be fixed

Toxicable:
What do you think about supporting a single base tsconfig that the generated one extends?
I think that would be the most popular use case, having a tsconfig in the root of ya project that does things like `noImplicitThis` or `module` then individual targets which may override an option or two

alexeagle:
yeah let me add a demo of extending a user-authored tsconfig, it's already possible but it needs the same path mangling to point to source tree

alexeagle:
Supports generated .ts sources now

alexeagle:
I don't think we can move the write_tsconfig helper to be related to TsConfigInfo, because we need to read all the properties in the macro where we predict the outputs from tsc.

PR #2132, opened 2020-08-21
fix(karma): allow custom browsers to specify args (fixes #595)

### LINKED ISSUE #595: karma_web_test ignores browser launch flags
# 🐞 bug report

### Affected Rule

The issue is caused by the rule: `karma_web_test`

### Is this a regression?

No

### Description

Browser definitions in `rules_webtesting` ([chromium-local.json](https://github.com/bazelbuild/rules_webtesting/blob/master/browsers/chromium-local.json) as an example) have args that can be passed to the browser when launching. These are not used by the `karma.conf.js` which ultimately sets the `flags` used by the Karma launcher: https://github.com/bazelbuild/rules_nodejs/blob/master/packages/karma/karma.conf.js#L293.

## 🔬 Minimal Reproduction

The e2e tests reproduce this; in `e2e/karma` we see that Chromium is launched without either of the arguments in the `chromium-local.json`.

## 🌍  Your Environment

**Operating System:**

```
ProductName:	Mac OS X
ProductVersion:	10.14.3
BuildVersion:	18D109
```

```
Build label: 0.23.1
Build time: Mon Mar 4 10:40:32 2019 (1551696032)
Build timestamp: 1551696032
Build timestamp as int: 1551696032
```
**Rules version (SHA):**

```
2be3c7a0794f7771c0317e673799fbc1320eb036
```

**Anything else relevant?**

The context for needing this might help! On macOS Mojave, running Chromium when non-headless needs `--no-sandbox` otherwise `Chromium Helper` crashes and tabs can't render. I didn't dig into it _too_ much, so there's likely a smaller subset of features we could disable to get it running.

applmak:
Today, the args as specified in the various manifests of the rules_webtesting
browsers never make it into the generated karma.conf.js. This results in these
arguments never being used when launching Chrome, preventing customization of
browsers such as window size, enabling remote debugging, or other flag-based
options. This PR fixes this by reading those arguments from the manifest, and
adding them to the browsers list in the generated karma.conf.js.

This PR also includes a small (manual) test that specifies an additional
argument in a custom browser, which then can be seen to be used via:
```
VERBOSE_LOGS=1 bazel run packages/karma/test/karma:testing_custom_chrome
```
Note the appearance of the additional flags in the new debug output.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

See description above.

Issue Number: #595

## What is the new behavior?
See description above.

Perhaps! This might surprise people. If anyone has been using custom browsers up to now, they will suddenly see new arguments on their command-lines. However, given that the feature didn't really work before, I don't think that this is a big deal.

- [ ] Yes
- [x] No


packages/karma/karma.conf.js
```diff
@@ -321,13 +321,21 @@ try {
                   webTestNamedFiles['CHROMIUM']}' in runfiles`);
             }
           }
+          // Read any additional chrome options (as specified by the
+          // rules_webtesting manifest).
+          const chromeOptions = (webTestMetadata['capabilities'] || {})['goog:chromeOptions'];
+          const additionalArgs = (chromeOptions ? chromeOptions['args'] : []).filter(arg => {
```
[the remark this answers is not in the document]
applmak: Done.

packages/karma/test/karma/BUILD.bazel
```diff
@@ -41,3 +43,9 @@ karma_web_test_suite(
         "requirejs-config.js",
     ],
 )
+
+custom_browser(
```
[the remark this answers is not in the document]
applmak: I had to make the generated .conf.js a 'predeclared' output so as to write this test. I hope that's okay.

applmak:
Ping!

packages/karma/test/karma/BUILD.bazel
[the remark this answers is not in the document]
alexeagle: yeah that looks fine. Note you could also add the file to an Output Group to give a way to select the dynamic (non-predeclared) output

packages/karma/test/karma/karma.conf.js.golden
```diff
@@ -0,0 +1,454 @@
+// Karma configuration
```
alexeagle, line 1: looks brittle that this golden is so long, probably more than we meant to assert on in this case.

I'd like to make a convenience rule for `golden_content_test` or sth so we can just say "make sure this line (± whitespace) appears in the file"

this is okay for now if the content is stable

packages/karma/test/karma/BUILD.bazel
```diff
@@ -41,3 +44,16 @@ karma_web_test_suite(
...
+    name = "custom_chrome",
+    browser = "@io_bazel_rules_webtesting//browsers:chromium-local",
+    metadata = "custom_chrome.json",
+)
+
+generated_file_test(
+    name = "test_custom_chrome_karma_conf",
+    src = "karma.conf.js.golden",
+    # Refers to the output from ts_project above
```
alexeagle, line 57: oops copy pasta

applmak: Fixed.

packages/karma/test/karma/karma.conf.js.golden
applmak: It is stable. Should I file a bug about (perhaps) amending generated_file_test to include substring match capability?

applmak:
Ah, looks like it's not as stable as I would have wished. I'll make that improvement to generated_file_test first.

packages/karma/test/karma/karma.conf.js.golden
applmak: I have added that capability. PTAL.

applmak:
Yep! It looks like I will need help. I'm not sure of the mechanism that seems to be excluding the :testing_chromium-local from buildkite/rules-nodejs-nodejs, but when it tried to run :testing_custom_chrome, it errors with a missing .so.

applmak:
Ping!

alexeagle:
Hmm, I don't have time to dig in right now. The error for anyone following along:

```
ERROR [launcher]: Cannot start ChromeHeadless
  | /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/sandbox/linux-sandbox/1531/execroot/build_bazel_rules_nodejs/bazel-out/k8-fastbuild/bin/packages/karma/test/karma/testing_custom_chrome.sh.runfiles/io_bazel_rules_webtesting/third_party/chromium/chromium.out/chrome-linux/chrome: error while loading shared libraries: libatk-bridge-2.0.so.0: cannot open shared object file: No such file or directory
```

Could this be due to the custom_chrome.json enabling some options that cause Chromium to dynamic-link more libraries than before? We only installed a minimal set on our CI workers - ideally you'd find a way to assert that the custom browser args are being honored, but without introducing new dependencies on the underlying machine.

(Also WDYT about the substring mode accepting a string attribute with the content to match, rather than (or in addition to) accepting a golden file input? In a lot of cases you just want the equivalent of `assert_contains` which would mean a lot of tiny one-line files)

applmak:
So, these tests have never worked on CircleCI. The CircleCL script executes:
```
bazel test --flaky_test_attempts=3 --build_tests_only --local_test_jobs=12 --show_progress_rate_limit=5 --curses=yes --color=yes --terminal_columns=143 --show_timestamps --verbose_failures --jobs=30 --announce_rc --experimental_repository_cache_hardlinks --disk_cache= --sandbox_tmpfs_path=/tmp --experimental_build_event_json_file_path_conversion=false --build_event_json_file=/tmp/tmperwd7qg1/test_bep.json --test_tag_filters=-e2e,-examples,-manual,-browser:chromium-local,-browser:firefox-local,-cypress --test_env=HOME --test_env=USE_BAZEL_VERSION --sandbox_writable_path=/var/lib/buildkite-agent/.cache/bazelisk -- //... /
```

which removes tests that are tagged browser:chromium-local. 

Running
```
bazel query --output=build packages/karma/test/karma:all
```
shows the generated test rule as:
```
web_test(
  name = "testing_custom_chrome",
  tags = ["browser:custom_chrome", "custom_chrome", "ibazel_notify_changes", "native"],
  generator_name = "testing",
  generator_function = "karma_web_test_suite",
  generator_location = "packages/karma/test/karma/BUILD.bazel:19:1",
  size = "large",
  browser = "//packages/karma/test/karma:custom_chrome",
  config = "@io_bazel_rules_webtesting//web:default_config",
  data = [],
  launcher = "//packages/karma/test/karma:testing_wrapped_test",
  test = "//packages/karma/test/karma:testing_wrapped_test",
)
```

which means it does not get excluded from the circlecl integration test.

So: either, I can add a misleading tag to the test suite (something like browser:chromium-local), or I can extend the circlecl tag exclusion to include browser:custom_chrome. LMK what you want.

applmak:
I found the configuration at .buildci/presubmit.yml and updated it to include an exclusion for tests with the tag 'browser:custom_chrome'.

PTAL.

applmak:
Ping!

applmak:
Pong!

applmak:
Pingigidipong!

alexeagle:
CI rerun still failed, trying a rebase on stable/HEAD to see if it's green

alexeagle:
Thanks for pinging :)

PR #2139, opened 2020-08-24
Ts config ts project

alexeagle:


packages/typescript/internal/ts_project.bzl
```diff
@@ -432,7 +436,7 @@ def ts_project_macro(
         extra_deps.append("_validate_%s_options" % name)
 
     typings_out_dir = declaration_dir if declaration_dir else out_dir
-    tsbuildinfo_path = ts_build_info_file if ts_build_info_file else tsconfig[:-5] + ".tsbuildinfo"
+    tsbuildinfo_path = ts_build_info_file if ts_build_info_file else name + ".tsbuildinfo"
```
gregmagolan, line 439: Should this be derived from the `ts_build_info_file` attr just added in https://github.com/bazelbuild/rules_nodejs/pull/2138?

alexeagle: it is

gregmagolan: 🤦

Issue #2140, opened 2020-08-24
ts_project#extends should be a single label, not a list


alexeagle:
Now that you can use the `ts_config` rule to represent `tsconfig.json#extends` dependencies on other `tsconfig.json` files, the `extends[]` feature in `ts_project` is kinda wrong. You have to list the whole transitive array rather than just the one tsconfig file you directly reference. Also you can't tell locally from reading a `tsconfig.json` file what the correct value of `extends` should be

alexeagle:
Actually I think the right thing is for extends to become a single value instead of a list. If `tsconfig` is a dict, then you can't use the `ts_config` rule at the same time.

PR #2141, opened 2020-08-24
feat(typescript): deprecate ts_project#extends

alexeagle:
See #2140


alexeagle:
This shouldn't merge yet because we might want to use the `extends` attribute in a generated tsconfig

alexeagle:
will include the deprecation in #2130

PR #2142, opened 2020-08-25
chore: remove need for empty types[] workaround

alexeagle:


alexeagle:
sigh this is pretty big and hairy, not following up right now

Issue #1694, opened 2020-03-06
pkg_npm should support stamping with arbitrary stable/volatile keys 


joeljeske:
# 🚀 feature request

### Relevant Rules

`pkg_npm`

### Description

I want to be able to stamp multiple pieces of version information in npm archive. 

### Describe the solution you'd like

`pkg_npm` (and probably `pkg_web`) should allow values to use make-style variable replacements in the `substitutions` dictionary. The keys would still refer to strings in the source files to find and replace. 

```py
pkg_npm(
  substitutions = {
    # normal substitution
    "blaze": "bazel",
    # Make-style substitutions (from both volatile-keys and stable-keys)
    "GIT_HASH": "$(BUILD_SCM_HASH)",
    "HOST": "$(BUILD_HOST)",
    "GIT_BRANCH": "$(STABLE_SCM_BRANCH)", 
  },
  stamp = 1 # or 0,
)
```

### Describe alternatives you've considered

Writing a rule to manually copy the info from the version file into the relevant files.

alexeagle:
Yeah that makes a lot of sense. Do you have time to contribute it?

joeljeske:
Yea I can look into it hopefully. I would want to see your thoughts on #1693 and how it relates to this ticket. My needs would only be satisfied here if I could mark some targets as to never rely on stamping and some conditional on the stamp flag. 

 It seems like there might be place for a generic rule or task that makes the stamping easy for plain text interpolations as opposed to duplicating this functionality in multiple places.

alexeagle:
Also I'd be curious why it's good to support both stable keys and volatile keys. In particular since contributing npm publishing to bazelisk I was wondering why they use the stable keys file, and whether we should be recommending users to do the same. @philwo if you have a minute to comment?

philwo:
@meisterT told me he can help us out here. I'm not very knowledgeable about this particular feature of Bazel.

joeljeske:
I would seem that the use of stable keys vs. volatile keys is up to the needs of the user/organization. Some orgs may want to re-use as much as possible and only ever stamp an artifact the first time its contents are generated, and others may want to re-stamp it when certain criteria is met (branch change, user/host change, etc...) 

It would seem to me that the stamping feature of Bazel is representative of the needs of Google but may not be very flexible for a wider variety of needs, especially when it comes to [remote caching of stamped builds](https://github.com/bazelbuild/bazel/issues/10075)

thesayyn:
[...]

PR #2157, opened 2020-08-30
feat(builtin): accept any stamp vars in pkg_npm

alexeagle:
See #1694

### DIFF
--- a/docs/stamping.md
+++ b/docs/stamping.md
@@ -7,17 +7,49 @@ toc: true
 # Stamping
 
 Bazel is generally only a build tool, and is unaware of your version control system.
-However, when publishing releases, you typically want to embed version information in the resulting distribution.
-Bazel supports this natively, using the following approach:
+However, when publishing releases, you may want to embed version information in the resulting distribution.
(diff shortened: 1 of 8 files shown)


docs/stamping.md
```diff
@@ -7,38 +7,63 @@ toc: true
...
-# This tells Bazel how to interact with the version control system
-# Enable this with --config=release
-build:release --stamp --workspace_status_command=./tools/bazel_stamp_vars.sh
+Some rules accept an attribute that uses the status variables.
+For example, in a `pkg_npm` you can use the `substitutions` attribute like:
+
+```python
+pkg_npm(
+    name = "npm_package",
+    substitutions = {"0.0.0-PLACEHOLDER": "{STABLE_GIT_COMMIT}"},
```
mattem, line 28: do we need to make it clearer that if `--stamp` isn't set, the placeholder will be left alone, rather than set to empty string? Or what happens if the variable isn't defined?

alexeagle: yes, that's a subtle behavior change in this PR actually, we used to replace it with `0.0.0` if there was no stamp info. I'll update the doc and the commit description

PR #2164, opened 2020-09-01
fix(cypress): no longer force the use of browserify

mrmeku:


mattem (approved):

alexeagle (approved):

alexeagle:
@mrmeku could you rebase? probably will go green since windows fail looks like a flake

mrmeku:
@alexeagle Done!

alexeagle (approved):

PR #2174, opened 2020-09-07
chore: try github actions version of stale bot

alexeagle:
The probot one never triggered despite being setup correctly AFAICT.
This bot seems better supported and I sow it working on bazel-watcher repo.


mattem (approved):
+1 as it does pull requests too

PR #2175, opened 2020-09-07
feat: add link_workspace_root to nodejs_binary, npm_package_bin, rollup_bundle, terser_minified, ts_project

gregmagolan:
Link the workspace root to the bin_dir to support absolute requires like 'my_wksp/path/to/file'.
If source files need to be required then they can be copied to the bin_dir with copy_to_bin.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/internal/linker/link_node_modules.bzl
+++ b/internal/linker/link_node_modules.bzl
@@ -56,16 +56,18 @@ def _link_mapping(label, mappings, k, v):
     else:
         return True
 
-def write_node_modules_manifest(ctx, extra_data = [], mnemonic = None):
+def write_node_modules_manifest(ctx, extra_data = [], mnemonic = None, link_workspace_root = False):
     """Writes a manifest file read by the linker, containing info about resolving runtime dependencies
 
     Args:
         ctx: starlark rule execution context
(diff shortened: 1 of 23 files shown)


alexeagle (approved):
It sucks that we have to add this attribute to a handful of rules. Maybe other rules will need it too?
We could hang it on a repository rule and wire that up somehow, but it might be too magical.
Any other ideas?

gregmagolan:
> It sucks that we have to add this attribute to a handful of rules. Maybe other rules will need it too?
> We could hang it on a repository rule and wire that up somehow, but it might be too magical.
> Any other ideas?

Not sure how it could be done as a repository rule. I also like being able to turn it on for specific rules instead of globally (if at all) as it can have unexpected side-effets.

kellycampbell:
Would be good to mention this new flag on the 2.0 upgrade wiki section related to it: https://github.com/bazelbuild/rules_nodejs/wiki#no-longer-link-the-workspace-root

alexeagle:
Thanks @kellycampbell good point

PR #2176, opened 2020-09-08
fix(builtin): don't set --preserve-symlinks-main by default

gregmagolan:
This option should not be on by default as it creates issues when running programs out of `node_modules/.bin` which are all symlinks and for these programs to work it is convention that node be able to resolve their symlinks to their actual `node_module` location.

Also, with our node patches, node will not be able to resolve a main symlink out of the sandbox anymore which was the original reason that `--preserve-symlinks-main` was set.

If users want to use this option they can still pass it to their program with `--node_options=--preserve-symlinks-main`.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No (since we have node patches enabled now)

### DIFF
--- a/internal/node/node_repositories.bzl
+++ b/internal/node/node_repositories.bzl
@@ -19,7 +19,6 @@ See https://docs.bazel.build/versions/master/skylark/repository_rules.html
 """
 
 load("//internal/common:check_bazel_version.bzl", "check_bazel_version")
-load("//internal/common:check_version.bzl", "check_version")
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
Sounds right to me, if you can figure out the Windows

gregmagolan:
> Sounds right to me, if you can figure out the Windows

Windows failure was a flake. Also showed up in https://github.com/bazelbuild/rules_nodejs/pull/2179.

PR #2177, opened 2020-09-08
feat: experimental rule for esbuild and example es-dev-server

mattem:
Experimenting with [esbuild](https://github.com/evanw/esbuild) for bundling in place of rollup or webpack, and [es-dev-server](https://www.npmjs.com/package/es-dev-server) to serve es module format without a bundle or (an explicit) babel step.

To serve:
```
cd examples/es
bazel run //:server
```

Output the bundle
```
cd examples/es
bazel build //:esbuild
```

A lot of the rule here is a duplication of collecting all the outputs from `ts_library`, same as `rollup_bundle`, perhaps this needs to be made easier, then the need for a rule may be removed here.


alexeagle:
how about using `ts_project` instead?

alexeagle:
do you want to promote from draft? or can we close it to clean up the PR board?

mattem:
I'll close this in favour of making packages/esbuild

PR #2201, opened 2020-09-17
fix(builtin): js_library: correctly propagate DeclarationInfos

duarten:
Transitive typings are added to typings_depsets, but we were checking
the typings array instead.

Signed-off-by: Duarte Nunes <duarte@hey.com>

## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently, a `ts_project` cannot depend on a `js_library` with transitive declaration files (such as when the `js_library` depends on another `ts_project`. This PR fixes this.

Issue Number: N/A

## What is the new behavior?

Now the `js_library` propagates declarations downstream.

- [ ] Yes
- [X] No

### DIFF
--- a/internal/js_library/js_library.bzl
+++ b/internal/js_library/js_library.bzl
@@ -214,7 +214,7 @@ def _impl(ctx):
 
     # Don't provide DeclarationInfo if there are no typings to provide.
     # Improves error messaging downstream if DeclarationInfo is required.
-    if len(typings):
+    if len(typings) or len(typings_depsets) > 1:
         providers.append(declaration_info(
             declarations = depset(transitive = typings_depsets),
             deps = ctx.attr.deps,
--- a/internal/js_library/test/transitive/transitive_declarations_test.bzl
+++ b/internal/js_library/test/transitive/transitive_declarations_test.bzl
@@ -1,0 +1,20 @@
+"Unit tests for js_library rule"
+
+load("@bazel_skylib//lib:unittest.bzl", "asserts", "unittest")
+load("//:providers.bzl", "DeclarationInfo")
+
+def _impl(ctx):
+    env = unittest.begin(ctx)
+    decls = []
+    for decl in ctx.attr.lib[DeclarationInfo].declarations.to_list():
+        decls.append(decl.basename)
+    asserts.equals(env, ctx.attr.declarations, decls)
+    return unittest.end(env)
+
+transitive_declarations_test = unittest.make(_impl, attrs = {
+    "declarations": attr.string_list(default = ["a.d.ts"]),
+    "lib": attr.label(default = ":b"),
+})
+
+def transitive_declarations_test_suite():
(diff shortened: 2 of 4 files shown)


internal/js_library/js_library.bzl
```diff
@@ -214,7 +214,7 @@ def _impl(ctx):
 
     # Don't provide DeclarationInfo if there are no typings to provide.
     # Improves error messaging downstream if DeclarationInfo is required.
-    if len(typings):
+    if len(typings_depsets):
```
alexeagle, line 217: I don't think this can be right, because it looks like typing_depsets is always at least length one.
(That suggests that even if this PR were green, we are missing a test for "don't provide DeclarationInfo if there are no typings")

packages/typescript/test/ts_project/bridge_js_library/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "compilerOptions": {
+    "declaration": true,
+    "types": [],
+    // Help TypeScript locate the a.d.ts file from dep.
+    // Note that it comes from a js_library which means the .d.ts file is copied to the bazel-out folder.
+    // Needed when running in a sandbox or remote.
+    "rootDirs": [
+      ".",
+      "../../../../../bazel-out/darwin-fastbuild/bin/packages/typescript/test/ts_project/bridge_js_library",
```
alexeagle, line 10: do we really need to get into this composite ts_project behavior to test the one-line change? I imagine there's a much simpler test that just involves
- add one .d.ts file
- expose it with a js_library
- pass that through another js_library so it should come via transitive
- assert that you get the original .d.ts provided in the DeclarationInfo (using a tiny testing rule to extract them I think)

alexeagle:
Thanks for jumping in with a fix for this!

packages/typescript/test/ts_project/bridge_js_library/tsconfig.json
duarten: Sure, I'll try to get rid of the `ts_project`. Is there a place where testing rules usually live?

alexeagle: it can just live right next to the test files

duarten: Updated :)

internal/js_library/test/transitive/transitive_declarations_test.bzl
```diff
@@ -0,0 +1,20 @@
+"Unit tests for js_library rule"
+
+load("@bazel_skylib//lib:unittest.bzl", "asserts", "unittest")
```
alexeagle, line 3: wow super fancy 🧐 
nice

alexeagle (approved):

PR #2204, opened 2020-09-21
docs(labs): produce stardoc API for labs

### LINKED ISSUE #2195: Labs package is missing from docsite
I'd like to point a user to `ts_proto_library` docs but we don't generate them.

alexeagle:
Fixes #2195


alexeagle:
@mrmeku the resulting doc looks confusing because all the docs are on `protobufjs_ts_library` while `ts_proto_library` has only the bare attributes. Do you have any time to clean it up?

WORKSPACE
```diff
@@ -99,6 +99,17 @@ load("@build_bazel_rules_typescript//:package.bzl", "rules_typescript_dev_depend
...
+load("//packages/labs:package.bzl", "npm_bazel_labs_dependencies")
+
+npm_bazel_labs_dependencies()
+
+load("@rules_proto//proto:repositories.bzl", "rules_proto_dependencies", "rules_proto_toolchains")
+
+rules_proto_dependencies()
+
+rules_proto_toolchains()
+
```
mattem, line 112: Was this re-ordering due to the rules_proto dep?

mattem (approved):

WORKSPACE
alexeagle: yes, it needs to come earlier than rules_go to avoid getting their version in the maybe()

PR #2209, opened 2020-09-27
feat(examples): change angular example to ts_project

jbedard:
@alexeagle, not sure who else was interested

So this got more complicated and took a year... :|
* prerendering was dropped because I don't have time to figure it out
* esbuild is used to merge spec files before running karma on them. This replaces concatjs, we can't simply use the compiled-ts and dynamically load all the files at runtime because those files might have cjs/amd etc.
* random other minor changes that were for some reason exposed, see comments on the diff

~~Mainly just a ~~copy~~ refactor of the angular example switching to `ts_project` with `@angular/compiler-cli` as the tsc. However running into a few issues...~~

* ~~ngc is not working with a path error:
```Error: Internal Error: absoluteFrom(src/app/home): path is not absolute"```~~ https://github.com/angular/angular/pull/36291 fixes this
* ~~trying to get a package/module_name referencing a `ts_project` and imported into another
```in deps attribute of ts_project rule //src/app/todos:todos: '//src/shared/material:material' does not have mandatory providers: 'DeclarationInfo'```~~

### CODE REVIEW

File: examples/angular_ngc/tools/angular_ts_project.bzl [hunk 1]
```diff
@@ -0,0 +1,12 @@
+"Shows how to enable both worker mode and use_angular_plugin to make a drop-in replacement for ng_module"
+
+load("@npm//@bazel/typescript:index.bzl", "ts_project")
+
+def ng_ts_project(name, tsconfig = "//:tsconfig.json", srcs = [], angular_assets = [], **kwargs):
+    ts_project(
+        name = name,
+        tsconfig = tsconfig,
+        tsc = "@npm//@angular/compiler-cli/bin:ngc",
```
alexeagle, line 9: I think it should use ngtsc?

examples/angular_ngc/tools/angular_ts_project.bzl
```diff
@@ -0,0 +1,12 @@
+"Shows how to enable both worker mode and use_angular_plugin to make a drop-in replacement for ng_module"
+
+load("@npm//@bazel/typescript:index.bzl", "ts_project")
+
+def ng_ts_project(name, tsconfig = "//:tsconfig.json", srcs = [], angular_assets = [], **kwargs):
+    ts_project(
```
alexeagle, line 6: you probably need declaration = True so that we know to produce .d.ts outputs

alexeagle (approved):

alexeagle:
hey @jbedard any update on getting the fix from @Toxicable on the pathing issue in the angular compiler?

jbedard:
@alexeagle the patch from @Toxicable seems to work, I've added a patch for that to this PR. I'm still getting the `js_library` `DeclarationInfo` issue though which I don't understand... any ideas there?

alexeagle:
I'll fetch this and repro locally

alexeagle (commented):
Can you move the files so this is an edit to the existing examples/angular (so that when it lands, we use `ts_project` instead of `ts_library` for it)

otherwise this appears as 73 new files and I can't tell what you changed. I don't want to increase from 3 to 4 angular examples, it's too confusing for users and too much CI load to test them

alexeagle:
okay so for your DeclarationInfo problem, let's diagnose starting from the target that's supposed to produce .d.ts

```
% bazel build src/shared/material:material-ng --output_groups=types
INFO: Invocation ID: 45b78a82-ddd9-4e99-aef1-a0e58ac3fd3e
INFO: Analyzed target //src/shared/material:material-ng (0 packages loaded, 0 targets configured).
INFO: Found 1 target...
Target //src/shared/material:material-ng up-to-date:
  dist/bin/src/shared/material/material.module.d.ts
  dist/bin/src/shared/material/material.module.d.ts.map
```

cool, there is .d.ts coming from that ng_ts_project rule. Let's see if it gets propagated through the js_library rule

```
% bazel build src/shared/material:material --output_groups=types
INFO: Invocation ID: 804d5698-39c1-488b-b34b-ff1efe93426d
INFO: Analyzed target //src/shared/material:material (80 packages loaded, 13410 targets configured).
INFO: Found 1 target...
Target //src/shared/material:material up-to-date:
  dist/bin/src/shared/material/material.module.d.ts
  dist/bin/src/shared/material/material.module.d.ts.map
```

BUT I had to make two fixes to get this result.
1) Internal: js_library forgot to propagate the OutputGroupsInfo so you couldn't ask it for "types", will fix
2) js_library propagates typings from `deps`, not `srcs`. changing your js_library to

```
js_library(
    name = "material",
    package_name = "@bazel/shorten",
    deps = [":material-ng"],
)
```
fixes it.

packages/typescript/internal/ts_project.bzl
```diff
@@ -48,13 +48,19 @@ def _join(*elements):
 def _ts_project_impl(ctx):
     arguments = ctx.actions.args()
 
-    generated_srcs = False
+    has_generated = False
```
alexeagle, line 51: ooh, this looks like a nice bugfix, if helpful I can repro in a standalone test case for ts_project and land as a prefactoring

jbedard: There's 2 things here:
* if there are no source files after 1+ generated files in `ctx.files.srcs` then the `fail` would never be hit, this was a bug
* the `--rootDir` was only absolute if using generated sources. I honestly don't recall why I did this or if it correct, but I assume I did it try to to fix/workaround https://github.com/angular/angular/issues/36290 - but maybe that is only an angular bug and not ts_project bug?

If you'd like to take this commit and add tests then go for it :+1:

jbedard:
I'm currently getting a couple errors:

```
error TS6059: File '/private/var/tmp/_bazel_jason.bedard/02195f90740dce00c87d7ba79d4dda35/sandbox/darwin-sandbox/40/execroot/examples_angular/src/app/todos/todos.module.ts' is not under 'rootDir' 'bazel-out/darwin-fastbuild/bin/src/app/todos'. 'rootDir' is expected to contain all source files.
```

```
ERROR: /tmp/tmp-16352IrbYU6I3u1uq/src/app/todos/BUILD.bazel:11:1: in ts_project rule //src/app/todos:todos:
	File "/tmp/tmp-16352IrbYU6I3u1uq/src/app/todos/BUILD.bazel", line 11
		ts_project(name = 'todos')
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/7c6ac0f00953126366d099f5c23b87a0/external/npm/@bazel/typescript/internal/ts_project.bzl", line 61, in _ts_project_impl
		fail(<1 more arguments>)
srcs cannot be a mix of generated files and source files: bazel-out/k8-fastbuild/bin/src/app/todos/todos.component.css.map, src/app/todos/todos.component.html
```

packages/typescript/internal/ts_project.bzl
jbedard: I've updated that commit to output a better error messages showing the last files found in the generated vs src folders, for example the error this PR is currently getting which I commented about below:
```
srcs cannot be a mix of generated files and source files: bazel-out/k8-fastbuild/bin/src/app/todos/todos.component.css.map, src/app/todos/todos.component.html
```

phelanma:
Hi, little question do you have a plan for this PR in the short term? (week~2 weeks period)

Just a "need to know", if I should continue to use `ts_library` over `ts_project`, as this(cli issue + internal fixes) seems like a block on the feature?

jbedard:
This is blocked atm and needs work I think both in angular and rules_nodejs, so for now I think you'll need to stick with `ts_library`. I'd just recommend putting the angular config into a macro so you can easily change it in the future, see the macro in this PR for an example.

literalpie:
Has there been any progress on this? Are there specific issues for what needs to be changed in rules_nodejs that I can watch?

jbedard:
@literalpie I think this was blocked on https://github.com/angular/angular/issues/36290 but I should really rebase and give this another shot...

examples/angular/tools/karma.conf.js
```diff
@@ -0,0 +1,414 @@
...
+  } catch (e) {
+    console.error(`Failed to extract ${archiveFile}`);
+    throw e;
+  }
+}
+
+/**
+ * Check if Chrome sandboxing is supported on the current platform.
+ */
+function supportChromeSandboxing() {
```
jbedard, line 115: Should we just delete all the complications here to make the example simpler?

literalpie:
Are there any updates on this? Are there new issues blocking it?

alexeagle:
It's blocked on getting the Pre-rendering to work.

jbedard:
@literalpie I have got this working in my own project using essentially the same macros as in the PR. So for simpler cases (no pre-rendering) it shouldn't be hard to setup.

alexeagle:
@jbedard maybe we just have to remove the pre-rendering for now? I'm not sure which is worse, staying on the older rule or losing the feature.

examples/angular/src/BUILD.bazel
```diff
@@ -176,67 +159,27 @@ terser_minified(
     src = ":bundle-es5",
 )
 
-# take the skeleton index.html and use it to render the root, /hello and /todos routes
-# this generates multiple index.html files, one for each requested route
-ng_prerender(
```
jbedard, line 181: Prerendering has been dropped in hopes of getting this merged...

examples/angular/src/BUILD.bazel
```diff
@@ -268,8 +209,8 @@ pkg_web(
 ngsw_config(
     name = "pwa",
     src = ":prodapp",
+    out = "pwa-config",
```
jbedard, line 212: This just avoids a "rule and output have same name" (or something like that) warning

examples/angular/tools/ngsw_config.bzl
```diff
@@ -11,27 +11,25 @@ def ngsw_config(name, config, index_html, src, out = None, **kwargs):
 
     nodejs_binary(
         name = ngsw_config_name,
-        data = ["@npm//@angular/service-worker", index_html, config, src],
```
jbedard, line 14: I think including these here is one thing that was causing an issue, although I don't fully understand why... including the `src` here (output of `ts_project`) was somehow causing the typescript compiler to re-run but without the correct tsconfig that adds the `paths`. So typescript was unnecessary re-compiling with the wrong tsconfig which failed with path errors.

examples/angular/tools/ngsw_config.bzl
```diff
@@ -11,27 +11,25 @@ def ngsw_config(name, config, index_html, src, out = None, **kwargs):
...
+        data = ["@npm//@angular/service-worker"],
         visibility = ["//visibility:private"],
         entry_point = "@npm//:node_modules/@angular/service-worker/ngsw-config.js",
     )
 
     cmd = """
 mkdir -p $@
 cp -R $(locations {TMPL_src})/. $@/
-cp $(location {TMPL_index}) $@/index.html
```
jbedard, line 22: This index file is already in `TMPL_src` so this was causing a "cannot overwrite readonly file" error on ubuntu. I believe it's the exact same file so this was unnecessary (or maybe pre-rendering previously required it 🤷).

alexeagle (approved):
woohoo! it's green

PR #2221, opened 2020-10-08
ci: tell Angular CLI not to update webdriver

alexeagle:
It can get ahead of the Chrome version on our CI machine

### DIFF
--- a/examples/angular_bazel_architect/BUILD.bazel
+++ b/examples/angular_bazel_architect/BUILD.bazel
@@ -104,7 +104,7 @@ architect_test(
     name = "e2e",
     args = [
         "frontend:e2e",
-        "--webdriverUpdate=false",
+        "--no-webdriverUpdate",
     ],
     configuration_env_vars = ["NG_BUILD_CACHE"],
     data = glob([
--- a/examples/angular_bazel_architect/BUILD.bazel
+++ b/examples/angular_bazel_architect/BUILD.bazel
@@ -102,7 +102,10 @@ architect_test(
 
 architect_test(
     name = "e2e",
-    args = ["frontend:e2e"],
+    args = [
+        "frontend:e2e",
+        "--webdriverUpdate=false",
+    ],
     configuration_env_vars = ["NG_BUILD_CACHE"],
     data = glob([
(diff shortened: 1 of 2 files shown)


examples/angular_bazel_architect/package.json
```diff
@@ -8,7 +8,8 @@
     "test": "bazel test :test",
     "lint": "bazel run :lint",
     "e2e": "bazel test :e2e",
-    "postinstall": "patch-package && ngcc --properties es2015 browser module main --first-only"
+    "postinstall": "patch-package && ngcc --properties es2015 browser module main --first-only",
+    "webdriver-update": "webdriver-manager update --standalone false --gecko false --versions.chrome 85.0.4183.38"
```
alan-agius4, line 12: Shouldn't you call `webdriver-update` post-install?

alexeagle: thx, added `&& yarn webdriver-update` after patch-package

alan-agius4 (approved):

examples/angular_bazel_architect/BUILD.bazel
```diff
@@ -102,7 +102,10 @@ architect_test(
 
 architect_test(
     name = "e2e",
-    args = ["frontend:e2e"],
+    args = [
+        "frontend:e2e",
+        "--no-webdriver-update",
```
alan-agius4, line 107: The architect CLI, doesn't transform kebabed cases args to camel. But the schema validation needs camel cased properties.

```suggestion
        "--webdriverUpdate=false",
```

alan-agius4 (approved):
LGTM, after the arg change

examples/angular_bazel_architect/BUILD.bazel
alexeagle: hmm still fails
```
Error: Schema validation failed with the following errors:
  Data path ".webdriverUpdate" should be boolean.
```

alan-agius4: I looked at this and I pushed a commit (I hope you don't mind) which should address the problem.

PR #2225, opened 2020-10-14
docs: escape HTML entities

kyliau:
Stardoc extracts the documentation from .bzl files and generate them
as Markdown, but some contents are in HTML table. For such content,
we need to escape HTML entities otherwise they'll be rendered as
HTML node. This not only messes up the actual docs page, but could also
lead to potential security vulnerability.
For example, see the `format` attribute of `rollup_bundle` on
https://bazelbuild.github.io/rules_nodejs/Rollup.html (end of page).

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/docs/Built-ins.html
+++ b/docs/Built-ins.html
@@ -767,7 +767,7 @@ Some test runners like Karma and Jasmine have custom rules with added features, 
 <p>Bazel always runs tests with a working directory set to your workspace root.
 If your test needs to run in a different directory, you can write a <code>process.chdir</code> helper script
 and invoke it before the test with a <code>--require</code> argument, like
-<code>templated_args = ["--node_options=--require=./$(rootpath chdir.js)"]</code>.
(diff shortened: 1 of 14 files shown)


mattem (approved):
The BK failure seems like a flake

alexeagle (commented):
I like getting the docsite to look right, but I'm also worried that we don't really know how we are supposed to format docs. it would be nice to just say "use markdown" and make the tooling do the right thing. Maybe we should try dropping the HTML tables? wdyt @mattem ?

mattem:
Yeah I agree. The split between markdown and HTML is confusing. I'd say either drop the post-process script and use a dedicated markdown -> HTML converter library, or change the doc format to output markdown tables, see how Jekyll handles it.

alexeagle (commented):
Looks like these spots all render correctly at HEAD since we generate markdown without html tables.

PR #2228, opened 2020-10-15
Fix docs

alexeagle:
Alternative to https://github.com/bazelbuild/rules_nodejs/pull/2225

Go back to pure markdown rather than HTML tables which cause endless compat problems.
Can probably throw away our post-process-docs.js too

Preview: https://alexeagle.github.io/rules_nodejs/

### DIFF
--- a/docs/_includes/footer.html
+++ b/docs/_includes/footer.html
@@ -13,6 +13,7 @@
 <script src="https://cdnjs.cloudflare.com/ajax/libs/anchor-js/3.2.0/anchor.min.js" type="text/javascript"></script>
 <script>
   // Automatically add anchors and links to all header elements that don't already have them.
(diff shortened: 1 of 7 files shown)


mattem:
Looks awesome, from a quick review some things I noticed:

* Aspects and providers are still in the tables format
* The 'Attributes' heading is removed off rules, but the 'Parameters' heading is still present for macros
* When rules have a large description (eg https://alexeagle.github.io/rules_nodejs/Labs.html), the break between the description and the rule signature is small, and the sections kinda blend into each other

alexeagle:
Thanks @mattem addressed that feedback and updated the preview, PTAL

mattem (approved):

mattem:
lgtm 🚀

PR #2230, opened 2020-10-16
chore: update to bazel 3.6

alexeagle:


alexeagle:
yeah need to update rules_docker to get the removal of the gflags dependency
https://github.com/bazelbuild/rules_docker/compare/v0.14.3..v0.14.4

mattem (approved):

PR #2232, opened 2020-10-17
fix(builtin): js_library supports --output_groups=types

alexeagle:
If providing DeclarationInfo we should also give this way for filegroup/bazel CLI to select them


alexeagle:
I should update docs too...

jbedard (approved):

PR #2233, opened 2020-10-17
chore: update rules_docker to 0.14.4

mattem:


alexeagle:
rules_docker pushes subtle breaking changes to their WORKSPACE setup in patch releases
https://github.com/bazelbuild/rules_nodejs/pull/2230/files#diff-eef72c516320aac9abc26ebacbf88d2b869bdea745e4a2c4288540fde58fb8e6 is one change I had to make
but it trips over some rules_go dependency spaghetti :(

mattem:
I just noticed that :( Adding a newer bazel_skylib to the top of the WORKSPACE allows it to run, is that coming from rules_go?

alexeagle (approved):

alexeagle:
Bazel could use a package manager, eh?
Or packages could stop having so many dependencies...

PR #2238, opened 2020-10-18
feat(builtin): flip the default of the strict_visibility flag on the npm and yarn install rules to True

### LINKED ISSUE #2199: Flip strict_visibility flag on npm / yarn install rules
#2193 introduced the `strict_visibility` flag on `npm_install` and `yarn_install` which currently defaults `False`. As part of 3.0.0, default this flag `True`

mattem:
This is a potential breaking change for uses in 3.0.0

closes #2199


alexeagle:
nice, I guess this is failing as expected
` ts_project rule //packages/typescript/test/ts_project/a:tsconfig: target '@npm//@babel/parser:parser' is not visible from target '//packages/typescript/test/ts_project/a:tsconfig'. `

mattem:
Yes, these seem like legitimate missing dependencies. Working though them

jbedard:
I was getting an error I think with karma and `tmp` when I tried enabling this. Not sure if you're seeing that I've? I'll try it again later and confirm if that was the one.

mattem:
Hmm, I haven't see that. I did see an issue here of having a dependency on a `file:./` reference in the `package.json` file, which I've fixed in this PR.

jbedard:
```
_karma_web_test rule //abc:test_wrapped_test: target '@npm//tmp:tmp' is not visible from target '//abc:test_wrapped_test'. Check the visibility declaration of the former target if you think the dependency is legitimate
```
https://github.com/bazelbuild/rules_nodejs/blob/2.2.2/packages/karma/karma_web_test.bzl#L457-L469

Although that looks like it's actually my fault and this rule caught it 😄 
https://bazelbuild.github.io/rules_nodejs/Karma.html#karma_web_test-peer_deps

jbedard:
A bit odd that `@bazel/karma` has a `tmp` peer dep though, is it not? Especially since it's also in the `karma/package.json` as a `dependency`... or is there a reason I'm not thinking of?

(sorry if this is getting off topic for this PR)

mattem:
I think the reason it's a peer dep is because it's listed here:
https://github.com/bazelbuild/rules_nodejs/blob/2.2.2/packages/karma/karma_web_test.bzl#L21
Although, I'm not sure what it's used for, and this also assumes the user has an `@npm` repository

mattem:
Those integration tests could use the --keep_going flag!

mattem:
🟢 finally.

This change does mean that users will have to add the deps used by karma_web_test to their own package.json file if they haven't already, these are added by the karma_web_test macro, so it's a little abstracted from the BUILD files a user would see.

jbedard:
> This change does mean that users will have to add the deps used by karma_web_test to their own package.json file if they haven't already, these are added by the karma_web_test macro, so it's a little abstracted from the BUILD files a user would see.

That was the issue I came across. I had all of them except `tmp` which seemed odd to me. All the karma/jasmine `peerDependencies` sort of make sense, to allow/force the user to define the version they want. But I think the `tmp` dep is entirely internal to `@bazel/karma`? I also found that if you don't use the exact version of `tmp` which `@bazel/karam` depends on (v0.1.0) then it was crashing for me. It seems like that one should be removed?

mattem:
`tmp` is used in the karma concatjs impl to create a unique temp file for the bundled files, but yes it's completely internal to `@bazel/karma`, not listed as a peerDep, so yeah it's kinda non-intuitive to have that dependency. Perhaps we can just remove it and generate the temp files in the karma package.

alexeagle:
I added a commit removing the "tmp" dependency. I don't think it's essential that we cleanup these files ourselves since they go in bazel's TEST_TMPDIR.

alexeagle:
@googlebot I consent.

alexeagle (approved):

jbedard:
Can the "tmp" dependency change also go into 2.x?

PR #2249, opened 2020-10-22
chore: replace @bazel_tools//platforms with @platforms//os

alexeagle:
enable incompatible_use_platforms_repo_for_constraints for examples

See https://github.com/bazelbuild/bazel/issues/8622
This is an expected breaking change in Bazel 4.0

It seems like we need to update rules_go in order to turn on the flag for our own build, still TODO


alexeagle:
replaced with https://github.com/bazelbuild/rules_nodejs/pull/2536

PR #2252, opened 2020-10-25
feat: add esbuild package

mattem:
Adds the `@bazel/esbuild` package, a thin, mildly opinionated wrapper around esbuild. The rule currently just calls the esbuild binary on the command line, further iterations of this could change to call the Go API, add workers and additional features.

Example added to `examples/esbuild`, along with unit tests under the package dir.

In a previous iteration this PR included a repo rule to fetch the esbuild tool, however this has now been removed and user is expected to provide the binary to the rule via the `tool` attribute. An example macro is included in the tests dir, and the examples show supporting multi-platforms.


lukasholzer:
@googlebot I consent.

mattem:
Seems there is a difference in path resolution behaviour between MacOS and Ubuntu in esbuild perhaps, will need to dig more into that. It passes on BuildKite though

mattem:
@lukasholzer I'm stuck here on the Windows build. Do you have access to a Windows machine to debug? If not I'll try and source something this week.

lukasholzer:
@mattem yea have a windows server via remote desktop :D – will give it a try the next days and try to fix it :)

mattem:
Woo, windows is 🍏 Needs examples and docs 📖

examples/esbuild/src/BUILD.bazel
```diff
@@ -0,0 +1,53 @@
...
+
+# create a single bundle of the JS sources
+esbuild(
+    name = "bundle",
+    entry_point = "main.ts",
+    minify = True,
+    # setting node as the platform will default us to CJS
+    platform = "node",
+    tool = select({
+        "@bazel_tools//src/conditions:darwin": "@esbuild_darwin//:bin/esbuild",
```
alexeagle, line 23: this is fine to be consistent with other places but I think we are supposed to use `@platforms` in the future https://github.com/bazelbuild/bazel/issues/8622

packages/esbuild/helpers.bzl
```diff
@@ -0,0 +1,108 @@
+"""
+Utility helper functions for the esbuild rule
+"""
+
+load("@bazel_skylib//lib:paths.bzl", "paths")
```
alexeagle, line 5: This is the first time we make users depend on skylib, but it doesn't seem necessary here. We aren't using a provider symbol that has to be shared, just need a bit of logic?
I think we should vendor the one function we need into our third_party

packages/esbuild/esbuild.bzl
```diff
@@ -0,0 +1,220 @@
...
+            doc = "An executable for the esbuild binary",
+        ),
+    },
+    implementation = _esbuild_impl,
+)
+
+def esbuild(name, splitting = False, **kwargs):
+    """esbuild helper macro around the `esbuild_bundle` rule
+
+    For a full list of attributes, see the `esbuild_bundle` rule
```
alexeagle, line 200: instead of the docs telling you to go find other docs, I think we should do our trick where there is only one name ("esbuild_bundle" I guess) which we document. Then this macro is "esbuild_bundle_macro" but in the public API we re-export it as "esbuild_bundle" again

packages/esbuild/esbuild.bzl
```diff
@@ -0,0 +1,220 @@
...
+
+    Args:
+        name: The name used for this rule and output files
+        splitting: If `True`, produce a code split bundle in an output directory
+        **kwargs: All other args from `esbuild_bundle`
+    """
+
+    if splitting == True:
```
alexeagle, line 208: maybe we should make this more similar to rollup_bundle and other rules with output_dir=True ?

packages/esbuild/esbuild.bzl
```diff
@@ -0,0 +1,220 @@
...
+            "no-remote-exec": "1",
+        },
+    )
+
+    return [
+        DefaultInfo(files = depset(outputs + [jsconfig_file])),
+    ]
+
+esbuild_bundle = rule(
+    attrs = {
```
alexeagle, line 107: there's no doc argument to the rule() function

packages/esbuild/esbuild.bzl
```diff
@@ -0,0 +1,220 @@
...
+            doc = "The bundle's entry point (e.g. your main.js or app.js or index.js)",
+        ),
+        "external": attr.string_list(
+            default = [],
+            doc = "A list of module names that are treated as external and not included in the resulting bundle",
+        ),
+        "format": attr.string(
+            values = ["iife", "cjs", "esm", ""],
+            mandatory = False,
+            doc = """The output format of the bundle, defaults to iife when platform is browser
```
alexeagle, line 139: could we add links to the esbuild documentation for each option that we pass through?

packages/esbuild/esbuild.bzl
```diff
@@ -0,0 +1,220 @@
...
+        ),
+        "srcs": attr.label_list(
+            allow_files = True,
+            default = [],
+            doc = """Non-entry point JavaScript source files from the workspace.
+
+You must not repeat file(s) passed to entry_point""",
+        ),
+        "target": attr.string(
+            default = "es2015",
```
alexeagle, line 181: feels like we're missing an `args` to let you pass arbitrary other arguments to esbuild

and once we have that, we should defend why Bazel needs to know about all of them. I think we only need to know the ones that are needed to predict output paths?

examples/esbuild/src/BUILD.bazel
mattem: Made https://github.com/bazelbuild/rules_nodejs/issues/2440 to track these

internal/linker/link_node_modules.bzl
```diff
@@ -19,7 +19,7 @@ def _debug(vars, *args):
         print("[link_node_modules.bzl]", *args)
 
 # Arbitrary name; must be chosen to globally avoid conflicts with any other aspect
-_ASPECT_RESULT_NAME = "link_node_modules__aspect_result"
+ASPECT_RESULT_NAME = "link_node_modules__aspect_result"
```
gregmagolan, line 22: If we're depending on this outside of this file, lets change the name to `MODULE_MAPPINGS_ASPECT_RESULTS_NAME` perhaps to match `module_mappings_aspect`.

Also, lets ensure that the module mappings we're depending on will be forward compatible to 4.0 since I'll be removing the map to tuple (type, path) and make it just just a map to path

mattem:
rebased on stable

packages/esbuild/esbuild_repo.bzl
```diff
@@ -0,0 +1,48 @@
+"""
+Helper macro for fetching esbuild versions for internal tests and examples in rules_nodejs
+"""
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+_PLATFORM_SHA = {
```
gregmagolan, line 7: looks like this is unsused

examples/esbuild/WORKSPACE
```diff
@@ -0,0 +1,66 @@
...
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "build_bazel_rules_nodejs",
+    sha256 = "6142e9586162b179fdd570a55e50d1332e7d9c030efd853453438d607569721d",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/3.0.0/rules_nodejs-3.0.0.tar.gz"],
+)
+
+_ESBUILD_VERSION = "0.8.34"
```
gregmagolan, line 28: use helper `esbuild_dependencies()` function here?

packages/esbuild/esbuild_repo.bzl
```diff
@@ -0,0 +1,48 @@
...
+    "darwin_64": "3bf980b5175df873dd84fd614d57722f3b1b9c7e74929504e26192d23075d5c3",
+    "linux_64": "9dff3f5b06fd964a1cbb6aa9ea5ebf797767f1bd2bac71e084fb0bbefeba24a3",
+    "windows_64": "826cd58553e7b6910dd22aba001cd72af34e05c9c3e9af567b5b2a6b1c9f3941",
+}
+
+_VERSION = "0.8.34"
+
+def esbuild_dependencies():
```
gregmagolan, line 9: Could pass in the version & a dict of platform => SHA as a param (with default values for v0.8.34) so make it easier to use. Then docs could be updated to suggest using this helper.

examples/esbuild/WORKSPACE
mattem: We can't here, as that would cause the eager fetch of the npm workspace to use in the example WORKSPACE

packages/esbuild/esbuild_repo.bzl
mattem: Nice catch, removed

internal/linker/link_node_modules.bzl
mattem: Renamed to `MODULE_MAPPINGS_ASPECT_RESULTS_NAME`

packages/esbuild/helpers.bzl
```diff
@@ -0,0 +1,127 @@
+"""
+Utility helper functions for the esbuild rule
+"""
+
+# From https://github.com/bazelbuild/bazel-skylib/blob/master/lib/paths.bzl
+def _dirname(p):
```
gregmagolan, line 6: Looks like we already vendor this function in third_party `third_party/github.com/bazelbuild/bazel-skylib/lib/paths.bzl`. You can load it from 
```
load("//third_party/github.com/bazelbuild/bazel-skylib:lib/paths.bzl", "paths")
```
and use it with
```
paths.dirname
```
`internal/node/node_repositories.bzl` has an example of this

examples/esbuild/WORKSPACE
gregmagolan: Right. The ol' eager fetching.

packages/esbuild/esbuild_repo.bzl
gregmagolan: From your answer above, this would introduce eager loading for users which is not great

gregmagolan (approved):

gregmagolan:
🌮  very nice

mattem:
🚀

PR #2257, opened 2020-10-27
fix: npm_package.pack should work in windows os

JiaLiPassion:
`npm_package.pack` rule not work in Windows OS since it will generate a `run_npm.sh` which is not work in Windows, so this PR will generate a `bat` file to make it workable.


JiaLiPassion:
@googlebot I consent

JiaLiPassion:
I am not sure why `buildkite` CI is red....

alexeagle (approved):
thanks!

Issue #2256, opened 2020-10-27
Upgrade our examples to npm7


alexeagle:
Mostly to make sure it works with rules_nodejs

alexeagle:
Also turn on yarn frozen_lockfile at the same time (by backing out this change https://github.com/bazelbuild/rules_nodejs/commit/b6a8cbb78473df25756c6021b71a291fbeb5a8b5#diff-da7e70fd50ce8f04f6ac7d53c8a254acfabac83a73dc10a6c16dfe81875b563f)

alexeagle:
Need https://github.com/bazelbuild/rules_nodejs/issues/2339 first

PR #2268, opened 2020-11-04
feat: add Node v15.0.0

mattem:
Updates the Vue and vendored_node to Node v15.0.1 to use npm 7

reference: https://github.com/bazelbuild/rules_nodejs/issues/2256


mattem:
It seems that npm7 doesn't create the transitive `node_modules` folders when a dependency is referenced via `file:`. In the vendored_node examples test (and I guess the others), `@bazel/jasmine` is referenced this way, so the BUILD file generation script doesn't find all the dependencies needed on disk. In this case, we miss `jasmine-reporters` which is a transitive dep via `@bazel/jasmine`

mattem:
I filed an issue on the npm repo https://github.com/npm/cli/issues/2145

mattem:
This works correctly if the parent `package.json` file references a `tgz` file for the dependency. We could change the bazel integration tests to run the `.pack` target before running the test targets, or modify the `pkg_npm` rule to have the `.pack` tar as a build output (currently only available via `bazel run //foo.pack`

mattem:
Can make the required changes to the bazel integration tests after #2339 lands

mattem:
closed for https://github.com/bazelbuild/rules_nodejs/pull/2366

PR #2274, opened 2020-11-09
Fix select with condition on darwin.

comius:
Bazel is changing how @bazel_tools/conditions:darwin is implemented.
With old implementation based on flags, it was ok to have darwin and
darwin_x86_64 in a select. This was based on flag --cpu=darwin and
--cpu=dawin_x86_64.

New implementation is based on constraints, where "darwin" means OS (and
any CPU), while "darwin_x86_64" mean only specific CPU. As such they
cannot be used in the same select, because the selection would be
ambiguous.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/toolchains/node/BUILD.bazel
+++ b/toolchains/node/BUILD.bazel
@@ -78,7 +78,6 @@ alias(
     name = "toolchain",
     actual = select({
         "@bazel_tools//src/conditions:darwin": "@nodejs_darwin_amd64_config//:toolchain",
-        "@bazel_tools//src/conditions:darwin_x86_64": "@nodejs_darwin_amd64_config//:toolchain",
         "@bazel_tools//src/conditions:linux_aarch64": "@nodejs_linux_arm64_config//:toolchain",
         "@bazel_tools//src/conditions:linux_s390x": "@nodejs_linux_s390x_config//:toolchain",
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
thanks!

Wyverald:
rules_nodejs still seems to be failing: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1761#3e9ef6b7-1c7c-490e-b51d-85777d2f8c2f @comius

comius:
> rules_nodejs still seems to be failing: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1761#3e9ef6b7-1c7c-490e-b51d-85777d2f8c2f @comius

The buildkite is using `master` branch, whereas the patch has been merged into `stable`.
Since `stable` is more recent and designated as default, I guess we should be using it?

@alexeagle could you confirm?

alexeagle:
That's right, we no longer use master. I would delete it but not all incoming refs get redirected.

comius:
> That's right, we no longer use master. I would delete it but not all incoming refs get redirected.

I proposed a change in CI to use stable branch - https://github.com/bazelbuild/continuous-integration/pull/1064

PR #2278, opened 2020-11-10
extract function: ts_project rootDir calculation

alexeagle:


packages/typescript/internal/ts_project.bzl
```diff
@@ -76,6 +70,32 @@ def _join(*elements):
         return "/".join(segments)
     return "."
 
+def _calculate_root_dir(ctx):
+    some_generated_path = None
+    some_source_path = None
+    root_path = None
+    allow_js = True
```
jbedard, line 77: This is `True` 100% of the time? 🤷

jbedard (approved):

jbedard:
@googlebot I consent.

packages/typescript/internal/ts_project.bzl
alexeagle: this rule doesn't have the allow_js from the user, because it's consumed by the macro, so we can't tell here if the user expected TypeScript to emit from .js inputs. However your logic in this location always considered .js and .jsx to be subject to the single-root-dir constraint. I think it's fine since our intent is to allow generated .css/.html and not generated .js

jbedard: More just debating if it's worth declaring a var vs just passing `True` directly to the method it's passed to. Doesn't really matter, but putting in a (non-final/const) variable just makes me assume it might change 🤷

alexeagle: ah, I see. It's for documentation - yeah I would have rather done `(/*allow_js*/True)` if starlark had inline comment syntax.

PR #2178, opened 2020-09-08
fix(examples): fix jest example on windows

lukasholzer:
Fixes #1454

image

## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?

- [ ] Yes
- [ ] No

PR #2279, opened 2020-11-10
chore: only create runfiles trees when needed

alexeagle:
should improve performance

this is split out from https://github.com/bazelbuild/rules_nodejs/pull/2178 since it seems unrelated and probably causing a test to fail


lukasholzer:
@googlebot I consent.

PR #2291, opened 2020-11-17
perf(cypress): pack cypress runfiles into a single tar

mrmeku:
Currently the external cypress workspace contains a file groups with tens of thousands of file attached. 
This change changes that file group to have a single tar archive instead to drastically increase the speed the runfiles tree for a cypress test can be created.


examples/cypress/package.json
```diff
@@ -11,9 +11,10 @@
         "@bazel/typescript": "^2.3.0",
         "@types/node": "14.0.13",
         "cypress": "^4.8.0",
-        "typescript": "3.9.5"
+        "typescript": "3.9.5",
+        "tar": "6.0.5"
```
alexeagle, line 15: the user shouldn't have to depend on `tar` right? if the cypress code requires it, should come from its own dependency

packages/cypress/internal/install-cypress.js
```diff
@@ -10,30 +10,36 @@
...
 const nodePath = process.argv[1];
 const cwd = process.cwd();
 const cypressBin = process.argv[2];
 
-// Sandboxing doesn't work on windows, so we can just use the global cypress cache.
-if (process.platform === 'win32') {
-  installGlobalCypressCache();
-  process.exit(0);
-}
+const tarPath = join(cypressBin.split('node_modules')[0], 'node_modules', 'tar', 'index.js')
```
alexeagle, line 25: are you mutating the node_modules tree? seems like it would be better to put that somewhere else

packages/cypress/test/package.json
```diff
@@ -5,6 +5,7 @@
         "express": "4.17.1",
         "cypress": "^5.2.0",
         "@types/node": "14.0.14",
-        "typescript": "3.9.5"
+        "typescript": "3.9.5",
+        "tar": "6.0.5"
```
alexeagle, line 9: would be nice to avoid an extra dependency here too

examples/cypress/package.json
mrmeku: Done

packages/cypress/internal/install-cypress.js
mrmeku: Resolved

packages/cypress/test/package.json
mrmeku: This dependency is necessary for `//packages/cypress/test` due to a convoluted test setup procedure (since `@bazel/cypress` isn't actually installed we still need its dep on tar to be listed somewhere)

alexeagle: yup that makes sense

alexeagle (approved):

Issue #2290, opened 2020-11-16
Proposal: ambient_srcs and declared_modules


DavidANeil:
## Current

Currently `ts_library` generates a tsconfig file that contains all of the transitive inputs in the `"files"` attribute.  When tsc encounters this attribute it is forced to load, parse, and bind the source file.

## Proposal

The core of this proposal is to simply not do that.  By only including the current project's sources in `"files"` then tsc will automatically resolve imports to discover new files, then only those used files will be loaded from disk.

Most projects would not compile under that scheme.  To create a workaround for this breaking change, two new attributes will be added to `ts_library` and `js_library`: `ambient_srcs` and `declared_modules`

`ambient_srcs` is a list of sources that contain ambient type declarations, which means they neither import nor export any types.
These are things such as `@types/node/globals.d.ts`, they affect the global scope simply by being loaded.
All `ambient_srcs` of deps (TODO: direct or transitive?) will always be included in the `"files"` attribute, so that they can affect the global scope.

`declared_modules` is a `string_list_dict` that contains a mapping from `"module-name"` to any source files that declare that module.  Each entry in `declared_modules` will be added to `"paths"` attribute of tsconfig.  This allows source declaration files that only declare modules to be loaded lazily as a regular entry in `srcs` rather than eagerly as `ambient_srcs`.
Declared modules are a type of ambient declaration that do not affect the global type scope, only the import scope.
That is, if `@types/node/fs.d.ts` has been loaded with its
```typescript
declare module "fs"
```
then tsc will know to resolve an import from `"fs"` to the type information in that declared module.
Instead of relying on the file already being loaded, `declared_modules` allows us to put an entry in the `"paths"` attribute of the tsconfig to teach the compiler how to resolve the import.

## Unblocks

Since this saves tsc from needing to open every source fill of every transitive dependency this unblocks the usage of `unused_inputs_list` which can drastically reduce the action graph in many real-life use cases.

## Results

In practice the above changes have reduced build times for one of our applications from ~2200s to ~1000s.
Adding `unused_inputs_list` further decreases that down to 7s when making changes to a file that is not depended on.

gregmagolan:
No longer in scope for rules_nodejs which only supplies the Node.js toolchain as of [v6.0.0](https://github.com/bazelbuild/rules_nodejs/releases/tag/v6.0.0-beta0).

Downstream canonical JavaScript + Node.js ruleset is now https://github.com/aspect-build/rules_js.

PR #2293, opened 2020-11-18
Ambient declarations, declared modules, and unused_inputs_list

DavidANeil:
PR Checklist

## PR Type

The declaration files for all transitive dependencies are included in the `"files"` attribute in the generated tsconfig.

## What is the new behavior?
Only the source files of the target being built plus the ambient declaration files of all transitive dependencies are included in the `"files"` attribute.
The `"paths"` attribute includes any declared modules.
Closes: #2290 

Also uses unused_inputs_list to reduce frequency of rebuilds.
Closes: #968

- [X] Yes

Many applications will not need to take any direct action.
Users may either want to use `ambient_srcs` and `declared_modules` to declare those types of inputs, when relevant.
Or they may specify `use_precise_ts_files = False` on their `ts_library` to entirely disable the new functionality.


third_party/github.com/bazelbuild/rules_typescript/internal/tsc_wrapped/cache.ts
```diff
@@ -72,7 +72,7 @@ class Cache<T> {
     // Drop half the cache, the least recently used entry == the first entry.
     this.debug('Evicting from the', this.name, 'cache...');
     const originalSize = this.map.size;
-    let numberKeysToDrop = originalSize / 2;
+    let numberKeysToDrop = Math.ceil(originalSize / 2);
```
mattem, line 75: 👍

examples/react_webpack/package.json
```diff
@@ -10,6 +10,7 @@
         "css-loader": "^3.2.0",
         "csstype": "^2.2.0",
         "http-server": "^0.11.1",
+        "protobufjs": "^6.8.8",
```
alexeagle, line 13: this seems like a bug. the react_webpack example doesn't depend on protobufjs so it should come transitively from `@bazel/typescript`  or `@bazel/worker` I suspect?

internal/js_library/js_library.bzl
```diff
@@ -32,10 +32,12 @@ load(
 )
 
 _ATTRS = {
+    "ambient_srcs": attr.label_list(allow_files = True),
```
alexeagle, line 35: your feature is for `ts_library` but not needed for `ts_project`. That makes it feel undesirable to me for the layering to require that js_library know about this stuff.
Imagine I'm just following the TypeScript documentation. This wouldn't make any sense to me.

What would this PR look like if you didn't change `js_library`? Maybe it shakes out one of these ways:
1. there's some example you can't do like provide a `.js`/`.d.ts` pair exporting an ambient, but we could add that in a followup (just to defer this discussion and make it easier to land a first PR). But since `ts_library` already has a `module_name` field I don't see why `js_library` would be needed in such a case
2. these two new field could just be tucked into DeclarationInfo (but I guess the srcs one couldn't)
3. users have to put a `ts_library` as the srcs of a `js_library` to make this work

internal/npm_install/generate_build_file.ts
```diff
@@ -60,6 +60,11 @@ const BAZEL_VERSION = args[6];
 const PUBLIC_VISIBILITY = '//visibility:public';
 const LIMITED_VISIBILITY = `@${WORKSPACE}//:__subpackages__`;
 
+const isModuleRegExp = /^export|^import(?!declare globa)/m;
```
alexeagle, line 63: intentional misspelling `globa` instead of `global`?

internal/npm_install/generate_build_file.ts
```diff
@@ -60,6 +60,11 @@ const BAZEL_VERSION = args[6];
...
+const declaresGlobalRegExp = /^declare global/m;
+const isNotOnlyDeclaredModuleRegExp = /^\w(?!eclare module)/m;
```
alexeagle, line 65: intentional missing `d`?

internal/npm_install/generate_build_file.ts
alexeagle, line 63: could these be declared closer to the usage site?

internal/npm_install/generate_build_file.ts
```diff
@@ -866,7 +874,8 @@ function filterFiles(files: string[], exts: string[] = []) {
  * false otherwise. If the package contains `*.metadata.json` and a
  * corresponding sibling `.d.ts` file, then the package is considered to be APF.
  */
-function isNgApfPackage(pkg: Dep) {
+function
+isNgApfPackage(pkg: Dep) {
```
alexeagle, line 878: why all these added linebreaks?

internal/npm_install/generate_build_file.ts
```diff
@@ -965,7 +1047,19 @@ ${printJson(pkg)}
 # Files that are part of the npm package not including its nested node_modules
 # (filtered by the 'included_files' attribute)
 filegroup(
-    name = "${pkg._name}__files",${pkgFilesStarlark}
+    name = "${pkg._name}__nonambient_files",${nonAmbientFilesStarlark}
+)
+
+# Files that have side-effects and must always be loaded
+filegroup(
+    name = "${pkg._name}__ambient_files",${pkgAmbientFilesStarlark}
```
alexeagle, line 1055: nit: are they "scripts" and not "modules" in TS-lingo? maybe `ambient_scripts`?

internal/npm_install/test/golden/unidiff/BUILD.bazel.golden
```diff
@@ -40,7 +47,8 @@ js_library(
 )
 js_library(
     name = "unidiff__contents",
-    srcs = [":unidiff__files", ":unidiff__nested_node_modules"],
+    srcs = [":unidiff__nonambient_files", ":unidiff__nested_node_modules"],
+    ambient_srcs = [":unidiff__ambient_files"],
```
alexeagle, line 51: this partly answers my question about why `js_library` has to change, but I wonder if these attributes are only used privately or if it's something users will need to use in some case

internal/providers/declaration_info.bzl
```diff
@@ -26,14 +26,16 @@ Note: historically this was a subset of the string-typed "typescript" provider.
     # TODO(alexeagle): The ts_library#deps attribute should require that this provider is attached.
     # TODO: if we ever enable --declarationMap we will have .d.ts.map files too
     fields = {
+        "ambient_declarations": "A depset of typings files for ambient srcs",
```
alexeagle, line 29: I feel like throughout, we need a bit more explanation of what is an "ambient src".
Is this the same term of art used in someplace else so at least it's searchable?

packages/typescript/internal/ts_project.bzl
```diff
@@ -246,7 +246,7 @@ def _ts_project_impl(ctx):
     # Don't provide DeclarationInfo if there are no typings to provide.
     # Improves error messaging if a ts_project needs declaration = True
     if len(typings_outputs) or len(ctx.attr.deps):
-        providers.append(declaration_info(depset(typings_outputs), ctx.attr.deps))
+        providers.append(declaration_info(depset(typings_outputs), depset([]), ctx.attr.deps))
```
alexeagle, line 249: I wonder if the function signature could be changed so `depset([])` is the default for this new param and existing callers don't have to change

packages/typescript/internal/build_defs.bzl
```diff
@@ -155,12 +155,17 @@ def _compile_action(ctx, inputs, outputs, tsconfig_file, node_opts, description
...
+        arguments.append("@@" + unused_inputs_list.path)
+        action_outputs.append(unused_inputs_list)
+
         # Spawn a plain action that runs worker process with no linker
         ctx.actions.run(
             progress_message = "Compiling TypeScript (%s) %s" % (description, ctx.label),
             mnemonic = "TypeScriptCompile",
             inputs = action_inputs,
             outputs = action_outputs,
+            unused_inputs_list = unused_inputs_list,
```
alexeagle, line 168: do we also need to plumb the unused inputs list in the non-worker case below?

alexeagle (commented):
did a first pass mostly over the public API to start

internal/js_library/js_library.bzl
alexeagle: after reading the rest, I think maybe it's 4) we need this just for third-party libraries but there isn't a usecase for first-party so we should document these new attrs as internal use only, not meant as public API

examples/react_webpack/package.json
DavidANeil: I never could figure this one out, it seemed like this should have come from a transitive dep, but yarn wasn't installing it, maybe because yarn doesn't understand deps on local packages properly?
I'll try removing it again to see if there is a different fix.

internal/js_library/js_library.bzl
DavidANeil: If someone is trying to properly describe a `.d.ts` file using `js_library` they would need to use these attributes.
I think it is more common for people to write code as modules, but I don't think this is quite internal only.

internal/npm_install/generate_build_file.ts
DavidANeil: Unintentional

internal/npm_install/generate_build_file.ts
DavidANeil: Intentional.  It is consumed by the `\w`, that is this matches lines that start with a word-character but aren't immediately followed by `eclare module`

packages/typescript/internal/build_defs.bzl
DavidANeil: It needs the list of all inputs in order to derive which are unused.  Workers have that by default so I only enabled it for them.
The non-worker code behaves sanely, but just can't take advantage of it.
We could pass in the names of all of the inputs through the tsconfig, but that can be added later, so I decided to not go gallivanting around for it.

examples/react_webpack/package.json
alexeagle: okay waiting on you for next review pass then, ping me when ready

purkhusid:
Why was this closed? The performance improvements were impressive.

DavidANeil:
It will be re-opened against the stable branch now that 3.0 is merged into stable.

Closed for now because I have been busier than normal the last month.  Should be re-opened before the new year.

PR #2300, opened 2020-11-23
Add a browsers package that vendors browser binaries through bazel

lukasholzer:
As discussed last time in the contributor meeting here is the first PR for the browser binary vendor. This PR should be a prerequisite for adding a [TestCafe](https://devexpress.github.io/testcafe/) web testing rule.

This should be used for web testing if you don't rely on web driver or selenium configuration and you only need a browser binary

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

Added a repository rule to vendor browser binaries through bazel.

- [ ] Yes
- [x] No


alexeagle (commented):
@gregmagolan and I discussed it - since there is no .js code in here and you never call `require` it's not necessary to ship an npm package `@bazel/browsers`

Also this is a feature users may want from lots of places, and it doesn't have a large footprint either in terms of how many bytes we ship, nor in performance of using rules_nodejs

So this feature should move into the builtin (/internal) rather than be a separate package

gregmagolan:
Nice. Had a sec to do a first pass. This is a good start.

Agree that it should end up as a built-in as there is no .js component.

A few things to consider with this,

1) May want to support extract at action time so only the archive is in runfiles. I believer rules_webtesting supports this but it has been a while since I looked at that code. That is a work-around for the "no spaces in runfiles" issue as chrome after extracted has a library folder with spaces in it.

2) For cross-platform remote execution we'd want separate repositories per browser & platform & toolchains to go with it. This also makes the repo rule more efficient as you'd only download the browsers in use

3) Add pre-mapped versions for browsers for convenience similar to how we do that for node versions in node_repositories

4) Perhaps add the option to provision webdriver so we can flip our existing rules off of rules_webtesting entirely

PR #2312, opened 2020-11-25
chore: remove old stamping

### LINKED ISSUE #2158: Cleanup old stamping
- Delete bazel_stamp_file from rollup#config_file substitutions (should only use bazel_version_file and bazel_info_file)
- Delete pkg_npm#replace_with_version

alexeagle:
BREAKING CHANGE:
- rollup_bundle config_file no longer has substitutions from a "bazel_stamp_file" - use bazel_version_file instead
- pkg_npm no longer has replace_with_version attribute, use substitutions instead

Fixes #2158

### DIFF
--- a/BUILD.bazel
+++ b/BUILD.bazel
@@ -80,9 +80,6 @@ pkg_npm(
         "BUILD.bazel",
         "LICENSE",
     ],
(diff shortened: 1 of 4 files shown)


gregmagolan (approved):
🌮

PR #2321, opened 2020-11-30
fix(typescript): fail the build when ts_project produces zero outputs

alexeagle:
BREAKING CHANGE:
any ts_project rule that produces no outputs must be fixed or removed


packages/typescript/internal/ts_project.bzl
```diff
@@ -597,7 +597,7 @@ def ts_project_macro(
         write_tsconfig(
             name = "_gen_tsconfig_%s" % name,
             config = tsconfig,
-            files = srcs,
+            files = [s for s in srcs if _is_ts_src(s, allow_js)],
```
gregmagolan, line 600: Is now a good time to add ts_src_extensions and js_src_extensions to attributes and make it configurable?

gregmagolan (approved):

packages/typescript/internal/ts_project.bzl
alexeagle: hmm what other values do you think a user might want to provide? I can't see why they would want such a config

PR #2324, opened 2020-12-02
refactor: default to nobazel_patch_module_resolver

### LINKED ISSUE #2125: --nobazel_patch_module_resolver should be the default
In my last change https://github.com/bazelbuild/rules_nodejs/commit/7d070ffadf9c3b41711382a4737b995f987c14fa#diff-ccd16b070808615570daf5b61d659258 I wrote
> As a later step we'll change default for patch_module_resolver to false (could be for 2.0 or 3.0 depending on how breaking it looks)

we should do this for 3.0

alexeagle:
This turns off our monkey-patches for require() for nodejs_binary, nodejs_test, and macros like jasmine_node_test.
Note it was previously disabled for npm_package_bin and generated index.bzl binaries in https://github.com/bazelbuild/rules_nodejs/pull/1440

BREAKING CHANGE:
we no longer patch the require() function, instead you should rely on the linker to make node modules resolvable at the standard location
if this breaks you, try opting-in to the require patches with templated_args = ["--bazel_patch_module_resolver"]

Fixes #2125

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -447,6 +447,8 @@ tasks:
     # TODO(gregmagolan): figure out how to install missing shared libs
     # Without this filter the @cypress external repository will be built and that build will fail due to missing shared libs.
     - "--build_tag_filters=-cypress,-pkg_npm.pack"
+    - "--define=NODE_DEBUG=module"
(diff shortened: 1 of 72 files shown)


alexeagle:
landed in https://github.com/bazelbuild/rules_nodejs/pull/2347

PR #2327, opened 2020-12-02
feat(builtin): yarn install use --frozen-lockfile as default

### LINKED ISSUE #941: yarn_install should use --frozen-lockfile
# 🚀 feature request

### Relevant Rules
`yarn_install`, `npm_install`

### Description

In the interest of build reproduciblity, if the lockfile is wrong or incomplete then the build should fail.  Running in these frozen modes is often faster as well.

* `yarn_install` should use `yarn install --frozen-lockfile`
* `npm_install` should use `npm ci` instead of `npm install`

This is of course assuming that the relevant tools are new enough to support those arguments. 
 * `npm` [added the `ci` option](https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable) in 5.7.1.  The first node release to ship with npm>5.5.1 was 8.12.0, which could be an issue for versions 8.11.1 or 8.9.1. which are still supported by the rules.
* `yarn` has had the `--frozen-lockfile` parameter since v0.19.0 - much older than any of the [supported versions](https://github.com/bazelbuild/rules_nodejs#installation-with-a-specific-supported-version-of-nodejs-and-yarn).

Make it optional?  Doesn't seem like a good idea, though.

lukasholzer:
To be more hermetic with the install of the dependencies use the frozen lockfile flag to install the exact version from the `yarn.lock` file.

To update a dependency use the vendored yarn binary with `bazel run @nodejs//:yarn upgrade <dep-name>`.

Fixes #941

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently the `yarn_install` rule will use the `yarn install` command and installs the latest version that is possible in the range of the `packag.json`

Issue Number: #941

## What is the new behavior?

The new behaviour is using the frozen lock file option for yarn installs, this means that it will install the exact version from the `yarn.lock` file

- [X] Yes
- [ ] No

### DIFF
--- a/internal/npm_install/npm_install.bzl
+++ b/internal/npm_install/npm_install.bzl
@@ -157,8 +157,8 @@ def _add_data_dependencies(repository_ctx):
     for f in repository_ctx.attr.data:
         to = []
         if f.package:
-            to += [f.package]
-        to += [f.name]
+            to.append(f.package)
+        to.append(f.name)
 
         # Make copies of the data files instead of symlinking
         # as yarn under linux will have trouble using symlinked
@@ -326,6 +326,13 @@ def _yarn_install_impl(repository_ctx):
     yarn = get_yarn_label(repository_ctx)
 
     yarn_args = []
+
+    # Set frozen lockfile as default install to install the exact version from the yarn.lock
+    # file. To perform an yarn install use the vendord yarn binary with:
+    # `bazel run @nodejs//:yarn install` or `bazel run @nodejs//:yarn install -- -D <dep-name>`
+    if repository_ctx.attr.frozen_lockfile:
+        yarn_args.append("--frozen-lockfile")
+
     if not repository_ctx.attr.use_global_yarn_cache:
         yarn_args.extend(["--cache-folder", str(repository_ctx.path("_yarn_cache"))])
     else:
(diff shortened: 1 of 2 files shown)


examples/worker/WORKSPACE
```diff
@@ -29,6 +29,11 @@ load("@build_bazel_rules_nodejs//:index.bzl", "yarn_install")
 
 yarn_install(
     name = "npm",
+    # Set this to true in a real world example (the default value is true), to have hermetic
+    # builds and use the `--frozen-lockfile`. We have to disable this here as the @bazel
+    # scoped packages are getting replaced in the package.json with a file url and the version
+    # won't match with the one in the yarn.lock file then.
+    frozen_lockfile = False,
```
alexeagle, line 36: adding these to the actual code isn't needed, it's a bunch of extra explanation for users, and when they `cd` into this folder to run the example, `frozen_lockfile` will work fine.
We have a plan to fix the integration test runner to use `.tar` file urls which I think will work with frozen lockfile. Therefore the bug is local to the integration test runner, so the workaround should live there too.

Here's the spot where we already modify this file as we copy it into the test fixture
https://github.com/bazelbuild/rules_nodejs/blob/3.x/internal/bazel_integration_test/test_runner.js#L222

Can you just make the change to `yarn_install` there? can be syntactically trivial edit like
`yarn_install(\n` -> `yarn_install(\n    frozen_lockfile=False,\n`

Then we shouldn't need to change any examples

lukasholzer: @alexeagle Thx for the hint – yea I prefer this a lot 👍  Change already applied ;)

alexeagle (approved):

internal/npm_install/npm_install.bzl
```diff
@@ -427,6 +434,20 @@ yarn_install = repository_rule(
...
+            default = True,
+            doc = """Use the `--frozen-lockfile` flag for yarn.
+
+Don’t generate a `yarn.lock` lockfile and fail if an update is needed.
+
+This flag enables an exact install of the version that is specified in the `yarn.lock`
+file. This helps to have reproduceable builds across builds.
+
+To update a dependency or install a new one run the `yarn install` command with the
+vendored yarn binary. `bazel run @nodejs//:yarn install`. You can pass the options like
```
ppodolsky, line 447: You've written here: use vendored yarn binary to update dependency. But it is impossible to do because running vendored yarn binary now requires actual yarn.lock firstly lmao

PR #2328, opened 2020-12-02
feat(builtin): use npm ci as default behaviour for installing node_modules

### LINKED ISSUE #159: Feature request: support `npm ci`
See [this blog post](http://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable) for an introduction to `npm ci`. It would be nice if rules_nodejs had a `npm_ci` command in addition to its `npm_install` command.

lukasholzer:
To be more hermetic with the install of the dependencies use npm ci to install the exact version from the package-lock.json file.

To update a dependency use the vendored npm binary with `bazel run @nodejs//:npm install <dep-name>`.

Fixes #159

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently, the npm_install rule will use the npm install command and installs the latest version that is possible in the range of the packag.json

Issue Number: #2327 159

## What is the new behavior?

The new behaviour is using the `npm ci` command to install the dependencies, this means that it will install the exact version from the package-lock.json file

- [x] Yes
- [ ] No


alexeagle (commented):
could you make the same change as for yarn frozen_lockfile - the examples shouldn't need to change, it's the integration test logic that's wrong

internal/npm_install/npm_install.bzl
```diff
@@ -303,6 +312,18 @@ npm_install = repository_rule(
 See npm CLI docs https://docs.npmjs.com/cli/install.html for complete list of supported arguments.""",
             default = [],
         ),
+        "npm_ci": attr.bool(
```
alexeagle, line 315: a boolean is a strange way to model a string.

what should a user expect that `npm_ci = False` means? I don't think it would be very clear that it means "use npm install instead"

maybe it should be

`"npm_command": attr.string(default = "ci", doc = "which npm command to run to install dependencies", values = ["ci", "install"])`

lukasholzer: A very good idea. I was not 100% happy with the naming but I don't have something better in mind. Your suggestion is a perfect solution, will adapt the changes :)

alexeagle:
@googlebot I consent.

alexeagle (approved):
😎

Issue #2131, opened 2020-08-21
Create a symlink for a package's BUILD.bazel file on "link:" node modules


mistic:
# 🐞 bug report

### Affected Rule

`yarn_install` or `npm_install`

### Is this a regression?

 No

### Description

Currently when using managed dependencies if we declare a `link:` in our `package.json` to a local package the resulting generated `BUILD.bazel` file would be a concatenation between a common `BUILD.bazel` content for a node_module plus the content of the `BUILD.bazel` file in the linked package source code, if any. That behaviour creates a couple of problems and make the rules_nodejs not compliant with the common node development environemnt.

That would be solvable if we create a symlink for the package source `BUILD.bazel` on the npm workspace instead of generating a file. If the package source doesn't contain a `BUILD.bazel` we can still create the symlink and log a warning to let the user know he must create a `BUILD.bazel` file in order to use `link:` packages as node_modules. A possible implementation for this feature would be to look for symlinked modules at https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/npm_install/generate_build_file.ts#L475 and filter those out to be processed in another function responsible for creating the symlinks for `${realpath}/BUILD.bazel`. In order to make that operation more user friendly the `BUILD.bazel` created by the user should only be needed to expose a target with the same name of the package so the implementation of that feature needs to rework a little what is being generated by https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/npm_install/generate_build_file.ts#L964

## 🔬 Minimal Reproduction

There is none provided. A simple bazel setup with managed dependencies and a `link:` on the package.json for some local package to be symlinked as a node_module would be enough.

## 🔥 Exception or Error

None

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
macOS 10.15.6
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
Bazelisk version: v1.5.0
Build label: 3.3.0
Build target: bazel-out/darwin-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Wed Jun 17 12:59:56 2020 (1592398796)
Build timestamp: 1592398796
Build timestamp as int: 1592398796
  </code>
</pre>

**Rules_nodejs version:**

(Please check that you have matching versions between WORKSPACE file and `@bazel/*` npm packages.)

<pre>
  <code>
rules_nodejs@2.0.3
@bazel/typescript@2.0.3
[...]

PR #2330, opened 2020-12-03
feat: create symlink for build files present on node modules installed with relative paths

mistic:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently if a local module installed with npm contains a BUILD file we generate the usual BUILD file content and just append the one found on the BUILD file. 

Issue Number: https://github.com/bazelbuild/rules_nodejs/issues/2131

## What is the new behavior?

With this PR the BUILD file inside an installed local module will be symlinked instead of generated and get its content appended.

- [x] Yes
- [ ] No

It is really unlikely that the current behaviour is being useful to someone who relies on having a BUILD file in a local installed npm module as it introduces a race condition. However, the migration path would be to rewrite their BUILD files inside such modules to contain the expected generated targets by npm_install/yarn_install rules like `*-module__contents` or `*--module__files` and also add the extra targets they have previously defined on the BUILD file. 

HINT: In order to don't have to manually writ `*-module__contents` or `*--module__files` we can remove the BUILD file, run the npm_install/yarn_install rules and just grasp the generated BUILD file content.


alexeagle:
Awesome, this looks good to me. Thanks! I think I agree that no one really should rely on the current behavior but @gregmagolan added that and has a better memory than I do.

internal/npm_install/generate_build_file.ts
```diff
@@ -194,11 +203,33 @@ js_library(
...
   if (pkg._files.includes('BUILD.bazel')) buildFilePath = 'BUILD.bazel';
+
+  // Recreate the pkg dir inside the node_modules folder
+  const nodeModulesPkgDir = `node_modules/${pkg._dir}`;
+  // Check if the current package dep dir is a symlink (which happens when we install a node_module
+  // with link:)
+  const isPkgDirASymlink =
+      fs.existsSync(nodeModulesPkgDir) && fs.lstatSync(nodeModulesPkgDir).isSymbolicLink();
+  // Check if the current package is also written inside the workspace
+  const isPkgInsideWorkspace = fs.realpathSync(nodeModulesPkgDir).includes(fs.realpathSync(`.`));
```
gregmagolan, line 220: Its a corner case, but if `symlink_node_modules = False` then yarn & npm are run in the root of the external repository. In that case, `link:` would be fairly useless as anything linked with a relative path would have to be copied over via that `data` attribute. Still, the comment here should reflect that `fs.realpathSync(`.`)` will not be the root of the workspace if `symlink_node_modules = False`

gregmagolan:
This looks great. I read through the issue and the code and agree that this is the right thing to do if the node_module is a symlink to inside the workspace. It's a clever way to consume 1st party deps. The only thing I'd add is to update the yarn_install & npm_install docs to document this feature & the reasoning.

As an aside, why use this method instead of `js_library` with `package_name` to link 1st party deps? End result of a link in node_modules to the 1st party dep seems the same tho the js_library with package_name would use the linker to achieve it.

mistic:
@gregmagolan Thanks for reviewing this as you are one of the few people well positioned and with enough knowledge to validate the idea. Regarding your question, `js_library` with `package_name` could be used to solve almost every situation. The problem arises when a local package consumed by the monorepo also gets published into the npm registry and starts being used by other `node_module` that you also have installed in your monorepo. That will cause a race condition that will make you not being able to use the linked node_module correctly as the generated BUILD file is not updated yet. If you run npm/yarn install a second time it will succeed because the BUILD file on the second run gots updated. Fixing the legacy behaviour of appending BUILD file contents and link instead will empower us to cover all the situations we could end up having in a node environment.

internal/npm_install/generate_build_file.ts
mistic: @gregmagolan have you any suggestion to solve that corner case?

mistic:
@alexeagle I've also added some content both to the `yarn_install` and `npm_install` rules about that new behaviour. Can you take a look and review it please?

alexeagle:
I'm still not sure this use case of bazel -> bazel dependency makes sense, even given your problem description I still think you should have a `js_library` with `package_name` as the dependency of any Bazel code. That should be orthogonal to whether some non-Bazel dependencies in your repo still depend on the npm-built package.

docs/Built-ins.md
```diff
@@ -947,6 +947,24 @@ Defaults to `True`
 
 Defaults to `3600`
 
+**LOCAL MODULES**
```
alexeagle, line 950: see the note at the top of this file - it's generated so your edits will be lost. You should instead change the stardoc blocks in the source file

mistic: @alexeagle You're right. Any idea about the best place to put those notes as they are not doc for any attribute?

mistic: I've updated the docs correctly on the source

internal/npm_install/generate_build_file.ts
alexeagle: Looking for approve from @gregmagolan to merge as-is

gregmagolan (approved):

Issue #1877, opened 2020-05-07
Remove install_bazel_dependencies


alexeagle:
As a consequence of making all packages loaded from their `@npm//@bazel/...` locations, we no longer need to call a repository rule to load from `@npm` to create new workspaces.

This is important to remove from users code as much as possible since it prevents lazy-installing npm dependencies only for builds that involve frontend. (eg in a monorepo with backend code those devs shouldn't wait for npm installs)

Also consider renaming functions like "npm_bazel_karma_dependencies"? Not sure if it's worth the churn though.

Also consider removing support for the "bazelWorkspaces" package.json key which is now vestigial and maybe no one uses

alexeagle:
Check that we removed `vendor_external` in all the places in pkg_npm rules

alexeagle:
- stop using `name = "npm"` in some/all of the examples/e2e to prove it works with no other changes needed

alexeagle:
working on this in #1921

alexeagle:
Okay, I merged c18a98dbffbdca1534223818a36f94c4b51f6db2 so we no longer recommend using install_bazel_dependencies, and warn if you still use it.

Verified `vendor_external` removed in appropriate places. I'll send another PR now to change some examples to not use "npm".

I think we should leave it in this state for one major so users (Angular ng_module in particular) have a chance to do a migration while getting advantages of our 2.0. So I'll bump the rest of this issue to 3.0 milestone. We should remove install_bazel_depnedencies and the `bazelWorkspaces` package.json key handling at that time.

smhmayboudi:
I receive this message:

- bazelisk --version: bazel 3.3.0
- node -v: v12.8.1
- yarn -v: 1.22.4

```
$ bazelisk build --platforms=@build_bazel_rules_nodejs//toolchains/node:linux_amd64 //apps/action/src:action_nodejs_image
INFO: Invocation ID: 48c402b2-deff-42b8-862e-808728e0f673
ERROR: Failed to load Starlark extension '@npm_bazel_typescript//:index.bzl'.
Cycle in the workspace file detected. This indicates that a repository is used prior to being defined.
The following chain of repository dependencies lead to the missing definition.
 - @npm_bazel_typescript
This could either mean you have to add the '@npm_bazel_typescript' repository with a statement like `http_archive` in your WORKSPACE file (note that transitive dependencies are not added automatically), or move an existing definition earlier in your WORKSPACE file.
ERROR: cycles detected during target parsing
[...]

PR #2335, opened 2020-12-08
chore: remove install_bazel_dependencies

mrmeku:
Fixes https://github.com/bazelbuild/rules_nodejs/issues/1877


mattem:
Should this target 3.x instead?

PR #2336, opened 2020-12-08
Remove jasmine-core as a peer dep for @bazel/concatjs

mattem:
Remove the `jasmine-core` peer dep from `@bazel/concatjs`. This PR also removes it from the examples that don't use `@bazel/jasmine` to test that it's not load bearing. 

With `strict_visibility` enabled now by default, previously this would have caused bazel visibility errors.

### DIFF
--- a/packages/concatjs/web_test/karma_web_test.bzl
+++ b/packages/concatjs/web_test/karma_web_test.bzl
@@ -21,7 +21,6 @@ load("@io_bazel_rules_webtesting//web/internal:constants.bzl", "DEFAULT_WRAPPED_
 KARMA_PEER_DEPS = [
     # NB: uncommented during pkg_npm
     #@external "@npm//@bazel/concatjs",
-    "@npm//jasmine-core",
     "@npm//karma",
     "@npm//karma-chrome-launcher",
     "@npm//karma-firefox-launcher",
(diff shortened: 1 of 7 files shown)


packages/concatjs/web_test/karma_web_test.bzl
```diff
@@ -21,7 +21,6 @@ load("@io_bazel_rules_webtesting//web/internal:constants.bzl", "DEFAULT_WRAPPED_
 KARMA_PEER_DEPS = [
     # NB: uncommented during pkg_npm
     #@external "@npm//@bazel/concatjs",
-    "@npm//jasmine-core",
     "@npm//karma",
```
gregmagolan, line 24: Further down in the file, there is a ```    # The files in the bootstrap attribute come before the require.js support.
    # Note that due to frameworks = ['jasmine'], a few scripts will come before
    # the bootstrap entries:
    # jasmine-core/lib/jasmine-core/jasmine.js
    # karma-jasmine/lib/boot.js
    # karma-jasmine/lib/adapter.js
    # This is desired so that the bootst``` comment.
Guessing `jasmine-core/lib/jasmine-core/jasmine.js` is no longer valid there?

gregmagolan (approved):

PR #2337, opened 2020-12-08
chore: remove install_bazel_dependencies

### LINKED ISSUE #1877: Remove install_bazel_dependencies

mrmeku:
Fixes #1877


.bazelci/presubmit.yml
```diff
@@ -16,10 +16,6 @@ tasks:
     - "//internal/npm_install/test:index"
     # Disabled due to https://github.com/bazelbuild/rules_nodejs/issues/1486
     #- "@fine_grained_deps_yarn//typescript/bin:tsc"
-    - "@bazel_workspace_a//:bin"
```
alexeagle, line 19: my only concern is that these tests might be giving us coverage for something. Maybe we can update them to not rely on install_bazel_dependencies? ppl should still be able to ship npm packages that include bazel rules

alexeagle: I patched the change locally to try that out and understand it better. These nested workspaces are specifically testing the package.json bazelWorkspaces install-as-separate-repository feature that's being removed here, so I think what you have is correct ;)

alexeagle (approved):

alexeagle:
🍕

PR #2341, opened 2020-12-11
refactor: explicitly use runfiles helper

alexeagle:
Broken out from https://github.com/bazelbuild/rules_nodejs/pull/2324/files


jbedard (approved):

WesleyYue:
Is there an example of how we can apply this fix without resorting to `bazel_patch_module_resolver` with ts_project compiled code? I have a minimal example here [reproducing the breakage](https://github.com/WesleyYue/bazel-vue-webpack-typescript/tree/js_module_not_found) (note it's the `js_module_not_found` branch). Is there something we can put in tsconfig.json to help tsc resolve the paths?

PR #2342, opened 2020-12-11
refactor(protractor): load protractor-utils from @bazel/protractor

alexeagle:
This makes it 'linkable' for local consumption, so we don't rely as much on the --bazel_patch_module_resolver
patched require() function. Evidently from the examples, this is non-breaking and users can continue to deep-import the helper, though it's no longer necessary

### DIFF
--- a/packages/protractor/BUILD.bazel
+++ b/packages/protractor/BUILD.bazel
@@ -62,7 +62,8 @@ nodejs_test(
 ts_library(
     name = "protractor",
     srcs = ["protractor-utils.ts"],
-    module_name = "@bazel/protractor/protractor-utils",
+    data = ["package.json"],
+    module_name = "@bazel/protractor",
     module_root = "protractor-utils",
     deps = ["@npm//@types/node"],
 )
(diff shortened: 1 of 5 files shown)


packages/protractor/test/protractor-utils/BUILD.bazel
```diff
@@ -28,5 +28,6 @@ jasmine_node_test(
     srcs = [":protractor_utils_tests_lib"],
     data = [
         ":fake-devserver",
+        "//packages/protractor",
```
jbedard, line 31: Why was this one required? Shouldn't that get pulled in with the `srcs`?

alexeagle: I think the answer is that the linker needs to see it so the package is resolvable, and the linker doesn't look at the srcs attribute.

alexeagle: Actually now that I comment it out, I think you're right that it's not needed, seems to pass the flag flip test without this line. Removed :)

jbedard (approved):

gregmagolan (approved):

PR #2343, opened 2020-12-11
refactor(typescript): ts_library plugins use custom compiler

alexeagle:
This is conceptually simpler than having the compiler resolve dependencies out of one compilation request that uses it.
It's also more compatible with the linker under worker sandboxing, where the compiler won't see the plugins at startup before the first compilation job is sent.

### DIFF
--- a/packages/typescript/test/angular_plugin/BUILD.bazel
+++ b/packages/typescript/test/angular_plugin/BUILD.bazel
@@ -1,0 +1,23 @@
+load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_binary")
+
+# Custom ts_library compiler that runs tsc_wrapped with angular/compiler-cli statically linked
+# This can be used with worker mode because we don't need the linker at runtime to make
+# the angular plugin loadable
+# Just a clone of @build_bazel_rules_typescript//internal:tsc_wrapped with added deps
+nodejs_binary(
+    name = "tsc_wrapped_with_angular",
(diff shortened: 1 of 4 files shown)


packages/typescript/test/angular_plugin/BUILD.bazel
```diff
@@ -0,0 +1,23 @@
+load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_binary")
+
+# Custom ts_library compiler that runs tsc_wrapped with angular/compiler-cli statically linked
+# This can be used with worker mode because we don't need the linker at runtime to make
+# the angular plugin loadable
+# Just a clone of @build_bazel_rules_typescript//internal:tsc_wrapped with added deps
```
jbedard, line 6: Why all the added deps?

I'm comparing to https://github.com/bazelbuild/rules_nodejs/blob/2.3.1/examples/angular/tools/BUILD.bazel ...

alexeagle: it's because this is using the first-party code directly rather than the `@bazel/typescript` npm package so the list of deps we repeat is https://github.com/bazelbuild/rules_nodejs/blob/3.x/third_party/github.com/bazelbuild/rules_typescript/internal/BUILD.bazel#L75-L83

mattem (approved):

gregmagolan (approved):

PR #2344, opened 2020-12-11
refactor: make dependencies on --bazel_patch_module_resolver explicit

alexeagle:
We are about to flip the default for this flag in #2125 so we need to explicitly opt-in in places where we need the behavior.
Ideally most of these should be fixed, this would be necessary if we ever want to deprecate the module resolver patches

### DIFF
--- a/packages/protractor/protractor_web_test.bzl
+++ b/packages/protractor/protractor_web_test.bzl
@@ -261,6 +261,8 @@ def protractor_web_test(
         entry_point = Label(protractor_entry_point),
         data = srcs + deps + data + [Label(d) for d in peer_deps],
         testonly = 1,
+        # TODO: make protractor binary not depend on monkey-patched require()
+        templated_args = ["--bazel_patch_module_resolver"],
         visibility = ["//visibility:private"],
     )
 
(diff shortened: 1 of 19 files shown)


packages/protractor/protractor_web_test.bzl
```diff
@@ -261,6 +261,8 @@ def protractor_web_test(
         entry_point = Label(protractor_entry_point),
         data = srcs + deps + data + [Label(d) for d in peer_deps],
         testonly = 1,
+        # TODO: make protractor binary not depend on monkey-patched require()
```
jbedard, line 264: That's actual protractor and something we can't fix?

jbedard:
I think most of these are just in internal BUILDs and tests? Except protractor, any others I didn't notice? Not sure if it's worth tracking those somewhere...?

packages/protractor/protractor_web_test.bzl
alexeagle: pretty sure it's our wrapper code. I figured our protractor package isn't that popular so i didn't spend much time here digging in to figure out why it's not working, but i think this is the main one to come back to and try to fix by 3.0

gregmagolan (approved):

PR #2348, opened 2020-12-11
fix(builtin): repair unintended breaking change

alexeagle:
In 2.3.1 we included https://github.com/bazelbuild/rules_nodejs/commit/bc36519086536e1576c7b0657437dccced397782#diff-0891b24ca97cb1c92991a8231fd0d58e42d9733bc0c329f7bd7cd1b40459d0e7R330 which changed the location of generated pkg_npm pack/publish targets to require a '.sh' suffix


alexeagle:
fyi @JiaLiPassion

alexeagle:
um actually I'm not sure about the breaking change, discussing right now in slack, sorry for the noise @JiaLiPassion

alexeagle:
https://github.com/angular/angular/blob/master/packages/bazel/src/ng_package/ng_package.bzl#L799-L800 is what we broke, but it's not a rules_nodejs public API so nothing to do here I think

PR #2350, opened 2020-12-14
chore(docs): remove old rules_typescript changes from 3.0.0-rc.0 changelog

jbedard:
This drops all the commits for changes in `third_party/github.com/bazelbuild/rules_typescript`. I assume we want to drop all of them? Most are ancient, but there's a couple newer ones that might actually be new changes?

* https://github.com/bazelbuild/rules_nodejs/commit/8bd9734
* https://github.com/bazelbuild/rules_nodejs/commit/02aac76

I also couldn't find where the "PiperOrigin-RevId: 221634093" came from but I assume that's super old from rules_typescript.


alexeagle (approved):
thanks!

PR #2355, opened 2020-12-18
chore: flip default of new bazel_run_from_execroot

alexeagle:
Want to see how green this is, ideally the flag flip is included in 3.0.0

/cc @duarten

### DIFF
--- a/internal/node/launcher.sh
+++ b/internal/node/launcher.sh
@@ -180,7 +180,7 @@ EXIT_CODE_CAPTURE=""
 RUN_LINKER=true
 NODE_PATCHES=true
 PATCH_REQUIRE=false
-FROM_EXECROOT=false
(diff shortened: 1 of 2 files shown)


alexeagle:
@duarten it's pretty red, I see all the first-party entrypoints (those declared in the current workspace) need the old behavior.

maybe the flag is better named "third_party_entry_point" but if that's really the difference, it's probably easy to detect if the workspace of the entry point is the current workspace. So maybe we can drop the flag for 3.x and just set the correct value always? Have any time in the next couple days to poke at this?

duarten:
@alexeagle I do have some time; I'll look into it! Seems like a nice solution.

alexeagle:
Thanks, let's check in sometime tomorrow. Hoping to cut 3.0.0 on Tuesday.

Issue #1886, opened 2020-05-12
Remove node_modules attribute from all rules


gregmagolan:
Users should use node_modules_library if they want to self manage node_modules and pass the target to deps.

TODO(@gregmagolan): need a little investigation to come up with the right recommendations.

alexeagle:
I spent an hour or two trying to tear these out and didn't succeed. Maintainers are stretched thin on time right now so punting this to next breaking release.

alexeagle:
fixed by https://github.com/bazelbuild/rules_nodejs/commit/c2927afa9a5d77652783f94eada80baf1c9be68f

PR #2356, opened 2020-12-19
refactor(builtin): remove node_modules attribute from nodejs_binary, nodejs_test & ts_library

gregmagolan:
Resolves https://github.com/bazelbuild/rules_nodejs/issues/1886

BREAKING CHANGE:

We removed the node_modules attribute from `nodejs_binary`, `nodejs_test`, `jasmine_node_test` & `ts_library`.

If you are using the `node_modules` attribute, you can simply add the target specified there to the `data` or `deps` attribute of the rule instead.

For example,

```
nodejs_test(
    name = "test",
    data = [
        "test.js",
        "@npm//:node_modules",
    ],
    entry_point = "test.js",
)
```

or

```
ts_library(
    name = "lib",
    srcs = glob(["*.ts"]),
    tsconfig = ":tsconfig.json",
    deps = ["@npm//:node_modules"],
)
```

We also dropped support for filegroup based node_modules target and removed `node_modules_filegroup` from `index.bzl`.

If you are using this feature for user-managed deps, you must now use a `js_library` target
with `external_npm_package` set to `True` instead.

For example,

```
js_library(
    name = "node_modules",
    srcs = glob(
        include = [
            "node_modules/**/*.js",
            "node_modules/**/*.d.ts",
            "node_modules/**/*.json",
            "node_modules/.bin/*",
        ],
        exclude = [
            # Files under test & docs may contain file names that
            # are not legal Bazel labels (e.g.,
            # node_modules/ecstatic/test/public/中文/檔案.html)
            "node_modules/**/test/**",
            "node_modules/**/docs/**",
            # Files with spaces in the name are not legal Bazel labels
            "node_modules/**/* */**",
            "node_modules/**/* *",
        ],
    ),
    # Provide ExternalNpmPackageInfo which is used by downstream rules
    # that use these npm dependencies
    external_npm_package = True,
)

nodejs_test(
    name = "test",
    data = [
        "test.js",
        ":node_modules",
    ],
    entry_point = "test.js",
)
```

See `examples/user_managed_deps` for a working example of user-managed npm dependencies.


jbedard:
So much beautiful deleting 😄 

I also got started on this but looks like you got further. I was having trouble getting the md => html tool running for some reason.

packages/concatjs/_README.md
```diff
@@ -117,19 +117,3 @@ browser_repositories(
     firefox = True,
 )
 ```
-
-## Installing with self-managed dependencies
```
alexeagle, line 121: isn't there some new version of this guidance? are we trying to deprecate "self-managed dependencies"?

Do they just need to move //:node_modules to data?

WORKSPACE
```diff
@@ -53,22 +53,6 @@ yarn_install(
     environment = {
         "SOME_USER_ENV": "yarn is great!",
     },
-    # The @npm//:node_modules_filegroup generated by manual_build_file_contents
-    # is used in the //packages/typescript/test/reference_types_directive:tsconfig_types
-    # test. For now we're still supporting node_modules as a filegroup tho this may
-    # change in the future. The default generated //:node_modules target is a js_library
-    # rule which provides NpmPackageInfo but we have not yet dropped support for
```
alexeagle, line 60: this looks like good text for the BREAKING CHANGE message
> dropped support for filegroup based node_modules target

e2e/packages/BUILD.bazel
```diff
@@ -35,10 +35,9 @@ nodejs_test(
     data = [
         "npm_determinism.spec.js",
         "@e2e_packages_npm_install//:node_modules/jsesc/package.json",
+        "@e2e_packages_npm_install_duplicate_for_determinism_testing//:node_modules",
     ],
     entry_point = ":npm_determinism.spec.js",
-    node_modules = "@e2e_packages_npm_install_duplicate_for_determinism_testing//:node_modules",
-    # TODO(gregmagolan): fix this test on windows
     tags = ["fix-windows"],
```
alexeagle, line 41: no more TODO to fix-windows?

examples/user_managed_deps/BUILD.bazel
```diff
@@ -38,15 +43,14 @@ nodejs_binary(
...
-jasmine_node_test(
+nodejs_test(
     name = "test",
-    srcs = glob(["*.spec.js"]),
-    node_modules = "//:node_modules",
-    tags = ["no-local-jasmine-deps"],
     # user-managed deps don't expose the provider the linker needs to make require work
-    templated_args = ["--bazel_patch_module_resolver"],
-    deps = [
+    # templated_args = ["--bazel_patch_module_resolver"],
```
alexeagle, line 49: leftover comment

index.bzl
```diff
@@ -63,21 +63,6 @@ COMMON_REPLACEMENTS = {
     "\"@bazel/([a-zA-Z_-]+)\":\\s+\"(file|bazel)[^\"]+\"": "\"@bazel/$1\": \"0.0.0-PLACEHOLDER\"",
 }
 
-def node_modules_filegroup(packages, patterns = [], **kwargs):
```
alexeagle, line 66: looks like also good to document in BREAKING CHANGE

internal/js_library/js_library.bzl
```diff
@@ -54,6 +54,10 @@ _ATTRS = {
         They will be copied into the package bin folder if needed.""",
         allow_files = True,
     ),
+    "npm_package": attr.bool(
+        doc = """If set to true, js_library will provide NpmPackageInfo.""",
```
alexeagle, line 58: maybe more text here - users should basically never set this unless making their own user-managed third-party dependencies?

internal/npm_install/generate_build_file.ts
```diff
@@ -1009,13 +1012,15 @@ js_library(
...
+    npm_package = True,
     srcs = [":${pkg._name}__files", ":${pkg._name}__nested_node_modules"],${namedSourcesStarlark}
     visibility = ["//:__subpackages__"],
 )
 
 # Typings files that are part of the npm package not including nested node_modules
 js_library(
-    name = "${pkg._name}__typings",${dtsStarlark}
+    name = "${pkg._name}__typings",
+    npm_package = True,${dtsStarlark}
```
alexeagle, line 1023: seeing this API seems a little wrong. If I'm making a `js_library` in my own code why wouldn't I say it's an npm package?
Should it be named differently? if users aren't meant to call it, could be a longer name like `third_party_npm_package` 
Another possibility is a different rule name like external_js_library ("external" is nice echo from rules_jvm_external) - it would be implemented in the same file with js_library so they share code

WORKSPACE
gregmagolan: +1

index.bzl
gregmagolan: +1

e2e/packages/BUILD.bazel
gregmagolan: whoops. didn't mean to remove that thanks.

packages/concatjs/_README.md
gregmagolan: Restored with updated instructions

examples/user_managed_deps/BUILD.bazel
gregmagolan: cleaned up

internal/npm_install/generate_build_file.ts
gregmagolan: Discussed offline & renamed to external_npm_package. Also renamed related provided to ExternalNpmPackageInfo to match.

internal/js_library/js_library.bzl
gregmagolan: Added more

alexeagle (approved):
🧇

PR #2358, opened 2020-12-21
feat: automate bazel_run_from_execroot

duarten:
Replace bazel_run_from_execroot by turning it on or off automatically,
based on the workspace to which the nodejs binary belongs. If it belongs
to the current workspace, use the old behavior. If it belong to a third
party workspace, use the behavior of bazel_run_from_execroot.

Signed-off-by: Duarte Nunes <duarte@hey.com>

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


duarten:
cc @alexeagle. Did the most straightforward thing. Works fine for tests running third party nodejs bins, and it the unit tests passed locally.

duarten:
Investigating the redness.

duarten:
It seems like the paths change quite a bit when running with `actions.run()/genrule`, as opposed to using `bazel run`. The latter runs out of the execroot, but the former runs out of the sandbox.

For the failing test case, the terser binary is at `bazel-out/host/bin/external/npm/terser/bin/terser.sh.runfiles/npm/node_modules/terser/bin/terser`. For the flag to work, it would need to be above `bazel-out/{arch}/bin/internal/linker/test/...`, which would mean it should be in `./node_modules`. But that directory is empty. 

I can detect this case and fallback to the old behavior. In any case, specifying `bazel_run_from_execroot` will break for binaries executed during the build.

What do you think?

alexeagle:
Looks like your path difference is host vs target platform. It's correct that terser is run on the host machine so it's built in the host configuration. 

> In any case, specifying bazel_run_from_execroot will break for binaries executed during the build.

Are you saying you haven't found a way to set a correct value for this flag in all cases? We obviously can't just break people. Maybe there's a correct way to determine this. I'd have to spend some time diving into the code though, as you've observed the path logic is really subtle.

It doesn't have to make it in 3.0.0

duarten:
Yes, if the flag is set for targets that are executed during the build (now I'm not sure if it's just related to the host configuration or whether it's related to the sandbox or something else), then it doesn't work correctly. Maybe the linker can help here, or maybe the way Bazel structures the paths is just incompatible with the node module resolution mechanism. I'll also dig more into this.

PR #2361, opened 2020-12-22
fix(builtin): don't pass kwargs to generated_file_test .update binary

alexeagle:
The test itself just runs diff, so it's always fast. This prevents warnings under bazel test --test_verbose_timeout_warnings
Note that the **kwargs are passed to both the binary and test targets here, so users can't pass this attribute


gregmagolan (approved):
Could pass size to just the test target as well so it is user configurable?

PR #2362, opened 2020-12-22
build: match bazel-toolchains version with bazel version

mattem:
Matches the `bazel-toolchains` version used for rbe with the bazel version to remove the DEBUG warning messages in the output


alexeagle (approved):

alexeagle:
seems like we could include this in the bazel version test, maybe in a followup

PR #2366, opened 2020-12-23
feat: add support for NodeJS 15.x

### LINKED ISSUE #2256: Upgrade our examples to npm7

### LINKED ISSUE #2339: pkg_npm should produce a tar output, not just runnable .pack target

mattem:
Adds support for NodeJS ^15.0.0

Main issue was captured in https://github.com/npm/cli/issues/2145, due to a breaking change in how npm handles transitive node_modules of file linked paths.
The included change here generates a tar file for the package and replaces the file path to the tar into the `package.json` under test.

closes #2339 
closes #2256

### DIFF
--- a/examples/vendored_node/WORKSPACE
+++ b/examples/vendored_node/WORKSPACE
@@ -26,10 +26,10 @@ http_archive(
 )
 
 http_archive(
-    name = "vendored_node_10_12_0",
-    build_file_content = """exports_files(["node-v10.12.0-linux-x64/bin/node"])""",
-    sha256 = "4eba2e9a6db95745b769915d58e57df6ca6724ec1f023f76556fce30ceca2367",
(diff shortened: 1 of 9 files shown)


alexeagle (approved):
sweet! I think this unblocks some other fix that was tripping over our npm package overrides in the integration test - something about a workaround...

Ah yes, I think we could un-hack here:
https://github.com/bazelbuild/rules_nodejs/commit/b6a8cbb78473df25756c6021b71a291fbeb5a8b5#diff-da7e70fd50ce8f04f6ac7d53c8a254acfabac83a73dc10a6c16dfe81875b563f
(and the corresponding one for `npm ci` https://github.com/bazelbuild/rules_nodejs/commit/1d650fbb4e383b20ce895dc815d07a74dc4c1b7e#diff-da7e70fd50ce8f04f6ac7d53c8a254acfabac83a73dc10a6c16dfe81875b563fR229
)

maybe as a follow-up cleanup tho

PR #2370, opened 2020-12-25
fix(examples): angualr universal_server

flolu:
Since rules_nodejs v3 the `--bazel_patch_module_resolver` flag is required to run the universal server

### DIFF
--- a/examples/angular/src/BUILD.bazel
+++ b/examples/angular/src/BUILD.bazel
@@ -311,4 +311,5 @@ nodejs_binary(
         ":universal_server_lib",
     ],
     entry_point = ":server.ts",
+    templated_args = ["--bazel_patch_module_resolver"],
 )


alexeagle (approved):
thanks for noticing!
could we add a test for this so we don't break it again?
also, would it be better to change server.ts to not rely on patched require() ? maybe it should use a runfiles helper

flolu:
**re:1**
What would you suggest as a test? Probably something similar to the `protractor_web_test_suite`s in `/e2e`, but spinning up the `universal_server` instead of the `prodserver`?

**re:2**
I'm actually not sure how to do this. Do you have any suggestions/examples for me?

alexeagle:
I'll merge as-is. How would you test angular universal outside of Bazel? Don't need a browser, I would guess domino is sufficient like the examples/kotlin test.

PR #2389, opened 2021-01-08
feat: support for nested node_modules in linker

gregmagolan:
This PR adds support for nested `node_modules` folders in the linker, which allows targets to depend on npm packages from multiple `yarn_install` & `npm_install` repository rules (max one per directory in the workspace).

This is an opt-in feature which is enabled by specifying `package_path` in `yarn_install` or `npm_install` for nested `package.json` files. Setting `package_path` informs the linker to link the external npm repository to a `node_modules` folder in the `package_path` sub-directory specified.

For example, given two yarn_install rules:

```
yarn_install(
    name = "npm",
    yarn_lock = "//:yarn.lock",
)

yarn_install(
    name = "npm_subdir",
    package_json = "//subdir:package.json",
    package_path = "subdir",
    yarn_lock = "//subdir:yarn.lock",
)
```

a target may depend on npm packages from both,

```
jasmine_node_test(
    name = "test",
    srcs = ["test.js"],
    deps = [
        "@npm//foo",
        "@npm_subdir//bar",
    ],
)
```

and the linker will link multiple 3rd party node_modules folders,

```
/node_modules => @npm//:node_modules
/subdir/node_modules => @npm_subdir/:node_modules
```

making the 3rd party npm deps `foo` & `bar` available at

```
/node_modules/foo
/subdir/node_modules/bar
```

This feature is opt-in as `package_path` currently defaults to `""`.  In a future major release (4.0.0), `package_path` will default to the directory the `package.json` file is in, which will turn it on by default.

This feature is an alternative to yarn workspaces & npm workspaces, which also lay out `node_modules` in multiple directories from multiple `package.json` files.

To date, targets have been limited to depending on and resolving form a single `yarn_install` or `npm_install` workspace which was always linked to a `node_modules` folder at the root of the workspace. With this change, the linker is able to link to both the root of the workspace as well as sub-directories in the workspace so targets may now resolve 3rd party npm deps from multiple `package.json` files which correspond to multiple `node_modules` folders in the tree.

1st party deps are still always linked to the root `node_modules` folder as they were before this change.

_NB: depending on multiple npm workspaces will not be supported by ts_library as its resolution logic is limited to a single node_modules root_

_NB: pre-linker node require patches (deprecated as of 3.0.0) will only include root node_modules; no nested node_modules shall included in require patch resolution_


internal/linker/index.js
```diff
@@ -393,28 +365,73 @@ function visitDirectoryPreserveLinks(dirPath, visit) {
...
+                else {
+                    log_verbose('no root npm workspace node_modules folder to link to; creating node_modules directory in', process.cwd());
+                    yield mkdirp('node_modules');
+                }
+            }
+        }
+        const rootNpmWorkspace = roots ? roots[''] : undefined;
+        if (!rootNpmWorkspace) {
+            log_verbose('no root npm workspace; creating node_modules directory in ', process.cwd());
+            yield mkdirp('node_modules');
```
Aghassi, line 432: Should this also set the `rootNpmWorkspace` since I assume you are now creating one?

internal/linker/link_node_modules.bzl
```diff
@@ -162,7 +149,7 @@ def _get_module_mappings(target, ctx):
     # pkg_npm but not vice-verse at the moment since ts_library cannot handle directory artifacts as
     # deps.
     if hasattr(linkable_package_info, "_tslibrary") and linkable_package_info._tslibrary:
-        mr[0] = "_tslibrary"
+        mr[0] = "__tslibrary__"
```
Aghassi, line 152: Not a change, but `mr[0]` reads in my head as "Mr. Zero" and it sounds like a Bond villain

internal/node/launcher.sh
```diff
@@ -259,20 +254,68 @@ if [[ "$PWD" == *"/bazel-out/"* ]]; then
...
-    # and under runfiles it is the legacy external runfiles path
-    export BAZEL_PATCH_GUARDS="${BAZEL_PATCH_GUARDS},${BAZEL_PATCH_ROOT}/${BAZEL_NODE_MODULES_ROOT},${PWD}/external/${BAZEL_NODE_MODULES_ROOT}"
+if [[ -n "${BAZEL_NODE_MODULES_ROOTS:-}" ]]; then
+  # BAZEL_NODE_MODULES_ROOTS is in the format "<path>:<workspace>,<path>:<workspace>"
+  # (e.g., "internal/linker/test:npm_internal_linker_test,:npm")
+  if [[ -n "${VERBOSE_LOGS:-}" ]]; then
+    echo "BAZEL_NODE_MODULES_ROOTS=${BAZEL_NODE_MODULES_ROOTS}"
   fi
+  OLDIFS="${IFS}"
+  IFS=","
```
Aghassi, line 284: What is IFS?

Aghassi (commented):
I made a mistake and reviewed some of the `js` files not knowing they are generated. Other than that, this looks good based on my limited knowledge

internal/node/launcher.sh
gregmagolan: bash magic https://bash.cyberciti.biz/guide/$IFS

internal/linker/index.js
```diff
@@ -393,28 +365,73 @@ function visitDirectoryPreserveLinks(dirPath, visit) {
...
-        if (!(yield exists(rootDir))) {
-            log_verbose('no third-party packages; mkdir node_modules at', root);
-            yield mkdirp(rootDir);
-        }
-        yield symlink(rootDir, 'node_modules');
-        process.chdir(rootDir);
+        log_verbose('startCwd:', startCwd);
+        const execroot = findExecroot(startCwd);
+        log_verbose('execroot:', execroot ? execroot : 'not found');
+        const isExecroot = startCwd == execroot;
```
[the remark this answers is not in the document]
gregmagolan: The == was intentional. I'll change to `const isExecroot = (startCwd == execroot);` to be clear. Thanks.

internal/linker/index.js
```diff
@@ -120,33 +120,24 @@ function symlink(target, p) {
         }
     });
 }
-function resolveRoot(root, startCwd, isExecroot, runfiles) {
+function resolveExternalWorkspacePath(workspace, startCwd, isExecroot, execroot, runfiles) {
```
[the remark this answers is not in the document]
gregmagolan: I believe there are comments in the .ts source

internal/linker/link_node_modules.bzl
gregmagolan: 🔫

internal/linker/link_node_modules.bzl
```diff
@@ -67,30 +58,35 @@ def write_node_modules_manifest(ctx, extra_data = [], mnemonic = None, link_work
...
-                if v[0] == "_tslibrary":
-                    v = ["execroot", v[1]]
                 _debug(ctx.var, "Linking %s: %s" % (k, v))
                 mappings[k] = v
 
+    # Simplify mappings before writing them to the manifest file.
+    # The extra complication is required in bzl to handle the __tslibrary__ special
+    # case since module_name in ts_library enables the linker. In a future major release,
+    # ts_library will get a package_name attribute to enable the linker and the __tslibrary__
+    # special case can be factored out.
```
alexeagle, line 86: make a 4.0 cleanup issue for this and link it here?

internal/linker/link_node_modules.ts
```diff
@@ -148,72 +148,38 @@ async function symlink(target: string, p: string): Promise<boolean> {
...
-    // Under execroot, the root will be under an external folder from the startCwd
-    // `execroot/my_wksp`. For example, `execroot/my_wksp/external/npm/node_modules`. If there is no
-    // root, which will be the case if there are no third-party modules dependencies for this
-    // target, set the root to `execroot/my_wksp/node_modules`.
-    return root ? `${startCwd}/external/${root}` : `${startCwd}/node_modules`;
+    // Under execroot, the npm workspace will be under an external folder from the startCwd
+    // `execroot/my_wksp`. For example, `execroot/my_wksp/external/npm`. If there is no
+    // npm workspace, which will be the case if there are no third-party modules dependencies for
+    // this target, npmWorkspace the root to `execroot/my_wksp/node_modules`.
+    return `${execroot}/external/${workspace}`;
```
alexeagle, line 161: does this interact with `--noenable_legacy_runfiles` in any way?

internal/linker/test/package.json
```diff
@@ -0,0 +1,6 @@
+{
+    "private": true,
+    "devDependencies": {
+        "semver": "1.0.0"
```
alexeagle, line 4: nit: maybe a comment that this version differs from the one in the root (and maybe also a comment in the root package.json to prevent us upgrading it to 1.0.0 by accident and losing the assertion)

internal/linker/test/yarn.lock
```diff
@@ -0,0 +1,8 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+semver@1.0.0:
```
alexeagle, line 5: this file is so short, it makes me wonder (in a separate PR) if we could make it work to say

```
http_file(
   name="npm_semver", urls = ["https://registry.yarnpkg.com/semver/-/semver-1.0.0.tgz#11f18a0c08ed21c988fc2b0257f1951969816615"]
)
external_npm_package(
    name = "npm__semver",
    tars = "that one ☝🏻 "
)
```
or sth so users could opt-out of using npm or yarn and use only Bazel. Probably an experiment we could do along with the spike I've been working on.

internal/npm_install/npm_install.bzl
```diff
@@ -79,6 +79,14 @@ fine grained npm dependencies.
         mandatory = True,
         allow_single_file = True,
     ),
+    "package_path": attr.string(
+        default = "",
+        doc = """If set, link the 3rd party node_modules dependencies under the package path specified.
+
+In most cases, this should be the directory of the package.json file so that the linker links the node_modules
+in the same location they are found in the source tree. In a future release, this will default to the package.json
```
alexeagle, line 87: make a 4.0 tracking issue for this and link it here

alexeagle (commented):
did a pass over this for the "shape" of the change and impacts on users, will do a final comb over the implementation logic when it's green

internal/bazel_integration_test/test_runner.js
```diff
@@ -340,7 +340,7 @@ for (const bazelCommand of config.bazelCommands) {
     'BASH_FUNC_is_absolute%%',
     'BASH_FUNC_rlocation%%',
     'BASH_FUNC_runfiles_export_envvars%%',
-    'BAZEL_NODE_MODULES_ROOT',
```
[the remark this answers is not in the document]
gregmagolan: doesn't seem likely as it is an internal detail to the linker and not part of the public API

internal/js_library/js_library.bzl
```diff
@@ -73,6 +73,12 @@ _ATTRS = {
         See `examples/user_managed_deps` for a working example of user-managed npm dependencies.""",
         default = False,
     ),
+    "external_npm_package_path": attr.string(
```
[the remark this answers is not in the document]
gregmagolan: There was the one use case where Tiago was manually setting up js_library targets that mimicked what yarn_install lays out. If we made `external_npm_package_path` internal then we'd want to make `external_npm_package` private as well.

internal/linker/link_node_modules.ts
gregmagolan: No. The path at the execroot is always `execroot/my_workspace/external/other_workspace`. Bazel doesn't layout an `execroot/other_workspace`

Aghassi:
Although I can't review this, I did test this internally and it works as expected so :shipit:

internal/linker/test/package.json
gregmagolan: root package.json would be a downgrade as it is currently at 5.6.0. 

I added a negative test that will assert break if someone happens to downgrade root to < 5.0.0
```
  it('should get semver from root pacakge.json (which is currently 5.6.0) if the is no transitive in 1p dep',
     () => {
       const onepb = require('@rules_nodejs/onepb');
       const major = onepb.semver().major(onepb.semverVersion());
       expect(major).toBeGreaterThanOrEqual(5);
     });
```

internal/linker/link_node_modules.bzl
gregmagolan: done

internal/linker/index.js
gregmagolan: Not used anywhere else.
Changed to
```
if (!roots || !roots['']) {
```
to be more clear on that

internal/npm_install/npm_install.bzl
gregmagolan: 👍

internal/linker/link_node_modules.bzl
```diff
@@ -67,30 +58,36 @@ def write_node_modules_manifest(ctx, extra_data = [], mnemonic = None, link_work
...
-            elif node_modules_root != possible_root:
-                fail("All npm dependencies need to come from a single workspace. Found '%s' and '%s'." % (node_modules_root, possible_root))
-
-        # ...first-party packages to be linked into the node_modules tree
+            path = dep[ExternalNpmPackageInfo].path
+            workspace = dep[ExternalNpmPackageInfo].workspace
+            if path in node_modules_roots:
+                other_workspace = node_modules_roots[path]
+                if workspace != other_workspace:
+                    fail("All npm dependencies at the path '%s' must to come from a single workspace. Found '%s' and '%s'." % (path, other_workspace, workspace))
```
alexeagle, line 73: nit: "must to come" -> "must come"

internal/linker/link_node_modules.bzl
```diff
@@ -127,18 +124,9 @@ def _get_module_mappings(target, ctx):
     for name in _MODULE_MAPPINGS_DEPS_NAMES:
         for dep in getattr(ctx.rule.attr, name, []):
             for k, v in getattr(dep, _ASPECT_RESULT_NAME, {}).items():
-                # A package which was reachable transitively via a *_binary or *_test
-                # rule is assumed to be in the runfiles of that binary,
-                # so we switch the linker target root.
-                # In theory we ought to be able to tell that the files really are
-                # propagated through the runfiles output of the target we are visiting
-                # but it seems that this info is only available in Bazel Java internals.
-                # TODO: revisit whether this is the correct condition after downstream testing
```
alexeagle, line 136: I like dropping the TODO - just hoping it's not a problem anymore?

internal/linker/link_node_modules.ts
```diff
@@ -136,72 +139,38 @@ async function symlink(target: string, p: string): Promise<boolean> {
...
+  }
+
+  if (!execroot) {
+    // This can happen if we are inside a nodejs_image or a nodejs_binary is run manually.
+    // Resolve as if we are in runfiles in a sandbox. This will be the case for nodejs_image.
```
alexeagle, line 157: nit: the last sentence is redundant

internal/linker/link_node_modules.ts
```diff
@@ -600,43 +531,143 @@ export type LinkerAliases = {
...
+        // root node_modules folder for 1st party deps
+        log_verbose(
+            'no root npm workspace node_modules folder to link to; creating node_modules directory in',
+            process.cwd());
+        await mkdirp('node_modules');
+      }
+    }
+  }
+
+  if (!roots || !roots['']) {
```
alexeagle, line 662: wondering why roots would be truthy and then we want to check if it has an empty string element. i guess there's some logic somewhere that treats no third-party modules as empty? feels like it should be fixed there (don't put an empty string in the roots)

internal/node/launcher.sh
```diff
@@ -257,24 +252,80 @@ if [[ "$PWD" == *"/bazel-out/"* ]]; then
   readonly rest=${PWD#*$bazel_out}
   readonly index=$(( ${#PWD} - ${#rest} - ${#bazel_out} ))
   if [[ ${index} < 0 ]]; then
-    echo "No 'bazel-out' folder found in path '${PWD}'!"
+    echo "No 'bazel-out' folder found in path '${PWD}'!" >&2
```
alexeagle, line 255: nit: I saw a trick in the bazel sources that was
`echo >&2 other arguments here`
which seems neat as a way to say "echo to stderr"
but if we like that we should probably make a separate change

internal/node/launcher.sh
```diff
@@ -257,24 +252,80 @@ if [[ "$PWD" == *"/bazel-out/"* ]]; then
...
+    echo "BAZEL_NODE_MODULES_ROOTS=${BAZEL_NODE_MODULES_ROOTS}" >&2
   fi
+  OLDIFS="${IFS}"
+  IFS=","
+  roots=(${BAZEL_NODE_MODULES_ROOTS})
+  IFS="${OLDIFS}"
+  for root in "${roots[@]}"; do
+    if [[ "${root}" ]]; then
+      OLDIFS="${IFS}"
+      IFS=":"
```
alexeagle, line 295: yeah I guess it's more likely portable than using `cut` ?

internal/node/launcher.sh
```diff
@@ -257,24 +252,80 @@ if [[ "$PWD" == *"/bazel-out/"* ]]; then
...
+        # (e.g., /private/.../execroot/build_bazel_rules_nodejs/internal/linker/test/node_modules)
+        export BAZEL_PATCH_ROOTS="${BAZEL_PATCH_ROOTS},${EXECROOT}/${root_path}/node_modules"
+        if [[ "${RUNFILES_ROOT}" ]]; then
+          # If in runfiles guard the node_modules location in runfiles as well
+          # (e.g., /private/.../execroot/build_bazel_rules_nodejs/bazel-out/darwin-fastbuild/bin/internal/linker/test/multi_linker/test.sh.runfiles/build_bazel_rules_nodejs/internal/linker/test/node_modules)
+          export BAZEL_PATCH_ROOTS="${BAZEL_PATCH_ROOTS},${RUNFILES_ROOT}/${BAZEL_WORKSPACE}/${root_path}/node_modules"
+        fi
+      fi
+      # TODO: the following guards on the external workspaces may not be necessary and could be removed in the future with care
+      if [[ "${root_workspace}" ]] && [[ "${root_workspace}" != "${BAZEL_WORKSPACE}" ]]; then
```
alexeagle, line 311: ideally feels like we should make an sh_test for this script, it has a lot of conditional logic and maybe amenable to unit test

internal/node/node.bzl
```diff
@@ -40,27 +40,19 @@ def _trim_package_node_modules(package_name):
...
-    if not node_modules_root:
-        # there are no fine grained deps but we still need a node_modules_root even if it is a non-existant one
-        node_modules_root = "build_bazel_rules_nodejs/node_modules"
-    return node_modules_root
+            path = d[ExternalNpmPackageInfo].path
+            workspace = d[ExternalNpmPackageInfo].workspace
+            if path in node_modules_roots:
+                other_workspace = node_modules_roots[path]
+                if other_workspace != workspace:
+                    fail("All npm dependencies at the path '%s' must to come from a single workspace. Found '%s' and '%s'." % (path, other_workspace, workspace))
```
alexeagle, line 53: "must to come" again - I guess this is a copy pasta we ought to clean up

internal/providers/external_npm_package_info.bzl
```diff
@@ -22,28 +22,41 @@ ExternalNpmPackageInfo = provider(
...
+    if hasattr(ctx.rule.attr, "deps"):
+        for dep in ctx.rule.attr.deps:
+            if ExternalNpmPackageInfo in dep:
+                path = dep[ExternalNpmPackageInfo].path
+                workspace = dep[ExternalNpmPackageInfo].workspace
+                sources_depsets = []
+                if path in paths:
+                    path_entry = paths[path]
+                    if path_entry[0] != workspace:
+                        fail("All npm dependencies at the path '%s' must to come from a single workspace. Found '%s' and '%s'." % (path, workspace, path_entry[0]))
```
alexeagle, line 49: "must to come"

internal/providers/node_runtime_deps_info.bzl
```diff
@@ -39,22 +39,24 @@ do the same.
...
-                node_modules_root = possible_root
-            elif node_modules_root != possible_root:
-                fail("All npm dependencies need to come from a single workspace. Found '%s' and '%s'." % (node_modules_root, possible_root))
-    return node_modules_root
+            path = d[ExternalNpmPackageInfo].path
+            workspace = d[ExternalNpmPackageInfo].workspace
+            if path in node_modules_roots:
+                other_workspace = node_modules_roots[path]
+                if other_workspace != workspace:
+                    fail("All npm dependencies at the path '%s' must to come from a single workspace. Found '%s' and '%s'." % (path, other_workspace, workspace))
```
alexeagle, line 57: must to come so much copy paste!

packages/typescript/internal/build_defs.bzl
```diff
@@ -51,8 +51,6 @@ def _trim_package_node_modules(package_name):
         segments += [n]
     return "/".join(segments)
 
-# This function is similar but slightly different than _compute_node_modules_root
-# in /internal/node/node.bzl. TODO(gregmagolan): consolidate these functions
```
alexeagle, line 55: haha after all the duplication we don't want to commit to reducing it :)

alexeagle (approved):
So nice! great perseverence

WORKSPACE
```diff
@@ -57,6 +58,21 @@ yarn_install(
...
 
+yarn_install(
+    name = "npm_internal_linker_test_multi_linker",
+    package_json = "//internal/linker/test/multi_linker:package.json",
+    package_path = "internal/linker/test/multi_linker",
+    yarn_lock = "//internal/linker/test/multi_linker:yarn.lock",
+)
+
+yarn_install(
+    name = "onepa_npm_deps",
```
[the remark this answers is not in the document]
gregmagolan: onepa - 1p package "a"
onepb - 1p package "b"
🤷‍♂️

internal/linker/link_node_modules.bzl
gregmagolan: The distinction isn't important anymore since we don't treat them any differently when passing the modules to the loader as we now rely on runfiles.resolve to resolve to the right place

internal/linker/link_node_modules.bzl
gregmagolan: 🌮

internal/linker/link_node_modules.ts
gregmagolan: This can happen if we are inside a nodejs_image or a nodejs_binary is run manually.

internal/node/node.bzl
gregmagolan: 🌮

internal/providers/external_npm_package_info.bzl
gregmagolan: 🌮

internal/providers/node_runtime_deps_info.bzl
gregmagolan: 🌮

packages/typescript/internal/build_defs.bzl
gregmagolan: Yep 🤷 . This ts_library impl is no longer reducible as it does the old single workspace check & not the new single workspace / path check.

internal/node/launcher.sh
gregmagolan: Seems like it has grown complicated enough to warrant one yeah. I'll make an issue and throw in in the 4.0 milestone so we don't forget about it

Issue #2390, opened 2021-01-08
allow_js in ts_project does not contain js files in outputs for jsx inputs


noveyak:
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule:  ts_project

### Is this a regression?

I don't think so

### Description

When compiling .jsx files, tsc produces corresponding .js files but these are missing in the output of the rule

## 🔬 Minimal Reproduction

I'm hoping this is relatively easy to understand, I can create a small reproduction case but any .jsx input to ts_project with allow_js should reproduce

## 🔥 Exception or Error

No error and just no .js output

Looking at the ts_project rule, it seems like the outputs are created here

```
    if not emit_declaration_only:
        js_outs.extend(_out_paths(srcs, out_dir, root_dir, False, ".js"))
    if source_map and not emit_declaration_only:
        map_outs.extend(_out_paths(srcs, out_dir, root_dir, False, ".js.map"))
    if declaration or composite:
        typings_outs.extend(_out_paths(srcs, typings_out_dir, root_dir, allow_js, ".d.ts"))
    if declaration_map:
        typing_maps_outs.extend(_out_paths(srcs, typings_out_dir, root_dir, allow_js, ".d.ts.map"))

```

It seems like passing in `allow_js` instead of `False` to _out_paths for js_outs is enough to fix this problem but I'm not sure if there are other implications/undesirable effects of this change

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
ubuntu 16
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
bazel 3.7.0
  </code>
</pre>

**Rules_nodejs version:** 

(Please check that you have matching versions between WORKSPACE file and `@bazel/*` npm packages.)

<pre>
  <code>
3.0
  </code>
</pre>

**Anything else relevant?**

alexeagle:
This would be a good issue to contribute if you'd like.

Add a test case under /packages/typescript/test/ts_project and make the code change you suggested, and see if everything else is green. `ts_project` has good coverage so if it's green it should be safe.

mattem:
closed with #2398

stefan-dimitrov:
The fix does not completely work - I cannot see the `.js` files in `bazel-out/`. It is there in the sandbox, but the jsx syntax is not converted - remains as in source (when using `jsx: "react"` in the tsconfig.json).

I'm running rules_nodejs version 4.4.5

Relevant tsconfig options:
```json
{
[...]

PR #2398, opened 2021-01-13
fix(typescript): capture js files in outputs of ts_project if allow_js

noveyak:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

.js files are not captured as outputs of ts_project if allow_js is set to true and .js files are passed as inputs

https://github.com/bazelbuild/rules_nodejs/issues/2390

## What is the new behavior?
This change will make .js files outputs of ts_project if allow_js is true when js files are passed in

- [x] Yes
- [ ] No

Although this behavior is probably desired, this will break anyone who currently uses `ts_project` with `allow_js`  and passes in js files to generate typings and does not pass in an `out_dir` because bazel will say that file is both an input and output. This is technically the same behavior as running `tsc` on js files without an out_dir because it will refuse to overwrite the file but a change in how the ts_project rule would have worked in the past

## Other information
First attempt at contributing here, it seemed like the existing test already had a comment that implied this was the expected behavior so I mostly just modified the test. Let me know if there is anything that isn't quite right.

### DIFF
--- a/packages/typescript/internal/ts_project.bzl
+++ b/packages/typescript/internal/ts_project.bzl
@@ -668,7 +668,7 @@ def ts_project_macro(
     typing_maps_outs = []
 
     if not emit_declaration_only:
-        js_outs.extend(_out_paths(srcs, out_dir, root_dir, False, ".js"))
+        js_outs.extend(_out_paths(srcs, out_dir, root_dir, allow_js, ".js"))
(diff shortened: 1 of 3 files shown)


alexeagle (approved):
thanks!

PR #2401, opened 2021-01-14
docs(typescript): add `host` configuration to `rootDirs`

dgp1130:
A `host` root directory is necessary for TypeScript to resolve imports for a tool executed at build time with `cfg = "host"`.

This is necessary for:

```
some_rule = rule(
    implementation = _some_rule_impl,
    attrs = {
        "_binary": attr.label({
            default: "//some/nodejs_binary/tool/built/with/ts_project",
            executable: True,
            cfg = "host",
        }),
    },
)
```

Because Bazel will build the tool in the `host` configuration, meaning generated *.js files will be dropped into `bazel-out/host/bin/...`, and `tsc` needs to know to look for dependent `ts_project()` outputs in that directory.

Arguably a similar change is also necessary for `cfg = "exec"`, however files would be in `bazel-out/k8-opt-exec-*/bin`, which can't be easily hard-coded in a tsconfig, so that's a problem to be tackled separately.

## PR Checklist

## PR Type

- [X] Documentation content changes

- [ ] Yes
- [X] No

## Other information

While not a breaking change, many users are likely to have copy-pasted this content from the docs and will not easily discover this information or update their `tsconfig.json` file. This could add additional confusion due to a difference between new projects which copy-paste this content, and old projects which already did.

### DIFF
--- a/packages/typescript/internal/ts_project.bzl
+++ b/packages/typescript/internal/ts_project.bzl
@@ -388,6 +388,7 @@ def ts_project_macro(
     > "compilerOptions": {
     >     "rootDirs": [
     >         ".",
+    >         "../../bazel-out/host/bin/path/to",
     >         "../../bazel-out/darwin-fastbuild/bin/path/to",
     >         "../../bazel-out/k8-fastbuild/bin/path/to",
     >         "../../bazel-out/x64_windows-fastbuild/bin/path/to",


mattem:
Thanks, I think it's okay to point users to updated docs as an issue is discovered

mattem (approved):

PR #2416, opened 2021-01-23
feat(typescript): create a better ts_project worker

mrmeku:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

The last ts project worker I made was riddled with race bugs and overall didn't work that well.
 Additionally it was not windows compatible.
Issue Number: N/A

## What is the new behavior?
This variant of the ts_project worker uses the createWatchCompilerHost api directly rather than calling out to the typescript cli as a child process. It does not rely on program pausing and is thus windows compatible.

- [ ] Yes
- [x] No


packages/typescript/internal/ts_project.bzl
```diff
@@ -641,26 +640,19 @@ def ts_project_macro(
...
                 Label("//packages/typescript/internal/worker:worker"),
-                Label(worker_tsc_bin),
-                Label(worker_typescript_module),
                 tsconfig,
             ],
             entry_point = Label("//packages/typescript/internal/worker:worker_adapter"),
             templated_args = [
-                "$(execpath {})".format(Label(worker_tsc_bin)),
                 "--project",
                 "$(execpath {})".format(tsconfig),
```
Wenqer, line 651: According to the https://docs.bazel.build/versions/3.7.0/persistent-workers.html#implementing-persistent-workers, such configurations should be passed as startup options (arguments with @ prefix). Could you please share why you decide to pass the project argument for every worker invocation? We want to reuse the same worker instance across different packages, or such an approach has another purpose?

packages/typescript/internal/ts_project.bzl
```diff
@@ -641,26 +640,19 @@ def ts_project_macro(
                 # Users get this dependency transitively from @bazel/typescript
                 # but that's our own code, so we don't.
                 "@npm//protobufjs",
+                "@npm//typescript",
```
Wenqer, line 643: Is there any way to not remove an ability to provide a custom typescript compiler package? In our company, we are required to use [ttypescript](https://github.com/cevek/ttypescript) in some places.

If it will be too much for this PR, will you accept an external contribution for such a feature, or we just should patch it on our end?

Wenqer (commented):
Hey! Thanks for working on it! We recently started a migration to the Bazel, and this feature is the last significant blocker for us.
I left several comments/questions. I will hope they do not create much burden for you.

packages/typescript/internal/ts_project.bzl
mrmeku: There would definitely be a way to still support cusotm typescript providers. We would just need to replace `import * as ts from 'typescript` with a `const ts = require(SOME_ARGUMENT)`. I'd love to leave that as a followup if possible

packages/typescript/internal/ts_project.bzl
mrmeku: I pass it only for the standalone varaint of the worker process. I don't think its necessary, but something is wrong with my current logic if I omit it

lencioni:
When I tried out the existing implementation of supports_workers, I (and others) have run into a problem where it does not find npm dependencies. Some more information is captured here: https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1609246740191500

> For me, when I add supports_workers = True, node_modules does not resolve during build time. And run with traceResolution: true shows that they just do not exist. However, when supports_workers is False, everything works fine.
> 
> According to what I debugged so far, the node_modules folder does not exist in my setup when I run ts_project with supports_workers = True.
> 
> With enabled sandbox and supports_workers = False, everything works fine.

@mrmeku Do you think your new implementation will handle this better? Can we add a test case that covers this scenario?

packages/typescript/internal/ts_project.bzl
mrmeku: Fixed it!

mrmeku:
> When I tried out the existing implementation of supports_workers, I (and others) have run into a problem where it does not find npm dependencies. Some more information is captured here: https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1609246740191500
> 
> > For me, when I add supports_workers = True, node_modules does not resolve during build time. And run with traceResolution: true shows that they just do not exist. However, when supports_workers is False, everything works fine.
> > According to what I debugged so far, the node_modules folder does not exist in my setup when I run ts_project with supports_workers = True.
> > With enabled sandbox and supports_workers = False, everything works fine.
> 
> @mrmeku Do you think your new implementation will handle this better? Can we add a test case that covers this scenario?

Yes, I think that this type I got the logic right so that there shouldn't be any missing npm dependencies. The package does rely on typescript being a peer dependency so that's a requirement, but nothing else is required.

examples/react_webpack/BUILD.bazel
```diff
@@ -16,6 +16,9 @@ sass(
 ts_project(
     # Experimental: Start a tsc daemon to watch for changes to make recompiles faster.
     supports_workers = True,
+    # Show an example of how to run the daemon with a custom tsc.
+    typescript_module = "@npm//ttypescript",
```
alexeagle, line 20: "module" doesn't mean anything in this context. I think it should be package like https://bazelbuild.github.io/rules_nodejs/Built-ins.html#npm_package_bin

packages/typescript/internal/ts_project.bzl
```diff
@@ -506,14 +500,13 @@ def ts_project_macro(
             For example, `tsc = "@my_deps//typescript/bin:tsc"`
             Or you can pass a custom compiler binary instead.
 
-        worker_tsc_bin: Label of the TypeScript compiler binary to run when running in worker mode.
+        typescript_module: Label of the package containing all data deps of tsc.
 
-            For example, `tsc = "@my_deps//node_modules/typescript/bin/tsc"`
-            Or you can pass a custom compiler binary instead.
+            For example, `tsc = "@my_deps//typescript"`
```
alexeagle, line 505: example is wrong it says "tsc"

packages/typescript/internal/worker/worker_adapter.ts
```diff
@@ -0,0 +1,162 @@
...
+      argsFilePath = argsFilePath.slice(1)
+    }
+    const args = fs.readFileSync(argsFilePath).toString().split('\n');
+    emitOnce(args).finally(() =>
+      cachedWatchedProgram?.close()
+    );
+  }
+}
+
+main();
```
alexeagle, line 162: nit: make it testable with `if (require.main === module)` ?

packages/typescript/internal/worker/worker_adapter.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+function emitOnce(args: string[]) {
+  const watchProgram = getWatchProgram(args);
+
+  if (consolidateChangesCallback) {
+    consolidateChangesCallback();
+  }
+
+  return new Promise((res) => {
```
alexeagle, line 115: maybe it wants to be an `async` function?

alexeagle (commented):
nice! excited to try this

examples/react_webpack/BUILD.bazel
alexeagle: or maybe just `typescript = "@npm//typescript`

packages/typescript/internal/ts_project.bzl
mrmeku: Fixed!

examples/react_webpack/BUILD.bazel
mrmeku: Fixed!

packages/typescript/internal/worker/worker_adapter.ts
mrmeku: Fixed!

packages/typescript/internal/worker/worker_adapter.ts
mrmeku: Yeah, good call

packages/typescript/internal/ts_project.bzl
```diff
@@ -642,25 +635,18 @@ def ts_project_macro(
...
+                Label(typescript_package),
+                Label("//packages/typescript/internal/worker:filegroup"),
                 tsconfig,
             ],
             entry_point = Label("//packages/typescript/internal/worker:worker_adapter"),
             templated_args = [
-                "$(execpath {})".format(Label(worker_tsc_bin)),
-                "--project",
-                "$(execpath {})".format(tsconfig),
-                # FIXME: should take out_dir into account
```
alexeagle, line 655: I think these FIXMEs still apply?

examples/react_webpack/tsconfig.json
```diff
@@ -4,6 +4,11 @@
         "lib": [
             "ES2015",
             "DOM"
+        ],
+        "plugins": [
+            {
+                "transform": "transformer-module"
```
alexeagle, line 10: does this do something? I'm expecting something to declare a dependency on this package

alexeagle (approved):

packages/typescript/internal/worker/worker_adapter.ts
```diff
@@ -0,0 +1,163 @@
...
+
+async function emitOnce(args: string[]) {
+  const watchProgram = getWatchProgram(args);
+
+  if (consolidateChangesCallback) {
+    consolidateChangesCallback();
+  }
+
+
+    workerRequestTimestamp = Date.now();
```
alexeagle, line 116: did something go wrong here? I thought this function returns a Promise, if not it shouldn't be async right?

packages/typescript/internal/ts_project.bzl
mrmeku: No, we now get the same parameters that are passed to the non-worker tsproject variant. Before we were reading startup args, but now we use the worker args

packages/typescript/internal/worker/worker_adapter.ts
mrmeku: Any async function implicitly returns a promise. See the await on line 117 below as an indicator that asynchronous behavior is happening

examples/react_webpack/tsconfig.json
mrmeku: I'll try to replicate what's being done here: https://github.com/cevek/ttypescript/tree/master/packages/ttypescript-examples

And I'll add a test to make sure the outputted .js is what we expect after the transformation

packages/typescript/internal/ts_project.bzl
mrmeku: So I have the logic here right to support ttypescript, but it turns out that they have a bug in which they aren't patching all the typescript APIs that are available. I've filed a bug for them: https://github.com/cevek/ttypescript/issues/110

examples/react_webpack/tsconfig.json
mrmeku: Got blocked by https://github.com/cevek/ttypescript/issues/110. Removing the transformation example for now

PR #2428, opened 2021-01-28
fix(builtin): only generate a .tar pkg_npm output when requested

mattem:
Adds an optional attribute to the `pkg_npm` macro to generate the `.tar` target. When set, the value is used as the filename, which must end in `.tgz`. If using this, the `pkg_npm` files must include a valid `package.json` file, otherwise npm will error.

### DIFF
--- a/internal/pkg_npm/pkg_npm.bzl
+++ b/internal/pkg_npm/pkg_npm.bzl
@@ -71,9 +71,18 @@ You can pass arguments to npm by escaping them from Bazel using a double-hyphen,
 
 `bazel run my_package.publish -- --tag=next`
 
-It is also possible to use the resulting tar file file from the `.pack` as an action input via the `.tar` label:
+It is also possible to use the resulting tar file file from the `.pack` as an action input via the `.tar` label.
+To make use of this label, the `tgz` attribute must be set, and the generating `pkg_npm` rule must have a valid `package.json` file
+as part of its sources:
 
 ```python
+pkg_npm(
+    name = "my_package",
+    srcs = ["package.json"],
+    deps = [":my_typescript_lib"],
+    tgz = "my_package.tgz",
+)
+
 my_rule(
     name = "foo",
     srcs = [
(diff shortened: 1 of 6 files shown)


internal/pkg_npm/pkg_npm.bzl
```diff
@@ -109,6 +118,12 @@ You can use values from the workspace status command using curly braces, for exa
 See the section on stamping in the [README](stamping)
 """,
     ),
+    "tgz": attr.string(
```
alexeagle, line 121: can it be an `attr.output` ?

alexeagle: oh wait I see, this attribute isn't actually used? this is just here to document how the macro works? smelly

internal/pkg_npm/pkg_npm.bzl
```diff
@@ -71,9 +71,18 @@ You can pass arguments to npm by escaping them from Bazel using a double-hyphen,
...
+It is also possible to use the resulting tar file file from the `.pack` as an action input via the `.tar` label.
+To make use of this label, the `tgz` attribute must be set, and the generating `pkg_npm` rule must have a valid `package.json` file
+as part of its sources:
 
 ```python
+pkg_npm(
+    name = "my_package",
+    srcs = ["package.json"],
+    deps = [":my_typescript_lib"],
+    tgz = "my_package_tgz",
```
alexeagle, line 83: I think I'd expect this to be `my_package.tgz` instead and not add the .tgz extension later? not sure if there's a reason it needs to be this way

internal/pkg_npm/pkg_npm.bzl
mattem: 👃 Yeah, perhaps we refactor it like you suggested to be like ts_project macro, for another PR as that seems an independent change

internal/pkg_npm/pkg_npm.bzl
alexeagle, line 83: this should be `.tgz` here right?

alexeagle (approved):

PR #2431, opened 2021-01-29
[docs] example of running webpack 5 with webpack-dev-server and babel under bazel

Aghassi:
This PR intends to show a way in which you can run Webpack + Webpack Dev Server + Babel with Hot Module Reloading using rules_nodejs and `ibazel`. Instructions are in the added readme, but the short and sweet of it is:

1. Navigate to the example folder
2. `yarn start`
3.  A browser will open, navigate to `localhost:8080`, open your chrome debugger for logs
4. Modify the file `hot.js` by changing the `innerHTML` element
5. Observe the browser update the page

**NOTE:** This currently requires `webpack-dev-server` `4.0.0-beta.0` to work. I tried with `latest` and it does not see the changes in the file system. I believe this to be in relation to `chokidar` dependency being updated between the two versions.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

- [ ] Yes
- [x] No


examples/webpack5/start.sh
```diff
@@ -0,0 +1,9 @@
+#! /bin/bash
+
+set -eux
+
+readonly workspace=$(bazel info workspace)
+# The cd here is not necessary, but if you were to run in a
+# nested package in a large workspace like a monorepo you can 
+# use the cd to set your working directory so webpack runs correctly
+bazel run --run_under="cd ${workspace} && " //:serve
```
mrmeku, line 9: Bazel always runs executables at `execRoot` (see https://docs.bazel.build/versions/master/output_directories.html)
Even in a directory with nested packages, you would need only point webpack to the right entry point within the nested directory structure. This can be accomplished using webpack's `--entry` flag or via the `entry` configuration option within your webpack config

examples/webpack5/BUILD.bazel
```diff
@@ -0,0 +1,13 @@
+load("@npm_deps//webpack:index.bzl", "webpack")
+
+webpack(
+    name = "serve",
+    args = [
+        "serve",
+    ],
+    data = glob(["src/**"]) + [
```
mrmeku, line 8: Dont you want "ibazel_notify_changes", here so that ibazel won't restart the devserver process?

examples/webpack5/start.sh
mrmeku: Alternatively you could just set `chdir` on the webpack target to get the same behavior

Aghassi: So this is a weird example because I ported it form internal code where the `--run_under` shifts the directory into a nested directory. This removes the need for subprocesses like webpack to have special paths past to them. I think the problem I'm trying to solve for here is that a lot of JS tooling assumes the `cwd` is where the project is. That's not true with Bazel, and thus makes setting things up kind of fragile. I'd rather meet the tooling where it currently is, than make the tooling meet Bazel where it is.

I'm not following how the `execroot` is better for the underlying file watcher. Wouldn't watching more unnecessary files be more expensive for webpack instead of only watching the files it is concerned with?

examples/webpack5/BUILD.bazel
Aghassi: Yes I need to add that. The one issue I hit on Friday was that `ibazel` for some reason invokes the process in a way that doesn't allow it to receive fs events, thus `webpack-dev-server` won't respond to files being modified. I haven't dug into why this is the case yet. But the example above (though it doesn't provide the rebuild benefit that `ibazel` would provide, it does allow the dev server to respond to file changes in the run tree.

Aghassi:
@mrmeku @gregmagolan helped me find a bug in `watchpack`. Fixing that and adding a debounce means this works with `ibazel` now 😄

Aghassi:
Fix for watchpack arrive in this version of webpack https://github.com/webpack/webpack/compare/v5.21.1...master `5.21.1`. I'll update this to remove the patch and just use normal node installs

Aghassi:
@gregmagolan @mrmeku I added babel to the ibazel bit so we can show how you monitor `src` files, transpile them, then have webpack pick up the output. Only downside right now is that we are setting an arbitrary timeout on the polling to debounce the ibazel notifications. Probably worth working with webpack maintainers to see if they will accept or suggest a way to have the webserver respond to some sort of change from stdin since ibazel provides that. For now though, this works.

examples/webpack5/README.md
```diff
@@ -0,0 +1,21 @@
...
+Will run the dev server in a browser at port http://localhost:8080/
+Open the debugger at that address and you will see HMR console logs
+To validate HMR is working with the dev server, modify the `hot.js` file
+in `src/` and save it while the server is running. This will cause a hot reload
+and you will see more console logs output.
+
+This requires a few things
+1. Webpack must be configured to have `resolve: { symlinks: false }`.
+2. Webpack `devServer` options must have `followSymlinks` set to `true`.
+3. There is a patch for Watchpack that fixes problem with watching symlinks on macOS. Once the patch is merged and taken in through the webpack ecosystem, we will remove the patch from this example.
```
gregmagolan, line 18:
```suggestion
```

examples/webpack5/WORKSPACE
```diff
@@ -0,0 +1,34 @@
...
+    name = "examples_webpack5",
+    managed_directories = {"@npm": ["node_modules"]},
+)
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "build_bazel_rules_nodejs",
+    sha256 = "f43932c681d745e9bfa2462bce769038c794058eb1d2ae7fa287083720836cce",
+    urls = ["https://github.com/aspect-dev/rules_nodejs-builds/raw/3.1.0+fc583524/build_bazel_rules_nodejs-snapshot_builds-snapshot.tar.gz"],
```
gregmagolan, line 25: update to latest release

examples/webpack5/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "regenerator-transform": "0.14.5",
+    "regexpu-core": "4.7.1",
+    "time-fix-plugin": "2.0.7",
+    "webpack": "5.21.1",
+    "webpack-cli": "4.4.0",
+    "webpack-dev-server": "4.0.0-beta.0",
+    "workbox-webpack-plugin": "6.0.2"
+  },
+  "resolutions": {
+    "**/watchpack": "2.1.1"
```
gregmagolan, line 30: "//": "add json comment about why this resolution is needed",

Aghassi:
Just as an update, I haven't taken up more work on this because of work priorities, but @sumwatshade helped me do an ibazel dev server with Webpack 5 over here https://github.com/Aghassi/bazel-module-federation/blob/main/src/client/BUILD.bazel#L16

alexeagle:
EOL over here in rules_nodejs

PR #2433, opened 2021-01-30
refactor: move protobufjs to userspace

alexeagle:
The Labs package has this old and crufty protobufjs_ts_library rule which is not really documented.
The grpc-based one was meant to replace it.

Instead of providing a rule in @bazel/labs, just add an example showing how this can be done in userspace with a macro.
(there's also a one-liner rule to convert ProtoInfo to DefaultInfo so that the macro can interop with `proto_library`)

NOTE: this is not a breaking change because the @bazel/labs package comes with no stability guarantee.
If you're using protobufjs_ts_library you can duplicate the macro from the example to keep using it.


docs/examples.md
```diff
@@ -129,9 +129,21 @@ Example at [examples/kotlin](https://github.com/bazelbuild/rules_nodejs/tree/sta
 
 ### Protocol Buffers and gRPC
 
-Note: this is considered a "labs" feature in rules_nodejs, so support and stability are not great. gRPC is still a WIP.
+Note: this is under active development. Come chat in the #javascript channel on Slack to get the latest.
+Support and stability are not great but expected to improve.
 
-See [examples/protocol_buffers](https://github.com/bazelbuild/rules_nodejs/tree/stable/examples/protocol_buffers)
+There are many alternative implementations for protobuf and RPC.
```
mrmeku, line 135: nit: GRPC rather than RPC

examples/protobufjs/defs.bzl
```diff
@@ -0,0 +1,102 @@
+"Illustrate how to wrap pbjs and pbts from protobufjs"
+
+load("@build_bazel_rules_nodejs//:index.bzl", "js_library")
+
+# TODO switch to protobufjs-cli when its published
+# https://github.com/protobufjs/protobuf.js/commit/da34f43ccd51ad97017e139f137521782f5ef119
+load("@npm//protobufjs:index.bzl", "pbjs", "pbts")
+load("@rules_proto//proto:defs.bzl", "ProtoInfo")
+
+# these deps are needed even tho they are not declared as dependencies of
```
mrmeku, line 10: This comment needs a little bit of love :)

mrmeku (approved):

docs/examples.md
alexeagle: Actually this is intentional, there are other RPC transports that are proto compatible. From the proto team's perspective, gRPC is just one option (and for example protobufjs produces service stubs that are transport-agnostic)

examples/protobufjs/defs.bzl
alexeagle: kk

PR #2435, opened 2021-02-01
fix: relative data paths in yarn_install & npm_install when symlink_node_modules=False and package.json is not at root

gregmagolan:
This fixes `yarn_install` & `npm_install` to copy `package.json` & lock files into the external repository to a folder that corresponds to the package.json's workspace folder and the package manager is then run out that location so that relative paths to `data` files are preserved (same as they would be outside of bazel) if the `package.json` is _not_ at the root of the WORKSPACE.

If you had a nested `package.json` and were using `yarn_install` or `npm_install` with `symlink_node_modules = False,` and passing in `data` that you were referencing during install, then you would have had to use the work-around of absolute workspace paths to the `data` files. This fix will allow you to use the same relative paths that you would use if you were running `yarn` or `npm` outside of bazel in this case.

For example, if the package.json file is located at `my/nested/package.json` then
it will end up at `_/my/nested/package.json` in the external repository.

When `symlink_node_modules` is `False`, `data` files are copied into the same tree, so that relative paths in
`package.json` that refer to `data` files are preserved.

A `"postinstall": "patch-package --patch-dir patches"` in a nested `package.json` file
expects a `patches` directory relative to the `package.json` file will now work with
`symlink_node_modules = False`.

```
yarn_install(
    name = "my_nested_npm_deps",
    package_json = "//my/nested:package.json",
    yarn_lock = "//my/nested:yarn.lock",
    data = ["//my/nested:patches/jest-haste-map+24.9.0.patch"],
    symlink_node_modules = False,
)
```

Additional fix is that `data` files are now copied to external repository with `mkdir -p && cp -f` instead of `rtcx.template({})` trick. The latter is slower & does not copy over binary files correctly. We must copy `data` files and _not_ symlink them since a `package.json` file with `file:path/to/data` will fail if `path/to/data` is a symlink.

### DIFF
--- a/tools/fine_grained_deps_yarn/yarn.lock
+++ b/tools/fine_grained_deps_yarn/yarn.lock
@@ -751,7 +751,7 @@ klaw@1.3.1:
   optionalDependencies:
     graceful-fs "^4.1.9"
 
-"local-module@link:tools/npm_packages/local_module/yarn":
+"local-module@link:../npm_packages/local_module/yarn":
   version "0.0.0"
   uid ""
 
--- a/e2e/packages/npm1/package.json
+++ b/e2e/packages/npm1/package.json
@@ -8,6 +8,6 @@
     "tmp": "0.1.0"
   },
   "scripts": {
-    "postinstall": "node ./postinstall.js"
+    "postinstall": "node ../postinstall.js"
   }
 }
--- a/e2e/packages/npm2/package.json
+++ b/e2e/packages/npm2/package.json
@@ -8,6 +8,6 @@
     "tmp": "0.1.0"
   },
   "scripts": {
-    "postinstall": "node ./postinstall.js"
+    "postinstall": "node ../postinstall.js"
   }
 }
--- a/e2e/packages/yarn1/package.json
+++ b/e2e/packages/yarn1/package.json
@@ -8,6 +8,6 @@
     "tmp": "0.1.0"
   },
   "scripts": {
-    "postinstall": "node ./postinstall.js"
+    "postinstall": "node ../postinstall.js"
   }
 }
--- a/e2e/packages/yarn2/package.json
+++ b/e2e/packages/yarn2/package.json
@@ -8,6 +8,6 @@
     "tmp": "0.1.0"
   },
   "scripts": {
-    "postinstall": "node ./postinstall.js"
+    "postinstall": "node ../postinstall.js"
   }
 }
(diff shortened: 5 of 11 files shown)


tools/fine_grained_deps_yarn/yarn.lock
```diff
@@ -751,7 +751,7 @@ klaw@1.3.1:
   optionalDependencies:
     graceful-fs "^4.1.9"
 
-"local-module@link:tools/npm_packages/local_module/yarn":
```
alexeagle, line 754: yup this definitely looked wrong before, didn't match npm semantics

alexeagle (approved):

pshields:
Hey there, I've been trying out `rules_nodejs` again and ran into an issue this evening when attempting to build a `ts_project` on Windows. It looks to me like my Windows installation doesn't know how to handle the `mkdir` command added in this PR. Is it possible that this change broke Windows support in some way?

Example trace:
```
ERROR: An error occurred during the fetch of repository 'npm':
        File "C:/users/pshie/_bazel_pshie/ltf3lztm/external/build_bazel_rules_nodejs/internal/npm_install/npm_install.bzl", line 462, column 15, in 
_yarn_install_impl
                _copy_file(repository_ctx, repository_ctx.attr.package_json)
        File "C:/users/pshie/_bazel_pshie/ltf3lztm/external/build_bazel_rules_nodejs/internal/npm_install/npm_install.bzl", line 179, column 17, in 
_copy_file
                fail("mkdir -p %s failed: \nSTDOUT:\n%s\nSTDERR:\n%s" % (dirname, result.stdout, result.stderr))
Error in fail: mkdir -p _ failed:
STDOUT:

STDERR:
java.io.IOException: ERROR: src/main/native/windows/process.cc(202): CreateProcessW("mkdir" -p _): The system cannot find the file specified.       
 (error: 2)
[...]
INFO: Elapsed time: 0.197s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
```

alexeagle:
It's possible we broke this, but we do have a windows CI and have ts_project tests on there. Any idea why those didn't run the mkdir?

pshields:
Hm, looks like I can't run the tests successfully on a Windows machine either, with a similar build error complaining about `mkdir`. I'll plan to log an issue for this since it seems to affect the stable branch HEAD. I think my Windows install is fairly standard, so if it's a problem with my Windows setup, maybe I can contribute some documentation about how to avoid this issue. If it's an actual issue with rules_nodejs, I'm not sure why the CI didn't catch it, unless perhaps the appropriate directories already existed and so `mkdir` was never called?

pshields:
I've logged #2458 as a bug report related to the `mkdir` error I'm encountering on Windows.

PR #2439, opened 2021-02-01
fix(typescript): allow up to typescript 4.2, add tests for 3.7-4.1

lencioni:
This commit relaxes the semver range of the typescript peerDependency to
allow for v4.2 and earlier. I chose this version because 4.1 has been
released, and we are expecting 4.2 to be released soon. TypeScript does
not follow semver, so this could probably be a bit higher, but for now
this is an improvement over the current range.

To support this, I added some e2e tests for some of the newer versions
that are now covered in the new semver range. This should help ensure
that we don't break anyone using these versions as changes are made.


alexeagle (approved):
thank you!!

PR #2446, opened 2021-02-03
fix(docs): fix regex that replaces //packages with @bazel

alexeagle:
It was producing wrong outputs that dropped one path segment.
Also add a better error if the terser_bin attribute is missing (when not in a repo named @npm)

### DIFF
--- a/tools/stardoc/post-process-docs.js
+++ b/tools/stardoc/post-process-docs.js
@@ -7,11 +7,10 @@ const content = readFileSync(md, {encoding: 'utf8'});
 // @npm is not the required name, but it seems to be the common case
 // this reflects the similar transformation made when publishing the packages to npm
 // via pkg_npm defined in //tools:defaults.bzl
-const out = content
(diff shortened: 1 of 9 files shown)


tools/stardoc/post-process-docs.js
```diff
@@ -7,11 +7,10 @@ const content = readFileSync(md, {encoding: 'utf8'});
 // @npm is not the required name, but it seems to be the common case
 // this reflects the similar transformation made when publishing the packages to npm
 // via pkg_npm defined in //tools:defaults.bzl
-const out = content
-  .replace(/(?:@.*)*?\/\/packages\/([^:"\s]*)/g, (str, pkg) => {
-    const parts = pkg.split('/');
-    return `@npm//@bazel/${parts[parts.length - 1]}`;
-  });
+const out = content.replace(/(?:@.*)*?\/\/packages\/([^/:"\s]*)/g, (str, pkg) => {
```
alexeagle, line 10: only change here was adding slash to the not-next characters group
`[^:"\s]` -> `[^/:"\s]`

mattem (approved):

PR #2448, opened 2021-02-04
Ensure using the nodejs source-map support

joeljeske:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes
- [X] Other... Please describe:

Performance enhancement

Issue Number: N/A

## What is the new behavior?

When using libraries like js-dom that provide a browser-like api inside nodejs, the source map support can be tricked into thinking that it is running in a browsers, which will attempt to use `XMLHttpRequest` instead of `require('fs')`. Depending on implementation of the XMLHttpRequest (presumably loaded by js-dom or alternative), this can be much slower to load source maps from disk, causing a higher latency run of nodejs. 

Even if the user is using a browser-like experience within nodejs, the nodejs implementation of source map support should always be superior, thus we should hint to the source map support to always use the nodejs fs module to load source maps. 

For me, I noticed this slow down in test runs, and found an 80% speedup across some runs. Other's mileage may vary.

- [ ] Yes
- [X] No

### DIFF
--- a/internal/node/require_patch.js
+++ b/internal/node/require_patch.js
@@ -491,7 +491,7 @@ module.constructor._resolveFilename =
 try {
   const sourcemap_support_package = path.resolve(
       process.cwd(), '../build_bazel_rules_nodejs/third_party/github.com/source-map-support');
-  require(sourcemap_support_package).install();
+  require(sourcemap_support_package).install({ environment: 'node' });
 } catch (_) {
   log_verbose(`WARNING: source-map-support module not installed.
     Stack traces from languages like TypeScript will point to generated .js files.`);


mattem:
Thanks. Windows failure looks like a flake, rebase to rerun?

alexeagle:
Interesting, the file you changed is no longer used by default since the `--bazel_patch_module_resolver` flag defaults to False. That means we should probably document how users should use `--require source-map-support` instead to get stack traces pointing back to original sources. (Or possibly do it for them in the launcher.sh, since we vendored the source-map-support package and always have it available)

@joeljeske do you have any time for that?

alexeagle (approved):

PR #2449, opened 2021-02-05
feat: add generate_local_modules_build_files flag to yarn_install and npm_install rules

mistic:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

The current expected behaviour was the one introduced by https://github.com/bazelbuild/rules_nodejs/pull/2330
However there was a regression and after reanalysing it I believe we will be better served by hide that feature behind a flag other than try to proceed with it in an auto discoverable logic.

## What is the new behavior?
With that new behaviour if you set `generate_local_modules_build_files=False` in both `yarn_install` or `npm_install` rules, BUILD files for `symlinked node_modules` will be also symlinked instead of auto generated.

- [ ] Yes
- [x] No


mistic:
@alexeagle do you think we can get this fix in for the next minor?

alexeagle:
It depends if it's merged before then 😊
It seems kinda hacky so maybe will need some discussion during review

mistic:
👍 I've kept the previous feature logic, just removed the current check to automatically determine if the feature should run (as it is not currently working as expected) and just used a variable to control it from the rule instead

mistic:
Solved the conflicts with the upstream

mattem (approved):

internal/npm_install/npm_install.bzl
```diff
@@ -44,6 +44,29 @@ repository so all files that the package manager depends on must be listed.
...
         default = {},
     ),
+    "generate_local_modules_build_files": attr.bool(
+        default = True,
+        doc = """Enables the BUILD files auto generation for local modules installed with `file:` (npm) or `link:` (yarn)
+
+When using a monorepo it's common to have modules that we want to use locally and
+publish to an external package repository. This can be achieved using a `js_library` rule
+with a `package_name` attribute defined inside the local package `BUILD` file. However,
+if the project relies on the local package dependency with `file:` (npm) or `link:` (yarn) to be used outside Bazel, this
```
ljharb, line 54: npm also supports `link:` and yarn also supports `file:`, so i'm not sure why these parentheticals are here?

PR #2454, opened 2021-02-08
docs(rollup): mention including the rollup library in devDependencies

pshields:
This is a possible docs update to mention including `rollup` itself as a dev dependency, since I ran into errors when I tried to use the rollup rules without having the `rollup` library itself installed.

### DIFF
--- a/docs/Rollup.md
+++ b/docs/Rollup.md
@@ -17,7 +17,7 @@ The Rollup rules run the [rollup.js](https://rollupjs.org/) bundler with Bazel.
 
 ## Installation
 
-Add the `@bazel/rollup` npm package to your `devDependencies` in `package.json`.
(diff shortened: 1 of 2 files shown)


mattem:
Thanks for the PR. That file is generated, are you able to make the same change in the source so it doesn't get overwritten when the docs are regenerated on the next release? The source file is: https://github.com/bazelbuild/rules_nodejs/blob/stable/packages/rollup/install.md

pshields:
> Thanks for the PR. That file is generated, are you able to make the same change in the source so it doesn't get overwritten when the docs are regenerated on the next release? The source file is: https://github.com/bazelbuild/rules_nodejs/blob/stable/packages/rollup/install.md

Sure, thanks for the pointer - I didn't realize the markdown files were also generated from elsewhere in the repo. I made the change there, too.

alexeagle (approved):

PR #2461, opened 2021-02-11
docs: update release instructions to include version change in docs selector

mattem:


alexeagle (commented):
There are a bunch of subtle and untested ordering dependencies in the release process. I'll try this method today and if it works then will approve/merge

mattem:
Yeah this isn't really ideal, but it might be enough to at least work around the issue until we find any time for doc pages

alexeagle:
actually we don't plan to have separate docsite for 3.1 vs 3.0 so I'll just change it to 3.x

mattem:
In that case, I'll update https://github.com/bazelbuild/rules_nodejs/pull/2460

PR #2465, opened 2021-02-16
chore: bump clang-format version

mattem:
Bumps the clang-format version used for formatting. While doing the docs work, the previous version does not handle the `??` operator within arrow functions correctly and produces broken code after formatting, this version fixes that.


alexeagle (approved):
We should switch to prettier at some point, maybe if we care enough

PR #2475, opened 2021-02-18
fix: use ':' instead of '=' for esbuild 'external' argument

mattsoulanille:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

esbuild passes externals as `--external=the_dependency` instead of `--external:the_dependency` [as documented](https://esbuild.github.io/api/#external).
Issue Number: N/A

## What is the new behavior?
Pass externals as `--external:the_dependency`.

- [ ] Yes
- [x] No

## Other information
I named the test directory `external-flag` since I thought `external` might be confused with external dependencies.

The test uses `fs` as the external dependency and runs a Jasmine test to check if `console.log(fs)` appears in the bundle. This isn't strictly necessary since the bundle will fail to build if `fs` is not available, but I thought it would be best to have a test target in the PR. It might be more correct to check for `var fs = __toModule(require("fs"));`, but I think that may change depending on esbuild's implementation. Maybe it's safe to check for `require("fs")`. What do you think?

### DIFF
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -50,7 +50,7 @@ def _esbuild_impl(ctx):
     args.add_joined(["--log-level", "info"], join_with = "=")
     args.add_joined(["--metafile", metafile.path], join_with = "=")
     args.add_all(ctx.attr.define, format_each = "--define:%s")
-    args.add_all(ctx.attr.external, format_each = "--external=%s")
+    args.add_all(ctx.attr.external, format_each = "--external:%s")
 
     # disable the error limit and show all errors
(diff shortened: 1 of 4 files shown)


mattem (approved):
Thanks!
There are some golden file tests in there already that would fail if the implementation within esbuild changed, which I had the same thought about regarding correctness, however as the version of esbuild is pinned in the tests here, and if it was bumped we'd have to manually verify the changes to the golden file, I deemed it okay.

In this case, like you say the bundle would fail if external wasn't set, so I think this test is fine, we can expand it later as needed.

PR #2476, opened 2021-02-18
fix(esbuild): add link_workspace_root for workspace absolute imports

### LINKED ISSUE #2474: esbuild can't resolve workspace absolute imports
# 🐞 bug report

### Affected Rules

- `ts_library`
- `esbuild`

### Is this a regression?

No.

### Description

I may be configuring something wrong, but `esbuild` can't seem to find transitive `ts_library` dependencies within the workspace.

## 🔬 Minimal Reproduction

I have a minimal reproduction at https://github.com/shosti/bazel-rulesjs-repro.  There's a top-level `ts_library` (`:a`) that includes `lib/b.ts` as a dependency (imported as `"myworkspace/lib/b"`). Running `bazel build //:a` succeeds, which indicates that the import paths are correct as far as `tsc` is concerned.

There's also an `esbuild` bundle (`:bundle`) that depends on `:a`. Running `bazel build //:bundle` fails with an import error message.

## 🔥 Exception or Error

<pre><code>
 > ../../../../../execroot/myworkspace/bazel-out/k8-fastbuild/bin/a.mjs: error: Could not resolve "myworkspace/lib/b" (mark it as external to exclude it from the bundle)
    1 │ import foo from "myworkspace/lib/b";
      ╵                 ~~~~~~~~~~~~~~~~~~~

1 error
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Ubuntu 20.10
  </code>
</pre>

<pre>
  <code>
Build label: 3.5.0
Build target: bazel-out/k8-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Wed Sep 2 20:11:46 2020 (1599077506)
Build timestamp: 1599077506
Build timestamp as int: 1599077506
  </code>
</pre>

**Rules_nodejs version:** 

<pre>
  <code>
3.2.0
  </code>
</pre>

**Anything else relevant?**

jbedard:
Fixes #2474

### DIFF
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -15,6 +15,9 @@ def _esbuild_impl(ctx):
     # how to resolve custom package or module names
     path_alias_mappings = dict()
 
+    if (ctx.attr.link_workspace_root):
+        path_alias_mappings.update(generate_path_mapping(ctx.workspace_name, "."))
+
     for dep in ctx.attr.deps:
         if JSEcmaScriptModuleInfo in dep:
             deps_depsets.append(dep[JSEcmaScriptModuleInfo].sources)
@@ -152,6 +155,10 @@ and cjs when platform is node. If performing code splitting, defaults to esm.
 
 See https://esbuild.github.io/api/#format for more details
         """,
+        ),
+        "link_workspace_root": attr.bool(
+            doc = """Link the workspace root to the bin_dir to support absolute requires like 'my_wksp/path/to/file'.
+    If source files need to be required then they can be copied to the bin_dir with copy_to_bin.""",
         ),
         "minify": attr.bool(
             default = False,
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -11,6 +11,7 @@ module.exports = {
         'concatjs',
         'create',
         'cypress',
+        'esbuild',
         'examples',
         'jasmine',
         'labs',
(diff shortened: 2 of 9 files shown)


packages/esbuild/esbuild.bzl
```diff
@@ -15,6 +15,9 @@ def _esbuild_impl(ctx):
     # how to resolve custom package or module names
     path_alias_mappings = dict()
 
+    if (ctx.attr.link_workspace_root):
+        path_alias_mappings.update(generate_path_mapping(ctx.workspace_name, "/".join(ctx.build_file_path.split("/")[:-1])))
```
mattem, line 19: This would add:

```
"build_bazel_rules_nodejs": [
    "packages/esbuild/test/typescript/index.mjs",
    "packages/esbuild/test/typescript"
],
"build_bazel_rules_nodejs/*": [
    "packages/esbuild/test/typescript/*"
]
```

to the path mappings. Wouldn't these need to point to the root of the workspace, which would be `'./*'` rather than the root of where the `esbuild` file is?

jbedard: Yeah you're right... the import examples in the test should be `build_bazel_rules_nodejs/packages/esbuild/test/typescript/...subfolder...`. I think that's actually what I originally had but was getting errors and confused (all because of a silly typo), and this is what came out of the confusion in the end...

jbedard: Updated, I think that's right now?

jbedard: For the test this now creates:
```
"paths": {
            "build_bazel_rules_nodejs": [
                "./index.mjs",
                "."
            ],
            "build_bazel_rules_nodejs/*": [
                "./*"
            ]
        }
```

mattem: Yeah that looks correct, nice one!

mattem (approved):

PR #2477, opened 2021-02-18
fix(esbuild): add --preserve-symlinks flag by default and npm module mappings

mattem:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

It was possible for esbuild to break out of the sandbox and find the `.ts` file inputs, this could be seen if the `.ts` file was generated, as it would live in bazel-out/

This also fixes an issue where external npm modules may not have gotten resolved correctly when bundling

Issue Number: N/A

## What is the new behavior?
Bumps the tested version of esbuild and adds the `--preserve-symlinks` flag by default. 

- [ ] Yes
- [ ] No
- [x] kinda

This will require uses to bump the version of esbuild used in their projects

### DIFF
--- a/examples/esbuild/WORKSPACE
+++ b/examples/esbuild/WORKSPACE
@@ -25,12 +25,12 @@ http_archive(
     urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/3.2.0/rules_nodejs-3.2.0.tar.gz"],
 )
 
-_ESBUILD_VERSION = "0.8.34"
+_ESBUILD_VERSION = "0.8.48"
 
 http_archive(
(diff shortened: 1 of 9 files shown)


commitlint.config.js
```diff
@@ -11,6 +11,7 @@ module.exports = {
         'concatjs',
         'create',
         'cypress',
+        'esbuild',
```
jbedard, line 14: I have this in its own commit in #2476 if we feel like merging that first 🤷

mattem: Yup can rebase

alexeagle (approved):

PR #2494, opened 2021-02-28
Update esbuild version in docs

bduffany:
* Sync the esbuild version in the docs so that it matches the deps file. The version mentioned in the README did not support the `--preserve-symlinks` flag, which looks like it is required by the latest version of these esbuild rules (I got a build error stating that `--preserve-symlinks` was not a valid flag)
* Add reminders to keep these docs in sync
* Also add reminders to bump SHAs -- I tried bumping the esbuild version, and forgot to update the SHAs, but this resulted in Bazel silently using the old archive. (BTW, instead of using the `_ESBUILD_VERSION` variable, I think it might be better to hard-code the full URLs without string interpolation, so that the SHA is right in your face as you're updating the URL, and you immediately think "ah right, I need to update the SHA too." -- WDYT?)

### DIFF
--- a/packages/esbuild/_README.md
+++ b/packages/esbuild/_README.md
@@ -17,8 +17,9 @@ yarn add -D @bazel/esbuild
 
 Add an `http_archive` fetching the esbuild binary for each platform that you need to support. 
 
+<!-- IMPORTANT: Keep this in sync with ./esbuild_repo.bzl -->
 ```python
-_ESBUILD_VERSION = "0.8.48"
+_ESBUILD_VERSION = "0.8.48"  # reminder: update SHAs below when changing this value
 http_archive(
     name = "esbuild_darwin",
     urls = [
--- a/packages/esbuild/_README.md
+++ b/packages/esbuild/_README.md
@@ -18,7 +18,7 @@ yarn add -D @bazel/esbuild
 Add an `http_archive` fetching the esbuild binary for each platform that you need to support. 
 
 ```python
-_ESBUILD_VERSION = "0.8.34"
+_ESBUILD_VERSION = "0.8.48"
 http_archive(
     name = "esbuild_darwin",
     urls = [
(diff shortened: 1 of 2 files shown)


packages/esbuild/_README.md
```diff
@@ -17,16 +17,17 @@ yarn add -D @bazel/esbuild
 
 Add an `http_archive` fetching the esbuild binary for each platform that you need to support. 
 
+<!-- IMPORTANT: Keep this in sync with ./esbuild_repo.bzl -->
```
mattem, line 20: Is this comment ignored by the docs generation?

mattem:
>  I think it might be better to hard-code the full URLs without string interpolation, so that the SHA is right in your face as you're updating the URL

Perhaps a dict with platform -> sha? Or like you commented on the other diff, have some tooling to update.
Also, while I agree we should be providing this as an example, the end user is free to express the esbuild repos in their workspace how they want.

packages/esbuild/_README.md
bduffany: Good point, removed.

bduffany:
> Perhaps a dict with platform -> sha? Or like you commented on the other diff, have some tooling to update.
> Also, while I agree we should be providing this as an example, the end user is free to express the esbuild repos in their workspace how they want.

Sounds good, I left as-is, but kept the reminder about updating SHAs when bumping the version

bduffany:
Just realized something -- could the README instead point people to call the `esbuild_dependencies` function in their WORKSPACE? (and maybe accept a param like `platforms = ["darwin", "linux"]`

mattem:
We specifically don't publish the `esbuild_repo.bzl` file in the npm distro as referencing it in a WORKSPACE would cause an eager fetch of all npm dependencies.

mattem (approved):

mattem:
If you wanted something locally, https://gist.github.com/mattem/61bc215c60e03b72ccc3b879262de258#file-es_build-bzl-L96 which will fetch the binary for the platform, but it would suffer from the same eager fetching problem if we distribute it as part of the npm package.

PR #2495, opened 2021-03-02
feat(examples): update angular_bazel_architect to version 11

alan-agius4:


mattem (approved):
Awesome 🚀 Reminds me we should update the other [Angular example](https://github.com/bazelbuild/rules_nodejs/tree/stable/examples/angular) too

alan-agius4:
Yeah, If I find sometime I'll try to do them in the coming days.

examples/angular_bazel_architect/src/app/app-routing.module.ts
```diff
@@ -4,5 +4,5 @@ import {RouterModule, Routes} from '@angular/router';
 
 const routes: Routes = [];
 
-@NgModule({imports: [RouterModule.forRoot(routes)], exports: [RouterModule]})
+@NgModule({imports: [RouterModule.forRoot(routes, { relativeLinkResolution: 'legacy' })], exports: [RouterModule]})
```
saxicek, line 7: I think this should be reverted as there are no routes affected by the related bug IMO.

alan-agius4: @saxicek, agreed. Mind opening a PR? thanks.

saxicek: Ok, opened https://github.com/bazelbuild/rules_nodejs/pull/2530.

PR #2506, opened 2021-03-05
fix(esbuild): set correct base url when rule is at root

### LINKED ISSUE #2503: esbuild: error: Could not resolve "react" (mark it as external to exclude it from the bundle)
This issue follows [this issue](https://github.com/bazelbuild/rules_nodejs/issues/2501).

I was able to get it past the `--preserve-symlinks` error, but now I have a new error and I'm not sure what to do about it.

I feel like the npm modules should be able to be bundled. It kind of defeats the purpose for me if I can't bundle the npm modules.

### Error

```
INFO: Invocation ID: 3c7403f6-0e2d-4c23-90b5-4da687d6b7c5
INFO: Analyzed target //:bundle (2 packages loaded, 8 targets configured).
INFO: Found 1 target...
ERROR: workspace/BUILD.bazel:27:8: Bundling Javascript index.mjs [esbuild] failed (Exit 1) esbuild failed: error executing command external/esbuild_darwin/bin/esbuild --bundle bazel-out/darwin-fastbuild/bin/index.mjs --sourcemap --preserve-symlinks '--platform=browser' '--target=es2015' '--log-level=info' ... (remaining 6 argument(s) skipped)

 > bazel-out/darwin-fastbuild/bin/index.mjs: error: Could not resolve "react" (mark it as external to exclude it from the bundle)
    1 │ import React from "react";
      ╵                   ~~~~~~~

 > bazel-out/darwin-fastbuild/bin/index.mjs: error: Could not resolve "react-dom" (mark it as external to exclude it from the bundle)
    2 │ import ReactDOM from "react-dom";
      ╵                      ~~~~~~~~~~~

2 errors
Target //:bundle failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 0.516s, Critical Path: 0.20s
INFO: 2 processes: 2 internal.
```

### Things I tried

(reference my setup [here](https://github.com/bazelbuild/rules_nodejs/issues/2501))

* created `es5_sources` filegroup and used that in `deps` instead of `:lib`
* added `@npm//react` and `@npm//react-dom` to `deps`
* changed the `devmode_` and `prodmode_` module and target to `esnext` and `es2015`, respectively

mattem:
When the esbuild rule is in the root `BUILD.bazel` file, `rule_path` is empty, splitting on this results in a list with one empty string, which is then translated into one path segment. When `rule_path` is empty, the rule is already in the correct base path.

This also adds an npm dependency to the example to exercise node_module resolution when bundling.

closes #2503

### DIFF
--- a/examples/esbuild/package.json
+++ b/examples/esbuild/package.json
@@ -7,6 +7,9 @@
         "tslib": "1.9.0",
         "typescript": "3.5.3"
     },
+    "dependencies": {
+        "chalk": "4.1.0"
+    },
     "scripts": {
         "test": "bazel test //..."
     }
--- a/examples/esbuild/package-lock.json
+++ b/examples/esbuild/package-lock.json
@@ -1,219 +1,6 @@
 {
-    "name": "esbuild",
-    "lockfileVersion": 2,
     "requires": true,
-    "packages": {
-        "": {
-            "devDependencies": {
-                "@bazel/esbuild": "^3.2.1",
-                "@bazel/typescript": "^3.2.1",
-                "@types/node": "12.6.3",
-                "tslib": "1.9.0",
-                "typescript": "3.5.3"
-            }
-        },
-        "node_modules/@bazel/esbuild": {
-            "version": "3.2.1",
-            "resolved": "https://registry.npmjs.org/@bazel/esbuild/-/esbuild-3.2.1.tgz",
-            "integrity": "sha512-8Y2Vd9KMxWfL9KzF9qbrUGHJZiJsee6xpT+dUCd29zYJA9Ddu4WCeNtS24Ejbm1y3NUgGPXh1nLgTQ/yIfCYng==",
-            "dev": true
-        },
-        "node_modules/@bazel/typescript": {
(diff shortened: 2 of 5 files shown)


examples/esbuild/package.json
```diff
@@ -7,6 +7,9 @@
         "tslib": "1.9.0",
         "typescript": "3.5.3"
     },
+    "dependencies": {
+        "chalk": "4.1.0"
```
jbedard, line 11: Should a package-lock.json also be changing...?

alexeagle: yeah looks like we should be using `npm ci` when we install deps for examples

alexeagle (approved):
LGTM other than the problem @jbedard points out

examples/esbuild/package.json
mattem: Nice catch @jbedard, I should have regenerated the lock file, will update. 
Currently we can't use `npm ci` during install of the deps in integration tests as the package replacement gets in the way, or is that not what you meant?

PR #2508, opened 2021-03-06
fix(rollup): include externals config in worker cache key

jbedard:
This works around https://github.com/rollup/rollup/issues/3874 by not using the old cache at all when this property changes. Seems weird we have to work around such a thing but I sure wasted a lot of time wondering why changes in the config file weren't being picked up by rollup.


mattem (approved):

alexeagle (approved):
nice find! good job digging to the root cause and documenting the workaround with a link to upstream issue 😎

PR #2509, opened 2021-03-06
feat(esbuild): add max_threads setting to limit number of threads used

mattem:
Adds a `max_threads` to allow limiting the max number of threads used by esbuild. Internally this sets the `GOMAXPROCS` environment variable.

This can be useful if running many esbuild rule invocations in parallel, which has the potential to cause slowdown.

Adds a simple test case to demonstrate the usage, however it's hard to assert.


packages/esbuild/esbuild.bzl
```diff
@@ -168,6 +174,13 @@ See https://esbuild.github.io/api/#format for more details
             doc = """Link the workspace root to the bin_dir to support absolute requires like 'my_wksp/path/to/file'.
     If source files need to be required then they can be copied to the bin_dir with copy_to_bin.""",
         ),
+        "max_threads": attr.string(
```
jbedard, line 177: Any reason to not make this an `int`?

mattem: Yeah I guess it could be, it's got to be converted to a string for `env` then, but it might make a nicer API. I don't have a strong opinion either way

jbedard: I figured converting int => int-string is better then simply sending "asdlfkj" through 🤷

packages/esbuild/test/node/BUILD.bazel
```diff
@@ -0,0 +1,39 @@
...
+    env = {
+        "ESBUILD_TEST": "YES",
+    },
+    stdout = "out.txt",
+    tool = ":bin",
+)
+
+generated_file_test(
+    name = "test",
+    src = "out.golden.txt",
```
alexeagle, line 37: sometime we should make a version of generated_file_test where the expected content can be inline in the BUILD file, in this case you'd rather keep the "YES" next to the other "YES"

alexeagle (approved):

jbedard (approved):

packages/esbuild/test/node/BUILD.bazel
mattem: Yes

PR #2512, opened 2021-03-07
fix: tsconfig validator fails on chained tsconfig references

bduffany:
Fixes an issue (#2513) that the validator was always treating `extends` as a plain old file instead of reading `TsConfigInfo` to get transitive deps. See the added test case, which shows a realistic example that would fail to validate due to missing `tsconfig.base.json` (referenced by `tsconfig.node.json`, which itself is referenced by `main/tsconfig.json`)

(Note: one workaround in case anyone stumbles upon this PR is to use `validate = False`).


alexeagle:
Thanks for the help, do you think you'll be able to get the tests green?

bduffany:
Fixed most of the failing tests. There's a few that are failing (on Windows only)... would it be possible to re-run them?

bduffany:
Friendly ping -- would it be possible to re-run the Windows tests to see whether the failures are specific to this PR?

alexeagle:
Hey sorry for the delay, juggling lots of things lately. You can always come ping me on slack if we're mostly done with a PR and a bit of attention is needed.

Yes those windows failures are legit. Actually they are failures in non-sandboxed mode, which you could reproduce on any platform, it's just that Windows is always in that mode.

What's happening is the typescript compiler does it's implicit look-around-the-disk to find ambient typings it can add to the program. When it does this, it finds typings we *don't* want to be included, they introduce conflicts with other typings. This is why we have `types: []` in most tsconfig's to disable this feature and we are explicit about the typings we want in each compile.

alexeagle:
@googlebot I consent.

alexeagle (approved):
thanks!

bduffany:
@alexeagle Thanks for finishing this off!

Issue #2482, opened 2021-02-21
Support prefix stripping in js_library


aaliddell:
# 🚀 feature request

### Relevant Rules

`js_library`

### Description

It would be useful to have the ability to strip a prefix from the require path when using a `js_library`.

For example, with the following project layout:

```
- WORKSPACE
- BUILD
- module/
-- file.js
```

In BUILD, we have a `js_library` declared like so:

```starlark
js_library(
    name = "some_library",
    srcs = ["module/file.js"],
    package_name = "module",
)
```

When using the `js_library`, you would presently have to do:

```js
require('module/module/file.js')
```

### Describe the solution you'd like

A mechanism to tell `js_library` a path that should be stripped/hidden, such that libraries can be within subdirectories without that appearing in the require path. Such as:

```starlark
js_library(
    name = "some_library",
    srcs = ["module/file.js"],
    package_name = "module",
    strip_prefix = "module",  # Example name
)
```

With this, the require becomes:

```js
require('module/file.js')
```

I think this could be achieved through changing the `path` on the `LinkablePackageInfo` to point to the directory, rather than the local package root:

https://github.com/bazelbuild/rules_nodejs/blob/2c2cc6eec42b0f4f805fb3063920107b78f821fd/internal/js_library/js_library.bzl#L225-L231

e.g. line 226 becomes:

```starlark
path = "/".join([p for p in [ctx.bin_dir.path, ctx.label.workspace_root, ctx.label.package, ctx.attr.strip_prefix] if p])
```

This would be a backwards compatible change, since in the absence of the `strip_prefix` attr being set, the behaviour would be unchanged. This is similar to `includes` on `cc_library` or `strip_import_prefix` on `proto_library`.

### Describe alternatives you've considered

- Tried patching the generated `LinkablePackageInfo` in a subsequent rule and forwarding other providers, but this doesn't work since the old provider is picked up by an aspect and results in 'conflicting mapping'. This would also break as soon as a new provider is added to rules_nodejs.
- Putting the BUILD file in the `module` folder is not feasible, as that folder is an output from prior rules rather than a static folder as in the above example.

mattem:
Closed via #2519

PR #2519, opened 2021-03-09
Add strip_prefix attr to js_library

aaliddell:
Solution to #2482

Open questions:
- What is the preferred name for this attr? I have called it `strip_path` but perhaps you have a preference for a different name.
- How would you like this tested? A new directory in `/e2e` or is there an existing one that it is suitable to add to?

## PR Checklist

- [x] Docs have been added / updated (for bug fixes / features): Docstrings and attr doc have been updated

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

See issue Number: #2482

## What is the new behavior?

The `LinkablePackageInfo` provider can now point into a subdirectory without that subdirectory being present in the import, as detailed in the issue above.

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/internal/js_library/js_library.bzl
+++ b/internal/js_library/js_library.bzl
@@ -228,8 +228,24 @@ def _impl(ctx):
 
     if ctx.attr.package_name:
         path = "/".join([
-            p for p in [ctx.bin_dir.path, ctx.label.workspace_root, ctx.label.package, ctx.attr.strip_prefix] if p
+            p for p in [ctx.bin_dir.path, ctx.label.workspace_root, ctx.label.package] if p
         ])
+
+        # Strip a prefix from the package require path
+        if ctx.attr.strip_prefix:
+            path += "/" + ctx.attr.strip_prefix
+
+            # Check that strip_prefix contains at least one src path
+            check_prefix = "/".join([p for p in [ctx.label.package, ctx.attr.strip_prefix] if p])
+            prefix_contains_src = False
+            for file in all_files:
+                if file.short_path.startswith(check_prefix):
+                    prefix_contains_src = True
+                    break
+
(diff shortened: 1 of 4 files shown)


internal/js_library/js_library.bzl
```diff
@@ -223,7 +227,9 @@ def _impl(ctx):
     ]
 
     if ctx.attr.package_name:
-        path = "/".join([p for p in [ctx.bin_dir.path, ctx.label.workspace_root, ctx.label.package] if p])
+        path = "/".join([
+            p for p in [ctx.bin_dir.path, ctx.label.workspace_root, ctx.label.package, ctx.attr.strip_path] if p
```
alexeagle, line 231: do you think we can do any decent error messaging if you have a typo or misunderstanding, and the strip_prefix doesn't match against one or more of the srcs?

alexeagle (commented):
Hi @aaliddell !

+1 for your justification about why such a feature is needed.

I think I'd name it `strip_prefix`, having worked with `pkg_tar` lately. At minimum we want to convey that the part of the path that's stripped is a leading part.

To test this I think you can just add an example under `internal/js_library/test/strip_prefix` that looks like your sample in #2482 and assert that a program can require from the path.

internal/js_library/js_library.bzl
aaliddell: I have added a check that at least one source is within the stripped path. However, an alternative would be to fail if any src is not in the stripped path. Any preference?

aaliddell:
I've renamed to `strip_prefix` and added a test in the proposed location. The failure in buildkite seems unrelated to the changes?

alexeagle (approved):
Thanks!

PR #2528, opened 2021-03-14
feat(esbuild): add 'sourcemap' option to configure sourcemap generation

mattsoulanille:
## PR Checklist

I'm pretty sure docs are auto-generated from the `doc` kwarg of attributes. If not, then docs have not been updated.

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently, there is no way to have esbuild include inlined sourcemaps in the output bundle.

Issue Number: N/A

## What is the new behavior?
By passing `sourcemap_inline = True` to an `esbuild` rule named `bundle`, sourcemaps will be generated inline in `bundle.js` as well as externally in `bundle.js.map`.

- [ ] Yes
- [x] No

## Other information

The motivation behind this feature is to make it easier to use esbuild as part of a development server. Inline sourcemaps are easier to correctly set up than external sourcemaps.

The argument is named `sourcemap_inline` instead of `sourcemap` since it can only add inline sourcemaps to the bundle. We can't let users directly change the `sourcemap` argument to esbuild, since that argument can turn off the generation of the external sourcemap file, which is a required output of the rule.


mattsoulanille:
Actually, this doesn't solve the sourcemaps problem I was having with sourcemaps on my dev server, and it doesn't seem to be necessary. Setting `sourcemap_content = True` is what I needed.

I'm still going to open this PR in case this feature is useful, but feel free to close it.

packages/esbuild/esbuild.bzl
```diff
@@ -54,8 +54,13 @@ def _esbuild_impl(ctx):
     args = ctx.actions.args()
 
     args.add("--bundle", entry_point.path)
-    args.add("--sourcemap")
     args.add("--preserve-symlinks")
+
+    if ctx.attr.sourcemap_inline:
```
mattem, line 59: Do you think `sourcemap_inline` implies we shouldn't be generating the external souremaps too? 
Perhaps we want a `sourcemap` attr to be an enum? eg `sourcemap = "inline"` that we can resolve to the right attributes in the macro for output files?

mattsoulanille: I don't think we're allowed to not generate external sourcemaps, since [I think the rule expects them to be generated when calling esbuild](https://github.com/bazelbuild/rules_nodejs/blob/e7d9614e65335eea42c390640c81cf20ba1c23f8/packages/esbuild/esbuild.bzl#L289), but I'm also not that familiar with writing rules. If the sourcemap file isn't actually required by Bazel, we could have a `sourcemap` enum that directly sets the esbuild [`sourcemap` arg](https://github.com/bazelbuild/rules_nodejs/blob/e7d9614e65335eea42c390640c81cf20ba1c23f8/packages/esbuild/esbuild.bzl#L289).

If we do always need to generate sourcemaps, we could restrict the `sourcemap` enum to `external` or `both`, which might be more intuitive than `sourcemap_inline`. What do you think?

mattem: We can change that by having a conditional on around here: https://github.com/bazelbuild/rules_nodejs/blob/e7d9614e65335eea42c390640c81cf20ba1c23f8/packages/esbuild/esbuild.bzl#L95
and tweak the macro too to provide the correct `output_map` name.

Bazel itself has no interest in the sourcemaps, but we will need to declare that we are going to generate one so it knows to expect the output file, that's essentially what the line in the link above is doing.

Therefore we can also include `inline` in the sourcemap enum, something like this would do it I think:
```
diff --git a/packages/esbuild/esbuild.bzl b/packages/esbuild/esbuild.bzl
index f5d9aa59..66299e5c 100644
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -54,7 +54,12 @@ def _esbuild_impl(ctx):
     args = ctx.actions.args()
 
     args.add("--bundle", entry_point.path)
-    args.add("--sourcemap")
+
+    if len(ctx.attr.sourcemap) > 0:
+        args.add_joined(["--sourcemap", ctx.attr.sourcemap], join_with = "=")
+    else:
+        args.add("--sourcemap")
+
     args.add("--preserve-symlinks")
     args.add_joined(["--platform", ctx.attr.platform], join_with = "=")
     args.add_joined(["--target", ctx.attr.target], join_with = "=")
@@ -87,8 +92,13 @@ def _esbuild_impl(ctx):
         args.add_joined(["--outdir", js_out.path], join_with = "=")
     else:
         js_out = ctx.outputs.output
+        outputs.append(js_out)
+
         js_out_map = ctx.outputs.output_map
-        outputs.extend([js_out, js_out_map])
+        if ctx.attr.sourcemap != "inline":
+            if js_out_map == None:
+                fail("output_map must be specified if sourcemap is not set to 'inline'")
+            outputs.append(js_out_map)
 
         if ctx.attr.format:
             args.add_joined(["--format", ctx.attr.format], join_with = "=")
@@ -244,6 +254,10 @@ See https://esbuild.github.io/api/#target for more details
             cfg = "exec",
             doc = "An executable for the esbuild binary",
         ),
+        "sourcemap": attr.string(
+            values = ["external", "inline", "both"],
+        )
     },
     implementation = _esbuild_impl,
     doc = """Runs the esbuild bundler under Bazel
@@ -270,9 +284,14 @@ def esbuild_macro(name, output_dir = False, **kwargs):
             **kwargs
         )
     else:
+        output_map = None
+        sourcemap = kwargs.get("sourcemap", None)
+        if sourcemap != "inline":
+            output_map = "%s.js.map" % name
+
         esbuild(
             name = name,
             output = "%s.js" % name,
-            output_map = "%s.js.map" % name,
+            output_map = output_map,
             **kwargs
         )

```

mattsoulanille: This looks good, and it's nicer since it matches esbuild's args better. I might have time to implement this tomorrow.

mattem:
lgtm thanks! 🙏

mattem (approved):

PR #2536, opened 2021-03-16
chore: incompatible_use_platforms_repo_for_constraints

alexeagle:
See https://github.com/bazelbuild/bazel/issues/8622

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -52,3 +52,5 @@ build:docker-sandbox --experimental_docker_image=gcr.io/cloud-marketplace/google
 # Incompatible flags to run with
 build --incompatible_no_implicit_file_export
 build --incompatible_restrict_string_escapes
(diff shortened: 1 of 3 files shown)


gregmagolan:
NB: new API available since 0.28 and our current minimum bazel version is 2.1.0 (as defined in `internal/node/node_repositories.bzl`)

gregmagolan (approved):

alexeagle:
Thanks for checking that compatibility :)

aranguyen (commented):
@alexeagle I notice this commit is not yet in any recent release? Is there plan to have it in the next release? If it is already part of a release, please advise. I'm pushing for the flag flip and bazel is currently loading an older version of rules_nodejs (2.2.2) which still has the lingering references to @bazel_tools//platforms.

PR #2537, opened 2021-03-17
chore: upgrade to Bazel 4

alexeagle:
Also add integration testing against previous two major releases, so we don't accidentally force users to upgrade


alexeagle:
hmm, macos examples now take 51min to run. maybe we are doing too much testing?

mattem:
Can we run the multiple bazel versions examples tests in parallel on different machines?

mattem (approved):

PR #2538, opened 2021-03-17
fix(builtin): always install source-map-support

### LINKED ISSUE #2520: Sourcemaps is broken after v3 upgrade
# 🐞 bug report

### Affected Rule
The issue is caused by the rule: `nodejs_binary`

### Is this a regression?

 Yes, the previous version in which this bug was not present was: 2.3.1

### Description

After upgrading the rules_nodejs from `2.3.1` to `3.2.2` the sourcemaps have stopped working. In case the program is throwing an error,

Before:
```
Error: here
    at Object.<anonymous> (index.ts:1:7)
```

After:
```
throw new Error("here");
^

Error: here
    at Object.<anonymous> (/private/var/tmp/_bazel_mshah/fa060aa929cca939406f1ed94045b096/execroot/example/bazel-out/darwin-fastbuild/bin/bin.sh.runfiles/example/index.js:2:7)
```

As you can see, before this upgrade, the file name displayed in logs is a `.ts` file and it points to original source, but after the upgrade, the logs contain reference to the compiled `.js` file.

## 🔬 Minimal Reproduction

Link to repo: https://github.com/mitul45/bazel-example

To see actual stack-trace, [uncomment lines](https://github.com/mitul45/bazel-example/blob/main/WORKSPACE#L13-L17) to use `rules_nodejs` `v2.3.1` and downgrade `@bazel/typescript` to `2.3.1` (`yarn upgrade @bazel/typescript@2.3.1`)

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
macOS Catalina
10.15.7
  </code>
</pre>

<pre>
  <code>
/git/bazel-example (main)  >> yarn bazel version
yarn run v1.22.5
$ /Users/mshah/git/bazel-example/node_modules/.bin/bazel version
Bazelisk version: v1.7.5
Build label: 4.0.0
Build time: Thu Jan 21 07:38:50 2021 (1611214730)
Build timestamp: 1611214730
Build timestamp as int: 1611214730
✨  Done in 1.39s.
  </code>
</pre>

**Rules_nodejs version:**

<pre>
  <code>
3.2.2
  </code>
</pre>

**Anything else relevant?**
Nope

alexeagle:
We previously did this when patch_module_resolver was the default, because it had some extra (unrelated) logic at the end to install it.
When we flipped the default in 3.0, we accidentally dropped the feature.

This just reinstates it using the source-map-support/register feature.
Similarly to our prior implementation, we just call our own vendored source-map-support so users don't have to think about it.

Also fix missing docs for npm_package_bin attrs stdout,stderr,exit_code_out

Fixes #2520


alexeagle:
@joeljeske this is a followup to our discussion https://github.com/bazelbuild/rules_nodejs/pull/2448#issuecomment-780657109

mattem (approved):
Test failure looks like a flake on Windows

PR #2545, opened 2021-03-22
feat(esbuild): add output_css flag to esbuild()

dae:
When esbuild encounters an 'import "./foo.css"' reference in normal (non-splitting)
mode, it will place the contents of that file in a .css file next to the
main output .js file.

This patch adds a flag output_css, which when set to True, will declare
the created .css file as another output, so it can be consumed by other
rules.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -100,6 +100,9 @@ def _esbuild_impl(ctx):
                 fail("output_map must be specified if sourcemap is not set to 'inline'")
             outputs.append(js_out_map)
 
+        if ctx.outputs.output_css:
+            outputs.append(ctx.outputs.output_css)
+
         if ctx.attr.format:
             args.add_joined(["--format", ctx.attr.format], join_with = "=")
 
@@ -219,6 +222,10 @@ See https://esbuild.github.io/api/#splitting for more details
             mandatory = False,
             doc = "Name of the output source map when bundling",
         ),
+        "output_css": attr.output(
+            mandatory = False,
+            doc = "Name of the output css file when bundling",
+        ),
         "platform": attr.string(
             default = "browser",
             values = ["node", "browser", "neutral", ""],
@@ -273,7 +280,7 @@ For further information about esbuild, see https://esbuild.github.io/
     """,
 )
 
-def esbuild_macro(name, output_dir = False, **kwargs):
+def esbuild_macro(name, output_dir = False, output_css = False, **kwargs):
     """esbuild helper macro around the `esbuild_bundle` rule
 
     For a full list of attributes, see the `esbuild_bundle` rule
@@ -281,6 +288,8 @@ def esbuild_macro(name, output_dir = False, **kwargs):
     Args:
         name: The name used for this rule and output files
         output_dir: If `True`, produce a code split bundle in an output directory
+        output_css: If `True`, declare name.css as an output, which is the
+                    case when your code imports a css file.
         **kwargs: All other args from `esbuild_bundle`
     """
 
@@ -304,5 +313,6 @@ def esbuild_macro(name, output_dir = False, **kwargs):
             name = name,
             output = output,
             output_map = output_map,
+            output_css = None if not output_css else "%s.css" % name,
             **kwargs
         )
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -28,6 +28,9 @@ def _esbuild_impl(ctx):
         elif hasattr(dep, "files"):
             deps_depsets.append(dep.files)
 
+        if DefaultInfo in dep:
+            deps_depsets.append(dep[DefaultInfo].data_runfiles.files)
+
         if NpmPackageInfo in dep:
             deps_depsets.append(dep[NpmPackageInfo].sources)
             npm_workspaces.append(dep[NpmPackageInfo].workspace)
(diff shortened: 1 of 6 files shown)


mattem:
Are there other outputs esbuild will generate? CSS maps for example, or others unrelated to css.
I wonder if we can infer the css outputs depending on the inputs, or allow the user to declare `additional_outputs` where other output files can be declared.

dae:
(force-pushed to remove the two unnecessary link_workspace_root lines)

A brief scan of the docs seem to indicate it's just the .css file for now, and --summary does not show other unexpected files being created. I'm new to esbuild though, so it's possible there's some other feature I missed.

They have an issue about sourcemap support for css: https://github.com/evanw/esbuild/issues/519

dae:
I don't think we can infer the outputs, because esbuild changes its behaviour depending on whether imports are found in the files it processes, so we'd need to parse the files ourselves to know what it's going to do. additional_outputs would work, with the minor downside that it's inconsistent - the caller gets the ".js" appended to the name for free, but is responsible for appending ".css" itself

mattem (changes requested):
It would be good to put some content in the css files and check them with a golden file test, just so we know that _something_ is coming out from the declared file.

mattem (approved):

dae:
I've rebased on latest main and added a fix for esbuild 0.11, and buildkite's passing. Not sure why circleci is complaining - we had a similar issue in a previous PR: https://github.com/bazelbuild/rules_nodejs/pull/2564.

packages/esbuild/esbuild.bzl
```diff
@@ -273,14 +283,16 @@ For further information about esbuild, see https://esbuild.github.io/
     """,
 )
 
-def esbuild_macro(name, output_dir = False, **kwargs):
+def esbuild_macro(name, output_dir = False, output_css = False, **kwargs):
```
mattem, line 286: Now after the rebase, what happens if the user passes `output = "foo.js"` when the rule is named `bar`? I'm guessing esbuild will output `foo.css` and bazel wont find the output?

I guess now we have the same issue with `output_map` // cc @mattsoulanille

mattsoulanille: I don't think we can rely on the output file being split_extension()-able. I'm pretty sure the user is allowed to provide a file that has no extension at all, but I haven't tested it.

@mattem I think we need to remove the output_map arg. Esbuild (as far as I can tell) doesn't allow you to specify where the sourcemap gets created and always bases the path off of the output filename. The rule implementation also only uses output_map to tell bazel about the sourcemap and doesn't even pass it to esbuild. We should probably compute the path in the rule implementation instead of passing it as an arg.

dae:
```
dae@dip:~% esbuild --bundle foo.js --outfile=bar.js

  bar.css  24b
  bar.js   15b

⚡ Done in 2ms
dae@dip:~% esbuild --bundle foo.js --outfile=bar

  bar.css  24b
  bar      15b

⚡ Done in 2ms
```

So it looks like stripping the extension could work. I have no preference either way, so please just let me know whether you'd prefer the current boolean behaviour be fixed, or it changed to accept a filename directly.

mattem: Ah so looks like esbuild will always correctly identify the filename and append `.css`, so if there is an extension we can strip it and use it for the `output_css` name, rather than a boolean. It would be good if the two properties in the API had the same semantics.

@mattsoulanille True, however it also serves as a label that can be then addressed directly, eg a user could have `//foo:bar.js.map` as an input to a downstream rule. This comes in handy when you need to route both the `.js` and `.js.map` somewhere, but the `.css` to another rule rather than depending on the full `/foo:bundle` target.

What we should do though is remove the `.metadata` and `.jsconfig` from the default output 😅

dae: Sorry, I'm a bit confused - you talk about stripping, but also removing the boolean, whereas I would have thought stripping is only useful in the boolean case. Which would you prefer? :-)

mattem: Oops sorry. Upon thinking about it a bit more, perhaps we don't do anything in the macro at all, and leave it completely up to the user to provide the correct name via `output_css` from the rule, and pass it with everything else via `**kwargs`. This way, if esbuild changes to allow control over the resulting css filename there's no change to our API, we would just have to pass an additional flag with the existing `output_css` name that we get, and the user can simply customize it by changing this existing attribute. wdyt? Sorry, I know we've gone around the houses here!

dae: No problem, done. If you'd like all the changes squashed into a single commit, please let me know.

packages/esbuild/test/css/BUILD.bazel
```diff
@@ -0,0 +1,67 @@
...
+)
+
+esbuild(
+    name = "with_css",
+    args = [
+        "--keep-names",
+        "--resolve-extensions=.mjs,.js",
+    ],
+    entry_point = "main.ts",
+    output_css = True,
```
mattem, line 54: This should be `main.css` now right?

dae: Sorry, made the change in haste. Have pushed a fix.

mattem:
Thanks! 🚀

mattem (approved):

PR #2546, opened 2021-03-22
Update jasmine-reporters to fix xmldom vulnerability

bduffany:
https://github.com/larrymyers/jasmine-reporters/pull/202 updated the xlmdom dependency to a version that doesn't have the security vulnerability https://github.com/advisories/GHSA-h6q6-9hqw-rwfv -- bumping the version here to pull in that dep.


alexeagle (approved):
we should set renovate to do these for us - I just don't want to get overwhelmed by PRs...

PR #2547, opened 2021-03-22
fix(builtin): add transitive typings to runfiles provider produced by js_library

mattem:
Add transitive typings to runfiles provider produced by `js_library`. Allows downstream consumers that require the `.d.ts` files at runtime (eg typechecking) to access the files via the runfiles tree.

### DIFF
--- a/internal/js_library/js_library.bzl
+++ b/internal/js_library/js_library.bzl
@@ -212,7 +212,9 @@ def _impl(ctx):
             files = depset(transitive = files_depsets),
             runfiles = ctx.runfiles(
                 files = all_files,
-                transitive_files = depset(transitive = files_depsets),
+                transitive_files = depset(
+                    transitive = files_depsets + typings_depsets,
+                ),
             ),
         ),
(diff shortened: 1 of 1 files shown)


Aghassi (approved):

alexeagle:
I'm surprised this is useful, what's the use case? a typescript program deps on a binary and wants to typecheck against some library code contained in that binary?

mattem:
Yeah, internally running `tsc_test` to perform typechecking, where one of the dependencies is via a `js_library` which outputs the required `.d.ts` files. There is an alternative method, `js_library` outputs types to an output group `types`, so it's possible to do:

```
filegroup(
    name = "types_dts",
    srcs = [
        ...
    ],
    output_group = "types",
)

tsc_test(
    name = "typecheck",
    args = [
        "--noEmit",
        "--incremental",
        "false",
    ],
    chdir = package_name(),
    data = [
        ":types_dts",
        ...
    ],
)
```

mattem:
`//examples:examples_protobufjs` flake

alexeagle (approved):

PR #2548, opened 2021-03-22
fix(typescript): update documentation now that ts_project is recommended

alexeagle:
We don't deprecate ts_library yet


packages/typescript/_README.md
```diff
@@ -0,0 +1,129 @@
...
+(Possibly replacing `@npm` with the name of the repository where you installed dependencies)
+
+Then call it, using the [`npm_package_bin`](Built-ins#npm_package_bin) documentation.
+
+Here is an example:
+https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/node/test/BUILD.bazel#L491-L507
+
+### ts_project
+
+`ts_project` simply runs `tsc --project`, with Bazel knowing which outputs to expect based on the TypeScript compiler options, and with interoperability with other TypeScript rules via a Bazel Provider (DeclarationInfo) that transmits the type information.
```
mattem, line 37: Link to provider docs?

packages/typescript/_README.md
```diff
@@ -0,0 +1,129 @@
...
+    # but has performance penalty in that Bazel must evaluate it.
+    srcs = glob(["*.ts"]),
+    deps = ["//path/to/other:library"],
+)
+```
+
+Here, `//path/to/other:library` is another target in your repo that produces TypeScript typings (for example, another `ts_project` rule). To use third-party libraries from npm, first install them (likely using `npm_install` or `yarn_install` rules) then add those to the `deps` as well:
+
+```starlark
+ts_library(
```
mattem, line 104: Did you mean `ts_project`?

mattem (approved):

packages/typescript/_README.md
```diff
@@ -0,0 +1,129 @@
...
+
+Like we do for any npm package that exposes a binary, rules_nodejs will see your dependency on
+`typescript` and will generate an `index.bzl` file allowing you to run `tsc`.
+To use it, add the load statement `load("@npm//typescript:index.bzl", "tsc")` to your BUILD file.
+(Possibly replacing `@npm` with the name of the repository where you installed dependencies)
+
+Then call it, using the [`npm_package_bin`](Built-ins#npm_package_bin) documentation.
+
+Here is an example:
+https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/node/test/BUILD.bazel#L491-L507
```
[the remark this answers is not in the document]
alexeagle: I'd like to point to the right spot in the file too, it's very long. I pinned to 3.2.2 rather than stable so we don't break the link

packages/typescript/_README.md
alexeagle: yeah - it's so hard to get the links right tho, so they work in both markdown and on the rendered docsite, let's try...

packages/typescript/_README.md
alexeagle: oops!

PR #2555, opened 2021-03-26
chore: update terser dependency

alexeagle:


flolu:
- `--beautify` has been removed (https://github.com/terser/terser/blob/master/CHANGELOG.md#v560) that's why `//packages/terser/test/debug:test` fails
- And `//packages/terser/test/user_config:test` fails because the `terser_config.json` seems to be not used. No matter what I change in the config, it does not affect the output.

PR #2563, opened 2021-03-28
fix(esbuild): files not being found when building external repo

dae:
An esbuild rule that ran fine in a local repo (eg bazel build //:my_esbuild_rule)
fails when built externally (eg bazel build @other_workspace//:my_esbuild_rule).
By prepending the workspace root to the bin path, the resulting base
URL works correctly in both local and remote repo invocations, allowing
esbuild output to be consumed by a different repo.

This change fixes remote repos in Bazel 4.0.0. 3.7.0 seems to require a
workspace_root.replace("external/", "") call first in order
to get the correct location. So while this patch should not break anything
that was not already broken, it only fixes/adds support for remote repos
in 4.0+.

## PR Checklist

### DIFF
--- a/packages/esbuild/helpers.bzl
+++ b/packages/esbuild/helpers.bzl
@@ -86,8 +86,9 @@ def write_jsconfig_file(ctx, path_alias_mappings):
         File object reference for the jsconfig file
     """
 
-    # The package path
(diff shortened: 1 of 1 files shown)


dae:
I think a rebase went wrong, will try this again later today.

mattem (approved):

PR #2564, opened 2021-03-28
feat(esbuild): support location expansion in esbuild args

dae:
When an esbuild rule is passed an argument like "--inject:path/in/repo.js",
this works when building in the local workspace, as esbuild is invoked
with the correct working directory. But if the esbuild rule is in
a remote workspace (eg bazel build @other_workspace//path/in:esbuild_rule),
then the path is no longer valid.

By expanding $(location ...) references in provided arguments, it allows
callers of the rule to pass arguments like the following, which work
in both local and remote repo cases:

esbuild(args = ["--inject:$(location //path/in:repo.js)"], ...)

## PR Checklist

### DIFF
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -109,7 +109,7 @@ def _esbuild_impl(ctx):
     args.add_joined(["--tsconfig", jsconfig_file.path], join_with = "=")
     inputs.append(jsconfig_file)
 
-    args.add_all(ctx.attr.args)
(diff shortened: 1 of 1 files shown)


mattem (approved):

mattem:
Thanks! Wonder what's going on the circleci tests 🤔  @alexeagle Any ideas?

dae:
I have linked CircleCI to my GitHub account in the past, but don't use it currently, and don't recall ever doing so for rules_nodejs.  🤔

PR #2565, opened 2021-03-28
fix: fetch node toolchain based on exec platform, not target platform

dae:
Currently rules_nodejs can't be used when cross-compiling to a different
platform, if that platform does not have a node toolchain available.
This means if you want to cross-compile a project that includes JS
files, you need to do it in multiple steps:

- build the JS files with the default host platform
- build the other products with a target platform specified, copying
the generated JS files in

The fact that it's currently set to target makes me wonder if this
use case was just not considered at the time, or whether there's some
other workflow that changing to exec will break. Thoughts?

## PR Checklist

### DIFF
--- a/toolchains/node/BUILD.bazel
+++ b/toolchains/node/BUILD.bazel
@@ -103,7 +103,7 @@ alias(
 
 toolchain(
     name = "node_linux_amd64_toolchain",
-    target_compatible_with = [
+    exec_compatible_with = [
         "@platforms//os:linux",
         "@platforms//cpu:x86_64",
     ],
@@ -113,7 +113,7 @@ toolchain(
 
 toolchain(
     name = "node_linux_arm64_toolchain",
-    target_compatible_with = [
+    exec_compatible_with = [
         "@platforms//os:linux",
         "@platforms//cpu:aarch64",
     ],
@@ -123,7 +123,7 @@ toolchain(
 
 toolchain(
     name = "node_darwin_amd64_toolchain",
-    target_compatible_with = [
+    exec_compatible_with = [
         "@platforms//os:osx",
         "@platforms//cpu:x86_64",
     ],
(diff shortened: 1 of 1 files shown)


dae:
Ok, judging by the test failures, I gather this is working as intended at the moment - perhaps for building native node libraries? Presumably I can accomplish the same result by registering a custom toolchain in my workspace file, so I'm guessing that's the recommended way to handle cross compilation?

(please feel free to close this)

dae:
If "register your own toolchain" is the correct answer here, would you be happy to accept the following patch? Currently the toolchain type is marked as private, so I can't register an exec-bound toolchain in my workspace.

```patch
diff --git a/toolchains/node/BUILD.bazel b/toolchains/node/BUILD.bazel
index 9bfab568..20ffc3ef 100644
--- a/toolchains/node/BUILD.bazel
+++ b/toolchains/node/BUILD.bazel
@@ -70,7 +70,10 @@ filegroup(
 )

 # node toolchain type
-toolchain_type(name = "toolchain_type")
+toolchain_type(
+    name = "toolchain_type",
+)

 # Allow targets to use a toolchains attribute, such as sh_binary and genrule
 # This way they can reference the NODE_PATH make variable.
```

If this is not desirable, is there another solution?

alexeagle:
Sorry for the delay!

> would you be happy to accept the following patch
Yes! I think you ought to be able to pass your own toolchain, regardless whether we get this PR green.

I don't really understand the toolchain exec-bound vs target-bound. Maybe we can find someone else who does? Or else I'll need to make some time to read up on this.

dae:
From digging through the sources a bit more, my guess is the current target-bound behaviour is intended to support being able to create nodejs_binary()s that will run on a different architecture to the host machine. #2591 makes it fairly easy to support cross compilation, so I'm happy to close this one.

PR #2567, opened 2021-03-29
Propagate esbuild `no-remote-exec` from tags

siggisim:
## PR Checklist

I've tested this change with a local patch, and it works great with or without RBE enabled. Happy to add tests if it makes sense for a change like this.

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently esbuild `execution_requirements` are hardcoded to disable remote execution. When running a build with remote execution enabled, the esbuild action will always run locally.

If running with `--spawn_strategy=remote`, you get the following error:
```
ERROR: /Users/siggi/Code/buildbuddy/app/BUILD.bazel:13:8: Bundling Javascript app/app.mjs [esbuild] failed: No usable spawn strategy found for spawn with mnemonic Action.  Your --spawn_strategy, --genrule_strategy and/or --strategy flags are probably too strict. Visit https://github.com/bazelbuild/bazel/issues/7480 for migration advice
```

Issue Number: N/A

## What is the new behavior?
If the `no-remote-exec` tag is set on the esbuild rule, that value is propagated down to the action's `execution_requirements`.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -115,15 +115,17 @@ def _esbuild_impl(ctx):
     if ctx.attr.max_threads > 0:
         env["GOMAXPROCS"] = str(ctx.attr.max_threads)
 
+    execution_requirements = {}
+    if "no-remote-exec" in ctx.attr.tags:
+        execution_requirements = {"no-remote-exec": "1"}
+
     ctx.actions.run(
         inputs = inputs,
         outputs = outputs,
         executable = ctx.executable.tool,
         arguments = [args],
         progress_message = "%s Javascript %s [esbuild]" % ("Bundling" if not ctx.attr.output_dir else "Splitting", entry_point.short_path),
-        execution_requirements = {
-            "no-remote-exec": "1",
-        },
+        execution_requirements = execution_requirements,
         mnemonic = "esbuild",
         env = env,
     )


packages/esbuild/esbuild.bzl
```diff
@@ -115,15 +115,17 @@ def _esbuild_impl(ctx):
     if ctx.attr.max_threads > 0:
         env["GOMAXPROCS"] = str(ctx.attr.max_threads)
 
+    execution_requirements = {}
+    if "no-remote-exec" in ctx.attr.tags:
```
alexeagle, line 119: is this a breaking change, in that users will have to add this tag after the next release?

@mattem I'm not clear why no-remote-exec was required here, do you know?

siggisim: Only if those users have RBE enabled, and don't want this particular action to happen on RBE for whatever reason. While the behavior may change (for those users), it shouldn't really break anyone's build.

Discussed this a little on Friday with @mattem on Slack, he wasn't sure why that execution_requirements was set initially:
https://bazelbuild.slack.com/archives/CEZUUKQ6P/p1616796872102700

I did test this change on RBE and the rule works just fine remotely.

Happy to do this in a different way if you prefer and make disabling RBE the default - just seems like the tag approach is somewhat standard.

mattem: I had a think about this, and I think it's just some copypasta from when I copied it from evertz internal.
If this is working as expected on RBE then it should be fine, but there is the point about breaking change, albeit small imo.

siggisim: Let me know what you think the best way to proceed is - happy to make any changes or mark the PR as introducing a breaking change if you prefer.

Would love to get this landed, but can always stick with the local fork approach until we're ready!

alexeagle: If we were really worried about the breaking change, we'd make your new attribute default match the current behavior. But I don't think we are, plus the esbuild package is quite new so I'm less concerned about dormant code that depends on it.

alexeagle (approved):

PR #2568, opened 2021-03-29
feat: introduce package for runfile helpers

devversion:
Introduces a new package called `@bazel/runfiles` that
provides runfile helpers for Bazel with NodeJS.

The benefit of providing the helpers as a separate package
is that people can explicitly specify a dependency on
the helpers and also can get typings. The current best practice
of using a require w/ an environment variable is untyped and
prone to mistakes (and doesn't fit well into the Node/NPM ecosystem).


devversion:
This is an initial proposal for how the `@bazel/runfiles` package could look like. cc. @alexeagle @gregmagolan

internal/linker/test/link_node_modules.spec.ts
```diff
@@ -1,9 +1,12 @@
 import * as fs from 'fs';
 import * as path from 'path';
 
-import * as linker from '../link_node_modules';
+import {Runfiles} from '../../runfiles/index';
 import {LinkerAliases, LinkerTreeElement} from '../link_node_modules';
 
+// The linker is imported through it's checked-in file. We do this because the import to
+// the runfile helpers from the linker file will always resolve to the checked-in file too.
+const linker: typeof import('../link_node_modules') = require('../index.js');
```
alexeagle, line 9: ooh so clever

packages/runfiles/index.js
```diff
@@ -0,0 +1,6 @@
+const runfileHelper = require(process.env['BAZEL_NODE_RUNFILES_HELPER']);
```
alexeagle, line 1: this is an interesting choice. @gregmagolan had the same idea - but then you have some strange version skew coupling. If you upgrade `@bazel/runfiles` package by one version, you still get the code from whatever build_bazel_rules_nodejs package you're running. I suspect it's better if the sources move over here to packages/runfiles and users get the versioned thing. then internally we have a way to continue using it from our builtin code

packages/typescript/checked_in_ts_project.bzl
```diff
@@ -31,19 +37,43 @@ def checked_in_ts_project(name, src, checked_in_js = None, **kwargs):
...
+            input = "%s_types" % name,
+        )
+
+        generated_file_test(
+            name = "%s_check_types" % name,
+            generated = "%s_types_out" % name,
+            src = checked_in_types,
+        )
+
+def append_skip_formatting_header(name, input):
```
alexeagle, line 70: love this refactoring. You clearly learned things from Jeremy Elbourn :)

maybe some bits of this could land as a prefactoring to make the review easier?

alexeagle (commented):
to prove the PR is complete, could you also update examples to use the new package?

e.g. here https://github.com/bazelbuild/rules_nodejs/blob/stable/examples/closure/test.js#L1
and a spot in examples/create-react-app as well

packages/runfiles/index.js
devversion: That makes sense. I'll update the PR.

alexeagle: is this actually done? still seems like the runfiles implementation is coming out of the builtin instead of the npm package

devversion: I was thinking if someone uses `@bazel/runfiles` it will be the helpers specific to the installed version of `@bazel/runfiles`. But if someone uses the environment variable (which we should just deprecate in the future IMO), they will get the version that is built-in.

devversion: The actual runfile implementation is now in `packages/runfiles`

packages/index.bzl
```diff
@@ -23,6 +23,7 @@ NPM_PACKAGES = ["@bazel/%s" % pkg for pkg in [
     "labs",
     "protractor",
     "rollup",
+    "runfiles",
```
alexeagle, line 26: would you add it to .commitlint.js as well so you can feat(runfiles) ?

alexeagle (commented):
why the rollup upgrade? if that's needed here, would you mind doing that as a separate prefactoring PR to make the review more tractable?

packages/index.bzl
devversion: :+1: Good catch.

alexeagle (approved):
ah, it's so pretty now :)

PR #2574, opened 2021-03-30
feat(typescript): add support for "jsx: preserve" compiler option

duarten:
When tsc is instructed to preserve jsx, the emitted files have a .jsx
extension for .tsx and .jsx inputs, which means ts_project needs to be
aware of the option to properly define outputs.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

The current behavior causes the `ts_project` rule to fail as it is expecting different outputs.

Issue Number: N/A

## What is the new behavior?

In the new behavior, `ts_project` is aware of the `jsx` compiler option and sets output files accordingly.

- [ ] Yes
- [X] No

### DIFF
--- a/packages/typescript/internal/ts_project.bzl
+++ b/packages/typescript/internal/ts_project.bzl
@@ -269,6 +269,7 @@ def _validate_options_impl(ctx):
         allow_js = ctx.attr.allow_js,
         declaration = ctx.attr.declaration,
         declaration_map = ctx.attr.declaration_map,
+        preserve_jsx = ctx.attr.preserve_jsx,
         composite = ctx.attr.composite,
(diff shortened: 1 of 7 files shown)


duarten:
@alexeagle Trying to get it to green :) I actually just refactored the approach a bit, there's now a "preserve_jsx" attribute instead of "jsx", since we just care about the "preserve" option.

alexeagle (approved):
at first my reaction was "it's semantically similar to just map jsx=foo from tsconfig into BUILD" but then thinking more, the common case of jsx=react would force people to add a line to their BUILD file. So I like your boolean attribute more.
Thanks for the fix!!

PR #2576, opened 2021-03-30
fix(jasmine): transitive specs are no longer added to the test suite

a-ignatov-parc:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Jasmine runner adds all transitive specs to active execution, which results in some specs running multiple times. 

This can be especially problematic when there are multiple `ts_projects` with test specs that depend on each other. For example:

```
ts_project(name="proj_a") → jasmine_node_test(name="test_a") // will run specs from proj_a
↓
ts_project(name="proj_b") → jasmine_node_test(name="test_b") // will run specs from proj_a and proj_b
↓
ts_project(name="proj_c") → jasmine_node_test(name="test_c") // will run specs from proj_a, proj_b and proj_c
```

Issue Number: N/A

## What is the new behavior?

Only specs from direct deps are added to active execution

- [x] Yes, with opt-in flag
- [ ] No

### DIFF
--- a/packages/jasmine/jasmine_node_test.bzl
+++ b/packages/jasmine/jasmine_node_test.bzl
@@ -25,7 +25,9 @@ def _js_sources_impl(ctx):
     depsets = []
     for src in ctx.attr.srcs:
         if JSModuleInfo in src:
-            depsets.append(src[JSModuleInfo].sources)
+            provider = src[JSModuleInfo]
+            files = provider.direct_sources if ctx.attr.use_direct_specs else provider.sources
+            depsets.append(files)
         if hasattr(src, "files"):
             depsets.append(src.files)
     sources = depset(transitive = depsets)
@@ -48,6 +50,7 @@ _js_sources = rule(
         "srcs": attr.label_list(
             allow_files = True,
         ),
+        "use_direct_specs": attr.bool(),
     },
     outputs = {
         "manifest": "%{name}.MF",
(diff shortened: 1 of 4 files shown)


packages/jasmine/jasmine_node_test.bzl
```diff
@@ -25,7 +25,7 @@ def _js_sources_impl(ctx):
     depsets = []
     for src in ctx.attr.srcs:
         if JSModuleInfo in src:
-            depsets.append(src[JSModuleInfo].sources)
+            depsets.append(src[JSModuleInfo].direct_sources)
```
alexeagle, line 28: technically this is a breaking change, since users who were relying on nested packages having their tests run will now just get nothing running those tests. I think maybe we have to hold this change for a 4.x branch, unless we want to add some opt-in flag for it.

alexeagle:
after some thought I agree the old behavior is a bug

packages/jasmine/jasmine_node_test.bzl
a-ignatov-parc: > unless we want to add some opt-in flag for it.
I would prefer this option. Our tests are broken at the moment, so it would be great to get this fix faster than waiting for 4.x

a-ignatov-parc: I've updated the fix a bit:
1. Added `use_direct_specs` opt-in flag.
2. With this flag, only specs from direct sources of `srcs` will be added to execution. Specs from `deps` no longer taken into account.

alexeagle (approved):
nice, thanks!!

PR #2579, opened 2021-03-31
feat(builtin): document how nodejs_binary#entry_point can use a direc…

alexeagle:
…tory


internal/node/test/dir_entry_point/BUILD.bazel
```diff
@@ -0,0 +1,96 @@
...
+    ],
+    tool = "run_a",
+)
+
+#############
+# Test Case 2
+# Without the pkg_npm, we should still be able to run the program.
+# But we need an adapter rule that gives us the equivalent of the "main" field
+
+package_main(
```
alexeagle, line 76: directory_entry_point should be the name

gregmagolan (approved):

PR #2591, opened 2021-04-07
feat: make node toolchain_type public so new toolchains can be added

dae:
Enables rules_nodejs to be used in cross-compilation contexts, where
the target architecture is not supported by rules_nodejs.

rules_nodejs's default toolchains are limited via target_compatible_with,
which means they will not work when cross-compiling to Android or iOS
with --platforms for example. With this change, users can define their
own toolchain that is limited by execution platform instead of target
platform.

For example, if you're targeting an unsupported platform, but building
on a Mac, you can place the following in the local WORKSPACE:

    register_toolchains("//ts/node_toolchain")

And the following in ts/node_toolchain/BUILD.bazel:

    toolchain(
        name = "node_toolchain",
        exec_compatible_with = [
            "@platforms//os:osx",
            "@platforms//cpu:x86_64",
        ],
        toolchain = "@nodejs_darwin_amd64_config//:toolchain",
        toolchain_type = "@build_bazel_rules_nodejs//toolchains/node:toolchain_type",
    )

Closes #2565

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

### DIFF
--- a/toolchains/node/BUILD.bazel
+++ b/toolchains/node/BUILD.bazel
@@ -70,7 +70,10 @@ filegroup(
 )
 
 # node toolchain type
-toolchain_type(name = "toolchain_type")
+toolchain_type(
+    name = "toolchain_type",
+    visibility = ["//visibility:public"],
+)
 
 # Allow targets to use a toolchains attribute, such as sh_binary and genrule
 # This way they can reference the NODE_PATH make variable.


alexeagle (approved):
Thanks! We'll need to add some docs too, your PR description here is nice but not very discoverable later.

PR #2594, opened 2021-04-08
feat(esbuild): allow ts / tsx files in esbuilds srcs

mattem:
Esbuild can take `.ts` files as an input and compile to js, this change allows that, and shows that you can plumb together both `esbuild`, `ts_project` and `js_library` in novel ways to compile Typescript and bundle it.

It's a little tricky to handle the `entry_point`, as we resolve a `.ts` entrypoint to its js for the user. This change means that if a `.ts` file is passed to `entry_point` then it's expected to be in `srcs`, rather than `deps`.

As a breaking change in 4.0.0, I'd like to propose that we _don't_ automagically resolve to a `.js` input, and the user has to specify the correct extension, or at least provide a flag to flip to the new behaviour, or back to the `legacy_resolve_ts_entry_points` (I like verbose names). 

closes #2594

### DIFF
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -4,7 +4,7 @@ esbuild rule
 
 load("@build_bazel_rules_nodejs//:providers.bzl", "JSEcmaScriptModuleInfo", "JSModuleInfo", "NpmPackageInfo", "node_modules_aspect")
 load("@build_bazel_rules_nodejs//internal/linker:link_node_modules.bzl", "MODULE_MAPPINGS_ASPECT_RESULTS_NAME", "module_mappings_aspect")
-load(":helpers.bzl", "filter_files", "generate_path_mapping", "resolve_js_input", "write_jsconfig_file")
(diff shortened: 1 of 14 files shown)


mattem:
`//examples:examples_protobufjs` 😠

packages/esbuild/esbuild.bzl
```diff
@@ -246,9 +246,7 @@ See https://esbuild.github.io/api/#sources-content for more details
         "srcs": attr.label_list(
             allow_files = True,
             default = [],
-            doc = """Non-entry point JavaScript source files from the workspace.
-
-You must not repeat file(s) passed to entry_point""",
+            doc = """Source files to be made avialble to esbuild""",
```
alexeagle, line 249: nit: spelling

alexeagle (approved):

mattem:
Need to add an extra test in here. I think I've broken the case where if you have only JS sources in `esbuild` `srcs` attr, then the `entry_point` wont be resolved.

Issue #2586, opened 2021-04-05
esbuild: sub-dependencies resolved incorrectly


ofagbemi:
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: esbuild

### Is this a regression?

No

### Description

Let's say that I'm bundling some code that depends on `module-a@1` and `module-b`. `module-b` depends on a different version of `module-a` — let's say `module-a@2`. In the bundle, `module-a` gets resolved to `module-a@1`, even if we never import `module-a` directly and it only actually gets imported via `module-b`.

I think this might be happening because the [esbuild rule uses the tsconfig `paths` field](https://github.com/bazelbuild/rules_nodejs/blob/d10e17cc4dc0e5f3753595c74134e2cc8dc4a8b9/packages/esbuild/esbuild.bzl#L44). The module is getting resolved to `@npm/node_modules/module-a` instead of to `@npm/node_modules/module-b/node_modules/module-a`.

My real-world example is trying to use `next` — which depends on `styled-jsx -> stylis@3` — and `@emotion/css` — which depends on `@emotion/cache -> stylis@4` — together.

## 🔬 Minimal Reproduction

Here's a repo that repros: https://github.com/ofagbemi/esbuild-test

In the example, module-1 includes `stylis@3` and `@emotion/css`, which includes `stylis@4`. To see the issue,
1. run `bazel build :bundle`
2. open `bazel-bin/module-1/bundle.js`
3. notice that only one copy of `stylis.js` actually gets bundled — the one at `external/npm/node_modules/stylis/stylis.js` (can ctrl/cmd-f for "stylis.js")

Note that esbuild handles this fine on its own. You can compare the above to the output of `npx esbuild module-1/index.ts --bundle --platform=node > test-bundle.js` — esbuild properly includes both `node_modules/stylis/stylis.js` and `node_modules/@emotion/cache/node_modules/stylis/dist/umd/stylis.js`.

I think the problem is with the way the Bazel esbuild rule sets the [tsconfig `paths` field](https://github.com/bazelbuild/rules_nodejs/blob/d10e17cc4dc0e5f3753595c74134e2cc8dc4a8b9/packages/esbuild/esbuild.bzl#L44) — you can observe a similar issue if you set paths to something like `{ "*": "./node_modules/*" }`.

## 🔥 Exception or Error

n/a

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
MacOS Catalina v10.15.7
  </code>
[...]

PR #2601, opened 2021-04-08
fix(esbuild): use run_node to call esbuild via a launcher rather than invoking directly

mattem:
Use `run_node` to call esbuild via a launcher rather than invoking directly. This addresses two issues:

* using `run_node` allows us to invoke the linker fist. We no longer have to teach esbuild how to resolve `node_modules`, fixing https://github.com/bazelbuild/rules_nodejs/issues/2586
* Opens up the possibility of using esbuild via an API rather than directly, allowing us to add support for plugins if needed

Currently 1st party sources are still resolved though the `jsconfig.json` file, as this does change the bundle output slightly. 

closes https://github.com/bazelbuild/rules_nodejs/issues/2586


alexeagle (approved):

mattem:
Windows 😡

PR #2603, opened 2021-04-13
fix(concatjs): update karma to 6.3.2 and fix #2093

### LINKED ISSUE #2093: [karma-server]: UncaughtException:: file.detectType is not a function
Please see the Karma project's https://github.com/karma-runner/karma/issues/3537#, where they say the issue is with rules_nodejs.

mattsoulanille:
## PR Checklist

No new tests added, but one was fixed to work with the change.
Docs shouldn't need to be updated for this change.

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #2093 

## What is the new behavior?

Fixes #2093. Updates karma packages to latest versions.

- [ ] Yes
- [x] No

The old version of karma seems to still work with these changes (n = 2 tests), but I haven't extensively tested it.


packages/concatjs/index.ts
```diff
@@ -4,6 +4,7 @@
  */
 import * as crypto from 'crypto';
 import * as fs from 'fs';
+import * as File from 'karma/lib/file';
```
mattsoulanille, line 7: This is somewhat hacky. We're probably not supposed to import from `lib`

mattem:
Should we be bumping the peer dep here: https://github.com/bazelbuild/rules_nodejs/blob/stable/packages/concatjs/package.json#L25 ? Or this this API available in 4.0.0 too

mattsoulanille:
Nice catch. I forgot that there are separate `package.json` files for the `pkg_npm` targets. 

We might not actually have to bump the peer dependency. The only key missing in the new [File](https://github.com/karma-runner/karma/blob/master/lib/file.js) object that is present in the current implementation's object is `content`, which the current one sets to the empty string. Karma 4 seems to work fine without it, at least with our tests in `//packages/concatjs/web_test/test/karma_typescript/...` and `//packages/concatjs/web_test/test/karma/...`.

We could still bump it to 5.1.0 just in case. What do you think?

mattem:
As the tests pass I think this shows it's backwards compatible - we could bump the version as part of the upcoming 4.0.0 milestone release if we wanted, then it's a clear breaking point.

mattem (approved):

mattsoulanille:
Sounds good to me. Thanks!

PR #2615, opened 2021-04-19
fix(builtin): properly parse status file value containing spaces

josephperrott:
Based on Bazel's documentation, values provided through the workspace_status_command can contain
whitespace on the same line the key is defined on:

> The key names can be anything but they may only use upper case letters and underscores. The
> first space after the key name separates it from the value. The value is the rest of the line
> (including additional whitespaces).

The `parseStatusFile` utility function is updated to rely on a regex for performing this parsing to
match the described behavior.

### DIFF
--- a/internal/pkg_npm/packager.js
+++ b/internal/pkg_npm/packager.js
@@ -61,21 +61,32 @@ function unquoteArgs(s) {
 }
 
 /**
- * The status files are expected to look like
- * BUILD_SCM_HASH 83c699db39cfd74526cdf9bebb75aa6f122908bb
- * BUILD_SCM_LOCAL_CHANGES true
- * STABLE_BUILD_SCM_VERSION 6.0.0-beta.6+12.sha-83c699d.with-local-changes
- * BUILD_TIMESTAMP 1520021990506
- *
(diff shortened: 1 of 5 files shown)


alexeagle:
failure looks legit
```
ERROR: /home/circleci/rules_nodejs/internal/pkg_npm/test/BUILD.bazel:82:8: Executing genrule //internal/pkg_npm/test:gen_test_pkg_pack failed: (Exit 1): bash failed: error executing command /bin/bash -c ... (remaining 1 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox bash failed: error executing command /bin/bash -c ... (remaining 1 argument(s) skipped)

npm ERR! Invalid version: "{BUILD_SCM_VERSION}"
```

alexeagle (commented):
This occurs in a couple other places, do you mind fixing those too?

`internal/pkg_web/assembler.js`
a few `rollup.config.js` files
`install.md`

alexeagle (approved):
looks like we ought to come back with a refactoring later that makes this function available through our built-in somehow. thanks for fixing!

PR #2643, opened 2021-04-29
fix(protractor): unable to specify `server` as configurable attribute

devversion:
Currently it is not possible to specify the `protractor_web_test`
server using a configurable attribute. i.e. using `select`.

This happens because the `server` attribute is accessed directly
in the macro so that Bazel is unable to respect the `select`. We
fix this by no longer accessing the server in the macro.

### DIFF
--- a/packages/protractor/protractor_web_test.bzl
+++ b/packages/protractor/protractor_web_test.bzl
@@ -110,6 +110,15 @@ def _protractor_web_test_impl(ctx):
     )
 
     runfiles = [configuration] + configuration_sources + on_prepare_sources
+    server_files = depset()
+
+    # If a server has been specified, add it to the runfiles together with it's required runfiles. This is necessary
+    # as the test executable references the server executable as per `TMPL_server` and executes it.
+    if ctx.executable.server:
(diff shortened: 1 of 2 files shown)


packages/protractor/protractor_web_test.bzl
```diff
@@ -110,6 +110,15 @@ def _protractor_web_test_impl(ctx):
     )
 
     runfiles = [configuration] + configuration_sources + on_prepare_sources
+    server_files = depset()
+
+    # If a server has been specified, add it to the runfiles together with it's required runfiles. This is necessary
+    # as the test executable references the server executable as per `TMPL_server` and executes it.
+    if ctx.executable.server:
+        server_files = depset(
```
alexeagle, line 118: nit : I would have named it `server_runfiles` but not important enough to fix

alexeagle (approved):

packages/protractor/protractor_web_test.bzl
devversion: yup. very true. I will send a follow-up!

Issue #2647, opened 2021-04-30
Allow adding generated ESM files to package generated by pkg_npm


SkaillZ:
# 🚀 feature request

### Relevant Rules

`pkg_npm`

### Description

The `pkg_npm` rule currently doesn't include any generated ESM files. Since ESM files are often included in NPM packages in addition to the CommonJS files, there should be an option to generate a package with ESM files included.

### Describe the solution you'd like

An attribute that allows including ESM files could be added.

See: #2648

PR #2648, opened 2021-04-30
feat(builtin): allow bundling ESM output with the pkg_npm rule

SkaillZ:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #2647 

## What is the new behavior?

An attribute has been added that allows generating ESM files.

- [ ] Yes
- [x] No

### DIFF
--- a/internal/pkg_npm/pkg_npm.bzl
+++ b/internal/pkg_npm/pkg_npm.bzl
@@ -6,7 +6,7 @@ If all users of your library code use Bazel, they should just add your library
 to the `deps` of one of their targets.
 """
 
-load("//:providers.bzl", "DeclarationInfo", "JSModuleInfo", "LinkablePackageInfo", "NODE_CONTEXT_ATTRS", "NodeContextInfo")
+load("//:providers.bzl", "DeclarationInfo", "JSEcmaScriptModuleInfo", "JSModuleInfo", "JSNamedModuleInfo", "LinkablePackageInfo", "NODE_CONTEXT_ATTRS", "NodeContextInfo")
 
 _DOC = """The pkg_npm rule creates a directory containing a publishable npm artifact.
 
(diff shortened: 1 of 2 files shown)


SkaillZ:
@googlebot I signed it!

internal/pkg_npm/pkg_npm.bzl
```diff
@@ -98,6 +98,10 @@ PKG_NPM_ATTRS = dict(NODE_CONTEXT_ATTRS, **{
         doc = """Other targets which produce files that should be included in the package, such as `rollup_bundle`""",
         allow_files = True,
     ),
+    "include_esm": attr.bool(
```
alexeagle, line 101: this is great, do you know why we need this flag though? we don't have flags for the other JS flavors. Would something go wrong if we just made this always true for everyone in the next release?

alexeagle (commented):
thanks!

internal/pkg_npm/pkg_npm.bzl
SkaillZ: Sorry for the late reply! I don't think it would break anything so I've removed the flag.

alexeagle (approved):
thanks!

internal/pkg_npm/pkg_npm.bzl
```diff
@@ -295,11 +295,15 @@ def _pkg_npm(ctx):
         # Only collect DefaultInfo files (not transitive)
         deps_files_depsets.append(dep.files)
 
-        # All direct & transitive JavaScript-producing deps
-        if JSModuleInfo in dep:
-            deps_files_depsets.append(dep[JSModuleInfo].sources)
+        # All direct and transitive deps that produce CommonJS modules
+        if JSNamedModuleInfo in dep:
```
alexeagle, line 299: I'm concerned about dropping `JsModuleInfo` - I'm pretty sure it's possible that we get a dep that has JS files only in that provider but not in either of these two you've added. I think we should have all three. Would be nice if there was a red test result on this PR making me more confident...

SkaillZ: I've added the `JSModuleInfo` back, but I'm not sure where this provider is created. If you could give me an example of a rule where it's used, I could add a test case with it as well.

alexeagle (approved):
thanks!

PR #2658, opened 2021-05-06
feat(builtin): add js_library JSEcmaScriptModuleInfo support

jbedard:


alexeagle (approved):

packages/rollup/rollup_bundle.bzl
```diff
@@ -254,7 +254,8 @@ def _rollup_bundle(ctx):
     for dep in ctx.attr.deps:
         if JSEcmaScriptModuleInfo in dep:
             deps_depsets.append(dep[JSEcmaScriptModuleInfo].sources)
-        elif JSModuleInfo in dep:
```
jbedard, line 257: This seems to fix the `internal/pkg_npm/test:test_pkg` test, and I don't understand why..?

Why is the [esbuild dep collection](https://github.com/bazelbuild/rules_nodejs/blob/3.4.2/packages/esbuild/esbuild.bzl#L19-L39) so different from the [rollup version](https://github.com/bazelbuild/rules_nodejs/blob/3.4.2/packages/rollup/rollup_bundle.bzl#L254-L265)? Without this change `pkg_npm/test:test_pkg` fails because the outputted files have slight changes (a single `exports.X = undefined (or void?)` and the sourcemap change).

jbedard: The other failing before this change was `packages/rollup/test/integration`. Somehow this only failed on CI and not locally though!? The error was
```
Compilation failed Error: '@far/a' is imported by packages/rollup/test/integration/foo.js, but could not be resolved – treating it as an external dependency
```

I also don't understand this one, both why it only failed on CI (not locally) and why this change fixes it?

PR #2659, opened 2021-05-06
fix(builtin): linker incorrectly resolves workspace `node_modules` for windows

devversion:
Depending on whether there are generated files being stored in
`bazel-out/bin/external/npm` or not, the linker might resolve the
workspace node modules incorrectly for Windows. e.g.

consider the manifest being the followed:

```
npm/karma/bin/_karma.module_mappings.json C:/users/paul/<..>/execroot/angular_material/bazel-out/<..>/bin/external/npm/<..>`
npm/node_modules/@angular/core/index.js C:/users/paul/projects/material2/node_modules/@angular/core/index.js
```

The linker currently on Windows will resolve the NPM workspace to the
`bazel-out` directory. This is incorrect as on Windows, the node modules
are not symlinked as with linux/macOS. Hence the node modules are not
symlinked to the execroot properly and tests/builds fail due to NodeJS
imports being "faulty".

The fix is to simply use the runfile helpers to resolve the workspace
node modules. There is no need in resolving the workspace root if we
can just directly narrow the lookup to the workspace node modules.

This is needed as the runfile resolution with manifest naively looks
for the first entry starting with `npm` in the manifest. This breaks
the assumption that the resolved workspace path refers to a directory
that also contains the `node_modules` folder (as seen in the example
above).

### DIFF
--- a/internal/linker/index.js
+++ b/internal/linker/index.js
@@ -111,24 +111,25 @@ function symlink(target, p) {
         }
     });
 }
-function resolveExternalWorkspacePath(workspace, startCwd, isExecroot, execroot, runfiles) {
+function resolveWorkspaceNodeModules(workspace, startCwd, isExecroot, execroot, runfiles) {
     return __awaiter(this, void 0, void 0, function* () {
+        const targetManifestPath = `${workspace}/node_modules`;
         if (isExecroot) {
-            return `${execroot}/external/${workspace}`;
+            return `${execroot}/external/${targetManifestPath}`;
         }
         if (!execroot) {
-            return path.resolve(`${startCwd}/../${workspace}`);
+            return path.resolve(`${startCwd}/../${targetManifestPath}`);
         }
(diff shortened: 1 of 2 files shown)


internal/linker/index.js
```diff
@@ -309,7 +309,7 @@ function main(args, runfiles) {
                     }
                     else {
                         log_verbose(`no npm workspace node_modules folder under ${packagePath} to link to; creating node_modules directories in ${process.cwd()} for ${packagePath} 1p deps`);
-                        yield mkdirp('${packagePath}/node_modules');
+                        yield mkdirp(`${packagePath}/node_modules`);
```
gregmagolan, line 312: 😱

gregmagolan (approved):
Nice fix. Thank you.
Windows CI is currently very flaky. I'll hit retry on it.

alexeagle:
Thanks Paul, awesome fix 😎

PR #2662, opened 2021-05-07
fix(builtin): account for racy deletion of symlink in linker

JoostK:
On Windows, the linker has been prone to race conditions with respect to
symlinking packages into the `node_modules` tree, causing ENOENT errors
that causes the target to fail. Since the intention of `symlinkWithUnlink`
is to unlink the directory anyway, we can simply ignore ENOENT errors for
the directory that would be about to be unlinked.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

We've been having quite a lot of Windows-only flakes where the linker fails with an ENOENT error.
And example can be found [in this angular/angular job](https://app.circleci.com/pipelines/github/angular/angular/32399/workflows/23aa632b-f939-4a28-8c3a-f4038c3a8c68/jobs/980823).

## What is the new behavior?

The ENOENT error that occurs due to a race condition no longer causes the linker to fail.

- [ ] Yes
- [x] No

## Other information

There might continue to be a window for race conditions. Looking at the call site:

https://github.com/bazelbuild/rules_nodejs/blob/768d352d2fea8ce76e1d0cb5f0bde86c5f73c222/internal/linker/link_node_modules.ts#L653-L664

it becomes clear that `isLeftoverDirectoryFromLinker` is executed between obtaining the stats and actually running `symlinkWithUnlink`, and since`isLeftoverDirectoryFromLinker` performs a directory walk it may provide ample opportunity for the stats to get stale.


JoostK:
@devversion I _hope_ this will address the linker failure we're seeing in angular/angular.

JoostK:
This would unfortunately not fix [this class of failure](https://app.circleci.com/pipelines/github/angular/angular/32646/workflows/23902cbd-7044-4fb2-90e1-5708a9282092/jobs/984434).

devversion:
@JoostK yeah, this might improve the odds of having successful builds. I'm specifically speaking of "odds" here because I think conceptually Windows and Bazel w/ concurrency will never work out 100% here. The reason is that multiple actions/tests operate on the same `node_modules` with multiple instances of the linker running. The linker can always break `node_modules` for an active test in favor of linking the node modules for a second test being spawned in parallel

devversion:
I think one idea to solve the concurrency for windows would be to just link the `node_modules` local to the actions being executed, and then setting the `NODE_PATH`. Surely just a thought.. I'm not sure if that's worth the effort, but that _could_ avoid the race conditions in a reasonable way.

PR #2663, opened 2021-05-08
feat(esbuild): add support for multiple entry points

jbedard:
Very similar to the rollup one, including copying (almost identically) the `_desugar_entry_point_names` util. The main difference is that rollup allows each entry point to have its own output file specified (see "Chunks can be named by adding an = to the provided value:" under https://rollupjs.org/guide/en/#input), where in esbuild it seems that you specify a output filename pattern which all outputted files follow: https://esbuild.github.io/api/#entry-names

In the future we can potentially add support for that `--entry-names` param, but I guess that can always be done manually today with the arbitrary `args`.

Note that the `output_dir` is now defaulted to `True` in the macro if `entry_points` is used. I'm not 100% sure if that is correct or the best way to do it? Are code splitting and `entry_points` starting to overlap a bit too much?

### DIFF
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -5,7 +5,7 @@ esbuild rule
 load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_binary")
 load("@build_bazel_rules_nodejs//:providers.bzl", "JSEcmaScriptModuleInfo", "JSModuleInfo", "NpmPackageInfo", "node_modules_aspect", "run_node")
 load("@build_bazel_rules_nodejs//internal/linker:link_node_modules.bzl", "MODULE_MAPPINGS_ASPECT_RESULTS_NAME", "module_mappings_aspect")
-load(":helpers.bzl", "filter_files", "generate_path_mapping", "resolve_entry_point", "write_jsconfig_file")
+load(":helpers.bzl", "desugar_entry_point_names", "filter_files", "generate_path_mapping", "resolve_entry_point", "write_jsconfig_file")
 
 def _esbuild_impl(ctx):
     # For each dep, JSEcmaScriptModuleInfo is used if found, then JSModuleInfo and finally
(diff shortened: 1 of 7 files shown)


packages/esbuild/esbuild.bzl
```diff
@@ -308,10 +337,11 @@ def esbuild_macro(name, output_dir = False, **kwargs):
         entry_point = Label("@build_bazel_rules_nodejs//packages/esbuild:launcher.js"),
     )
 
-    if output_dir == True:
```
mattem, line 311: This will cause an error when `output_dir` is `True` as it will pass `entry_points = False`, where `entry_points` is expecting a label list, but will get a boolean.

Perhaps something like the following (`None` is falsy and the default on `label_list` is an empty list)
```
if output_dir == True or entry_points:
    esbuild(
            name = name,
            output_dir = True,
            launcher = _launcher,
            **kwargs
        )
```

It can then be removed from the macro's signature

packages/esbuild/esbuild.bzl
```diff
@@ -168,9 +187,19 @@ See https://esbuild.github.io/api/#define for more details
             doc = "A list of direct dependencies that are required to build the bundle",
         ),
         "entry_point": attr.label(
-            mandatory = True,
             allow_single_file = True,
-            doc = "The bundle's entry point (e.g. your main.js or app.js or index.js)",
+            doc = """The bundle's entry point (e.g. your main.js or app.js or index.js)
+
+This is just a shortcut for the `entry_points` attribute with a single entry.
+Either this attribute or `entry_point` must be specified, but not both.
```
mattem, line 194: I think this is copy / paste from rollup, but nit, I don't really like the word "just" 😛 

```suggestion
This is a shortcut for the `entry_points` attribute with a single entry.
Specify either this attribute or `entry_point`, but not both.
```

mattem (changes requested):
> Note that the output_dir is now defaulted to True in the macro if entry_points is used

We can't predict all the output files for multiple entry_points?

packages/esbuild/esbuild.bzl
```diff
@@ -7,6 +7,22 @@ load("@build_bazel_rules_nodejs//:providers.bzl", "JSEcmaScriptModuleInfo", "JSM
 load("@build_bazel_rules_nodejs//internal/linker:link_node_modules.bzl", "MODULE_MAPPINGS_ASPECT_RESULTS_NAME", "module_mappings_aspect")
 load(":helpers.bzl", "filter_files", "generate_path_mapping", "resolve_entry_point", "write_jsconfig_file")
 
+def _desugar_entry_point_names(entry_point, entry_points):
```
mattem, line 10: Perhaps this could go in `helpers.bzl`?

jbedard:
> We can't predict all the output files for multiple entry_points?

We can predict the output file for each entry point like when there's a single entry point. But we can not predict the shared chunks created for the same reasons as when doing code splitting (it's a `chunk-[hash].js` file, or might be a different format/location based on the `--entry-names` arg).

mattem (approved):

PR #2665, opened 2021-05-09
chore: update bazel/create package

alexeagle:
- put deps in a separate file
- don't name WORKSPACE.bazel due to bug


packages/create/index.js
```diff
@@ -152,19 +167,21 @@ workspace(
...
-)
+load("//tools:bazel_deps.bzl", "fetch_dependencies")
+
+fetch_dependencies()
 
 ${pkgMgr === 'yarn' ? yarnInstallCmd : npmInstallCmd}`;
 
-  write('WORKSPACE.bazel', workspaceContent);
-  write('.bazelignore', `node_modules
+  write('tools/BUILD.bazel', '# No targets in this Bazel package\n')
```
mattem, line 176: nit: "No targets in this Bazel package" makes it seem to me that it's not valid to put targets in the root package, rather than saying that this package is currently empty?

packages/create/index.js
```diff
@@ -152,19 +167,21 @@ workspace(
...
+  write('tools/bazel_deps.bzl', bazelDepsContent);
+  // Don't name it WORKSPACE.bazel since there's a bug with managed_directories
```
Wenqer, line 178: @alexeagle could you please share more details about this bug? Does it mean that "workspace" file should be used always without .bazel extension, or it affects only bazel/create package?

alexeagle: @gregmagolan is tracking down the bug which hasn't been filed upstream yet. It affects everyone, not just this bazel/create package.

packages/create/index.js
alexeagle: fixed

PR #2668, opened 2021-05-11
feat(cypress): cypress executable toolchain

mrmeku:
This is a major refactor the cypress_repository rule
Prior to this change, the cypress CLI from NPM was invoked
to download the corresponding Cypress binary.

This was bad for a number of reasons. Namely it was non cachable
and did not match bazel's semantics for grabbing the right binary
for a particular platform / architecture

The new cypress_repositories rule, downloads binaries for all
platforms from Cypress's CDN and uses bazel's select logic
to use the appropriate one given a particular configuration.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

- [x] Yes
- [ ] No

The old repository rule `cypress_repository` has been removed and in its place is a new `cypress_repositories` rule. User's will swap out these rules and add a `version` attribute to their new `cypress_repositories` workspace rule with a version which matches their package.json version of cypress.


alexeagle:
Was just discussing today with @mattem how esbuild has a similar fetch problem. Seems like they could use a similar solution maybe with tool chains?

mrmeku:
@alexeagle I'm a bit less familiar with toolchains than maybe I should be. Could you give me a brief overview about how one might incorporate this type of fetch problem with toolchains?

Would a toolchain circumvent the need for the `select` statements I created in an external repository?

purkhusid:
Would love to see this merged, anything blocking this at the moment?

packages/cypress/internal/cypress_repositories.bzl
```diff
@@ -0,0 +1,113 @@
...
+    """
+    Repository rule used to install cypress binary.
+
+    Args:
+        name: Name of the external workspace where the cypress binary lives
+        version: Version of cypress binary to use. Should match package.json
+        linux_urls: (Optional) URLs at which the cypress binary for linux distros of linux can be downloaded. If omitted, https://cdn.cypress.io/desktop will be used.
+        linux_sha256: (Optional) SHA-256 of the linux cypress binary
+        darwin_urls: (Optional) URLs at which the cypress binary for darwin distros of linux can be downloaded. If omitted, https://cdn.cypress.io/desktop will be used.
+        darwin_sha256: (Optional) SHA-256 of the darwin cypress binary
```
mattem, line 38: Does Cypress have more native binaries than just Linux, Darwin and Windows? eg arm vs amd? (esbuild does, so I'm not sure this pattern scales)

mrmeku:
@googlebot I consent.

packages/cypress/internal/cypress_repositories.bzl
mrmeku: They have the following binaries: 
"darwin-x64", "win32-ia32", "linux-x64", "win32-x64"

I opted not to include the win32-ia32 out of laziness

index.bzl
```diff
@@ -53,6 +55,10 @@ js_library = _js_library
 directory_file_path = _directory_file_path
 # ANY RULES ADDED HERE SHOULD BE DOCUMENTED, see index.for_docs.bzl
 
+# Structs for creating custom rules
+nodejs_binary_struct = _nodejs_binary_struct
```
alexeagle, line 59: I wonder if these should be private APIs to start with (just omit from here and reach into /internal in the impl to start)

also "struct" feels like a strange name to me since it describes the datatype rather than the content. Maybe nodejs_binary_internals or nodejs_binary_impl ? what's the naming convention in that ruleset we're borrowing the concept from?

alexeagle (approved):

index.bzl
mrmeku: Updated from *_struct to *_kwargs as is used in rules_go: https://github.com/bazelbuild/rules_go/blob/4cd45a2ac59bd00ba54d23ebbdb7e5e2aed69007/go/private/rules/test.bzl#L179

PR #2674, opened 2021-05-14
chore: replace clang-format with prettier

alexeagle:
the husky and lint-staged integration give a nice auto-formatting-on-commit model, so we don't have to think as hard.

Also clang-format is clearly Google obscureware and everyone uses Prettier for this.


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx lint-staged
```
Aghassi, line 4: Fair warning, this takes the latest version I believe which means you may hit some drift issues if they update it in a breaking way.

alexeagle: Yeah seems like too many problems here. Also no file suppression comment tsc can add?? I'll probably give up on this one

PR #2676, opened 2021-05-16
feat: allow esbuild cwd for subprocess to be set

Aghassi:
This change is to allow users to set where the esbuild process is
invoked relative to the root of the repo. This is useful in the case
of monorepos in which packages may be nested and not invoked from
root. This is important to have as it helps with relative path resolution
when an entry_point is passed ot ESBuild. Without this, ESBuild will fail
to resolve relative imports.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

Allow ESBuild rule to set the CWD of the `spawn` subprocess it invokes to allow ESBuild to run from non repo root locations.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -135,6 +135,9 @@ def _esbuild_impl(ctx):
     launcher_args = ctx.actions.args()
     launcher_args.add("--esbuild=%s" % ctx.executable.tool.path)
 
+    if ctx.attr.cwd:
+        launcher_args.add("--cwd=%s" % ctx.attr.cwd)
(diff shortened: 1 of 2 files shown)


Aghassi:
@mattem We can circle back on this next week in relation to https://github.com/evanw/esbuild/issues/1287

Aghassi:
Going to close this because we determined the root of the issue is the esbuild follows node resolutions with modules correctly, where as webpack does try both path and module resolutions. This is not the solution

PR #2678, opened 2021-05-17
WIP: npm post install patches fix

zoidbergwill:
I have tried so many things.

Turning yarn_install symlinks off
Turning patch following symlinks on

Various other things

I think the real problem is that the pre/post install patches might only work with some configuration of npm/yarn install that we're not enforcing, but I'm not 100% sure.

Basically I have a patch I try to apply, and it doesn't work, using very similar config to the examples in this repo's WORKSPACE + BUILD.bazel.

I've tried a couple things without much success, unfortunately.

### DIFF
--- a/internal/npm_install/npm_install.bzl
+++ b/internal/npm_install/npm_install.bzl
@@ -237,6 +237,7 @@ data attribute.
 })
 
 def _apply_patches(repository_ctx, patches):
+    package_json_dir = repository_ctx.path(repository_ctx.attr.package_json).dirname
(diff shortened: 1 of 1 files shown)


zoidbergwill:
@googlebot I fixed it.

zoidbergwill:
So I probably should've tried to make a issue instead of a pull request, sorry about that.

This PR also does terrible bad things.

Fortunately `patch-package` in the `package.json` does what we need, but I think it'd be better to use `post_install_patches` instead eventually, if possible.

alexeagle:
> So I probably should've tried to make a issue instead of a pull request, sorry about that.

sounds good, I guess you don't want this to be reviewed and merged then, right?

PR #2682, opened 2021-05-17
Reduce ci load

alexeagle:
It makes our CI slower to run all examples against all Bazel versions. Other rulesets don't seem to test backcompat at all, so it's hard to imagine anyone is actually running on Bazel 2 and using recent rules releases

### DIFF
--- a/index.bzl
+++ b/index.bzl
@@ -88,11 +88,7 @@ BAZEL_VERSION = "4.0.0"
 # accidentally forcing users to update their LTS-supported bazel.
 # These are the versions used when testing nested workspaces with
 # bazel_integration_test.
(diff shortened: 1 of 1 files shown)


mrmeku (approved):
Not sure the com_google_protobuf upgrade belongs here.

Should we remove "3.6.0" as well? Its not an LTS release so I'm sure we should care about supporting it

alexeagle:
that upgrade showed up because the 4.x branch fell behind, I gotta start doing regular cherry picks now :/

alexeagle:
I think you're probably right that when we ship a major we should drop compat with as many old Bazel versions as we can, which would be anything no longer in LTS

PR #2684, opened 2021-05-18
refactor(builtin): do not create unnecessary providers in js_library

jbedard:
... not sure about this, but I found it quite odd seeing all these providers come out of `js_library` when only 1-2 went in. Is there any reason to output all the empty ones? `DeclarationInfo` was intentionally left out if empty, why not the rest?

Wouldn't this effect cases where these providers are looked for in a priority order such as [esbuild](https://github.com/bazelbuild/rules_nodejs/blob/3.5.0/packages/esbuild/esbuild.bzl#L23-L26), [rollup](https://github.com/bazelbuild/rules_nodejs/blob/3.5.0/packages/rollup/rollup_bundle.bzl#L258-L261) etc?

### DIFF
--- a/internal/js_library/js_library.bzl
+++ b/internal/js_library/js_library.bzl
@@ -169,18 +169,17 @@ def _impl(ctx):
         # every single file on bin should be added here
         src_all_files.append(file)
 
-    files_depset = depset(src_all_files)
-    js_files_depset = depset(src_js_files)
-    named_module_files_depset = depset(src_named_module_files)
-    typings_depset = depset(src_typings)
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
I moved this to the 4.x branch since it can be a breaking change that the providers go away, now a `js_library` isn't a legal dependency in some locations a user might have put it.

jbedard:
Any idea why all the angular example integration tests are failing? I don't think those even use `js_library`...?

jbedard:
Closing and will instead look at making collecting of providers more consistent across rules.

PR #2685, opened 2021-05-18
fix(esbuild): provide JSModuleInfo of output bundle

jbedard:
Seems like other rules such as rollup and terser provide this, and it makes sense. I'm not sure of the greatest way to test it though? All 3 could probably use such a test...

### DIFF
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -149,8 +149,14 @@ def _esbuild_impl(ctx):
         tools = [ctx.executable.tool],
     )
 
+    outputs_depset = depset(outputs)
+
     return [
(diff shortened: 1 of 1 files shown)


mattem:
Seems good to me, even with a test I'm not sure exactly what we'd be testing, as at this point handling providers is bazel internal, there is no starlark logic here and we have existing tests to verify the right output.

mattem (approved):

jbedard:
I think a test would need to confirm that the provider interacts with another rule consuming the provider and make sure the integrate properly. However most rules consuming this provider fallback to the default info anyway so the tests would probably pass either way >_>

I still think it's worth the consistency with rollup+terser etc 🤷

PR #2692, opened 2021-05-24
fix(esbuild): prefer finding entry_point files in deps rather than srcs

mattem:
When using esbuild with `js_library`, the entry_point may be part of the deps, therefore the path will end up in `bazel-out/` inside the sandbox where esbuild is running, however the entrypoint path will be in the sources path, leaving any relative path inputs to no longer be relative to the entrypoint file.

The reordering here forces the `resolve_entry_point` method to prefer the file that resides in the `bazel-out/` path in the sandbox.

A possible future breaking change would be to force users to list the entrypoint file(s) in either srcs or deps for the correct pathing, or remove `srcs` from esbuild, meaning all input files must be in the output tree.

### DIFF
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -41,7 +41,9 @@ def _esbuild_impl(ctx):
     entry_points = desugar_entry_point_names(ctx.file.entry_point, ctx.files.entry_points)
 
     deps_inputs = depset(transitive = deps_depsets).to_list()
-    inputs = filter_files(entry_points) + ctx.files.srcs + deps_inputs
+
+    # TODO(mattem): 4.0.0 breaking change, entry_points must exist in deps, and are not considered additional srcs
+    inputs = deps_inputs + ctx.files.srcs + filter_files(entry_points)
 
     metafile = ctx.actions.declare_file("%s_metadata.json" % ctx.attr.name)
     outputs = [metafile]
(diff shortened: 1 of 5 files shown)


jbedard:
> A possible future breaking change would be to force users to list the entrypoint file(s) in either srcs or deps for the correct pathing, or remove srcs from esbuild, meaning all input files must be in the output tree.

👍 - I often find it confusing when a bundle-type rule has both srcs and deps. I would general lean towards deps but 🤷

mattem:
Yeah I agree it can be confusing. The intention was to support where esbuild can take ts sources directly, these would go in `srcs`, rather than via a _something_ and into `deps`. There is an example of this in the tests, but no idea if anyone holds it this way (I guess we have to assume they do). 

I'd also like to support using esbuild without the `--bundle` flag, passing in TS files and having it output JS (currently each input file has to be listed explicitly as an entry point, esbuild won't just output JS from a single TS entry)

joneshf:
A minor hiccup with removing `srcs`, is that there'd have to be some alternative way to [inject](https://esbuild.github.io/api/#inject) a file. What I currently do (not sure if it's the right way) is add the file to `srcs` and then add an entry to the `args` attribute: `"--inject:$(execpath process-shim.js)"`.

I assume that is `srcs` was removed, the rule would need to grow an `inject` attribute that could take files and deal with them accordingly.

alexeagle (approved):

alexeagle:
@joneshf I don't think anything would change for you under that proposal other than anything you have in `srcs` you would move to `deps`

PR #2697, opened 2021-05-25
Fix copy_file on Windows

cocreature:
In general you cannot rely on mkdir and cp being in path here and this
below up for us on CI. Given that this is a repository rule you also
cannot rely on something like rules_posix to resolve them otherwise so
this PR hacks around this in a rather crude but working way.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

rules_nodejs fails on Windows if msys mkdir or cp are not in path.

## What is the new behavior?

No dependency on msys.

- [ ] Yes
- [x] No

### DIFF
--- a/internal/npm_install/npm_install.bzl
+++ b/internal/npm_install/npm_install.bzl
@@ -335,8 +335,9 @@ def _copy_file(repository_ctx, f):
     to_segments = to.split("/")
     if len(to_segments) > 1:
         dirname = "/".join(to_segments[:-1])
+        args = ["mkdir", "-p", dirname] if not is_windows_os(repository_ctx) else ["cmd", "/c", "if not exist {dir} (mkdir {dir})".format(dir = dirname.replace("/", "\\"))]
         result = repository_ctx.execute(
-            ["mkdir", "-p", dirname],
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
Thanks! The logic is isolated enough I'm not too worried about hackiness

cocreature:
Hm something seems off with the copy stuff on CI. Interestingly we did not hit this on our own CI.

alexeagle:
Can you reproduce locally on windows?
I don't want Bazel team to yell at us about CI utilization

mattem:
Isn't it expected that the user has installed MSYS2 (as per the bazel windows setup, https://docs.bazel.build/versions/master/install-windows.html#installing-compilers-and-language-runtimes) and these utilities are on the PATH? This is why CI doesn't have issues here.

cocreature:
Alright I think I got it to work now. Sorry again for the noise and CI usage. The fix was to switch from `copy` to `xcopy`. `copy` uses `+` to represent file appends and there doesn’t seem to be a great way to escape this without breaking other things.

PR #2704, opened 2021-05-26
feat(esbuild): add support for toolchains

mattem:
**BREAKING CHANGE**

Add toolchain support for esbuild, removing the need for the `tool` attribute. Users can load the `esbuild_repositories` macro and call it within their `WORKSPACE`, this will by default define esbuild binaries for Windows, MacOS and Linux (all amd64).

TODO:

- [x] Update docs
- [x] Update esbuild version updater script

Thanks to @dae for the discussion in #2592 and the prototype repo

Relnotes:

`esbuild` package now uses toolchains, and the `tools` attribute has been removed. By default, toolchains for Windows, Mac (`amd64`, `arm64`) and Linux (`amd64`, `arm64`) are included.

See the docs for `configure_esbuild_toolchain` for defining custom toolchains

Add the following to the `WORKSPACE` file (changing `@npm` as required), see the docs for an alternate load method

```
load("@npm//@bazel/esbuild:esbuild_repositories.bzl", "esbuild_repositories")

esbuild_repositories()
```


packages/esbuild/test/tests.bzl
```diff
@@ -4,10 +4,10 @@ load("//packages/esbuild:index.bzl", _esbuild = "esbuild")
 
 def esbuild(**kwargs):
     _esbuild(
-        tool = select({
-            "@bazel_tools//src/conditions:darwin": "@esbuild_darwin//:bin/esbuild",
-            "@bazel_tools//src/conditions:linux_x86_64": "@esbuild_linux//:bin/esbuild",
-            "@bazel_tools//src/conditions:windows": "@esbuild_windows//:esbuild.exe",
-        }),
+        # tool = select({
```
mattem, line 7: Remove this

mattem:
Oh, well, we seem to be at an impasse @googlebot

mattem:
Yeah... that showed you 🤖

dae:
Re the update script, any objections to adding esbuild-darwin-arm64 and esbuild-linux-arm64 as well? I can add them in a follow-up PR if that's easier.

mattem:
Makes sense to me, I'll add them in here

packages/esbuild/esbuild_repositories.bzl
```diff
@@ -0,0 +1,108 @@
+"""
+Helper macro for fetching esbuild versions for internal tests and examples in rules_nodejs
```
alexeagle, line 2: it's also used by end-users right?

packages/esbuild/esbuild_repositories.bzl
```diff
@@ -0,0 +1,108 @@
...
+"""
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@build_bazel_rules_nodejs//internal/common:check_version.bzl", "parse_version")
+
+_VERSION = "0.12.1"
+_MIN_VERSION = "0.12.0"
+_MAX_VERSION = "0.13.0"
+
+DARWIN_AMD64 = "darwin_amd64"
```
alexeagle, line 12: what about darwin_arm64? maybe a note here about what platforms are supported by esbuild and how we chose the list to mirror?

packages/esbuild/esbuild_repositories.bzl
```diff
@@ -0,0 +1,108 @@
...
+            mandatory = True,
+        ),
+    },
+)
+
+def esbuild_repositories(name = "esbuild_repositories", version = _VERSION, shas = ESBUILD_SHAS):
+    """Helper for fetching and setting up the esbuild versions and toolchains
+
+    Args:
+        name: currently unused
```
alexeagle, line 44: if the user passes some other name, is that illegal since you rely on this name elsewhere? should we check?

packages/esbuild/index.docs.bzl
```diff
@@ -16,57 +16,14 @@ or using yarn
 yarn add -D @bazel/esbuild
 ```
 
-Add an `http_archive` fetching the esbuild binary for each platform that you need to support. 
+The esbuild binary is fetched from npm automaticly and exposed via toolchains. Add the `esbuild_repositories` rule to the `WORKSPACE`:
```
alexeagle, line 19: nit: spelling "automatically"

alexeagle (approved):
pretty simple change, the user WORKSPACE is so much nicer now :)

packages/esbuild/esbuild_repositories.bzl
mattem: Yes, see comment in this diff

packages/esbuild/esbuild_repositories.bzl
mattem: No, as the doc states the name is currently unused, it's not relied on anywhere. There is a buildifier warning when there is no name attr.

mattem:
Allowing users to specify their own esbuild version and adding plugin support don't mix well, so I think I'm going to remove that part, which simplifies this feature quite a bit.

packages/esbuild/esbuild_repositories.bzl
mattem: Done, added 🦾

packages/esbuild/esbuild_packages.bzl
```diff
@@ -0,0 +1,70 @@
...
+        ),
+        "linux_arm64": struct(
+            sha = _LINUX_ARM64_SHA,
+            urls = [
+                "https://registry.npmjs.org/esbuild-linux-arm64/-/esbuild-linux-arm64-%s.tgz" % _VERSION,
+            ],
+            binary_path = "bin/esbuild",
+            exec_compatible_with = [
+                "@platforms//os:linux",
+                "@platforms//cpu:x86_64",
```
dae, line 55: I think this needs to be aarch64. I had a user report that arm64 did not work as an alias in these constraints - not sure if that also applies to the darwin_amd64 build as well.

mattem: I put the wrong one for linux here (copy paste). Does `arm64` work? I don't have a machine to test this.

dae: Me either :-) But the user reported arm64 did not work: https://github.com/ankitects/esbuild_toolchain/pull/2

mattem: Oh intersting, I see, thanks

PR #2705, opened 2021-05-26
chore: setup pre-commit hook for buildifier

alexeagle:


package.json
```diff
@@ -137,8 +133,7 @@
     },
     "husky": {
         "hooks": {
-            "pre-commit": "check-clang-format \"yarn format\"",
-            "pre-push": "yarn bazel:format-check && yarn bazel:lint",
+            "pre-push": "check-clang-format \"yarn format\"",
```
mattem, line 136: Does this include buildifier fixes? I think my biggest pain point from the last version was the different checks happening on commit and push.

alexeagle: no, husky is no longer involved with buildifier at all

mattem (approved):

PR #2706, opened 2021-05-26
chore: change clang-format hook to pre-commit

alexeagle:
No longer stitches it with npm ecosystem or with husky. We continue to have no CI verification either, formatting isn't that consequential


alexeagle:
I'm not sure this is working right. I can't get it to repair trailing commas and missing semicolons like the old version did. I confirmed that it reads our config file (it silently fails if it doesn't understand the config)
Might bail on this again like I did for prettier, just not worth the time

PR #2714, opened 2021-05-28
feat(typescript): produce JSNamedModuleInfo

alexeagle:
This allows concatjs use cases to depend on ts_project targets rather than ts_library.
Update one example to illustrate

### DIFF
--- a/examples/web_testing/BUILD.bazel
+++ b/examples/web_testing/BUILD.bazel
@@ -1,22 +1,42 @@
+load("@build_bazel_rules_nodejs//:index.bzl", "js_library")
 load("@npm//@bazel/concatjs:index.bzl", "karma_web_test_suite")
-load("@npm//@bazel/typescript:index.bzl", "ts_config", "ts_library")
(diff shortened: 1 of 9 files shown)


gregmagolan (approved):

alexeagle:
@gregmagolan one thought about this one before merging...

I think we could probably accomplish this without adding a feature to ts_project, by using this undocumented attribute
https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/js_library/js_library.bzl#L52-L58
basically each `ts_library` here would need to be wrapped in a js_library that takes the .js files as `named_module_srcs` and that causes it to expose a JSNamedModuleInfo. What's more, I think we could use  that amd shim generation tucked in `js_library` rather than add the triple-slash directives to all the TS source files.

OTOH it seems like we left those js_library features undocumented for a reason.

Which way do we want to document in the examples?

gregmagolan (approved):

alexeagle:
Replaced with #2729

PR #2717, opened 2021-05-28
refactor: convert from_source example to ts_project

alexeagle:
We no longer recommend ts_library so we shouldn't use it in examples

### DIFF
--- a/examples/from_source/index.bzl
+++ b/examples/from_source/index.bzl
@@ -1,22 +1,24 @@
 """Example package for generating stardoc from rules_nodejs at source"""
 
-load("@build_bazel_rules_nodejs//packages/typescript:index.bzl", "ts_library")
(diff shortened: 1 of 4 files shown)


gregmagolan (approved):

examples/from_source/index.bzl
```diff
@@ -1,6 +1,6 @@
 """Example package for generating stardoc from rules_nodejs at source"""
 
-load("@build_bazel_rules_nodejs//packages/typescript:index.bzl", "ts_library")
+load("@build_bazel_rules_nodejs//packages/typescript:index.bzl", "ts_project")
 
 def custom_ts_library(name, deps = [], **kwargs):
```
gregmagolan, line 5: custom_ts_project instead or refactor and get rid of the macro?

alexeagle: the macro is here to give coverage for the stardoc use case. I'll just fix the naming.

PR #2721, opened 2021-05-31
test(esbuild): add ts_project test

jbedard:


alexeagle (commented):
I think we should just replace the ts_library calls with ts_project. We'd like to deprecate ts_library and I don't think it's worth carrying forward the maintenance burden of keeping parallel test fixtures running.

jbedard:
I figured it's worth keeping around for now just to compare ts_project vs library and it also shows an example of where they differ as an example for people. I can just delete it instead if you want? 🤷

alexeagle:
yeah we should keep that testing since we determined it shouldn't be too hard to keep it working

packages/esbuild/test/ts_project/BUILD.bazel
```diff
@@ -0,0 +1,42 @@
...
+esbuild(
+    name = "bundle",
+    args = [
+        "--keep-names",
+        "--resolve-extensions=.mjs,.js",
+    ],
+    entry_point = "main.ts",
+    format = "esm",
+    deps = [":main"],
+)
```
mattem, line 29: Perhaps we should base this branch on 4.x, the API here has changed

jbedard: Done

mattem (approved):

alexeagle:
CI failure on windows still looks legit

```


01:39:08) ERROR: C:/b/bk-windows-trpz/bazel/rules-nodejs-nodejs/packages/esbuild/test/ts_project/BUILD.bazel:7:11: Compiling TypeScript project //packages/esbuild/test/ts_project:main [tsc -p packages/esbuild/test/ts_project/tsconfig.json] failed: (Exit 1): tsc.bat failed: error executing command
--


error
 TS5055: Cannot write file 
'C:/b/slwvhzqq/execroot/build_bazel_rules_nodejs/bazel-out/x64_windows-fastbuild/bin/packages/esbuild/test/ts_project/module-two/index.d.ts'
 because it would overwrite input file.
```

PR #2722, opened 2021-05-31
test(rollup): add ts_library, ts_project, alias-mapping tests

jbedard:


alexeagle:
You'll have an easier time having green PRs if you `pre-commit install` to get the buildifier hook, see https://github.com/bazelbuild/rules_nodejs/blob/stable/DEVELOPING.md#for-developers
(we recently changed it from husky)

alexeagle:
you can probably reproduce that windows failure locally by building with `--spawn_strategy=local`

jbedard:
Yeah that reproduces it, same with the esbuild version of the test in the other PR. Files are under `execroot/build_bazel_rules_nodejs` instead of `bazel-out`. Any ideas?

alexeagle (approved):

PR #2731, opened 2021-06-02
refactor: switch to using json.encode rather than to_json on structs

mattem:
**BREAKING CHANGE**

Switch to using `json.encode` rather than `to_json` on structs.
Can't yet flip `--incompatible_struct_has_no_methods` as other dependencies use the deprecated methods.

As the `json` and `proto` modules are only available in bazel 4.x (LTS) and above, this change removes support for any versions below 4.0.0.


alexeagle:
Also change the root spot (index.bzl) that lists the versions we test against?
And there's a docs spot that lists supported Bazel versions

alexeagle (approved):

PR #2732, opened 2021-06-02
refactor(esbuild): refactor esbuild rule to use the esbuild API

mattem:
**BREAKING CHANGE**

Refactors the esbuild rule to use esbuild's API, rather than calling the CLI directly. This serves as a prefactor PR for adding plugin support.

`args` now takes a dict (was `string_list`), where the keys are the arg names from the JS API for esbuild.
`define` now takes a dict (was `string_list`), following the format of the `define` esbuild JS API.

### DIFF
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -2,11 +2,11 @@
 esbuild rule
 """
 
-load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_binary")
+load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_binary", "params_file")
 load("@build_bazel_rules_nodejs//:providers.bzl", "ExternalNpmPackageInfo", "JSEcmaScriptModuleInfo", "JSModuleInfo", "node_modules_aspect", "run_node")
 load("@build_bazel_rules_nodejs//internal/linker:link_node_modules.bzl", "MODULE_MAPPINGS_ASPECT_RESULTS_NAME", "module_mappings_aspect")
 load("@build_bazel_rules_nodejs//packages/esbuild/toolchain:toolchain.bzl", "TOOLCHAIN")
-load(":helpers.bzl", "desugar_entry_point_names", "filter_files", "generate_path_mapping", "resolve_entry_point", "write_jsconfig_file")
+load(":helpers.bzl", "desugar_entry_point_names", "filter_files", "generate_path_mapping", "resolve_entry_point", "write_args_file", "write_jsconfig_file")
 
 def _esbuild_impl(ctx):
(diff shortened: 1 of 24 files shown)


packages/esbuild/esbuild.bzl
```diff
@@ -45,52 +45,54 @@ def _esbuild_impl(ctx):
 
     # TODO(mattem): 4.0.0 breaking change, entry_points must exist in deps, and are not considered additional srcs
```
alexeagle, line 46: is this TODONE?

packages/esbuild/esbuild.bzl
```diff
@@ -164,21 +171,26 @@ def _esbuild_impl(ctx):
...
+            doc = """A dict of extra arguments that are included in the call to esbuild, where the key is the argument name.
+Values are subject to $(location ...) expansion""",
         ),
-        "define": attr.string_list(
-            default = [],
-            doc = """A list of global identifier replacements.
+        "args_file": attr.label(
+            allow_single_file = True,
+            mandatory = False,
+            doc = "A JSON file containing additional arguments that are passed to esbuild. Note: only args or args_file may be set",
```
alexeagle, line 182: nit: only *one* of args or args_file

packages/esbuild/esbuild.bzl
```diff
@@ -337,12 +349,36 @@ def esbuild_macro(name, output_dir = False, **kwargs):
...
 
+    srcs = kwargs.pop("srcs", [])
+    deps = kwargs.pop("deps", []) + ["@esbuild_npm//esbuild"]
     entry_points = kwargs.get("entry_points", None)
+
+    args = kwargs.pop("args", {})
+    args_file = kwargs.pop("args_file", None)
+
+    if args and args_file:
+        fail("Both 'args' and 'args_file' attributes set, these are mutually excluside")
```
alexeagle, line 360: spelling: exclusive

packages/esbuild/test/css/main.ts
```diff
@@ -1,3 +1,5 @@
 import { dep } from "./dep";
 
 dep();
+
+console.log(process.cwd())
```
alexeagle, line 5: intentional?

alexeagle (approved):

packages/esbuild/esbuild.bzl
mattem: Not yet, will do in another PR

packages/esbuild/test/css/main.ts
mattem: Perhaps I'll split this into another test. The input has a shim defined for `process.cwd` so it should result in an empty string, the test asserts that the shim is picked up correctly.

mattem: Moved to the define test, make a little more sense there

PR #2739, opened 2021-06-03
refactor: use ts_project in cypress example

alexeagle:
We no longer recommend ts_library so we shouldn't use it in our examples


mrmeku (approved):

mrmeku:
Ping me when you merge this so I can update my other cypress PR

PR #2743, opened 2021-06-04
docs: update supported bazel versions in README

mattem:
Update the root README with the bazel version support choices.


alexeagle (approved):
thanks!

PR #2746, opened 2021-06-07
docs(rollup): explain that the stamp variable is magically replaced

alexeagle:


jbedard (approved):

packages/rollup/index.bzl
```diff
@@ -143,6 +143,7 @@ function parseStatusFile(p) {
   return results;
 }
 
+// This undefined variable will be replaced with the full path during the build.
```
jbedard, line 146: Could maybe say "the undefined 'bazel_version_file' variable" to be a bit more clear 🤷

packages/rollup/index.js
```diff
@@ -239,6 +239,6 @@ async function main(args) {
   }
 }
 
-if (require.main === module) {
-  main(process.argv.slice(2)).then(r => (process.exitCode = r));
-}
+if (require.main == module) {
+  (async () => process.exitCode = await main(process.argv.slice(2)))();
```
jbedard, line 243: Does this cause any actual change? How about the `==` change above?

alexeagle: no, just style to match other files while I was here

packages/rollup/index.bzl
alexeagle: that's a good idea but I don't want to wait for CI again. Maybe I should make CI faster? :)

jbedard: yes you should :P

PR #2749, opened 2021-06-08
feat: support dict style directory_file_path entry_point in nodejs_binary, nodejs_test & jasmine_node_test

gregmagolan:
This PR adds syntactical sugar to nodejs_binary, nodejs_test & jasmine_node_test entry_point such that you can specify a directory_file_path entry point with

```
entry_point = { "//directory/artifact:target": "file/path/main.js" }
```

Pre-factor to https://github.com/bazelbuild/rules_nodejs/pull/2698


alexeagle (approved):

internal/node/node.bzl
```diff
@@ -637,6 +638,13 @@ nodejs_binary = rule(
     ],
 )
 
+def nodejs_binary_macro(name, **kwargs):
+    nodejs_binary(
+        name = name,
+        entry_point = maybe_directory_file_path(name, kwargs.pop("entry_point", None)),
```
alexeagle, line 644: it's annoying that the stardoc generation now cannot take this macro into account. The simplest thing is to document in entry_point "even though we said it's a Label, surprise! it can also be a dict"
The hardest thing is to be like `ts_project` and move all the docs to the macro.

PR #2751, opened 2021-06-08
fix(builtin): allow for only stderr to be set on npm_package_bin

gregmagolan:
Technically, only stderr output is valid. I had this configuration briefly https://github.com/bazelbuild/rules_nodejs/pull/2698 but then refactored it out. This bug fix came out of that.


alexeagle:
@mattem and I are both surprised there's a use case for that

gregmagolan:
> @mattem and I are both surprised there's a use case for that

I wouldn't argue it is a good one, but you could technically write a test to assert on the stdout of a tool

mattem (approved):

PR #2760, opened 2021-06-09
fix: regex in label replacement for stardoc

bweston92:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

Warning with current regex pattern:
```
/ An unescaped delimiter must be escaped with a backslash (\)
```

### DIFF
--- a/tools/stardoc/post-process-docs.js
+++ b/tools/stardoc/post-process-docs.js
@@ -7,7 +7,7 @@ const content = readFileSync(md, {encoding: 'utf8'});
 // @npm is not the required name, but it seems to be the common case
 // this reflects the similar transformation made when publishing the packages to npm
 // via pkg_npm defined in //tools:defaults.bzl
-const out = content.replace(/(?:@.*)*?\/\/packages\/([^/:"\s]*)/g, (str, pkg) => {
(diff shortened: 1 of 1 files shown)


alexeagle:
It's correctly getting replaced in the checked-in docs
https://github.com/bazelbuild/rules_nodejs/blob/3.5.1/docs/TypeScript.md#typical-usage
so I think the regex is already working, when it gets run

alexeagle:
I also think that unescaped slash looks fishy. It's a bit annoying to verify, because the target that runs this post-processor also runs Jekyll to do the GH Pages rendering, and that requires having non-hermetic stuff installed on your machine via Ruby gems.
On the 4.x branch things are simpler as we've removed the GitHub pages site. Maybe try this change over there and verify whether it actually results in different docs output?
Or even easier just paste this into a web-based regex tester and check? Possibly in the JS regex syntax, when you're inside a character group you don't need to escape slashes?

alexeagle:
I think it's no longer needed as we're about to cut over to the 4.x branch

PR #2761, opened 2021-06-09
chore: update to bazel 4.1.0

alexeagle:


alexeagle:
replaced with smaller scope change https://github.com/bazelbuild/rules_nodejs/pull/2792

PR #2766, opened 2021-06-11
perf(cypress): export cypress files as a single symlink

mrmeku:
Rather than using a glob to expose the extract cypress files, we expose a symlink to the directory in which those files live. This has two main benefits:

- Reduces the amount of time to construct a runfiles tree for a cypress test
- Removes the need for --experimental_inprocess_symlink_creation on MacOS

### DIFF
--- a/examples/BUILD.bazel
+++ b/examples/BUILD.bazel
@@ -110,13 +110,15 @@ example_integration_test(
 
 example_integration_test(
     name = "examples_cypress",
-    bazel_commands = ["test --sandbox_debug //..."],
+    bazel_commands = ["test //..."],
     npm_packages = {
         "//packages/cypress:npm_package": "@bazel/cypress",
         "//packages/typescript:npm_package": "@bazel/typescript",
     },
     owners = ["@mrmeku"],
-    tags = ["cypress"],
+    tags = [
+        "cypress",
+    ],
 )
 
 example_integration_test(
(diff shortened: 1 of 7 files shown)


lencioni (approved):
Are there any drawbacks to this?

mrmeku:
There shouldn't be. I believe it has the exact same functionality as before

examples/BUILD.bazel
```diff
@@ -110,13 +110,15 @@ example_integration_test(
 
 example_integration_test(
     name = "examples_cypress",
-    bazel_commands = ["test --sandbox_debug //..."],
+    bazel_commands = ["test //..."],
```
alexeagle, line 113: nit: this is the default

packages/cypress/internal/cypress_repositories.bzl
```diff
@@ -45,17 +45,16 @@ def cypress_repositories(
         urls = windows_urls + [
             "https://cdn.cypress.io/desktop/{}/win32-x64/cypress.zip".format(version),
         ],
-        strip_prefix = "Cypress",
         build_file_content = """
 filegroup(
     name = "files",
-    srcs = glob(["**"]),
+    srcs = ["Cypress"],
```
alexeagle, line 51: nit: note that this is a directory used as an InputArtifact?

alexeagle (approved):

alexeagle:
My guess from the error message is this is fixable, maybe in nodejs_image. I wonder what was the previous path of the cypress binary compared with now - I imagine it was in cypress_linux either way. But now our toolchain definition means there's more than one file in that directory?

hey @purkhusid do you mind opening a separate issue for that? I only happened to notice this comment on a closed PR.

purkhusid:
Sure, issue created: https://github.com/bazelbuild/rules_nodejs/issues/2893

Issue #2338, opened 2020-12-08
`jest_test` fails to locate test files (when used with `ts_project`?)


chrisallenlane:
# 🐞 bug report

### Affected Rule
The issue likely pertains to `jest_test`, and possibly to `ts_project`.

### Is this a regression?
Unknown.

### Description
I am encounting surprising behavior when using `jest_test` with `ts_project`, and suspect I may have encountered a bug in `rules_node`.

I am attempting to create a TypeScript monorepo that utilizes `jest` for testing. I've been modelling my work after the [jest examples][], though I'm attempting to use `ts_project` instead of `ts_library`, as I've read that the former is preferred. With that said, `bazel test ...` is not performing as expected.

My project builds and runs successfully. However, when I run `bazel test ...`, `jest` appears to be unable to locate test files. 

I had suspected that this was due to a configuration error on my part. I have discovered, however, that the tests pass when run as follows:

```sh
cd bazel-bin/ts/lib/greet && ./ts_greet_test.sh
```

(`ts_greet_test` is the name of the test in question.)

Given that the `ts_greet_test.sh` file is (I believe) machine-generated, can this problem still be a mistake on my part, or may I have encountered an issue with `jest_test` (and perhaps `ts_project`)?

## 🔬 Minimal Reproduction
A minimal reproduction is available here:
https://github.com/sharpspring/jest-test-repro

## 🔥 Exception or Error
Verbose logs are available here:
https://gist.github.com/chrisallenlane/1058a1001d1b89032ceecdc5fa89bb7c

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
ProductName:	Mac OS X
ProductVersion:	10.14.6
BuildVersion:	18G3020
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
Build label: 3.7.1-homebrew
Build target: bazel-out/darwin-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Tue Nov 24 20:20:36 2020 (1606249236)
Build timestamp: 1606249236
Build timestamp as int: 1606249236
  </code>
</pre>

**Rules_nodejs version:**

<pre>
  <code>
2.3.0
  </code>
</pre>

**Anything else relevant?**
I am relatively new to `bazel`, and apologize if I've missed anything obvious here.

Thanks very much for your help.

[...]

PR #2768, opened 2021-06-11
chore: upgrade jest example to jest@17

cuyl:
replace the patch with setting enableSymlinks to true. ref: #2338


mrmeku (approved):
Make sure to run buildifier to format your build files. That'll get CI to pass

alexeagle (approved):
Woohoo! Thanks!
We also have a copy of that jest-haste-map in the /patches folder, I guess we use it internally somewhere as well - if you have time for a second PR for that.

PR #2772, opened 2021-06-11
feat(esbuild): filter ts declaration files by default

mrmeku:
Transitive targets which produce only declarations can be skipped,
making the round trip time from editing a .ts(x) file to seeing
that change bundled greately reduced since tsc is not invoked.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Esbuild includes .d.ts files emitted by ts_project which requires tsc be run before esbuild

- [x] Yes
- [ ] No

### DIFF
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -45,6 +45,9 @@ def _esbuild_impl(ctx):
 
     # TODO(mattem): 4.0.0 breaking change, entry_points must exist in deps, and are not considered additional srcs
     inputs = deps_inputs + ctx.files.srcs + filter_files(entry_points)
+
+    inputs = [d for d in inputs if not (d.path.endswith(".d.ts") or d.path.endswith(".tsbuildinfo"))]
+
     outputs = []
 
     args = dict({


packages/esbuild/esbuild.bzl
```diff
@@ -233,6 +236,13 @@ and cjs when platform is node. If performing code splitting or multiple entry_po
 See https://esbuild.github.io/api/#format for more details
         """,
         ),
+        "include_ts_declarations": attr.bool(
```
alexeagle, line 239: can we think of a use case where you'd need to set this to True? I'd rather not introduce the user-facing complexity of more configuration affordances if they're likely not useful and just out of FUD

mrmeku: One use case I've though of is if you want your JS to bundle to invoke TSC. For example, at this point I've made my one target filter out tsc outputs and not invoke the compiler but I've made a separate target invoke not filter those outputs out so that I can run that if I want the compiler to run during my dev cycle

alexeagle: I don't understand still. If your workflow is "I want to ensure my code type-checks" then why would you `bazel build :some-esbuild-target` to do that? The esbuild target still does nothing with the extra inputs.

mrmeku: Done

mattem (approved):

PR #2783, opened 2021-06-21
feat(builtin): add validate attribute on pkg_npm

### LINKED ISSUE #2782: Usability: no validation that pkg_npm#package_name matches package.json#name
Our intent with the `package_name` attribute on pkg_npm is so that our analysis phase can know this info without having to read the package.json file, which we can do only inside an action during the execution phase.
However it's easy for users to trip on this. In particular in pairing with @Smokrow just now, we didn't have the `package_name` attribute defined at all, silently making this package not linkable (it didn't provide LinkablePackageInfo)

alexeagle:
Currently just validates that the package_name attribute matches the name in package.json (if there is one)

Fixes #2782


gregmagolan (approved):
🌯

PR #2784, opened 2021-06-21
chore(builtin): flip default for pkg_npm#validate

alexeagle:
BREAKING CHANGE:
Just follow the printed instructions to either set the right package_name or disable the validation


gregmagolan (approved):
🌮

PR #2793, opened 2021-06-27
fix(builtin): fix npm_install & yarn_install post_install_patches when symlink_node_modules is enabled

gregmagolan:
Resolves https://github.com/bazelbuild/rules_nodejs/issues/2774.

It is recommended that user's do not do this since `node_modules` are re-used between runs of the repository rule and patches will run on node_modules files that may already be patched and fail to apply.

Updated docstring with this note and recommended that users use `patch-packages` instead when `symlink_node_modules` is enabled:

```
NB: If `symlink_node_modules` is enabled, the node_modules folder is re-used between executions of the
    repository rule. Patches may be re-applied to files in this case and fail to apply. A marker file
    `node_modules/.bazel-post-install-patches` is left in this mode when patches are applied. When the 
    marker file is detected, patch file failures are treated as WARNINGS. For this reason, it is recommended
    to patch npm packages with an npm tool such as https://www.npmjs.com/package/patch-package when
    `symlink_node_modules` is enabled which handles re-apply patching logic more robustly.
```

This PR turns off support for `pre_install_patches` with `symlink_node_modules` enabled since the patches will run in your workspace and patches would modify files in your clone. `pre_install_patches` runs before there is a `node_modules` folder so the limited use case of patching npm packages in `node_modules` does not apply in this case. As such, only with `symlink_node_modules` disabled is patching data files that are inputs to the `yarn install` or `npm install` non-destructive to the user workspace.


internal/npm_install/npm_install.bzl
```diff
@@ -375,24 +438,41 @@ def _add_scripts(repository_ctx):
...
+
+def _workspace_rerooted_path(repository_ctx, f):
+    segments = [_WORKSPACE_REROOTED_PATH]
     if f.package:
         segments.append(f.package)
     segments.append(f.name)
     return "/".join(segments)
 
-def _workspace_root_prefix(repository_ctx):
+def _workspace_root(repository_ctx):
```
alexeagle, line 453: it's a bit confusing that this is a "workspace root" where it's actually a package.json dirname, right? is that an npm workspace, a bazel workspace, or likely neither?

internal/npm_install/npm_install.bzl
```diff
@@ -298,7 +310,51 @@ data attribute.
...
+def _apply_post_install_patches(repository_ctx):
+    if len(repository_ctx.attr.post_install_patches) == 0:
+        return
+    if repository_ctx.attr.symlink_node_modules:
+        print("\nWARNING: @%s post_install_patches with symlink_node_modules enabled will run in your workspace and potentially modify source files" % repository_ctx.name)
+    working_directory = _workspace_root(repository_ctx) if repository_ctx.attr.symlink_node_modules else _WORKSPACE_REROOTED_PATH
+
+    # marker file used when repository_ctx.attr.symlink_node_modules is True
+    marker_file = "%s/node_modules/.bazel-post-install-patches" % repository_ctx.path(repository_ctx.attr.package_json).dirname
+    allow_failure = True if repository_ctx.attr.symlink_node_modules and _is_file(repository_ctx, marker_file) else False
```
alexeagle, line 348: since there's a bash dependency anyway, would it simplify the logic a bit to have a larger bash program for patches, with less starlark code? In particular maybe you don't need the write_file and is_file helpers in starlark since then you have the boilerplate of calling multiple subcommands

internal/npm_install/npm_install.bzl
```diff
@@ -298,7 +310,51 @@ data attribute.
...
+        return
+    if repository_ctx.attr.symlink_node_modules:
+        fail("pre_install_patches cannot be used with symlink_node_modules enabled")
+    _apply_patches(repository_ctx, _WORKSPACE_REROOTED_PATH, repository_ctx.attr.pre_install_patches)
+
+def _apply_post_install_patches(repository_ctx):
+    if len(repository_ctx.attr.post_install_patches) == 0:
+        return
+    if repository_ctx.attr.symlink_node_modules:
+        print("\nWARNING: @%s post_install_patches with symlink_node_modules enabled will run in your workspace and potentially modify source files" % repository_ctx.name)
```
alexeagle, line 343: since we are close to 4.0, could we simplify the allowed configurations a bit and make it a breaking change that disallows this behavior instead? I think we only introduced pre/post_install_patches for the uncommon use case of needing a patch installed for the bazel workflow while leaving the npm workflow untouched, right?

alexeagle (approved):

internal/npm_install/npm_install.bzl
gregmagolan: actually, this one happens to be the root of the user workspace which is where patches need to run from if symlink_node_modules is True. I'll ponder on better naming

internal/npm_install/npm_install.bzl
gregmagolan: Yes, it was intended to makes bazel-only workflow patches if and when supporting dual bazel & legacy builds.

The post_install_patches can also be used as a patch-package alternative but it gets complicated when symlink_node_modules is True

internal/npm_install/npm_install.bzl
gregmagolan: good idea. done.

internal/npm_install/npm_install.bzl
gregmagolan: Currently, using post_install_patches or pre_install_patches with symlink_node_modules True just fails without a good error message. This PR allows post_install_patches through and disallows pre_install_patches. Neither is a breaking change. We could alternately disallow post_install_patches as well if symlink_node_modules is True tho it looks like some users would like to use it in that shape given https://github.com/bazelbuild/rules_nodejs/issues/2774

internal/npm_install/npm_install.bzl
gregmagolan: renamed the set of functions and added some comments to make clearer

PR #2794, opened 2021-06-27
chore: update Bazel minimum version to LTS

alexeagle:
BREAKING CHANGE:
for our 4.0, rules_nodejs requires the current LTS of Bazel. This is a policy restriction to save us time tracking down support issues on older versions of Bazel, not a technical one. You can patch this check out locally if you really need to continue using older Bazel, but this puts you on unsupported track.


alexeagle:
@googlebot I fixed it.

alexeagle:
@googlebot I consent.

alexeagle:
@googlebot I consent.

gregmagolan (approved):

PR #2795, opened 2021-06-27
chore: update default node.js to current LTS

alexeagle:
This matches 'Recommended for most Users' on nodejs.org/en/download


alexeagle:
@mattem maybe we should update all the way to 16 as the default? It is not yet "active LTS" per https://nodejs.org/en/about/releases/ but it would make it easier for Mac developers who are going to need it on their new machines now.

mattem:
Hmm, I'd be tempted to keep it at LTS, or as a breaking change remove the default so we are not the ones to decide.

alexeagle:
@googlebot I fixed it.

gregmagolan (approved):

Issue #2450, opened 2021-02-05
Add package_name to ts_library so that module_name no longer enables linking


gregmagolan:
This is a breaking change so needs to be sequenced with 4.0.

It will allow us to remove the `__tslibrary__` special handling in `internal/linker/link_node_modules.bzl`:

``
    # Special case for ts_library module_name for legacy behavior and for AMD name work-around
    # Tag the mapping as "__tslibrary__" so it can be overridden by any other mapping if found.
    #
    # In short, ts_library module_name attribute results in both setting the AMD name (which
    # desired and necessary in devmode which outputs UMD) and in making a linkable mapping. Because
    # of this, you can get in the situation where a ts_library module_name and a downstream pkg_name
    # package_name conflict and result in duplicate mappings. This work-around will make this
    # situation work however it is not a recommended pattern since a ts_library can be a dep of a
    # pkg_npm but not vice-verse at the moment since ts_library cannot handle directory artifacts as
    # deps.
    if hasattr(linkable_package_info, "_tslibrary") and linkable_package_info._tslibrary:
        mr[0] = "__tslibrary__"

     ...

    # Simplify mappings before writing them to the manifest file.
    # The extra complication is required in bzl to handle the __tslibrary__ special
    # case since module_name in ts_library enables the linker. In a future major release,
    # ts_library will get a package_name attribute to enable the linker and the __tslibrary__
    # special case can be factored out.
    for k, v in mappings.items():
        mappings[k] = v[1]
```

PR #2799, opened 2021-06-30
feat: add package_name & package_path to ts_library

gregmagolan:
Resolves https://github.com/bazelbuild/rules_nodejs/issues/2450

BREAKING CHANGES:

`module_name` will no longer turn on linking for the `ts_library` target; instead `package_name` must now be specified to enable linking. `package_path` may also be specified to control the link location.

This PR includes a breaking change to simplify `module_mappings_aspect` in `internal/linker/link_node_modules.bzl` now that a `ts_library` special case is no longer needed.

NB: the alternative would be to remove linking support from `ts_library` entirely as a `js_library` or `npm_link` could be used instead, however, given that `ts_library` doesn't expose its `.js` output as the default output this would require users to use a `filegroup` with an output selector to maintain the same functionality as `ts_library` currently has with `module_name`. This PR keeps the BREAKING CHANGE to just adding a `package_name` attribute to make the `ts_library` `js` outputs linkable.

### DIFF
--- a/examples/angular/src/lib/shorten/BUILD.bazel
+++ b/examples/angular/src/lib/shorten/BUILD.bazel
@@ -6,6 +6,7 @@ package(default_visibility = ["//:__subpackages__"])
 # with a main field to show how to create a first-party npm lib with a package.json
 ts_library(
     name = "shorten",
+    package_name = "@bazel/shorten",
     srcs = ["index.ts"],
     module_name = "@bazel/shorten",
 )
--- a/examples/angular_view_engine/src/lib/shorten/BUILD.bazel
+++ b/examples/angular_view_engine/src/lib/shorten/BUILD.bazel
@@ -6,6 +6,7 @@ package(default_visibility = ["//:__subpackages__"])
 # with a main field to show how to create a first-party npm lib with a package.json
 ts_library(
     name = "shorten",
+    package_name = "@bazel/shorten",
     srcs = ["index.ts"],
     module_name = "@bazel/shorten",
 )
--- a/internal/linker/link_node_modules.bzl
+++ b/internal/linker/link_node_modules.bzl
@@ -39,29 +39,17 @@ def add_arg(args, arg):
 def _link_mapping(label, mappings, k, v):
     # Check that two package name mapping do not map to two different source paths
     package_name = k.split(":")[0]
-    source_path = v[1]
+    link_path = v
 
-    # Special case for ts_library module_name for legacy behavior and for AMD name work-around
-    # Mapping is tagged as "__tslibrary__".
-    # See longer comment below in _get_module_mappings for more details.
-    if v[0] != "__tslibrary__":
-        for iter_key, iter_values in mappings.items():
-            # Map key is of format "package_name:package_path"
-            # Map values are of format [deprecated, source_path]
-            iter_package_name = iter_key.split(":")[0]
-            iter_source_path = iter_values[1]
-            if iter_values[0] != "__tslibrary__" and package_name == iter_package_name and source_path != iter_source_path:
-                fail("conflicting mapping at '%s': '%s' and '%s' map to conflicting %s and %s" % (label, k, iter_key, source_path, iter_source_path))
+    for iter_key, iter_values in mappings.items():
+        # Map key is of format "package_name:package_path"
+        # Map values are of format [deprecated, link_path]
+        iter_package_name = iter_key.split(":")[0]
+        iter_source_path = iter_values
+        if package_name == iter_package_name and link_path != iter_source_path:
+            fail("conflicting mapping at '%s': '%s' and '%s' map to conflicting %s and %s" % (label, k, iter_key, link_path, iter_source_path))
 
-    # Allow __tslibrary__ special case to be overridden other matching mappings
-    if k in mappings and mappings[k] != v:
-        if mappings[k][0] == "__tslibrary__":
-            return True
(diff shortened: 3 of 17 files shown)


packages/esbuild/test/bundle/bundle.golden.txt
```diff
@@ -1,8 +1,53 @@
 (() => {
+  var __create = Object.create;
```
alexeagle, line 2: whoa what happened here?

examples/angular_view_engine/src/lib/shorten/BUILD.bazel
```diff
@@ -6,6 +6,7 @@ package(default_visibility = ["//:__subpackages__"])
 # with a main field to show how to create a first-party npm lib with a package.json
 ts_library(
     name = "shorten",
+    package_name = "@bazel/shorten",
```
alexeagle, line 9: when should we drop the view engine example? is that still used by anyone in Angular land? @alan-agius4 ?

examples/angular/src/lib/shorten/BUILD.bazel
```diff
@@ -6,6 +6,7 @@ package(default_visibility = ["//:__subpackages__"])
 # with a main field to show how to create a first-party npm lib with a package.json
 ts_library(
     name = "shorten",
+    package_name = "@bazel/shorten",
     srcs = ["index.ts"],
     module_name = "@bazel/shorten",
```
alexeagle, line 11: should the `module_name` be removed now? or is that still required as the AMD module name?

internal/linker/link_node_modules.bzl
```diff
@@ -165,50 +149,27 @@ def _get_module_mappings(target, ctx):
...
 
-    if hasattr(linkable_package_info, "package_path") and linkable_package_info.package_path:
-        mn = "%s:%s" % (linkable_package_info.package_name, linkable_package_info.package_path)
-    else:
-        # legacy (root linked) style mapping
-        # TODO(4.0): remove this else condition and always use "%s:%s" style
-        mn = linkable_package_info.package_name
-    mr = ["__link__", linkable_package_info.path]
-
-    # Special case for ts_library module_name for legacy behavior and for AMD name work-around
```
alexeagle, line 184: :tada:

alexeagle (approved):

examples/angular/src/lib/shorten/BUILD.bazel
gregmagolan: can't yet at it is tied to ts_library internals; this PR makes it so that module_name no longer turns on the linker as it was causing issues downstream

packages/esbuild/test/bundle/bundle.golden.txt
gregmagolan: I don't know. @mattem can you take a look at this golden bundle change?

examples/angular_view_engine/src/lib/shorten/BUILD.bazel
gregmagolan: good question; I kept it green in this PR but I'm not against removing it as it is tied to angular v9

packages/esbuild/test/bundle/bundle.golden.txt
mattem: The import now comes via the linker I'm guessing.
Previously esbuild was picking up the `mjs` file from `bazel-out` where it's now finding the `js` in `node_modules`.
Ideally you'd set `resolveExtensions` to prefer the `mjs` over the `js`:

```
    args = {
        "resolveExtensions": [
            ".mjs",
            ".js",
        ],
    },
```
Perhaps we could make that the default again (it was previously, but we removed it to remove as much customization and use vanilla esbuild as much as possible).

This would then result in:
```
(() => {
  // node_modules/lib/index.mjs
  var NAME = "bazel";
  // bazel-out/darwin-fastbuild/bin/packages/esbuild/test/bundle/a.mjs
  console.log(`Hello ${NAME}`);
})();
//# sourceMappingURL=bundle.js.map
```

From the bundle metadata from your diff:
```
"inputs": {
    "node_modules/lib/index.js": {
      "bytes": 864,
      "imports": []
    },
    "bazel-out/darwin-fastbuild/bin/packages/esbuild/test/bundle/a.mjs": {
      "bytes": 520,
      "imports": [
        {
          "path": "node_modules/lib/index.js",
          "kind": "import-statement"
        }
      ]
    }
```

From HEAD on 4.x:
```
"inputs": {
    "bazel-out/darwin-fastbuild/bin/packages/esbuild/test/bundle/index.mjs": {
      "bytes": 395,
      "imports": []
    },
    "bazel-out/darwin-fastbuild/bin/packages/esbuild/test/bundle/a.mjs": {
      "bytes": 520,
      "imports": [
        {
          "path": "bazel-out/darwin-fastbuild/bin/packages/esbuild/test/bundle/index.mjs",
          "kind": "import-statement"
        }
      ]
    }
```

In esbuild, we generate the package mappings into a tsconfig file to ensure we pick them up from bazel-out, not node_modules. We could change this though.
https://github.com/bazelbuild/rules_nodejs/blob/4.x/packages/esbuild/esbuild.bzl#L36

gregmagolan: brilliant. worked as predicted. thanks!

not sure about making it default as this would only be an issue with ts_library which produces both right?

PR #2801, opened 2021-07-01
fix(builtin): runfile resolution incorrect if entry starts similarly

devversion:
Given the following runfile manifest:

```
src/lib/README.md /c/users/paul/projects/test/src/lib/README.md
src/lib/README.md.update <..>/bazel-out/<..>
```

If a developer intends to resolve the `README.md` file using
the `@bazel/runfiles` package, or the patched module resolution,
they will end up with a wrong path (at least on Windows where
runfiles are not symlinked).

This happens because the `Runfiles.resolve` method continues
looking for entries that start similar (even if we have a full-match).
This logic helps with finding directories, but breaks resolution
with exact matches. This commit also corrects the method name
because technically this method is used for more than resolution
of directories.

BREAKING CHANGE: The `@bazel/runfiles` `lookupDirectory` method has been
removed. Use the `resolve` method instead

### DIFF
--- a/internal/linker/index.js
+++ b/internal/linker/index.js
@@ -145,7 +145,7 @@ function resolveWorkspaceNodeModules(workspace, startCwd, isExecroot, execroot, 
         if (!execroot) {
             return path.resolve(`${startCwd}/../${targetManifestPath}`);
         }
-        const fromManifest = runfiles.lookupDirectory(targetManifestPath);
+        const fromManifest = runfiles.resolve(targetManifestPath);
         if (fromManifest) {
             return fromManifest;
         }
--- a/internal/linker/link_node_modules.ts
+++ b/internal/linker/link_node_modules.ts
@@ -189,7 +189,7 @@ async function resolveWorkspaceNodeModules(
   // If we got a runfilesManifest map, look through it for a resolution
   // This will happen if we are running a binary that had some npm packages
   // "statically linked" into its runfiles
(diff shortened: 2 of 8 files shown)


alexeagle:
Thanks @devversion !

I think since we are about to ship a first beta for 4.x it would be easier to just target that branch, and go ahead and make it a breaking change rather than go through a deprecation period.

Also if you install pre-commit.com and run `pre-commit install` in the project, you'll get the buildifier automation that CI is yelling at you about here :)

devversion:
@alexeagle Sounds good. Updated! 

<sup>CLA bot weirdly started thinking for a second (when I switched the base branch) that I lack CLA.. (commits were properly signed)</sup>

alexeagle (approved):
awesome as always 🙏🏻

alexeagle:
CI failure looks maybe legit?

```


(16:00:10) ERROR: C:/b/bk-windows-w50h/bazel/rules-nodejs-nodejs/packages/esbuild/test/sourcemap/BUILD.bazel:24:8: Bundling Javascript packages/esbuild/test/sourcemap/main.ts [esbuild] failed: (Exit 1): _bundle_inline_esbuild_launcher.bat failed: error executing command
--


Error:
 ENOENT: no such file or directory, open 
'C:\b\uhe4ogw6\execroot\build_bazel_rules_nodejs\packages\esbuild\node_modules\esbuild\lib\main.js'
```

will do a third retry

PR #2776, opened 2021-06-16
chore: remove jest-haste-map patch

cuyl:

PR #2806, opened 2021-07-02
chore: remove jest-haste-map patch

mrmeku:


alexeagle:
Let's stay in the contributor's original PR #2776

PR #2808, opened 2021-07-05
refactor: rename API provider field 'type_blacklisted_declarations' t…

mattem:
BREAKING CHANGE

Renames the field `type_blacklisted_declarations` to `type_blocklisted_declarations` within the `DeclarationInfo` provider API

### DIFF
--- a/docs/Providers.md
+++ b/docs/Providers.md
@@ -15,7 +15,7 @@ Users should not load files under "/internal"
 **USAGE**
 
 <pre>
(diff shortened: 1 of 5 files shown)


mrmeku (approved):

alexeagle (approved):

alexeagle:
hard to reason about the impact of this breaking change.

on the one hand, I think this is only used in conjunction with tsickle for generating Closure Compiler externs, which almost no users do.

on the other hand, it would be very hard if you did use some other ruleset that produces DeclarationInfo, to get matching versions of everything.

mattem:
Hm, we have other breaking changes that produce the tricky version alignment issue, so I'm not sure this alone is would be the source of that (future Matt will regret saying this)

PR #2809, opened 2021-07-05
docs: link to the 'main' URL for bazel.build docs

mattem:
Updates all bazel.build links to point to `main` rather than `master` URL segment


mrmeku:
I thought our primary branch was called stable rather than main. What does main refer to?

mattem:
This is for the bazel.build docs not our own, it has no bearing on rnj branching names.

mrmeku (approved):

PR #2820, opened 2021-07-15
feat: add package_json_remove and package_json_replace attributes to yarn_install & npm_install

gregmagolan:
For easier modification of package.json over pre_install_patches.

Targetting 4.x as this uses json.encode & json.decode which is only available in bazel 4.0.0+.

Tests added. Docstrings are:

```
    "package_json_remove": attr.string_list(
        doc = """List of `package.json` keys to remove before running the package manager.

Keys are '.' separated. For example, a key of `dependencies.my-dep` in the list corresponds to the `package.json`
entry,

{
    "dependencies": {
        "my-dep": "..."
    }
}

This can be used, for example, during a migration to remove first party file: deps that are required
for the non-bazel build but should not be installed via the package manager in the bazel build since
they will be reference as bazel targets instead.

NB: removals specified are performed after preinstall_patches so if you are using both then the patch file should be relative
to the source `package.json`. Non-existant keys are silently ignored.""",
    ),
    "package_json_replace": attr.string_dict(
        doc = """Map of `package.json` keys to values to replace or create before running the package mangager.

Keys are '.' separated. For example, a key of `scripts.postinstall` corresponds to the `package.json`
entry,

{
    "scripts": {
        "postinstall": "..."
    }
}

This can be used, for example, during a migration to override npm scripts such as preinstall & postinstall.

NB: replaces specified are performed after preinstall_patches so if you are using both then the patch file should be relative
to the source `package.json`.""",
    ),
```

### DIFF
--- a/internal/npm_install/npm_install.bzl
+++ b/internal/npm_install/npm_install.bzl
@@ -287,6 +287,46 @@ All transitive dependencies are given limited visibility, enforcing that all dir
 listed in the `package.json` file.
 """,
     ),
+    "package_json_remove": attr.string_list(
+        doc = """List of `package.json` keys to remove before running the package manager.
+
+Keys are '.' separated. For example, a key of `dependencies.my-dep` in the list corresponds to the `package.json`
+entry,
+
+```
+{
+    "dependencies": {
+        "my-dep": "..."
+    }
+}
+```
+
+This can be used, for example, during a migration to remove first party file: deps that are required
+for the non-bazel build but should not be installed via the package manager in the bazel build since
+they will be reference as bazel targets instead.
+
+NB: removals specified are performed after preinstall_patches so if you are using both then the patch file should be relative
+to the source `package.json`. Non-existant keys are silently ignored.""",
+    ),
+    "package_json_replace": attr.string_dict(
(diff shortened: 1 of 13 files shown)


internal/npm_install/npm_install.bzl
```diff
@@ -496,6 +536,56 @@ def _copy_file(repository_ctx, f):
...
+    package_json_contents = repository_ctx.read(package_json_path)
+    repository_ctx.file(
+        package_json_path + ".orig",
+        content = package_json_contents,
+    )
+    package_json = json.decode(package_json_contents)
+
+    # process package_json_replace attr
+    for replace_key in repository_ctx.attr.package_json_replace.keys():
+        replace_value = repository_ctx.attr.package_json_replace[replace_key]
```
mattem, line 558: `.items()` returns a list of key, value tuples, eg:

```suggestion
    for replace_key, replace_value in repository_ctx.attr.package_json_replace.items():
```

mattem (approved):

internal/npm_install/npm_install.bzl
gregmagolan: nice. thanks

PR #2826, opened 2021-07-21
feat: add link_root & source_directories attributes to js_library

gregmagolan:
`link_root`: 
```
Path to link to from the root of the workspace.
        
If unset, the target's package name is used.
```

`source_directories`:
```
A hint that srcs have directories since this cannot be detected in the rule context.

If True, all srcs and deps should be source files so that the linker can link to the source tree.

When set, DeclarationInfo is always exported since we can't scan into directory sources for file extensions. In the future,
if we can detect that a source is a directory this can be removed.
```

These two attributes allow for more control over js_library targets including the ability to make an an npm package alias to a 3rd party npm package with exports_directories_only True or False with a `js_library` in `npm_install` or `yarn_install` `manual_build_file_contents`. See `npm_deps.bzl` for examples,

```
        links = {
            "semver-alias": "@fine_grained_deps_yarn//:semver-alias",
        },
        manual_build_file_contents = """
load("@build_bazel_rules_nodejs//:index.bzl", "js_library")
js_library(
    name = "semver-alias",
    package_name = "semver-alias",
    srcs = [
        "node_modules/semver/package.json",
        "node_modules/semver/semver.js",
        "node_modules/semver/bin/semver",
    ],
    link_root = "node_modules/semver",
    visibility = ["//semver-alias:__pkg__"],
)
""",
```

```
        links = {
            "semver-alias": "@fine_grained_deps_yarn_directory_artifacts//:semver-alias",
        },
        manual_build_file_contents = """
load("@build_bazel_rules_nodejs//:index.bzl", "js_library")
js_library(
    name = "semver-alias",
    package_name = "semver-alias",
    srcs = ["node_modules/semver"],
    link_root = "node_modules/semver",
    visibility = ["//semver-alias:__pkg__"],
    source_directories = True,
)
""",
```

and other fixes...

- `$node_modules_dir$` removed in favor of using `source_directories = True`
- if `source_directories = True` then `js_library` will like to sources (same as it does when `$node_modules$` is set)
- if `source_directories = True` or output directories detected then `js_library` will provide a `DeclarationInfo` (since we can't detect if there are .d.ts files within in directory) so it can be used as a dep to `ts_project`
- fix to `strip_prefix` contained files check to allow `strip_prefix` to work for `js_library` targets in external repositories

### DIFF
--- a/internal/js_library/js_library.bzl
+++ b/internal/js_library/js_library.bzl
@@ -70,8 +70,24 @@ If package_path is not set the this will be the root node_modules of the workspa
     ),
     "srcs": attr.label_list(allow_files = True),
     "strip_prefix": attr.string(
-        doc = "Path components to strip from the start of the package import path",
+        doc = """Path components to strip from the start of the package import path.
+
+This path is appended to the link_root when calculating the link path.""",
         default = "",
+    ),
+    "link_root": attr.string(
+        doc = """Path to link to from the root of the workspace.
+        
+If unset, the target's package name is used.""",
+        default = "<package_name>",
+    ),
+    "source_directories": attr.bool(
+        doc = """A hint that srcs have directories since this cannot be detected in the rule context.
+
+If True, all srcs and deps should be source files so that the linker can link to the source tree.
+
+When set, DeclarationInfo is always exported since we can't scan into directory sources for file extensions. In the future,
+if we can detect that a source is a directory this can be removed.""",
     ),
 }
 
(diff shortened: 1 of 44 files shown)


mattem:
Perhaps I've missed something, but doesn't [is_directory](https://docs.bazel.build/versions/main/skylark/lib/File.html#is_directory) provide the functionality that `source_directories` is hinting at?

PR #2830, opened 2021-07-24
fix: add missing SHA for rules_proto

mattem:


mattem:
Windows failure is a flake ❄️

PR #2837, opened 2021-07-30
Don't use false as a no-error value

behos:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

Use `null` as the non-error value for the wrapped callbacks instead of `false`.

- [ ] Yes
- [x] No

## Other information

The convention is to use `null` when calling back with no error, and
this convention is being enforced in `universalify` since
version 2.0.0.

`fs-extras` uses `universalify` to build `Promises` from the callbacks,
and the wrapper in `node_patches.js` breaks this integration.

### DIFF
--- a/internal/node/node_patches.js
+++ b/internal/node/node_patches.js
@@ -102,7 +102,7 @@ const patcher = (fs = fs__default['default'], roots) => {
                 return origReadlink(args[0], (err, str) => {
                     if (err) {
                         if (err.code === 'ENOENT') {
-                            return cb(false, stats);
+                            return cb(null, stats);
                         }
                         else if (err.code === 'EINVAL') {
                             // readlink only returns einval when the target is not a link.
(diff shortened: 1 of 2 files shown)


mattem:
Thanks for the PR! Are you able to sign the CLA so we can accept the patch?

mattem (approved):

behos:
> Thanks for the PR! Are you able to sign the CLA so we can accept the patch?

Yes, I'm resolving this through my org and will update the PR once I am added in our policy

behos:
@googlebot I signed it!

behos:
The windows_examples CI task is failing, but I can't tell if it's caused by this change or if it's a transient error. Is there a way to rerun it?

mattem:
Very likely a flake. I think only @alexeagle has the powers to hit rebuild, only other way is to rebase if possible.

mattem:
Thanks!

PR #2840, opened 2021-07-31
feat(esbuild): add support for plugins via supplying a configuration file

### LINKED ISSUE #2832: Esbuild should support plugins
# 🚀 ESBuild should support plugins 

### Relevant Rules

esbuild 

### Description

 

Esbuild has concept of plugins for varied experiences. The Bazel rules should support them, both custom plugins and ones from from NPM. 

Perhaps there could be a hook to customize the config passed to Esbuild, or maybe the specifying of a custom config file. 

 

 Have you considered any alternative solutions or workarounds?

none available

mattem:
Adds support for plugins via consuming a (mostly) complete configuration block.

Users can specify a configuration file along with its dependencies on the `esbuild_config` macro, which is then referenced on the `config` attr on the `esbuild` rule.

Mostly all properties are allowed to be set within this config file, except those that matter to output tracking etc.
Plugins can either be loaded from npm or be local sources in the repo.

closes #2832

### DIFF
--- a/packages/esbuild/esbuild.bzl
+++ b/packages/esbuild/esbuild.bzl
@@ -152,6 +152,14 @@ def _esbuild_impl(ctx):
         inputs.append(ctx.file.args_file)
         launcher_args.add("--user_args=%s" % ctx.file.args_file.path)
 
+    if ctx.attr.config:
+        configs = ctx.attr.config[JSEcmaScriptModuleInfo].sources.to_list()
+        if len(configs) != 1:
+            fail("Expected only one source file: the configuration entrypoint")
+
+        inputs.append(configs[0])
+        launcher_args.add("--config_file=%s" % configs[0].path)
+
     run_node(
         ctx = ctx,
         inputs = depset(inputs),
@@ -188,7 +196,7 @@ Values are subject to $(location ...) expansion""",
         "args_file": attr.label(
             allow_single_file = True,
             mandatory = False,
-            doc = "A JSON file containing additional arguments that are passed to esbuild. Note: only one of args or args_file may be set",
+            doc = "Internal use only",
         ),
         "define": attr.string_dict(
             default = {},
@@ -333,6 +341,13 @@ edge16, node10, esnext). Default es2015.
 See https://esbuild.github.io/api/#target for more details
             """,
         ),
+        "config": attr.label(
+            providers = [JSEcmaScriptModuleInfo],
+            mandatory = False,
+            doc = """Configuration file used for esbuild, from the esbuild_config macro. Note that options set in this file may get overwritten.
+See https://github.com/bazelbuild/rules_nodejs/tree/stable/packages/esbuild/test/plugins/BUILD.bazel for examples of using esbuild_config and plugins.
+            """,
+        ),
     },
     implementation = _esbuild_impl,
     toolchains = [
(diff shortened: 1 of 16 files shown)


mattem:
Oh windows...

packages/esbuild/esbuild.bzl
```diff
@@ -152,6 +152,14 @@ def _esbuild_impl(ctx):
         inputs.append(ctx.file.args_file)
         launcher_args.add("--user_args=%s" % ctx.file.args_file.path)
 
+    if ctx.attr.config:
+        configs = ctx.attr.config[JSEcmaScriptModuleInfo].sources.to_list()
+        if len(configs) > 1:
```
alexeagle, line 157: nit: `len(configs) != 1` just in case it's actually zero

packages/esbuild/esbuild.bzl
```diff
@@ -152,6 +152,14 @@ def _esbuild_impl(ctx):
...
+            fail("Expected only one source file the configuration entrypoint")
```
alexeagle, line 158: grammar, feels like it needs a colon after "file"

packages/esbuild/esbuild.bzl
```diff
@@ -367,12 +380,11 @@ def esbuild_macro(name, output_dir = False, splitting = False, **kwargs):
     deps = kwargs.pop("deps", []) + ["@esbuild_npm//esbuild"]
     entry_points = kwargs.get("entry_points", None)
 
-    args = kwargs.pop("args", {})
     args_file = kwargs.pop("args_file", None)
+    if args_file:
+        fail("Setting 'args_file' is not supported, set 'config_file' instead")
```
alexeagle, line 385: todo/issue filed to remove the attribute later?

packages/esbuild/helpers.bzl
```diff
@@ -8,6 +8,9 @@ TS_EXTENSIONS = ["ts", "tsx"]
 JS_EXTENSIONS = ["js", "jsx", "mjs"]
 ALLOWED_EXTENSIONS = JS_EXTENSIONS + TS_EXTENSIONS
 
+def make_deps_target(name):
```
alexeagle, line 11: is this abstraction worth the indirection?

packages/esbuild/launcher.js
```diff
@@ -22,22 +24,82 @@ function getEsbuildArgs(paramsFilePath) {
   }
 }
 
+async function processConfigFile(configFilePath) {
+  const fulllConfigFileUrl = pathToFileURL(join(process.cwd(), configFilePath));
```
alexeagle, line 28: extra "l" in full

packages/esbuild/test/plugins/BUILD.bazel
```diff
@@ -0,0 +1,57 @@
+load("//:index.bzl", "generated_file_test", "js_library", "nodejs_binary", "npm_package_bin")
```
alexeagle, line 1: mention the path to this test folder in the docs somewhere since it serves as an example

alexeagle (approved):

packages/esbuild/launcher.js
mattem: That's just how full it is

packages/esbuild/esbuild.bzl
mattem: Yeah, we could remove this in 5.0 - although it's kinda nice not to have all the verbosity of the config file just for one extra config option a user may want to pass.
What I really want here is a way to have the attribute private, but still settable via the macro, therefore hiding it from the docs.

mattem:
Angular examples flaked, unrelated.

PR #2841, opened 2021-07-31
chore: update nodejs version used

mattem:
Use a newer version of NodeJS, lets see what happens


alexeagle:
@googlebot I consent.

alexeagle (approved):

PR #2846, opened 2021-08-03
chore: enable tests on buildkite rbe_ubuntu1604

gregmagolan:


.bazelci/presubmit.yml
```diff
@@ -363,7 +363,9 @@ tasks:
     build_flags:
     # TODO(gregmagolan): figure out how to install missing shared libs
     # Without this filter the @cypress external repository will be built and that build will fail due to missing shared libs.
-    - "--build_tag_filters=-cypress,-pkg_npm.pack"
+    - "--build_tag_filters=-cypress,-examples,-e2e,-no-rbe,-browser:chromium-local,-browser:firefox-local"
```
mattem, line 366: Have all these others always been disabled, or is that new? (if so, add a comment as to why they are disabled?)

mattem (approved):

.bazelci/presubmit.yml
gregmagolan: Turns out we were missing
```
    test_targets:
    - "//..."
```
so we were not running any tests on RBE. I'll some comments as to why we're not running this subset

PR #2849, opened 2021-08-05
refactor: remove ExternalNpmPackageInfo

gregmagolan:
4.x WIP


gregmagolan:
Checkmate on this one. Trying to link per top-level node_modules folder runs into too many existing constraints in the legacy require patches, legacy module mappings aspects and in ts_library

gregmagolan:
I'm going to switch focus to the new rules_nodejs core and start on a greenfield approach for this so we can support more intelligent linking and pnpm style install in the future

PR #2853, opened 2021-08-11
fix(builtin): correctly calculate pkg._directDependency when a mapped node_module is used

mistic:
In a `package.json` we can declare something close to a mapped package like `"x": "npm:x-canary@^2.0.0-canary.14",` which will actually create `node_modules/x` but the node_module pkg name will be `x-canary`. Due to the latter, the current logic will not consider the pkg to be a `direct dependency` and as such will mark its visibility as being restricted when `strict_visibilty = True` is set. 

That PR also marks a given pkg as a `direct dependency` if the `pkg._dir` could be found listed as a dependency in the `package.json`

### DIFF
--- a/internal/npm_install/generate_build_file.ts
+++ b/internal/npm_install/generate_build_file.ts
@@ -661,7 +661,7 @@ export function parsePackage(p: string, dependencies: Set<string> = new Set()): 
 
   // set if this is a direct dependency of the root package.json file
   // which is later used to determine the generated rules visibility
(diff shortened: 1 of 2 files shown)


mistic:
The CI failure is not related with that PR as it is the same happening on `stable`

alexeagle (approved):
thanks!
Would be nice to have a test case for this as well, but can be in a follow-up

PR #2857, opened 2021-08-12
feat(esbuild): support stamping via config file

### LINKED ISSUE #2812: Support substitutions in esbuild
# 🚀 feature request

### Relevant Rules

The [esbuild](https://docs.aspect.dev/bazelbuild/rules_nodejs/3.6.0/packages/esbuild/readme) rule

### Description

It would be nice if the esbuild supported [stamping](https://docs.aspect.dev/bazelbuild/rules_nodejs/3.6.0/docs/stamping.html) like the `pkg_npm` & `pkg_web` rules do.

We utilise a Bazel monorepo and one of the great pluses of using Bazel is being able to rely on only rebuilding what is actually required. We can then utilise this to detect if certain services etc have actually changed and only deploy those that have, allowing much quicker deployments in addition to fast builds.

Unfortunately for Javascript bundles it is currently hard to embed up to date version information, specifically the volatile values from the workspace status command. For example, the current Git commit we treat as a volatile value as we don't want to trigger rebuilds of every package on every commit (whereas we have other stable values where we do want this).

It would be nice to inject selected stamping values into the `define` argument passed to esbuild. This could then perform substitutions in the code for any stamping values as normal, e.g. `--define:STATUS_COMMIT=abc123...` with src:

```javascript
const commit = 'STATUS_COMMIT';
``` 

This would be ideal as when an `esbuild` target requires to be rebuilt for any reason (e.g. change to upstream dependency, sources etc) then the compiled package would contain the up to date volatile status values. This would ensure version info about the Git commit & build time is accurate for a built package but avoid triggering needless rebuilds when the target isn't needing rebuilt- hence not triggering needless downstream deploys.

I'd suggested a `substitutions` argument for consistency with the existing rules but have no preferred names / formats for the arguments - as long as we could achieve the functionality described above.

I've not really looked into `pkg_web` as a possible option yet, this may be a potential work around (even if just on a single file) if this is deemed to costly/confusing to implement. However, it would be nicer to be in the `esbuild` rule itself if this were possible.

mattem:
Needs https://github.com/bazelbuild/rules_nodejs/pull/2840

closes #2812

### DIFF
--- a/packages/esbuild/launcher.js
+++ b/packages/esbuild/launcher.js
@@ -24,7 +24,7 @@ function getEsbuildArgs(paramsFilePath) {
   }
 }
 
-async function processConfigFile(configFilePath) {
+async function processConfigFile(configFilePath, existingArgs = {}) {
   const fullConfigFileUrl = pathToFileURL(join(process.cwd(), configFilePath));
   let config;
   try {
@@ -44,7 +44,6 @@ async function processConfigFile(configFilePath) {
   // These keys of the config can not be overriden
   const IGNORED_CONFIG_KEYS = [
     'bundle',
-    'define',
     'entryPoints',
     'external',
     'metafile',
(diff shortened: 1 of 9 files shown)


packages/esbuild/launcher.js
```diff
@@ -56,9 +55,31 @@ async function processConfigFile(configFilePath) {
...
+    if (value === null || value === void 0) {
+      return prev;
+    }
+
     if (IGNORED_CONFIG_KEYS.includes(key)) {
       console.error(`[WARNING] esbuild configuration property '${key}' from '${configFilePath}' will be ignored and overriden`);
+    } else if (MERGE_CONFIG_KEYS.includes(key) && existingArgs.hasOwnProperty(key)) {
+      // values from the rule override the config file
+      // perform a naive merge
+      if (Array.isArray(value)) {
```
alexeagle, line 72: this could use some unit tests

alexeagle (approved):

PR #2864, opened 2021-08-13
build: disable the angular architect example on Ubuntu

mattem:
Not sure what introduced the slowness, and it's reasonable when run locally, not sure what's going on here.

This is the first build where it went slow and red, but the commit is unrelated
https://buildkite.com/bazel/rules-nodejs-nodejs/builds/10103


mattem:
Hm, not convinced it's just that example, something else is going on

mattem:
Done in #2870

PR #2873, opened 2021-08-18
test: repro problem with fetching esbuild via http_archive

alexeagle:


docs/esbuild.md
```diff
@@ -22,31 +22,30 @@ or using yarn
...
-    strip_prefix = "package",
+npm_install(
+    name = "npm",
+    # @bazel/esbuild is a dependency in this package.json
+    package_json = "//:package.json",
+    package_lock_json = "//:package-lock.json",
 )
 
-load("@bazel_esbuild//:esbuild_repositories.bzl", "esbuild_repositories")
+load("@npm//@bazel/esbuild:esbuild_repositories.bzl", "esbuild_repositories")
```
mattem, line 39: This should be the rules_nodejs repo load not npm right?

packages/esbuild/BUILD.bazel
```diff
@@ -60,14 +60,29 @@ filegroup(
         "esbuild.bzl",
         "esbuild_config.bzl",
         "esbuild_packages.bzl",
-        "esbuild_repositories.bzl",
+        # Intentionally not shipping the repositories code via npm
+        # See comment below on :builtin_srcs
+        # "esbuild_repositories.bzl",
```
mattem, line 65: Can `esbuild_packages` also be removed from here?

packages/esbuild/esbuild_repositories.bzl
```diff
@@ -27,20 +30,23 @@ def esbuild_repositories(name = ""):
...
             # Install is run with ignore scripts so that esbuild's postinstall script does not run,
             # as we never use the downloaded binary anyway and instead set 'ESBUILD_BINARY_PATH' to the toolchains path.
             # This allows us to deal with --platform
             "--ignore-scripts",
         ],
         symlink_node_modules = False,
+        # BEGIN-INTERNAL
         package_path = "packages/esbuild",
+        # END-INTERNAL
+        #@external package_path = "external/" + pkg_label.workspace_name + "/@bazel/esbuild",
```
mattem, line 51: Sneaky, perhaps add a comment here?

packages/esbuild/BUILD.bazel
alexeagle: no, it's still loaded from toolchain/BUILD.bazel

docs/esbuild.md
alexeagle: oops, where did that come from? I thought I copy-pasted from the working example...

packages/esbuild/esbuild_repositories.bzl
alexeagle: done

docs/esbuild.md
alexeagle: oh, I bet I changed the wrong file and when accepting the new golden docs I overwrote it. Almost did it again :)

alexeagle: I see, it's getting auto-replaced when we publish the docs, because we always assume the packages are referenced from npm... will have to inhibit that replacement here somehow

PR #2882, opened 2021-08-23
fix(cypress): add js files from JSModuleInfo to plugin wrapper sources if available

mtl-wgtwo:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

The `cypress_web_test` implementation uses JSNamedModuleInfo to find the plugin source.  This is fine for typescript libraries,  but fails with the following error if a `js_library` is used to package the plugin source:

```
ERROR: /Users/mtl/Development/rules_nodejs.git/packages/cypress/test/BUILD.bazel:50:17: in cypress_web_test rule //packages/cypress/test:lib_test:
	File "/Users/mtl/Development/rules_nodejs.git/packages/cypress/internal/cypress_web_test.bzl", line 88, column 45, in _cypress_web_test_impl
		plugin_wrapper = _cypress_plugin_wrapper(ctx)
	File "/Users/mtl/Development/rules_nodejs.git/packages/cypress/internal/cypress_web_test.bzl", line 59, column 99, in _cypress_plugin_wrapper
		plugin_file = _filter_js(ctx.attr.plugin_file[JSNamedModuleInfo].direct_sources.to_list())[0]
Error: index out of range (index is 0, but sequence has 0 elements)
ERROR: Analysis of target '//packages/cypress/test:lib_test' failed; build aborted: Analysis of target '//packages/cypress/test:lib_test' failed
```

## What is the new behavior?

This PR extends the `_cypress_plugin_wrapper` to consider both `JSNamedModuleInfo` and `JSModuleInfo` sources, with priority given to the original behavior if there are both.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/cypress/internal/cypress_web_test.bzl
+++ b/packages/cypress/internal/cypress_web_test.bzl
@@ -13,7 +13,7 @@
 # limitations under the License.
 "E2E testing with Cypress"
 
-load("@build_bazel_rules_nodejs//:providers.bzl", "JSNamedModuleInfo")
+load("@build_bazel_rules_nodejs//:providers.bzl", "JSModuleInfo", "JSNamedModuleInfo")
 load("@build_bazel_rules_nodejs//internal/node:node.bzl", "nodejs_test_kwargs")
 
 ATTRS = dict(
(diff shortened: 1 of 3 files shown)


alexeagle:
FYI @jbedard ties in with your effort to make a consistent starlark library call for collecting info from the providers

alexeagle (approved):
thank you!!

jbedard:
This seems to add another alternative way of reading those providers? :/

This case is slightly different where it's trying to find a single file, not collect a list of deps. The use of `len(... .to_list())` is an example where it would be nice if the provider didn't even exist unless it had files, although I think we agreed that might cause more issues then solve.

PR #2902, opened 2021-08-31
fix: update `tsutils` to version `3.21.0`

alan-agius4:
`tsetse` - `check-return-value` rule is returning false positives due because the short-circuit code in https://github.com/bazelbuild/rules_nodejs/blob/9b454e38f7e2bbc64f75ee9a7dcb6ff45f1c7a12/third_party/github.com/bazelbuild/rules_typescript/internal/tsetse/rules/check_return_value_rule.ts#L60-L62 is not working as expected for certain usecases.

Example:

```ts
    get content() {
      try {
        return expect(harness.readFile(path)).withContext(`With file content for '${path}'`);
      } catch (e) {
        if (e.code !== 'ENOENT') {
          throw e;
        }

        // File does not exist so always fail the expectation
        return createFailureExpectation(
          expect(''),
          `Expected file content but file does not exist: '${path}'`,
        );
      }
    },
```

WIll cause tsetse to fail with the bellow;
```
              ~~~~~~~~~~
packages/angular_devkit/build_angular/src/testing/jasmine-helpers.ts:112:16 - error TS21222: [tsetse] return value is unused.
        See http://tsetse.info/check-return-value

112         return expect(Buffer.byteLength(harness.readFile(path))).withContext(
                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
113           `With file size for '${path}'`,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
114         );
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,7 @@
         "tmp": "0.1.0",
         "ts-lit-plugin": "1.1.9",
         "tsickle": "0.38.0",
-        "tsutils": "2.27.2",
+        "tsutils": "3.21.0",
         "typeorm": "0.2.18",
         "typescript": "~4.3.2",
         "unidiff": "1.0.1",
--- a/packages/concatjs/package.json
+++ b/packages/concatjs/package.json
@@ -19,7 +19,7 @@
     "dependencies": {
         "protobufjs": "6.8.8",
         "source-map-support": "0.5.9",
-        "tsutils": "2.27.2"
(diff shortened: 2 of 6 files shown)


alexeagle (approved):
thanks!

PR #2905, opened 2021-08-31
chore: upgrade rules_go version

alexeagle:
Prereq for cross-compilation from M1 mac, it seems


thesayyn:
I also experience this issue with M1. rules_typescript (ts_devserver) rely on some of the removed internal apis ig.

alexeagle:
yeah either we need intensive refactoring of how the devserver handles runfiles (IIRC there's no equivalent functionality still in that runfiles library) or just vendor the code we were originally using into our repo.

alexeagle:
https://github.com/bazelbuild/rules_go/pull/2076#issuecomment-520994853

> I'm not ready to figure out that refactoring right now so I'll just patch out this commit from our rules_go http_archive fetch. One day the merge conflict will bring someone back to this thread. Maybe me! :)

yup, here we are :/

PR #2911, opened 2021-09-02
feat(builtin): add support for using a local .nvmrc file for providing a node version

josephperrott:
Add `use_nvmrc` attribute on `node_repositories` to provide a Label pointing to the local .nvmrc file in the repository.


alexeagle (approved):
neat idea! thanks :)
only thing I'd wish for is a link to learn more about nvm in case a user is reading about this option but doesn't already know what that file does

Issue #2504, opened 2021-03-04
Cannot use filegroup for srcs in ts_project


tbezman:
# 🐞 bug report

### Affected Rule

ts_project

### Is this a regression?

I don't think so

### Description

When using `srcs` in `ts_project`, using a filegroup causes the rule to be unable to find any TypeScript files. I'm pretty sure this is because `ts_project_macro` passes along sources to `_out_paths` which does not expand the filegroup. 

https://github.com/bazelbuild/rules_nodejs/blob/stable/packages/typescript/internal/ts_project.bzl#L668

## 🔬 Minimal Reproduction

I've created a minimum repro in this repo https://github.com/tbezman/ts-project-repro . Note the comments over here https://github.com/tbezman/ts-project-repro/blob/master/some-package/BUILD.bazel#L18

## 🔥 Exception or Error

<pre><code>
ERROR: Traceback (most recent call last):
        File "/Users/terencebezman/Code/ts-project-repro/some-package/BUILD.bazel", line 15, column 11, in <toplevel>
                ts_project(
        File "/private/var/tmp/_bazel_terencebezman/d9e2260f83532a6e2ac3b769af8ce672/external/npm/@bazel/typescript/internal/ts_project.bzl", line 663, column 13, in ts_project_macro
                fail("""ts_project target "//{}:{}" is configured to produce no outputs.
Error in fail: ts_project target "//some-package:the_project" is configured to produce no outputs.

Note that ts_project must know the srcs in advance in order to predeclare the outputs.
Check the srcs attribute to see that some .ts files are present (or .js files with allow_js=True).
ERROR: Skipping '//some-package:the_project': no such target '//some-package:the_project': target 'the_project' not declared in package 'some-package' defined by /Users/terencebezman/Code/ts-project-repro/some-package/BUILD.bazel
WARNING: Target pattern parsing failed.
ERROR: no such target '//some-package:the_project': target 'the_project' not declared in package 'some-package' defined by /Users/terencebezman/Code/ts-project-repro/some-package/BUILD.bazel
INFO: Elapsed time: 0.138s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
</code></pre>

## 🌍  Your Environment

**Operating System: MacOS Big Sur**

**Output of `bazel version`:**

<pre>
  <code>
Bazelisk version: development
Build label: 4.0.0
[...]

PR #2913, opened 2021-09-03
Fixed ts_project json file outpath with root_dir

twheys:
When using the root_dir attr, the paths to json files are not modified from the input file paths in the same fashion that ts files are. I updated the definition of json_outs to use the same logic to produce an out file path that is used with ts files.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Currently, when setting the `root_dir` attr with a `ts_project` that includes a `.json` file in the `srcs`, the output file path is not generated relative to the root dir, as it is for ts and other files.

Issue Number: N/A

## What is the new behavior?

This PR fixes the generation of json output file paths to be relative to the root dir using the same logic as used with other file types.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/typescript/internal/ts_project.bzl
+++ b/packages/typescript/internal/ts_project.bzl
@@ -171,10 +171,14 @@ def _ts_project_impl(ctx):
     # tsc will only produce .json if it also produces .js
     if len(ctx.outputs.js_outs):
         pkg_len = len(ctx.label.package) + 1 if len(ctx.label.package) else 0
-        json_outs = [
-            ctx.actions.declare_file(_join(ctx.attr.out_dir, src.short_path[pkg_len:]))
+        json_file_names = [
+            _map_out_path(src.short_path[pkg_len:], ctx.attr.out_dir, ctx.attr.root_dir, {"*": ".json"})
             for src in ctx.files.srcs
             if src.basename.endswith(".json") and src.is_source
+        ]
+        json_outs = [
+            ctx.actions.declare_file(file_name)
+            for file_name in json_file_names
         ]
     else:
         json_outs = []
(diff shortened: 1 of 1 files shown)


twheys:
To add some clarity as to what exactly this patch fixes:

If I have a `ts_project` rule like this that uses `root_dir`, I expect that the output files are written with paths relative to `./server` in the local package directory. This is the case with the transpiled js files, but the json output files are being declared as `./server/.../file.json`, which results in the build rule failing because the json is output in the local directory. I fixed the declared output file path to match what is actually produced by the rule by using the same path logic that is applied to the transpiled js files.

```
ts_project(
    name = "server",
    srcs = glob([
            "server/**/*.ts",
            "server/**/*.tsx",
            "server/**/*.json",
    ]),
    root_dir = "server",
    tsconfig = ":server/tsconfig.json",
    ...
)
```

packages/typescript/internal/ts_project.bzl
```diff
@@ -171,11 +171,15 @@ def _ts_project_impl(ctx):
     # tsc will only produce .json if it also produces .js
     if len(ctx.outputs.js_outs):
         pkg_len = len(ctx.label.package) + 1 if len(ctx.label.package) else 0
-        json_outs = [
-            ctx.actions.declare_file(_join(ctx.attr.out_dir, src.short_path[pkg_len:]))
+        json_file_names = [
```
alexeagle, line 174: the asymmetry bothers me ... we have an attribute to the rule `js_outs` and the macro that calls the rule determines what output files to expect, so the calculation happens below

`js_outs.extend(_out_paths(srcs, out_dir, root_dir, allow_js, exts))`

can we do the same thing with json_outs - declare that as a new attribute on the rule so the two are handled symmetrically? or is there some reason you weren't able to do that?

twheys: Hi @alexeagle , thanks for the feedback. I agree with your point here. I think it would actually be best to move the calculation of outputs for the other fields into the rule body. That would fix the issue of not being able to use sources from other rules like `filegroup`, since you need to calculate exact list of the output files. I left a more detailed explanation as a comment on [this issue](https://github.com/bazelbuild/rules_nodejs/issues/2504#issuecomment-912456378).

I'd be happy to update this pull request to include those suggested changes, that would also address your comment here. I might just need some help testing to avoid any regression. Let me know what you think.

alexeagle: I'd prefer in this PR to follow the pattern that's already here, and keep that other issue orthogonal. Does it still work for you if you do the calculation in the macro?

twheys: Unfortunately, it won't work, because I need to resolve the individual input files in order to calculate the output files and this is not possible from within a macro, only from `ctx.files`

twheys: Let's continue this discussion on the other issue, since the outcome of this appears to be dependent on the output of that: https://github.com/bazelbuild/rules_nodejs/issues/2504

alexeagle: why can't the macro predict the output files? can't it determine that from the .json files listed in the srcs like it does for .ts?

again, for the sake of keeping the two orthogonal (this could land now, the other is trickier)

alexeagle: the other one landed, time to rebase and get this one done?

PR #2917, opened 2021-09-04
fix(typescript): write json srcs to generated tsconfig for resolveJso…

matthewjh:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

When using a `ts_project` rule with a generated tsconfig file, tsc errors when using `resolveJsonModule`, because the generated tsconfg file does not enumerate the .json inputs in the "files" array. For reasons unclear to me, tsc only enforces this on `composite` projects.

## What is the new behavior?

.json inputs are enumerated in generated tsconfig files.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/typescript/internal/ts_project.bzl
+++ b/packages/typescript/internal/ts_project.bzl
@@ -280,6 +280,7 @@ def _validate_options_impl(ctx):
         preserve_jsx = ctx.attr.preserve_jsx,
         composite = ctx.attr.composite,
         emit_declaration_only = ctx.attr.emit_declaration_only,
+        resolve_json_module = ctx.attr.resolve_json_module,
         source_map = ctx.attr.source_map,
         incremental = ctx.attr.incremental,
         ts_build_info_file = ctx.attr.ts_build_info_file,
@@ -310,6 +311,7 @@ validate_options = rule(
         "extends": attr.label(allow_files = [".json"]),
         "incremental": attr.bool(),
         "preserve_jsx": attr.bool(),
+        "resolve_json_module": attr.bool(),
         "source_map": attr.bool(),
         "target": attr.string(),
         "ts_build_info_file": attr.string(),
@@ -322,6 +324,9 @@ def _is_ts_src(src, allow_js):
     if not src.endswith(".d.ts") and (src.endswith(".ts") or src.endswith(".tsx")):
         return True
     return allow_js and (src.endswith(".js") or src.endswith(".jsx"))
+
+def _is_json_src(src, resolve_json_module):
+    return resolve_json_module and src.endswith(".json")
 
 def _replace_ext(f, ext_map):
     cur_ext = f[f.rindex("."):]
@@ -352,6 +357,7 @@ def ts_project_macro(
         declaration = False,
         source_map = False,
         declaration_map = False,
+        resolve_json_module = False,
         preserve_jsx = False,
         composite = False,
         incremental = False,
@@ -579,6 +585,8 @@ def ts_project_macro(
         allow_js: boolean; Specifies whether TypeScript will read .js and .jsx files. When used with declaration,
             TypeScript will generate .d.ts files from .js files.
 
+        resolve_json_module: boolean; Specifies whether TypeScript will read .json files.
+
         declaration_dir: a string specifying a subdirectory under the bazel-out folder where generated declaration
             outputs are written. Equivalent to the TypeScript --declarationDir option.
             By default declarations are written to the out_dir.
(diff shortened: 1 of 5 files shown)


packages/typescript/internal/ts_project.bzl
```diff
@@ -323,6 +323,9 @@ def _is_ts_src(src, allow_js):
         return True
     return allow_js and (src.endswith(".js") or src.endswith(".jsx"))
 
+def _is_json_src(src):
+    return src.endswith(".json")
```
matthewjh, line 327: I am including JSON srcs unconditionally. This _may_ be fine as `tsc` will error if JSON files are imported without `resolveJsonModule`.

On the other hand, they could be included conditionally on `resolveJsonModule` being set to `True` either in an attribute on the rule or nested in the tsconfig.compilerOptions map. This would be similar to how `allowJs` drives the inclusion of JS files.

matthewjh: I think unconditional inclusion may cause a problem if people are globing `**/*` and this sucks up incidental JSON files. I'm not sure how `tsc` behaves if JSON files are included in its compilation without being imported from any TS files. Whatever the case, it may become more or less lenient in future, so maybe we should play safe here.

Any thoughts?

alexeagle: We definitely need to handle those corner cases correctly. If you're not sure how `tsc` behaves then do that experiment, or add that scenario under packages/typescript/test/ts_project and let's verify?

matthewjh: ok, should be good now. I went down the route of:

1. adding a new attribute to the `ts_project` macro & rule, `resolve_json_module`
2. only including json srcs if `resolve_json_module` is true, as happens for `allow_js`
3. auto-including json srcs if `srcs` is None and `resolve_json_module` is true, as happens for `allow_js`

I did notice some inconsistency as to how some of the attributes are handled versus the equivalent inside `ts_config.compilerOptions`, but that is more general than this so I'd like to probe that in a separate PR.

packages/typescript/internal/ts_project.bzl
```diff
@@ -610,10 +614,14 @@ def ts_project_macro(
     """
 
     if srcs == None:
+        include = ["**/*.ts", "**/*.tsx", "**/*.js", "**/*.jsx"]
```
alexeagle, line 617: something is wrong with the logic here, you can never leave this block without including .js and .jsx where previously you could

packages/typescript/internal/ts_project.bzl
```diff
@@ -629,6 +637,7 @@ def ts_project_macro(
         declaration_map = compiler_options.setdefault("declarationMap", declaration_map)
         emit_declaration_only = compiler_options.setdefault("emitDeclarationOnly", emit_declaration_only)
         allow_js = compiler_options.setdefault("allowJs", allow_js)
+        resolve_json_module = compiler_options.setdefault("resolveJsonModule", resolve_json_module)
```
alexeagle, line 640: you should also update the validator to check that the attribute setting matches what's in the tsconfig file

alexeagle (commented):
thanks!

packages/typescript/internal/ts_project.bzl
matthewjh: thanks, fixed. are there any tests for the validator? i couldn't find any.

packages/typescript/internal/ts_project.bzl
matthewjh: good spot, thanks. i feel that internal details like this would be easier to catch with unit tests, so we don't have to figure out ways to make everything express through mini-integration tests (which is hard for this one, for instance). but i guess the tooling is not in place for this?

matthewjh: `generated_file_tests` for generated tsconfigs would have caught this. that's not a substitute for testing some of these internal pieces separately, but would be an improvement on what's there today. perhaps i can look at that later on in a different pr.

matthewjh:
@alexeagle this should be ready now.

packages/typescript/internal/ts_project.bzl
alexeagle: yes sometimes the "mini-integration test" is not the best choice. There's a way to assert on what actions are produced by a rule implementation but not run them (like running `bazel aquery` and asserting on the result" which I'd love to try.

maybe for this even a smaller test fixture using the testing libraries in skylib would work, but I haven't investigated those for a while.

packages/typescript/internal/ts_project.bzl
alexeagle: not for the negative case, no

alexeagle (approved):
thanks!

Issue #1306, opened 2019-10-26
Support Multiplexed TypeScriptCompile Worker with Node Worker Threads


JaredNeil:
# 🚀 feature request

### Relevant Rules

ts_library

### Description

Bazel is adding support for [multiplexed workers](https://github.com/bazelbuild/bazel/commit/6d1b9725b1e201ca3f25d8ec2a730a20aab62c6e) and Node now supports [worker threads](https://nodejs.org/api/worker_threads.html).
Combining the two we could get parallel compilation with a shared in-worker file cache.
This would be pretty cool and might have significant performance and resource benefits, so I thought I'd put it out here for discussion. Thoughts?

ashi009:
Cool! However, to make it possible, workers now need to use messages to access the shared cache, which could be the bottleneck here. Also, I doubt if typescript ASTs are serializable.

alexeagle:
I've been thinking/discussing this.

I think we should build a clean-room implementation of ts_library that brings in only the minimal bits - just run TypeScript with its composite projects feature - and puts them in the right place. For example I don't think the program should worry about protocol buffers - the `--require` shim in `bazel_require_script.js` should marshal/unmarshal the worker protocol.

@ashi009 I don't know if we need to share a cache between worker threads. The simplest thing is to run `tsc --watch` in each worker thread, and establish affinity between each tsconfig.json (program) and a particular worker. Each worker sees only that one program over its lifecycle.

@JaredNeil when would you like to start on this? I'm on vacation next week and we are shipping 1.0 end of November, so maybe we could collaborate on a branch during December?

JaredNeil:
I really haven't thought much about it. I just think it could save us a lot of memory when compiling all of our typescript if we could share the cache.

@ashi009, `postMessage` doesn't have to be able to JSON stringify the message. From [the docs](https://nodejs.org/api/worker_threads.html#worker_threads_port_postmessage_value_transferlist):
>Sends a JavaScript value to the receiving side of this channel. value will be transferred in a way which is compatible with the [HTML structured clone algorithm.](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm)

Is that clone going to cause a ton of overhead? Maybe.

@alexeagle Yeah, that could probably work. Not really a blocker for us, so whenever there is time for it.

alexeagle:
Discussed with @rictic today, still feels promising. We'd want it for google-internal too.

[...]

PR #2918, opened 2021-09-04
feat: decouple @bazel/worker from rules_typescript

thesayyn:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

`@bazel/worker` uses an old version of bazel worker protocol and coupled with `rules_typescript` which is not maintained anymore.
Since the used protocol is old, it lacks some of the new features such as `cancellation` and `multiplex-workers`.

Issue Number: https://github.com/bazelbuild/rules_nodejs/issues/1306

## What is the new behavior?

- Now the worker_protocol.proto is checked into the third-party directory and the corresponding code is generated by `protoc-gen-ts`. 
- New version of the package does not rely on just-in-time loading of the proto file and gets complete type safety.
- Package is now fully rewritten in typescript and there is no other third-party dependency other than `google-protobuf`.
- It is ready to handle `cancellation` but we have to introduce some breaking changes such as `runOneBuild` returning a cancel callback to handle the cancellation and taking a callback argument to call when the task is complete. similar to what we see on bundlers like webpack and rollup.
- A new argument (third) `console` is passed to `runOneBuild` callback to let the consumer have an API to write debug messages and feel right at home.

- [ ] Yes
- [x] No

## Other information
the new protocol file is generated by using [protoc-gen-ts](https://github.com/thesayyn/protoc-gen-ts) which is mostly maintained by me and a few open source contributors hence I am not sure if you would like to rely on a third-party project. might be a deal-breaker for you.

additionally, in this PR I have only changed things to make room for the incoming support for multiplex workers.


thesayyn:
I would like to hear your thoughts and the points that I should improve

alexeagle:
The variations of TS generation for .proto files is pretty confusing for sure. I like the comparison table on the bottom of your readme. https://github.com/thesayyn/protoc-gen-ts#alternatives

However you only use it as a build-time dependency within our repo to generate those files, not a runtime dependency for our users, so the bar is a lot lower on stability. We'll very rarely change the worker_protocol.proto so we'll rarely need to run the tool. If your repo disappears we could find another generator, or worst case we could edit the files by hand if a field or two gets added to the worker_protocol.proto in the future.

The bigger concern to figure out first is if this is a breaking change, then we can't land it yet. I'd rather not start a 5.x branch yet since I'll have to cherry-pick and manage two branches for 6 months until the next major - usually we only start that process a couple months before the release.

So I guess we'd like to explore how to make this non-breaking. Either by restoring back-compat or distributing it alongside the existing implementation.

thesayyn:
Hmm, then I would vote that we make this non-breaking change and make this reflect the current API.  Once we land this, I could send another PR which we could get landed in the next major release.  

I will send another commit to restore backward compatibility.

thesayyn:
Some of the tests are failing. will pick them tomorrow

thesayyn:
I just realized that I have to update a lot of packages to clean up the old way of using the worker package. Which is going to be fun though like meditation.

thesayyn:
Now some packages (those who has worker support) has a direct dependency on `@bazel/worker`.

thesayyn:
@alexeagle I would very much appreciate  get your thoughts on how I could make these packages depend on the worker package and make the tests pass

alexeagle:
I'm on mobile, but check the examples/BUILD file?

thesayyn:
Unfortunately `npm_packages` only replace occurrences in the root package and not dependants of the root package. So when `@bazel/typescript` it will be replaced by `npm_packages` but the dependencies of `@bazel/typescript` which is the dependency of `@bazel/typescript`

thesayyn:
@alexeagle ready for a review. and got all the tests passing

examples/BUILD.bazel
```diff
@@ -293,4 +329,7 @@ example_integration_test(
         "//packages/typescript:npm_package": "@bazel/typescript",
     },
     owners = ["@mattem"],
+    resolutions = {
```
alexeagle, line 332: maybe we should set this as the default value for the `resolutions` attribute so we don't have to remember where it has to be set? wouldn't do any harm to include it even for tests that don't require it, right?

examples/esbuild/WORKSPACE
```diff
@@ -25,12 +25,12 @@ http_archive(
     urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/4.0.0/rules_nodejs-4.0.0.tar.gz"],
 )
 
-load("@build_bazel_rules_nodejs//:index.bzl", "npm_install")
+load("@build_bazel_rules_nodejs//:index.bzl", "yarn_install")
 
-npm_install(
+yarn_install(
```
alexeagle, line 30: why switch to yarn? PRs should be minimal to make them easy to review. If you're making a change that looks orthogonal, either put in a separate PR, or anticipate my question and add a comment about why this change is required for your change

examples/from_source/package.json
```diff
@@ -3,7 +3,8 @@
   "private": true,
   "devDependencies": {
     "@types/node": "12.12.50",
-    "protobufjs": "6.10.1",
+    "google-protobuf": "^3.6.1",
```
alexeagle, line 6: if the example has to change, that means it's a breaking change...

packages/typescript/internal/ts_project.bzl
```diff
@@ -679,10 +679,18 @@ def ts_project_macro(
...
+                # worker package
                 "@npm//protobufjs",
                 # END-INTERNAL
                 Label(typescript_package),
                 Label("//packages/typescript/internal/worker:filegroup"),
+                # BEGIN-INTERNAL
+                # this is not needed when package since @bazel/typescript lists
+                # @bazel/worker as its dependency hence has access to it.
+                # this only needed when ts_project is run from the source.
+                Label("//packages/worker:library"),
```
alexeagle, line 692: nice

packages/worker/src/worker_protocol.ts
```diff
@@ -0,0 +1,333 @@
+/**
+ * Generated by the protoc-gen-ts.  DO NOT EDIT!
```
alexeagle, line 2: if you're going to check in a generated file, we should at least have a test that asserts it's up-to-date

how hard is it to leave the generated file in bazel-out and not check in as a source file?

thesayyn: there are some reasons behind this

- have IDE recognize the file 
- somehow rules_proto does not work on windows. it throws an error with a message "bazel-out is not an external or internal command" which probably has something to do with how rules_proto tries to invoke protoc. so the windows tests are going to fail. I have seen it before.
- generating this file every time is not an option since ts_project won’t pick it up. it might create cycling dependency between worker and typescript package since they are relying on each other to build (not sure though)

I would prefer to add a generated_file_test and make an assertion.  this test would be disabled on windows though because of the problem I have described above.

examples/from_source/package.json
thesayyn: I have no clue how I can make this non-breaking change... may be you have something on your mind.

examples/esbuild/WORKSPACE
thesayyn: oh, thats because npm ignores the `resolutions` object in the package.json file. so we needed to install the dependencies with yarn_install just like other examples. but if this is a deal-breaker, I will try to find a way to make it work for both yarn_install and npm_install

examples/BUILD.bazel
thesayyn: the only problem I see is having to build this package but bazel is smart enough to use from previous builds so that makes sense to put this as a default. will send a pr for it.

examples/from_source/package.json
thesayyn: also, I have no idea why this was needed in the first place. Shouldn’t rules_nodejs list this as a dependency even if the repository is consumed from the source? 

I only added this because the tests were yelling at me

I believe that the problem lies somewhere else.

thesayyn: Shouldn’t this be handled by rules_nodejs_dev_dependencies? perhaps something similar? as I don’t rely on the consumer's npm repository, I don’t see why this should be listed as a dependency by the consumer.

alexeagle: rules_nodejs_dev_dependencies is only for when you want to consume our rules from source. normal users shouldn't need it. but since this is the examples/from_source it actually does make sense here

thesayyn: should I add it to this example? or shall we make separate PR for this?

alexeagle: sorry it was my mistake, as it's building rules_nodejs from source it's expected that our dev dependencies leak through. So far we have the rules_nodejs_dev_dependencies for *starlark* dependencies, but we don't have anything for *npm* dependencies (in theory it should be the root package json but we don't actually publish `@bazel/javascript` we treat it as private)

packages/worker/src/metadata.ts
```diff
@@ -0,0 +1,27 @@
+export function readMetadata(buffer: Buffer, offset: number) {
```
alexeagle, line 1: add a file-level comment explaining what this metadata is

alexeagle (commented):
looking really close :)

thesayyn:
- i have the assertion test in place: `//packages/worker:copy_worker_protocol`
- I couldn’t put the update command comment into the generated file. the assertion fails. so instead I put it into the build file. I hope that's fine

examples/from_source/package.json
thesayyn: Hmm. That’s what I was expecting. so is this okay to have this change?

packages/worker/src/worker_protocol.ts
thesayyn: have the assertion test in place: //packages/worker:copy_worker_protocol

thesayyn: I couldn’t put the update command comment into the generated file. the assertion fails. so instead I put it into the build file. I hope that's fine

packages/worker/src/metadata.ts
thesayyn: I have already changed the file name to `size` and added the documentation to clear up confusion and not invent a slang word in the bazel world.

thesayyn:
I believe that this is ready for another review.

alexeagle (approved):
super nice 🌮

PR #2931, opened 2021-09-10
refactor: begin extracting a new core package

alexeagle:
Motivation is described by #2892 

Some advantages of this approach:
- We now follow the style guide at https://docs.bazel.build/versions/main/skylark/deploying.html
- Gives us a 'clean room' to eventually reduce the scope of the project.
- Allows others to develop their own competing nodejs rules without having to start from scratch.
- Allows the user to declare two different node versions since they control the name of
  the external repository we create

This is also a lot simpler. We don't do any transformations of our code in publishing the package, instead we now force users to take a dependency on bazel-skylib.


internal/node/node_repositories.bzl
```diff
@@ -19,710 +19,9 @@ See https://docs.bazel.build/versions/main/skylark/repository_rules.html
 """
```
thesayyn, line 19: Why did we decide to keep this one? 

I would expect nodejs//:repositories to be referenced everywhere but I believe this was done to keep the changes minimal.

thesayyn (commented):
so far looks good.

internal/node/node_repositories.bzl
alexeagle: yeah I extracted the primary rule from here, but to keep it non-breaking I had to keep the surrounding macro.

No users will install `rules_nodejs` workspace until we ask them to, which is a breaking change. So everything still has to be exposed in `build_bazel_rules_nodejs`, only private helper APIs can move.

gregmagolan (approved):
🌮

internal/node/node_repositories.bzl
thesayyn: Got it!

PR #2933, opened 2021-09-11
Core: add toolchains

alexeagle:
Note this gem of a comment:
https://docs.bazel.build/versions/main/skylark/deploying.html#registering-toolchains
>Note that in order to resolve toolchains in the analysis phase Bazel needs to analyze all toolchain targets that are registered. Bazel will not need to analyze all targets referenced by toolchain.toolchain attribute. If in order to register toolchains you need to perform complex computation in the repository, consider splitting the repository with toolchain targets from the repository with <LANG>_toolchain targets. Former will be always fetched, and the latter will only be fetched when user actually needs to build <LANG> code.

this is saying that with very careful layout of these rules into repositories, we only actually fetch node binaries for the requested platforms yay :)


nodejs/private/nodejs_repo_host_os_alias.bzl
```diff
@@ -0,0 +1,131 @@
...
+                platform = platform,
+                node_repository = node_repository,
+                compatible_with = meta.compatible_with,
+            )
+
+    # Base BUILD file for this repository
+    repository_ctx.file("BUILD.bazel", build_content)
+
+    # index.bzl file for this repository
+    repository_ctx.file("index.bzl", content = """# Generated by node_repositories.bzl
```
gregmagolan, line 110: nit: Generated by nodejs/private/nodejs_repo_host_os_alias.bzl

nodejs/private/nodejs_repo_host_os_alias.bzl
```diff
@@ -0,0 +1,131 @@
...
+    - A BUILD.bazel file declaring aliases to the host platform's node binaries
+      and toolchain definitions for all other platforms which can be registered or selected
+    - index.bzl containing some constants
+    """,
+    attrs = {
+        "include_toolchain_declarations": attr.bool(
+            doc = """The new rules_nodejs workspace declares its toolchains in the alias repo
+                but the legacy build_bazel_rules_nodejs workspace does not.""",
+        ),
+        "node_repository": attr.string(default = "nodejs"),
```
gregmagolan, line 127: `node_repository_prefix`?

nodejs/private/nodejs_repo_host_os_alias.bzl
```diff
@@ -0,0 +1,131 @@
...
+            "@platforms//os:linux",
+            "@platforms//cpu:ppc",
+        ],
+    ),
+}
+
+def _nodejs_host_os_alias_impl(repository_ctx):
+    node_repository = "@%s_%s" % (repository_ctx.attr.node_repository, os_name(repository_ctx))
+
+    build_content = """# Generated by node_repositories.bzl
```
gregmagolan, line 53: nit: Generated by nodejs/private/nodejs_repo_host_os_alias.bzl

gregmagolan (approved):
🌮

PR #2936, opened 2021-09-12
fix(typescript): exclude package.json and package-lock.json from default ts_project srcs

edsrzf:
This PR excludes `**/package.json` and `**/package-lock.json` from the default value for `srcs`. This seems like a very common case that would make the default unusable when `resolveJsonModule` is on.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

An NPM project using TypeScript with the `resolveJsonModule` set to true, not specifying `srcs` to the `ts_project` rule causes errors such as:

```
output 'package-lock.json' was not created
output 'package.json' was not created
```

Issue Number: N/A

## What is the new behavior?

Files matching the globs `**/package.json` or `**/package-lock.json` are excluded.

- [x] Yes
- [ ] No

This is a breaking change if anyone is relying on the existing behavior of including `package.json` or similarly-named files.

### DIFF
--- a/packages/typescript/internal/ts_project.bzl
+++ b/packages/typescript/internal/ts_project.bzl
@@ -481,14 +481,17 @@ def ts_project_macro(
 
         srcs: List of labels of TypeScript source files to be provided to the compiler.
 
-            If absent, defaults to `**/*.ts[x]` (all TypeScript files in the package).
+            If absent, the default is set as follows:
+            - Include `**/*.ts[x]` (all TypeScript files in the package).
+            - If `allow_js` is set, include `**/*.js[x]` (all JavaScript files in the package).
+            - If `resolve_json_module` is set, include `**/*.json` (all JSON files in the package), but exclude `**/package.json`, `**/package-lock.json`, and `**/tsconfig*.json`.
 
         deps: List of labels of other rules that produce TypeScript typings (.d.ts files)
 
         tsconfig: Label of the tsconfig.json file to use for the compilation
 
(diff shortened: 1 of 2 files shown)


alexeagle:
```


File "/workdir/packages/typescript/internal/ts_project.bzl", line 627, column 27, in ts_project_macro
--


Error in append: append() accepts no more than 1 positional argument but got 2
```

edsrzf:
🤦 

I guess this means that the default when `allow_js` is true has not been tested, since it did the same thing.

packages/typescript/internal/ts_project.bzl
```diff
@@ -618,10 +621,10 @@ def ts_project_macro(
         include = ["**/*.ts", "**/*.tsx"]
         exclude = []
         if allow_js == True:
-            include.append("**/*.js", "**/*.jsx")
+            include.extend(["**/*.js", "**/*.jsx"])
         if resolve_json_module == True:
             include.append("**/*.json")
-            exclude.append("**/tsconfig*.json")
+            exclude.extend(["**/package*.json", "**/tsconfig*.json"])
```
matthewjh, line 627: Might not `package*` be too broad? What about listing `package.json` and `package-lock.json` separately here?

edsrzf: You're probably right. I've made this update.

alexeagle (approved):
Nice, and thanks for the clear documentation update as well

PR #2942, opened 2021-09-14
fix(builtin): pkg_npm unable to copy files from transitioned actions

devversion:
The `pkg_npm` rule is currently using some basic computation to
determine output paths from action inputs. This logic is prone to
mistakes and also does not work well with inputs that have been
built as part of transitioned targets. e.g. The devserver binaries
for `@bazel/concatjs` are currently built using a 1:3 transition
where each devserver binary is built in a different Bazel output
directory fragment. e.g.

bazel-out/darwin-fastbuild-ST-05b09dfdd2e5
bazel-out/darwin-fastbuild-ST-b60b0173fed1
bazel-out/darwin-fastbuild-ST-fce3d0422d3b

https://unpkg.com/browse/@bazel/concatjs@4.1.0/bazel-out/

The `pkg_npm` rule needs to be able to properly compute
output paths regardless of the output directory fragment.
The solution to the problem is that the package rule should
pass relevant `short_path` information to the packager tool.
The shot path is relative to the containing root, allowing for
straightforward and _more reliable_ computation of the output
path with respect to the owning package declaring the `pkg_npm` rule.

e.g. `bazel-out/darwin-fastbuild-ST<..>/packages/core/index.js` should
end up being resolved to `<pkg-out>/index.js` if the owning package
name is `packages/core`. Previously this was computed by only stripping
off the leading default bazel output directory fragment, like:
`bazel-out/darwin-fastbuild/packages/core`

### DIFF
--- a/internal/pkg_npm/packager.js
+++ b/internal/pkg_npm/packager.js
@@ -25,6 +25,12 @@ const path = require('path');
 const isBinary = require('isbinaryfile').isBinaryFileSync;
 
 /**
+ * Type definition describing a file captured in the Bazel action.
+ * https://docs.bazel.build/versions/main/skylark/lib/File.html.
+ * @typedef {{path: string, shortPath: string}} BazelFileInfo
+ */
+
+/**
  * Create a new directory and any necessary subdirectories
  * if they do not exist.
  */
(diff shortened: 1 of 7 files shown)


devversion:
@alexeagle Finally got the NPM diff test green. I'm not sure what was going on there but it looked like the genrule (which exposed the tarball from the `.pack` target) has been cached and did not update properly. I have reworked the test to be Windows-compatible and more reliable. Now everything seems good.

alexeagle (approved):
amazing as always 🌹

alexeagle:
I'd like to cut a 4.1.1 that just includes this bugfix, but it will take extra time as we don't normally make release branches. okay to have this in 4.2.0 in the next week or so?

devversion:
@alexeagle works for me. It's easy to just not upgrade for now 👍 thx for reviewing and merging!

PR #2948, opened 2021-09-15
feat: introduce @bazel/webpack

thesayyn:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


packages/webpack/bin/BUILD.bazel
```diff
@@ -0,0 +1,17 @@
...
+the package is installed (possibly with different attribute values).
+"""
+
+load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_binary")
+
+nodejs_binary(
+    name = "webpack-worker",
+    data = [
+        "//packages/worker:library",
+        "@npm//webpack-cli",
```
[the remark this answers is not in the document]
thesayyn: Yes, this won’t work for them. this is only for testing and in case of usage from the source. For those who use this as a package, they will get the version that points to `webpack-cli` in the same repository as `@bazel/webpack` was installed. 

So,  //packages/webpack/bin:webpack-worker will be replaced via //@bazel/wepback/bin:webpack-worker as we have `bin` entries in package.json.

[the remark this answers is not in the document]
thesayyn: Also, since we peer dependency on `webpack-cli` it will be visible to worker package.

package.json
```diff
@@ -99,6 +99,9 @@
         "typeorm": "0.2.18",
         "typescript": "~4.3.2",
         "unidiff": "1.0.1",
+        "webpack-cli": "4.8.0",
```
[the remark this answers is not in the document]
thesayyn: I agree. We need to find a way to avoid adding dependencies of packages into the root package. Since I need this to run the generated_file_test, I can’t really remove it.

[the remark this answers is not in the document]
mrmeku: I think the point is moot considering all packages will be split out into their own repos in the future. If that changes we can address the problem, but in all likelyhood the packages will be removed as packages leave the monorepo

packages/webpack/bin/BUILD.bazel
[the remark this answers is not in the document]
mrmeku: I believe we want this binary to live within a macro such that we have one worker per target.

packages/webpack/test/worker/BUILD.bazel
```diff
@@ -0,0 +1,29 @@
...
+load("@build_bazel_rules_nodejs//:index.bzl", "generated_file_test")
+
+webpack(
+    name = "bundle_worker",
+    outs = ["app.bundle.worker.js"],
+    args = [
+        "./$(execpath index.js)",
+        "--mode",
+        "production",
+        # Discuss this: Whether we should automatically add this or not
```
mrmeku, line 11: Following the example of rollup and esbuild I think this style is great. It allows for the most amount of user customization

packages/webpack/test/worker/bundle.js
```diff
@@ -0,0 +1 @@
+(()=>{"use strict";console.log("hmm")})();
```
mrmeku, line 1: Do we want to check the bundle in?

mrmeku:
@googlebot I consent.

mrmeku (approved):

packages/webpack/test/worker/bundle.js
thesayyn: Not sure here. I am open to any suggestion that would let us assert that the rule can bundle the code correctly.

packages/webpack/internal/worker/webpack.config.js
```diff
@@ -0,0 +1,113 @@
...
+    };
+
+    process.on('message', gotInput);
+
+    return {
+      close: () => {
+        process.off('message', gotInput);
+      },
+      pause: () => {
+        process.off('message', gotInput);
```
mrmeku, line 86: If we do this during pause, how do we resume watching for messages?

packages/webpack/internal/worker/webpack.config.js
```diff
@@ -0,0 +1,113 @@
+const {join} = require('path');
+
+function notAvailable() {
+  throw new Error(`Watcher pooling or pausing is not available under bazel worker mode.`);
```
mrmeku, line 4: /s/pooling/polling

packages/webpack/internal/worker/worker_adapter.ts
```diff
@@ -0,0 +1,91 @@
...
+ */
+import * as worker from "@bazel/worker";
+import * as cp from 'child_process';
+import * as fs from 'fs';
+
+const webpackCli = require("webpack-cli");
+
+type IPCMessage = {
+  type: "built" | "error" | "ready"
+  [key: string]: unknown
```
mrmeku, line 20: I don't think you need this particular line

thesayyn: Nice catch!

packages/webpack/internal/worker/worker_adapter.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+function main() {
+  const MNEMONIC = 'webpack';
+  if (worker.runAsWorker(process.argv)) {
+    const webpackCliPath = require.resolve("webpack-cli/bin/cli.js");
+    let key: string;
+    let proc: cp.ChildProcess | undefined;
```
mrmeku, line 28: Might clean up the implementation to leave off the | undefined from the type def

packages/webpack/internal/worker/worker_adapter.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    worker.runWorkerLoop(async (args: string[], inputs: { [path: string]: string }) => {
+      return new Promise<boolean>((resolve, reject) => {
+        // We can not add this 
+        args = [...args,  "--watch"];
+        if (key != args.join("#") && proc) {
+          proc.kill("SIGKILL");
+        }
+
+        if (proc == undefined) {
```
mrmeku, line 38: nit: if (!proc)

packages/webpack/internal/worker/worker_adapter.ts
```diff
@@ -0,0 +1,91 @@
...
+       --strategy=${MNEMONIC}=worker`);
+    let argsFilePath = process.argv.pop()!;
+    if (argsFilePath.startsWith('@')) {
+      argsFilePath = argsFilePath.slice(1)
+    }
+    const args = fs.readFileSync(argsFilePath).toString().trim().split('\n');
+    // webpack cli assumes that the first two arguments are executable and scripts
+    // hence strips them out leading to corrupted arguments if we just pass the arguments
+    // that we receive from bazel via arg file. 
+    new webpackCli().run([process.argv, "", ...args]);
```
mrmeku, line 85: We have the empty string as an argument here?

packages/webpack/internal/worker/worker_adapter.ts
```diff
@@ -0,0 +1,91 @@
...
+  }
+}
+
+if (require.main === module) {
+  main();
+}
```
mrmeku, line 91: nit: this repo uses newlines at the end of files

packages/webpack/test/worker/bundle.js
mrmeku: We would make a test which consumes the output of the rule and makes some assertion on it

```
const {readFileSync} = require('fs');

const bundlePath = process.argv[3];

const got = readFileSync(bundlePath);
const want = `(()=>{"use strict";console.log("hmm")})();`
if (got !== want) {
  throw new Error("Bundle is incorrect");
}
```

mrmeku: But really, just having the build target be present is likely already sufficient. We can handle more elaborate testing in the `examples/react_webpack` example in a followup

packages/webpack/internal/worker/webpack.config.js
```diff
@@ -0,0 +1,113 @@
...
+        compiler.hooks.failed.tap('WorkerWatchPlugin', () => {
+          process.send({type: 'error'});
+        });
+        compiler.hooks.afterEnvironment.tap('WorkerWatchPlugin', () => {
+          process.send({type: 'ready'});
+        });
+      }
+    }
+  }]
+}
```
mrmeku, line 113: nit: newline

packages/webpack/internal/worker/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+    "extends": "../../../../tsconfig.json",
+    "compilerOptions": {
+        "moduleResolution": "node",
+        "module": "CommonJS",
+        "declaration": true
+    },
+}
```
mrmeku, line 8: nit: newline

packages/webpack/test/worker/bundle.js
thesayyn: I see that even if we write a test like above we would be inlining the bundle into the test. so I propose that we keep it the way it is. I agree on elaborate testing with an integration test in examples.

packages/webpack/internal/worker/webpack.config.js
thesayyn: It is called by webpack. it is pretty much like close but with a minor difference; the underlying file watchers are kept alive until the next watch call. Probably designed this way to benefit from a caching mechanism that watchpack has.

hence webpack will call the `watch` method for each subsequent build and `pause` the previous one before it starts another one.

thesayyn: See here

https://github.com/webpack/webpack/blob/562f17a8c04be426cb6ec9be474a37fb4dd2f6f7/types.d.ts#L11552  and https://github.com/webpack/webpack/blob/179d2e00fe0b901ff176f1c6dbbf6a17f1ff44ea/lib/Watching.js#L99

packages/webpack/internal/worker/webpack.config.js
thesayyn: Nice!

packages/webpack/internal/worker/worker_adapter.ts
thesayyn: it should have been the path of the script but I left it empty because webpack does not care if you or not. it just ignores the first two arguments in the array.

packages/webpack/bin/BUILD.bazel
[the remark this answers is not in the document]
thesayyn: This is how rollup works btw. but if you believe that this is not the right way then we could do the macro approach.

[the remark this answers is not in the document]
alexeagle: got it, this whole nodejs_binary is only for local/from-source use

packages/webpack/docs_test.js
```diff
@@ -0,0 +1,6 @@
+const actual = require('fs').readFileSync(
```
alexeagle, line 1: but there aren't actually any markdown files in the PR, seems like the docs feature isn't complete.

Not sure if we should finish it, or drop it for now and add when it gets more mature

packages/webpack/internal/webpack.bzl
```diff
@@ -0,0 +1,135 @@
...
+    # TODO: If the string has intentional spaces or if one or more of the expanded file
+    # locations has a space in the name, we will incorrectly split it into multiple arguments
+    return ctx.expand_location(s, targets = ctx.attr.data).split(" ")
+
+def _inputs(ctx):
+    # Also include files from npm fine grained deps as inputs.
+    # These deps are identified by the ExternalNpmPackageInfo provider.
+    inputs_depsets = []
+    for d in ctx.attr.data:
+        if ExternalNpmPackageInfo in d:
```
alexeagle, line 74: fyi @gregmagolan since you're hoping to remove it

packages/webpack/internal/worker/worker_adapter.ts
alexeagle: a better answer to that code review comment is to either add a comment or a named constant with the empty string as the value

packages/webpack/bin/BUILD.bazel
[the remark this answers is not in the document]
thesayyn: Correct.

packages/webpack/internal/worker/worker_adapter.ts
thesayyn: Noted! just sent a commit

packages/webpack/docs_test.js
thesayyn: not sure why this is here also. perhaps @mrmeku knows something

alexeagle:
For now we'll try to make the webpack rule stand alone, since we're not keen on expanding scope here. Can revisit promoting it later. https://github.com/aspect-dev/rules_webpack

PR #2956, opened 2021-09-17
build: update rules_webtesting to latest version supporting m1 browsers

devversion:
Updates to the latest version of `rules_webtesting` that comes with
M1 macOS browser support, RBE/exec platform selection fixes and better
caching of the browser (and tool) binaries.

https://github.com/bazelbuild/rules_webtesting/commit/6b2ef24cfe95d6e31d5e61414949ff8f0f184b13

### DIFF
--- a/common.bazelrc
+++ b/common.bazelrc
@@ -64,6 +64,11 @@ build:debug --compilation_mode=dbg
 # This prevents accidentally depending on this feature, which Bazel will remove.
 build --nolegacy_external_runfiles
 
+# Workaround for: https://github.com/bazelbuild/bazel/issues/4327.
+# The Chromium browser archive contains files with spaces that will
+# break runfile tree creation for browser tests.
+build --experimental_inprocess_symlink_creation
+
 # Turn on --incompatible_strict_action_env which was on by default
 # in Bazel 0.21.0 but turned off again in 0.22.0. Follow
 # https://github.com/bazelbuild/bazel/issues/7026 for more details.
--- a/common.bazelrc
+++ b/common.bazelrc
@@ -65,9 +65,10 @@ build:debug --compilation_mode=dbg
 build --nolegacy_external_runfiles
 
 # Workaround for: https://github.com/bazelbuild/bazel/issues/4327.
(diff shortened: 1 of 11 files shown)


common.bazelrc
```diff
@@ -64,6 +64,11 @@ build:debug --compilation_mode=dbg
 # This prevents accidentally depending on this feature, which Bazel will remove.
 build --nolegacy_external_runfiles
 
+# Workaround for: https://github.com/bazelbuild/bazel/issues/4327.
+# The Chromium browser archive contains files with spaces that will
+# break runfile tree creation for browser tests.
+build --experimental_inprocess_symlink_creation
```
alexeagle, line 70: this is the common file which is also exposed to users who run `npx @bazel/create` - is that intended?

alexeagle (approved):

common.bazelrc
devversion: Ops. I did not realize. That should't be in there IMO

devversion: @alexeagle I looked more into it. I'm still not 100% sure but I feel like this flag is reasonable to have here as it works around a bug affecting consumers relying on web testing.

The alternative would be to add this flag to all the `examples` and `e2e` examples relying on web testing. This is cumbersome though. Ideally I think we would recommend the `--nobuild_runfile_links` flag in the common config as it has various perf benefits and worked well in all of the Angular projects. This also fixes the issue quite nicely: See details here: https://github.com/bazelbuild/bazel/issues/4327#issuecomment-922106293. 

Unfortunately this flag does not work in this repo though because the export directory feature somehow relies on these runfile trees (which seems unexpected and unclear according to: https://github.com/bazelbuild/rules_nodejs/commit/a7200aaf526d2b648b3a2eaadc20b7c51b467275). All of this is a little tricky. Kind of unfortunate but I don't know a better solution.

alexeagle (approved):
okay I'll check in with @gregmagolan on progress understanding interaction between the runfiles and the exports_directories_only feature. This works as-is thanks

PR #2957, opened 2021-09-17
chore: add TODO for toolchain/genrule workaround

alexeagle:
https://github.com/bazelbuild/bazel/issues/14009 has a pretty good workaround which should be exposed by rules_nodejs so that genrules can work


alexeagle:
Nevermind, I found time already and TODID it in #2960

PR #2960, opened 2021-09-19
feat(builtin): expose a concrete toolchain for rules that don't under…

alexeagle:
…stand toolchain_type

This lets genrule interop with the node toolchain

Workaround for https://github.com/bazelbuild/bazel/issues/14009, thanks @katre


e2e/core/BUILD.bazel
```diff
@@ -34,37 +40,23 @@ write_file(
...
+# You can use the node toolchain together with a genrule().
+# This gives you complete control over starting the interpreter, but you also have to
+# manually handle module resolution.
+genrule(
+    name = "use_node_toolchain",
+    srcs = ["some.js"],
+    outs = ["genrule_out"],
+    cmd = "$(NODE_PATH) $(execpath some.js) $@",
+    toolchains = ["@node16_toolchains//:resolved_toolchain"],
+    tools = ["@node16_toolchains//:resolved_toolchain"],
```
katre, line 52: Do you actually need both of these? That's kind of ugly. Do you think runfiles provided via `toolchains` should also be available like `tools` are? Or is this clear enough?

nodejs/private/toolchains_repo.bzl
```diff
@@ -63,11 +63,37 @@ PLATFORMS = {
...
 
 def _impl(repository_ctx):
+    # Expose a concrete toolchain which is the result of Bazel resolving the toolchain
+    # for the execution or target platform.
+    # Workaround for https://github.com/bazelbuild/bazel/issues/14009
+    starlark_content = """# Generated by toolchains_repo.bzl
+
+load("@rules_nodejs//nodejs:toolchain.bzl", "node_toolchain_providers")
+
+def _impl(ctx):
```
katre, line 73: This re-declaration of `_impl` isn't really a re-declaration, because it's in a string constant, but it's very confusing.

I know we do it a lot in Bazel, but naming a rule implementation `_impl` is just asking for trouble and isn't very future-proof. I'd recommend using `_resolved_toolchain_impl` here (and fixing the previous use of `_impl` to something more specific, also).

nodejs/toolchain.bzl
```diff
@@ -25,6 +25,31 @@ May be empty if the target_tool_path points to a locally installed node binary."
...
+            nodeinfo = NodeInfo(
+                target_tool_path = toolchain_info.target_tool_path,
+                tool_files = toolchain_info.tool_files,
+            ),
+        ),
+        # Make the $(NODE_PATH) variable available in places like genrules.
+        # See https://docs.bazel.build/versions/main/be/make-variables.html#custom_variables
+        # Note that genrule seems to have a bug: this only works if the node_toolchain target
+        # itself is used by the toolchains attribute of genrule, not the toolchain_type.
+        platform_common.TemplateVariableInfo({
```
katre, line 48: Is it worth having the actual `node_toolchain` rule provide this (and any future templated variables), and having this function just fetch and forward the provider?

nodejs/toolchain.bzl
```diff
@@ -25,6 +25,31 @@ May be empty if the target_tool_path points to a locally installed node binary."
...
+
+       This is exposed as part of the workaround for https://github.com/bazelbuild/bazel/issues/14009
+    """
+    return [
+        DefaultInfo(
+            files = depset(toolchain_info.tool_files),
+            runfiles = ctx.runfiles(files = toolchain_info.tool_files),
+        ),
+        platform_common.ToolchainInfo(
+            nodeinfo = NodeInfo(
```
katre, line 39: The types are unclear to me here: Are you converting from one provider to another, or just duplicating the same provider type?

I would assume that it's the same provider type, and so you don't actually need to create new instances of `NodeInfo` (or even `ToolchainInfo`).

e2e/core/BUILD.bazel
illicitonion: I ran into exactly the same ugliness in https://github.com/bazelbuild/rules_perl/pull/27 :) It would be nice to implicitly add to runfiles, but not a huge deal

katre: Can one of you file an issue about this, so I don't lose track?

illicitonion: https://github.com/bazelbuild/bazel/issues/14014

nodejs/toolchain.bzl
alexeagle: "fetch and forward the provider" is what I could not figure out. In the impl of the `resolved_toolchain` rule I looked around at the 
`dir(ctx.toolchains["@rules_nodejs//nodejs:toolchain_type"])`
 and
`dir(ctx.toolchains["@rules_nodejs//nodejs:toolchain_type"].nodeinfo)`
and tried various ways to access the providers and forward them, but failed.

katre: Right, this was unclear and I apologize.

The result type from `ctx.toolchains[anything]` is a `ToolchainInfo` provider. In your case, this has a `nodeinfo` field which stores your `NodeInfo` provider.

My recommendation is that in the `node_toolchain` rule, you create both the `NodeInfo` and `TemplateVariableInfo` providers, and put them into the `ToolchainInfo` with different keys (`nodeinfo` and `template_variables` would work).

Then, in `resolved_toolchain`, you get back the `ToolchainInfo` (by calling `ctx.toolchains["@rules_nodejs//nodejs:toolchain_type"]`), and have `resolved_toolchain` return this:

```
    toolchainInfo = ctx.toolchains["@rules_nodejs//nodejs:toolchain_type"]
    return [
      toolchain_info,
      toolchain_info.nodeinfo,
      toolchain_info.template_variables,
    ]
```

That way, all three providers from `node_toolchain` are re-exported by `resolved_toolchain`.

illicitonion: This is exactly the pattern I implemented in https://github.com/bazelbuild/rules_perl/pull/27 in case it's easy to crib :)

alexeagle: you're very kind to give such hand-holding and I don't expect an apology at all! very appreciative, I don't know where you find the time to be so helpful 🙏🏻

gregmagolan (approved):
Nice! 🌮

PR #2980, opened 2021-09-23
refactor: use diff_test for docs and a single update command

alexeagle:


mattem:
Nice, this is much better ergonomics than the individual update targets

mattem (approved):

PR #2981, opened 2021-09-25
refactor: improve generate_build_file.ts performance

cuyl:
- use async API
- parallel runs `findPackages`

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

generate_build_file.ts takes 65~80s in my repo.

| function          | time       |
|---------------|---------|
| findPackages  | 60~70s |
| generate          | 7~9s     |

## What is the new behavior?

It improved to 25~35s.

- [ ] Yes
- [x] No


internal/npm_install/generate_build_file.ts
```diff
@@ -76,42 +75,58 @@ package(default_visibility = ["${visibility}"])
 }
 
 if (require.main === module) {
-  main();
```
jbedard, line 79: There's a lot of random missing semicolons throughout the file, this example creating an unnecessary diff...

internal/npm_install/generate_build_file.ts
```diff
@@ -76,42 +75,58 @@ package(default_visibility = ["${visibility}"])
...
+  main()
+}
+
+
+async function exists(p: string) {
+  return await fs.access(p, constants.F_OK | constants.W_OK).then(() => true, () => false);
```
jbedard, line 83: Why is the `catch` necessary unlike previously in the sync version?

internal/npm_install/generate_build_file.ts
```diff
@@ -76,42 +75,58 @@ package(default_visibility = ["${visibility}"])
...
 }
 
+// Avoid duplicate mkdir call when mkdir runs parallel.
```
jbedard, line 86: The `exists` avoids the `mkdir`, this avoids creating a (possibly unnecessary) promise to do that `exists`. Is creating this potentially giant `Map` worth avoid that extra promise and `exists` check?

internal/npm_install/generate_build_file.ts
```diff
@@ -76,42 +75,58 @@ package(default_visibility = ["${visibility}"])
...
+  await mkdirp(path.dirname(p));
+  await fs.symlink(target, p, 'file');
 }
 
 /**
  * Main entrypoint.
  */
-export function main() {
+export async function main() {
+
```
jbedard, line 129: There's also some unnecessary white space changes such as this one

internal/npm_install/generate_build_file.ts
```diff
@@ -255,8 +277,8 @@ function generatePackageBuildFiles(pkg: Dep) {
   const nodeModulesPkgDir = `node_modules/${pkg._dir}`;
   // Check if the current package dep dir is a symlink (which happens when we
   // install a node_module with link:)
-  const isPkgDirASymlink =
-      fs.existsSync(nodeModulesPkgDir) && fs.lstatSync(nodeModulesPkgDir).isSymbolicLink();
+  const isPkgDirASymlink = await fs.lstat(nodeModulesPkgDir)
+    .then(stat => stat.isSymbolicLink()).catch(() => false)
```
jbedard, line 281: Why is the `catch` necessary unlike previously?

internal/npm_install/generate_build_file.ts
```diff
@@ -440,29 +462,35 @@ def _maybe(repo_rule, name, **kwargs):
...
 }
 
 /**
  * Checks if a path is a file.
  */
-function isFile(p: string) {
-  return fs.existsSync(p) && fs.statSync(p).isFile();
+async function isFile(p: string) {
+  try {
+  return (await fs.stat(p)).isFile();
```
jbedard, line 481: Indentation.

Again this does the `catch` unlike previously?

Also why try/catch instead of `.catch` like the other spots?

internal/npm_install/generate_build_file.ts
```diff
@@ -476,64 +504,62 @@ function stripBom(s: string) {
...
-                fs.unlinkSync(fullPath);
-              }
-              return files;
-            }
-            return isDirectory ? files.concat(listFiles(rootDir, relPath)) : files.concat(relPath);
-          },
-          [])
-      // We return a sorted array so that the order of files
-      // is the same regardless of platform
-      .sort();
```
jbedard, line 536: Why move the sort elsewhere?

internal/npm_install/generate_build_file.ts
```diff
@@ -571,53 +597,51 @@ export function getDirectDependencySet(pkgJsonPath: string): Set<string> {
...
-                     .filter(f => isDirectory(f))
-                     // strip 'node_modules/' from filename
-                     .map(f => f.substring('node_modules/'.length));
+  const listing = await fs.readdir(p);
 
+  const scopes: string[] = []
+  for (let f of listing) {
+    if (f.startsWith('@')) {
+      f = path.posix.join(p, f)
+      if (await isDirectory(f)) {
```
jbedard, line 639: Worth making this one concurrent? (putting the promises into `scopes` and then `Promise.all` instead of chaining them)

internal/npm_install/generate_build_file.ts
```diff
@@ -571,53 +597,51 @@ export function getDirectDependencySet(pkgJsonPath: string): Set<string> {
 /**
  * Finds and returns an array of all packages under a given path.
  */
-function findPackages(p: string, dependencies: Set<string>) {
-  if (!isDirectory(p)) {
-    return [];
+async function findPackages(p: string, dependencies: Set<string>, pkgs: Array<Dep>) {
```
jbedard, line 600: Can we just `return await Promise.all` more like previously instead of passing the array in?

internal/npm_install/generate_build_file.ts
cuyl: [fs.exists](https://nodejs.org/dist/latest-v14.x/docs/api/fs.html#fs_fs_exists_path_callback) has been deprecated, so I use [fs. access](https://nodejs.org/dist/latest-v14.x/docs/api/fs.html#fs_fs_access_path_mode_callback) instead, `fs.acess` throws an error when the file doesn't match the   condition (visible and readable)

internal/npm_install/generate_build_file.ts
cuyl: "mkdirp" now runs in parallel and no longer has a strict execution order.
for example
```js
// async version
await Promise.all([mkdirp('@test/foo') , mkdirp('@test/bar')])

/*
  call1 exists('@test/foo') -> false
  call2 exists('@test/bar') -> false
  call1 exists('@test') -> false
  call2 exists('@test') -> false
  call1 fs.mkdir('@test') -> pass
  call2 fs.mkdir('@test') -> error
*/

// sync version with fs.existsSync and fs.mkdirSync
mkdirp('@test/foo')
mkdirp('@test/bar')
/*
  call1 exists('@test/foo') -> false
  call1 exists('@test') -> false
  call1 fs.mkdir('@test') -> pass
  call1 fs.mkdir('@test/foo') -> pass

  call2 exists('@test/bar') -> false
  call2 exists('@test') -> true
  call2 fs.mkdir('@test/bar') -> pass
*/
```

internal/npm_install/generate_build_file.ts
cuyl: If nodeModulesPkgDir is not existed, `fs.lstat` throws an error, so I catch the error instead of call existsSync before, It may reduce syscall.

internal/npm_install/generate_build_file.ts
cuyl: Same as isPkgDirASymlink.

jbedard: Lets make it the same as `isDirectory` though..? Doing `.catch(() => false)` instead of try/catch...

internal/npm_install/generate_build_file.ts
cuyl: I'll move it back.

internal/npm_install/generate_build_file.ts
cuyl: OK

internal/npm_install/generate_build_file.ts
jbedard: Wow, there were more of those then I realized... it wasn't just your PR lol

internal/npm_install/generate_build_file.ts
cuyl: I'm not sure, generate step didn't take much time.

cuyl: OK, use `Promise.all` now, you can check it.

jbedard: > I'm not sure, generate step didn't take much time.

If it makes the PR more complicated or you don't think it's worth it that's fine, we can always do it later.

cuyl:
> There's a lot of random missing semicolons throughout the file
@jbedard
A lot of random missing before as well, I fixed it all. maybe we could add an eslint rule to check `ts,js` code styles.

jbedard:
> maybe we could add an eslint rule to check ts,js code styles.

I agree with that one, would lead to less confusion on my part in PRs like this :)

internal/npm_install/generate_build_file.ts
```diff
@@ -571,52 +597,67 @@ export function getDirectDependencySet(pkgJsonPath: string): Set<string> {
...
  * Finds and returns an array of all packages under a given path.
  */
-function findPackages(p: string, dependencies: Set<string>) {
-  if (!isDirectory(p)) {
-    return [];
-  }
-
+async function findPackages(p: string, dependencies: Set<string>): Promise<Dep[]> {
   const pkgs: Dep[] = [];
+  const _findPackages = async (p: string) => {
```
jbedard, line 602: I *think* you can avoid the inner functions here and in `listFiles` basically doing the same as before your change, using `...` on the recursive call? Or was there another reason for the inner function?

```
  if (!await isDirectory(p)) {
    return [];
  }
  
  const pkgs: Dep[] = [];
  const listing = await fs.readdir(p);

  await Promise.all(listing.map(async f => {
    // filter out folders such as `.bin` which can create
    // issues on Windows since these are "hidden" by default
    if (f.startsWith('.')) return [];
    const pf = path.posix.join(p, f);
    
    if (await isDirectory(pf)) {
      if (f.startsWith('@')) {
        await findPackages(pf, dependencies);
      } else {
        pkgs.push(
          await parsePackage(pf, dependencies),
          ... await findPackages(path.posix.join(pf, 'node_modules'), dependencies)
        );
      }
    }
  }));
```

cuyl: I do this before I open this PR.
the problem is the `pkgs` need to be sort to ensure the order, So it just needs one `pkgs` instance and sort once through all inner recursive calls. 

I think it would be neater to move `pkgs.sort` into findPackages.

jbedard: Ah I see, that's why you moved it out originally? Sorry I missed that, I was just trying to reduce the diff since I thought it was an unnecessary change. Looks like I was wrong there?

cuyl: You are right, It is an unnecessary change, I just put the sort call into findPackages and listFiles in 12c687a, Sorry, I mixed the semi and refactor commit, It may be confused.

alexeagle:
thanks @cuyl this is exciting, I'll take a look when @jbedard is happy with it

internal/npm_install/generate_build_file.ts
jbedard: @cuyl did you want to try dropping these inner functions and putting the `.sort()` outside?

I know you did similar before so sorry if I'm flip-flopping here, I don't think I realized what was going on at first...

cuyl: OK, I have put the sort outside and rename the functions to more reasonable names.

internal/npm_install/generate_build_file.ts
```diff
@@ -79,78 +79,107 @@ if (require.main === module) {
...
  */
-function mkdirp(p: string) {
-  if (!fs.existsSync(p)) {
-    mkdirp(path.dirname(p));
-    fs.mkdirSync(p);
+async function mkdirp(p: string) {
+  let promise = mkdirPromiseMap.get(p);
+  if (!promise) {
+    promise = (async () => {
+      if (await exists(p)) return;
```
jbedard, line 106: How about switching back to `if (!async exists(p))`, or just add the `{...}` and put the return on a newline. More style issues that should be auto-enforced :/

jbedard: Or I guess we do this single-line `if (x) return;` eslewhere, I think the indentation is just off here? Just fixing the indentation is fine as well...

jbedard:
Looks awesome to me 🎉 

Still other things we can do concurrently but looks like you measured it and did the most important ones.

@alexeagle ...

alexeagle (approved):
thanks for the PR @cuyl and thanks  for reviewing @jbedard

PR #2982, opened 2021-09-27
feat(typescript): support for ESM variant of the Angular compiler plugin

devversion:
As of v13, the `@angular/compiler-cli` package will come as strict ESM
package. This means that the import currently in `tsc_wrapped` does
not work for v13+ of Angular. This commit adds an interop allowing for
both the ESM variant, and CJS variant of the Angular compiler to work.

### DIFF
--- a/third_party/github.com/bazelbuild/rules_typescript/internal/BUILD.bazel
+++ b/third_party/github.com/bazelbuild/rules_typescript/internal/BUILD.bazel
@@ -61,6 +61,7 @@ tsc(
     data = _TSC_WRAPPED_SRCS + [
         "//internal:tsconfig.json",
         "@npm//@types/node",
+        "@npm//@angular/compiler-cli",
         "@npm//protobufjs",
         "@npm//tsickle",
         "@npm//tsutils",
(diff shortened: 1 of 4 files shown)


third_party/github.com/bazelbuild/rules_typescript/internal/BUILD.bazel
```diff
@@ -61,6 +61,7 @@ tsc(
     data = _TSC_WRAPPED_SRCS + [
         "//internal:tsconfig.json",
         "@npm//@types/node",
+        "@npm//@angular/compiler-cli",
```
alexeagle, line 64: it feels a bit undesirable to have a compile-time dependency, but it does reflect the actual types of the tsc_wrapped program where we added the layering violation

alexeagle (approved):

third_party/github.com/bazelbuild/rules_typescript/internal/BUILD.bazel
devversion: It's only compile-them when tsc_wrapped is built as part of the `@bazel/typescript` NPM package, so it doesn't have any impact on consumers. 

The benefit is just a little type-safety here, compared to relying on`any`

PR #2987, opened 2021-09-28
fix(typescript): typescript downleveling breaking ESM variant of Angular v13 compiler

devversion:
Unfortunately my initial refactor to allow for the ESM variant of the
compiler-cli did not contain a workaround for an issue that is caused
by TypeScript. Currently the import is rewritten to a dynamic require,
preventing the ESM variant of the compiler-cli to work. This commit
fixes this in a similar way we have made `@angular/cli` compatible with
the ESM-variant of Angular v13 packages.


devversion:
@alexeagle This PR is now finally green. Looks like there was a fallback needed to `require` since there are some tests in the repo that rely on older NodeJS versions which do not support the dynamic `import` expression.

alexeagle (approved):

PR #2995, opened 2021-10-02
feat: bazel 4.1.0 upgrade

IgorMinar:
I've tried to update to 4.1.0 a while back and failed to finish the job because it got too complicated.

I realized that others could pick up where I left and finish this, so I'm sending a PR and a bit of hope 😄 .

The main motivation is that currently rules_nodejs don't work on arm64 platform and bazel 4.1+ is needed for that. This means that anyone on Apple's M1 machines can't use bazel.

Could someone pretty please finish this so that bazel works well on M1s? thank you!


IgorMinar:
turns out that I was pulling the wrong upstream branch (master). This has already been done in https://github.com/bazelbuild/rules_nodejs/pull/2792 

Thank you!

PR #2996, opened 2021-10-04
fix: use tree artifacts via copy_directory with exports_directories_only

gregmagolan:
Fixes the issue with exports_directories_only on RBE.

Removes `has_directories` from `ExternalNpmPackageInfo` since it is expected that source directories
are no longer used. This simplified `npm_umd_bundle.bzl` impl which was using this to filtering; it can now
just check `File.is_directory` since source directories are no longer passed in.

### DIFF
--- a/internal/js_library/js_library.bzl
+++ b/internal/js_library/js_library.bzl
@@ -149,9 +149,13 @@ def _impl(ctx):
     typings = []
     js_files = []
     named_module_files = []
+    has_directories = False
 
     for idx, f in enumerate(input_files):
         file = f
+
+        if file.is_directory:
+            has_directories = True
 
         # copy files into bin if needed
         if file.is_source and not file.path.startswith("external/"):
@@ -263,7 +267,7 @@ def _impl(ctx):
         ),
     ]
 
-    if ctx.attr.package_name == "$node_modules$" or ctx.attr.package_name == "$node_modules_dir$":
+    if ctx.attr.package_name == "$node_modules$":
         # special case for external npm deps
         workspace_name = ctx.label.workspace_name if ctx.label.workspace_name else ctx.workspace_name
         providers.append(ExternalNpmPackageInfo(
@@ -271,7 +275,6 @@ def _impl(ctx):
             sources = depset(transitive = npm_sources_depsets),
             workspace = workspace_name,
             path = ctx.attr.package_path,
-            has_directories = (ctx.attr.package_name == "$node_modules_dir$"),
         ))
     else:
         providers.append(LinkablePackageInfo(
@@ -290,7 +293,7 @@ def _impl(ctx):
             deps = ctx.attr.deps,
         ))
         providers.append(OutputGroupInfo(types = decls))
-    elif ctx.attr.package_name == "$node_modules_dir$":
+    elif has_directories:
         # If this is directory artifacts npm package then always provide declaration_info
         # since we can't scan through files
         decls = depset(transitive = files_depsets)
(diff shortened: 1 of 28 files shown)


internal/js_library/js_library.bzl
```diff
@@ -110,14 +110,14 @@ def _link_path(ctx, all_files):
...
         # Check that strip_prefix contains at least one src path
-        check_prefix = "/".join([p for p in [ctx.label.package, ctx.attr.strip_prefix] if p])
-        prefix_contains_src = False
-        for file in all_files:
-            if file.short_path.startswith(check_prefix):
-                prefix_contains_src = True
-                break
-        if not prefix_contains_src:
-            fail("js_library %s strip_prefix path does not contain any of the provided sources" % ctx.label)
+        # check_prefix = "/".join([p for p in [ctx.label.package, ctx.attr.strip_prefix] if p])
```
gregmagolan, line 113: strip_prefix logic needs fixing

internal/providers/external_npm_package_info.bzl
```diff
@@ -22,7 +22,6 @@ ExternalNpmPackageInfo = provider(
     doc = "Provides information about one or more external npm packages",
     fields = {
         "direct_sources": "Depset of direct source files in these external npm package(s)",
-        "has_directories": "True if any sources are directories",
```
gregmagolan, line 25: this is no longer needed since we no longer have this case as exports_directories_only now produced proper individually linkable js_library targets with package names instead of a js_library with a source directory and its package name set to the special case `$node_modules_dir$`

internal/npm_install/generate_build_file.ts
```diff
@@ -477,7 +509,17 @@ def _maybe(repo_rule, name, **kwargs):
...
+  let deps: Dep[] = [];
+  pkgs.forEach(pkg => {
+    deps = deps.concat(pkg._dependencies.filter(dep => !dep._isNested && !pkgs.includes(pkg)));
+  });
+  // filter out duplicate deps
+  deps = [...pkgs, ...new Set(deps)];
+
+  let buildFile = config.exports_directories_only ?
+      printScopeExportsDirectories(scope, deps) :
+      printScope(scope, deps);
```
gregmagolan, line 522:
```    
if (isSymbolicLink && config.exports_directories_only && path.basename(path.dirname(fullPath)) == '.bin') {
      // Delete any .bin symlinks if exports_directories_only is on since these can point to outside of the
      // tree artifact on linux which breaks that package's copy_directory.
      await fs.unlink(fullPath);
      continue
    }
```

nodejs/private/copy_directory.bzl
```diff
@@ -0,0 +1,104 @@
...
+        # remote-execution does not allow source directory inputs
+        # https://github.com/bazelbuild/bazel/commit/c64421bc35214f0414e4f4226cc953e8c55fa0d2
+        execution_requirements = {"no-remote-exec": "1"},
+    )
+
+def copy_bash(ctx, src, dst):
+    ctx.actions.run_shell(
+        tools = [src],
+        outputs = [dst],
+        command = "cp -rf \"$1/\" \"$2\"",
```
gregmagolan, line 42: `command = "rm -rf \"$2\" && cp -rf \"$1/\" \"$2\"",`

alexeagle (commented):
Had a quick scan over it, looking good. LMK when it's green and I'll do a more careful review

alexeagle:
@googlebot I consent.

matthewjh:
@gregmagolan what's the motivation behind this change?

alexeagle:
It fixes the exports_directories_only feature to work with RBE and with `--nobuild_runfile_links`

The feature originally depended on Bazel accidentally not noticing that inputs were directories, which happened to work in almost all cases. Now we make them TreeArtifacts which unfortunately bazel can't produce as InputArtifacts so a copy operation is needed.

matthewjh:
Thanks @alexeagle. I am about to set up RBE for my org. From what you say it sounds like I will have to turn off exports_directories_only for things to work, until this lands? Or does it only not work with both exports_directories_only  and `--nobuild_runfile_links`?

alexeagle:
Doesn't work with either, is my understanding

alexeagle:
🍿

gregmagolan:
> 🍿

Getting there. It looks like we have an execroot escape issue on Windows.

alexeagle (approved):

PR #3024, opened 2021-10-18
fix: simplify portion of linker and fix case where runfiles node_modules symlinks are missing under bazel run

gregmagolan:


alexeagle:
Yeah for the clean room I also was just fixing a thing that's under bazel run only, we'll need some outside-of-bazel bash test

Issue #266, opened 2018-08-02
Support yarn/npm7 workspaces


bugzpodder:
Would yarn workspaces work with this rule?
I was playing around with 0.10.0 recently but didn't get very far in trying to setup yarn_install.  Would a future version support this?

alexeagle:
@clydin from our team has been looking at yarn workspaces.

As I understand it, we have the same features as yarn workspaces without actually using it. You can have multiple directories with a `package.json` file, run a command to install all their dependencies, and publish individual directories to npm. But to have one package depend on another, you don't need package.json for this, as the `deps` of the consuming target can point directly to the output of the providing target.

tristanz:
@alexeagle I think you'll still run into problem supporting both yarn workspaces (e.g. for interactive development with livereload) and Bazel.  To support a local yarn install, you need the subpackages in your package.json, but this will break the yarn_install in WORKSPACE since those private subpackages won't be published to npm.

Do you have advice on how to support both a native yarn workflow and Bazel workflow from a monorepo with multiple packages?

Toxicable:
@tristanz I think the idea here is that you don't isntall your sub packages with a `yarn install` you just reference their outputs directly

trin0491:
If bazel supports yarn workspaces, could that be used to solve the following issue:

Assume a monorepo has two 'top-level' projects that are separately deploy able, e.g. 'app1' and 'app2' and they both depend upon a library 'lib1'.  Lib1 is typescript library with a package.json that has a dependency on a third party library, e.g. lodash.   Each of the applications also have package.json files with lodash but they also add another dependency, e.g. angular.  The two apps must have a consistent version for lodash but they may want to have different versions for angular at times, e.g. so they can upgrade independently.  Bazel supports supports multiple yarn_install/npm_install calls to create the workspaces but ts_library from app1 would prevent a dependency on lib1 because it will throw the 'All npm dependencies need to come from a single workspace" error and I don't think a single npm_install/yarn_install' workspace can contain two versions of angular so we can't use one package.json?

Is there some other solution to this or would we need the yarn workspace capability?  Thanks.

DylanVann:
There's probably a lot of people using Lerna / Yarn workspaces currently that would be interested in Bazel but need to keep workspaces working while using Bazel for some things.

[This is currently blocked](https://github.com/bazelbuild/rules_nodejs/issues/871#issuecomment-576871953).

migueloller:
[...]

PR #3051, opened 2021-10-29
feat: support multiple node_modules roots in require_patch

tcarrio:
Refs: #266

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

When referencing multiple `node_modules` from `(npm|yarn)_install` targets in a `nodejs_(binary|target)`, you won't be able to resolve modules from all `node_modules` roots.

For example, having a separate `yarn_install`:

```python
# setup stencil spec test dependencies
yarn_install(
    name = "npm_stencil_spec",
    exports_directories_only = True,
    frozen_lockfile = True,
    package_json = "//tools//stencil:spec/package.json",
    package_path = "tools/stencil/spec",
    yarn_lock = "//tools/stencil:spec/yarn.lock",
)
```
alongside the root:

```python
yarn_install(
    name = "npm",
    data = ["//:.npmrc"],
    exports_directories_only = True,
    frozen_lockfile = True,
    yarn_lock = "//:yarn.lock",
)
```

and referring to both of them in a test rule:

```python
nodejs_test(
    name = "test",
    args = [
        "test",
    ],
    chdir = native.package_name(),
    data = [
        "@npm//@stencil/core",
        "@npm_stencil_spec//jest",
    ],
    entry_point = stencil_cli_path,
)
```
would result in the test run not being able to access the `jest` module.

Issue Number: #266

## What is the new behavior?

Resolution of a `require` call will be done against all `node_modules` roots. The behavior is the same if it is not found.

- [ ] Yes
- [X] No

The prior behavior would have allowed for node_modules to be included without being able to be imported. Due to this, if anyone were relying on this undefined behavior to include `node_modules` data where importing that would have thrown an error, but I can't imagine a use case where this would be useful behavior to rely on.

## Other information

Some minor adjustments to test case names such as "mode" updated to "module".


tcarrio:
I will need to add some documentation around this feature. The limitation currently is that you must define the two node modules roots under different paths with `package_path` since they can't be combined into a single directory implicitly. This makes sense, but it would be nice to note it somewhere.

PR #3056, opened 2021-11-01
refactor: depend on bazel-skylib at runtime

alexeagle:
This allows us to always publish regular bzl_library targets. We can also stop carefully vendoring skylib and shipping our own copy.
Eventually we would run into a provider problem, where our vendored copy doesn't have the same provider symbol as the real repo.

BREAKING CHANGE:

build_bazel_rules_nodejs now depends on bazel_skylib.
You can install it in your WORKSPACE directly, or use our helper macro like so:

### DIFF
--- a/e2e/bazel_managed_deps/WORKSPACE
+++ b/e2e/bazel_managed_deps/WORKSPACE
@@ -25,6 +25,10 @@ http_archive(
     urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/4.4.0/rules_nodejs-4.4.0.tar.gz"],
 )
 
+load("@build_bazel_rules_nodejs//nodejs:repositories.bzl", "rules_nodejs_dependencies")
+
+rules_nodejs_dependencies()
+
 load("@build_bazel_rules_nodejs//:index.bzl", "yarn_install")
 
 yarn_install(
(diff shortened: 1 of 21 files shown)


e2e/bazel_managed_deps/WORKSPACE
```diff
@@ -25,6 +25,10 @@ http_archive(
     urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/4.4.0/rules_nodejs-4.4.0.tar.gz"],
 )
 
+load("@build_bazel_rules_nodejs//nodejs:repositories.bzl", "rules_nodejs_dependencies")
```
gregmagolan, line 28: I've seen this pattern around. rules_docker uses it
```
load("@build_bazel_rules_nodejs//nodejs:repositories.bzl", rules_nodejs_dependencies = "dependencies")
rules_nodejs_dependencies()
```
It avoids the unnecessary scoping in the rnj repo. Wdyt?

gregmagolan: rules_docker:
```
load(
    "@io_bazel_rules_docker//repositories:repositories.bzl",
    container_repositories = "repositories",
)
container_repositories()
```

alexeagle: I think it makes users use the re-assignment syntax in the load statement that they may not have seen, and is overall more complex to have a second name for the thing (some users might not re-assign)

alexeagle: rules_docker kinda needs it since they have so many language ecosystems to namespace?

gregmagolan (approved):
👾

Issue #3054, opened 2021-10-31
change in run_node behaviour from 4.4.1 to 4.4.2


dymart:
# 🐞 bug report

### Affected Rule

There seems to have been a change to the setup or execution of run_node. 
When updating to 4.4.2 the rule that uses run_node would no longer build correctly.
The rules work as expected under 4.4.1 and 4.4.0. 

### Is this a regression?

This seems to be a regression or a change in behaviour between minor releases. I'm not sure if the previous behaviour was taking advantage of unintentional functionality or if the change was intentional. 

When updating to 4.4.2 the rule that uses run_node would no longer build correctly.
The rules work as expected under 4.4.1 and 4.4.0. 

### Description

The rule is running a nextjs build with run_node. With 4.4.1 everything works as expected. When changing to 4.4.2 the build breaks. I'm not very familiar with what nextjs does for it's build process in the background but a change in the nodejs rules seems to have changed it's ability to complete a build.

## 🔬 Minimal Reproduction

A branch to demonstrate the problem has been created: https://github.com/ubiquitoustech/rules_nextjs/tree/debug
The debug branch is the one that includes the setup to debug the problem.
In the workspace under examples/with-next-css is the example build that breaks. 

from examples/with-next-css run:
$ bazel build prod

The workspace is setup to run on the broken commit initially.
https://github.com/ubiquitoustech/rules_nextjs/blob/50d111513d56111d5db3e03415f8471361def33a/examples/with-next-css/WORKSPACE#L72

Then if you switch the workspace to use the previous commit it builds.
You would need to comment out the above import and remove the comment from this one and build again.
https://github.com/ubiquitoustech/rules_nextjs/blob/50d111513d56111d5db3e03415f8471361def33a/examples/with-next-css/WORKSPACE#L55

This seems to be the commit that changes functionality: https://github.com/bazelbuild/rules_nodejs/commit/f6db6c918de7db1ebd41dc5aa08714d47f2d0545

The repo is large right now because there are some extra commits from rules_nodejs included in the external folder of examples/with-next-css

## 🔥 Exception or Error

The error is one from the nextjs build and I'm not sure what the root cause of that error would be. 

This seems to be the line in nextjs that is the error. https://github.com/vercel/next.js/blob/99abb8bfd7f662efb942763332b87f4348b8844d/packages/next/build/index.ts#L136

[...]

PR #3059, opened 2021-11-04
fix: #3054 regression in linker behavior in 4.4.2

gregmagolan:
resolves: https://github.com/bazelbuild/rules_nodejs/issues/3054

This restores the linker behaviour from 4.4.1.

The problem seems to be that after the refactor in 4.4.2, the linker is now creating a node_modules symlink at the root of the bazel-out tree where as before the refactor it wasn't.

In short, in 4.4.2 there is now node_modules symlinked at,

`execroot/<wksp>/bazel-out/darwin-fastbuild/bin/node_modules` -> `execroot/<wksp>/node_modules`

while in 4.4.1 this wasn't created.

The error for the nextjs build AFAICT is that it is resolving react to the two different node_modules trees in the build which is breaks the react build. React requires a single version at a single node_modules location. I've seen the error produced by next.js before due to this problem in another scenario:

```
Error occurred prerendering page "/500". Read more: https://nextjs.org/docs/messages/prerender-error
Error: Minified React error #321; visit https://reactjs.org/docs/error-decoder.html?invariant=321 for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
```

The `execroot/<wksp>/bazel-out/darwin-fastbuild/bin/node_modules` is redundant anyway since standard node_modules resolution will go down the tree and find `execroot/<wksp>/node_modules` without any extra help.

### DIFF
--- a/internal/linker/index.js
+++ b/internal/linker/index.js
@@ -337,9 +337,11 @@ function main(args, runfiles) {
             else {
                 yield mkdirp(execrootNodeModules);
             }
-            const packagePathBin = path.posix.join(bin, packagePath);
-            yield mkdirp(`${packagePathBin}`);
-            yield symlinkWithUnlink(execrootNodeModules, `${packagePathBin}/node_modules`);
+            if (packagePath) {
+                const packagePathBin = path.posix.join(bin, packagePath);
+                yield mkdirp(`${packagePathBin}`);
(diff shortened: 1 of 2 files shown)


alexeagle (approved):
🧙🏻

matthewt-codaio:
Thank you. We've been seeing this with 4.4.2 and had not yet understood the root cause.

PR #3083, opened 2021-11-17
fix(builtin): js_library exposed in external workspace not linked properly

devversion:
See individual commits

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -54,6 +54,11 @@ local_repository(
     path = "internal/pkg_npm/test/vendored_external",
 )
 
+local_repository(
+    name = "internal_linker_external_library_wksp",
+    path = "internal/linker/test/external_library/workspace",
+)
+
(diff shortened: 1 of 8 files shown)


packages/rollup/test/integration/far/a/BUILD.bazel
```diff
@@ -10,5 +10,5 @@ ts_project(
 js_library(
     name = "a",
     package_name = "@far/a",
-    deps = [":tsconfig"],
+    srcs = [":tsconfig"],
```
devversion, line 13: This is an example change to show how a clear separation of sources vs. actual runtime deps shoukd be declared.. allowing for better determination of sources in `js_library`. This would also become relevant for supporting e.g. transitioned outputs in a `js_library`. Right now the fact that both `deps` and `srcs` is considered is really magical and unexpected IMO.

Issue #3085, opened 2021-11-18
Apple M1 Mac (Darwin arm64) support for local development


CooperBills:
# Apple M1 Mac (Darwin arm64) support for local development

### Description

Currently, `rules_nodejs` fails to build with Apple M1 Macs (Darwin arm64). This issue is to track updates to rules_nodejs to support local development on Macs using M1 processors (Darwin arm64). 

### Changes to enable support:
- [ ] Bump Bazel to 4.2.0 (4.2.0 includes a bugfix to properly report the CPU architecture on macs, which rules_go needs. See: https://github.com/bazelbuild/bazel/issues/12671 for more details)
- [x] Update tsconfigs to search `darwin_arm64-fastbuild` 
    * Done: https://github.com/bazelbuild/rules_nodejs/pull/3087
- [x] Bump `@io_bazel_rules_webtesting//web/versioned:browsers-0.3.2.bzl` to `0.3.3`, which includes drivers and browsers for macos arm64 targets (see: https://github.com/bazelbuild/rules_webtesting/pull/427)
    * Done: https://github.com/bazelbuild/rules_nodejs/pull/3086
- [ ] Modify Cypress toolchain to allow running on mac arm64. Currently, Cypress does not have an arm64 build and officially only supports running Cypress under rosetta. The rules_nodejs toolchains lock Cypress to only x86, and would need to be opened up to allow running Cypress on M1 Macs (via rosetta).
- [x] Update `//packages/concatjs:docs_scrub_platform` to also scrub arm64 platform builds.
    * Done: https://github.com/bazelbuild/rules_nodejs/pull/3089

The above changes enable successful full builds (`bazel build //...`) and testing (`bazel test //...`), with ~97% tests passing (the last 3% is because headless firefox is failing with timeouts).

I have these changes locally, and will be opening PRs for them in the next day or so.

alexeagle:
Yes, users need Bazel 4.2.0, but I'm not sure how rules_nodejs has responsibility for that.

The other fixes sound right to me, thanks!

CooperBills:
Ah, to clarify: it's to bump the `.bazelversion` (and corresponding locations) in the repository so contributors will use 4.2.0 (It's currently 4.1.0).

CooperBills:
@alexeagle Looks like you already began bumping to 4.2.0 a month or so ago: https://github.com/bazelbuild/rules_nodejs/pull/2926

alexeagle:
[...]

PR #3088, opened 2021-11-19
fix: allow cypress to run on m1 macs

CooperBills:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Bazel fails to run Cypress tests with an M1 Mac (Darwin arm64) host, as it is unable to find a toolchain that doesn't limit the host to an x86 CPU.

Issue Number: https://github.com/bazelbuild/rules_nodejs/issues/3085

## What is the new behavior?
The CPU architecture compatibility restriction is removed in this PR, so Cypress will run using the x86 build under rosetta (note: Cypress' [official take](https://www.cypress.io/blog/2021/01/20/running-cypress-on-the-apple-m1-silicon-arm-architecture-using-rosetta-2/) is: "the way to run Cypress on Apple M1 ARM chip is by using Intel emulation with Rosetta 2, Apple's translation layer."). As of opening this PR, Cypress still does not have a native arm64 build.

- [ ] Yes
- [X] No

## Other information

Discussion/Questions:
* Is this fine, or should we be defining a separate toolchain_config for `darwin_arm64` with another appropriately named toolchain and `@cypress_*`external which really just pulls the x86 build again under the hood?
* Should we add any documentation stating that rosetta is required on the host? (it's very likely that any mac developer will already have it installed)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -112,6 +112,7 @@ load("//packages/cypress:index.bzl", "cypress_repositories")
 
 cypress_repositories(
     name = "cypress",
+    darwin_arm64_sha256 = "101a0ced77fb74b356800cb3a3919f5288d23cc63fdd39a0c500673159e954fc",
     darwin_sha256 = "101a0ced77fb74b356800cb3a3919f5288d23cc63fdd39a0c500673159e954fc",
     linux_sha256 = "d8ea8d16fed33fdae8f17178bcae076aaf532fa7ccb48f377df1f143e60abd59",
     version = "7.3.0",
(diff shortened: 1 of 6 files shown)


alexeagle:
> defining a separate toolchain_config for darwin_arm64 with another appropriately named toolchain and @cypress_*external which really just pulls the x86 build again under the hood

I think this is probably better, since it gives us the ability to warn about rosetta, and when/if they do release M1 binaries, it's a smaller change to modify that toolchain_config than to introduce a new one at that time.

Note that I'm hoping to spin out cypress to its own repository so that it could have independent maintainers, but probably not happening before 5.x.

Thanks!

alexeagle (approved):
Thanks!

this toolchain needs some cleanup to avoid eager-evaluation of the RHS of the `select()` statement. I think right now it will download for all platforms? Maybe even an extra download since you have another `http_archive` here now.

CooperBills:
Yeah, I'm pretty sure it will end up downloading another archive... Is that acceptable or would you like me to take a look at lazy-loading the cypress toolchains before merging this? Do you know of documentation or a good example I could look at for lazy evaluation of select statements for toolchains? (Only thing I could find was rules_rust's "proxy repositories": https://github.com/bazelbuild/rules_rust/blob/main/rust/repositories.bzl#L263-L264)

alexeagle:
https://github.com/aspect-build/bazel_rules_template/blob/main/mylang/private/toolchains_repo.bzl has some docs about that.

Pretty sure @mrmeku will make a new cypress rule from that template sometime soon, and we'll port over any fixes from here. So I think this is good to merge. Thanks!

PR #3096, opened 2021-11-24
fix: don't symlink execroot node_modules with a package_path when under bazel run

gregmagolan:
In the bazel run context there is no sandbox so node_modules not at the root (with a package_path in their `yarn_install` or `npm_install` rules) that as symlinked to a folder such as `execroot/<workspace>/<package_path>/node_modules` will end up in the user's workspace since `execroot/<workspace>/<package_path>` is a symlink into the user's workspace. This can lead to non-determinstic behavior and a corruption of the bazel managed node_modules if a user then runs `yarn install` or `npm install` outside of bazel. Those tools will follow the symlink that bazel left behind and operate in the bazel managed `node_modules` in the external repository created by `yarn_install` or `npm_install`.

With this change, the "primary" `node_modules` (where the linker symlinks 1p deps into and where other `node_modules` locations symlink to) for a node_modules within a package (ie, `.../<package_path>/node_modules`) is now the bin location since the execroot location may not exist in the `bazel run` context.

The node patches are upgraded to supported * segments in guarded roots so launcher can guard `bazel-out/*/bin/<package_path>/node_modules` directories.

### DIFF
--- a/internal/node/launcher.sh
+++ b/internal/node/launcher.sh
@@ -271,29 +271,24 @@ else
   RUNFILES_ROOT=
 fi
 
-# Tell the node_patches_script that programs should not escape the execroot
-export BAZEL_PATCH_ROOTS="${EXECROOT}"
-# Set all bazel managed node_modules directories as guarded so no symlinks may
-# escape and no symlinks may enter.
-# We always guard against the root node_modules where 1st party deps go.
-# (e.g., /private/.../execroot/build_bazel_rules_nodejs/node_modules)
-export BAZEL_PATCH_ROOTS="${BAZEL_PATCH_ROOTS},${EXECROOT}/node_modules"
+# Tell the node_patches_script that programs should not escape the execroot or
+# root node_modules
+# For example,
+#   .../execroot/build_bazel_rules_nodejs
+#   .../execroot/build_bazel_rules_nodejs/node_modules
+export BAZEL_PATCH_ROOTS="${EXECROOT},${EXECROOT}/node_modules"
 if [[ "${RUNFILES_ROOT}" ]]; then
-  # If in runfiles, guard the runfiles root itself
-  export BAZEL_PATCH_ROOTS="${BAZEL_PATCH_ROOTS},${RUNFILES_ROOT}"
-  # If in runfiles guard the node_modules location in runfiles as well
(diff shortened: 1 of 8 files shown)


packages/node-patches/test/fs/lstat.ts
```diff
@@ -38,7 +38,7 @@ describe('testing lstat', () => {
 
           const patchedFs = Object.assign({}, fs);
           patchedFs.promises = Object.assign({}, fs.promises);
-          patcher(patchedFs, [path.join(fixturesDir)]);
+          patcher(patchedFs, [path.join(fixturesDir)], []);
```
alexeagle, line 41: nit: would be nice to make a `const NO_GUARDS = []` so the meaning of the arguments is more obvious.

Alternatively maybe the function wants to have a default for this new argument so you don't need all these updates in the test

internal/node/launcher.sh
```diff
@@ -273,23 +273,15 @@ fi
 
 # Tell the node_patches_script that programs should not escape the execroot
 export BAZEL_PATCH_ROOTS="${EXECROOT}"
-# Set all bazel managed node_modules directories as guarded so no symlinks may
-# escape and no symlinks may enter.
-# We always guard against the root node_modules where 1st party deps go.
-# (e.g., /private/.../execroot/build_bazel_rules_nodejs/node_modules)
-export BAZEL_PATCH_ROOTS="${BAZEL_PATCH_ROOTS},${EXECROOT}/node_modules"
+# Make execroot/external a patch guard so that symlinks can't resolve into it
+export BAZEL_PATCH_GUARDS="${EXECROOT}/external"
```
alexeagle, line 277: feels like it would be nice to have some prose like a README that we keep up-to-date with the code changes? If/when we want to write a new linker we'll need all the goodies that you currently have in your head.

alexeagle (approved):

packages/node-patches/test/fs/lstat.ts
gregmagolan: Good idea. Will update.

internal/node/launcher.sh
gregmagolan: Indeed. I should write a few things down while its fresh. Detailing all of the different scenarios will take some effort. Sandbox vs no-sandbox. Windows vs posix. Runfiles vs execroot. Bazel run vs bazel build vs bazel test. etc...

packages/node-patches/test/fs/lstat.ts
gregmagolan: Ended up not going with an API change on that function .

PR #3108, opened 2021-12-02
fix(builtin): support mjs/cjs files as javascript files in `js_library`

devversion:
Allows for the exposal of `mjs` and `cjs` files in `js_library` targets.
Thsse are valid alternatives to the ambiguous `.js` extension that is
reliant on the closest `package.json` and its `type` field.


gregmagolan (approved):
👾

Issue #3113, opened 2021-12-04
Using Starlark build setting values as `esbuild` `define`s


fmeum:
# feature request
### Relevant Rules
I would like to add an attribute to the existing `esbuild` rule.
### Description
The existing `esbuild` rule offers a `define` attribute that can be used to replace global identifiers in the output bundle with strings from a dict that are subject to Make variable and location expansion.
Given that the more modern way to pass configurable settings into Bazel is the use of Starlark build settings (e.g. those defined in bazel_skylib), it would be very useful if `esbuild` would allow to globally replace identifiers with their values.
### Describe the solution you'd like
A label-keyed string dict attribute could be added to `esbuild`, where labels are expected to be Starlark build settings (providing bazel_skylib's `BuildSettingInfo`) and mapped to the identifier their value should replace in the resulting bundle.
A problem with this approach is that `esbuild` would need an honest, non-vendored dependency on `bazel_skylib` to access the `BuildSettingInfo` provider. If this is acceptable, I could provide a PR that implements this feature.
### Describe alternatives you've considered
Create a separate rule that consumes build settings and writes out an esbuild config that contains the relevant defines. This would however require lots of boilerplate and nested configs.

mattem:
Closed via #3122

PR #3122, opened 2021-12-06
feat(esbuild): make Starlark build settings usable as defines

fmeum:
Adds a `define_settings` attribute to the `esbuild` rules that allows using the values of Starlark build settings to globally replace
specified identifiers.

The value of the setting is automatically converted to a JS literal.

**Note:** This currently uses json.encode, which is only available from Bazel 5 on. If this addition is generally viewed favorably, I would look into "polyfilling" this functionality in Starlark.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #3113

## What is the new behavior?

The existing `esbuild` rule now has the `define_settings` attribute described in the issue.

- [ ] Yes
- [x] No

### DIFF
--- a/docs/esbuild.md
+++ b/docs/esbuild.md
@@ -102,10 +102,10 @@ This will create an output directory containing all the code split chunks, along
 **USAGE**
 
 <pre>
-esbuild(<a href="#esbuild-name">name</a>, <a href="#esbuild-args">args</a>, <a href="#esbuild-args_json">args_json</a>, <a href="#esbuild-config">config</a>, <a href="#esbuild-define">define</a>, <a href="#esbuild-deps">deps</a>, <a href="#esbuild-entry_point">entry_point</a>, <a href="#esbuild-entry_points">entry_points</a>, <a href="#esbuild-external">external</a>, <a href="#esbuild-format">format</a>,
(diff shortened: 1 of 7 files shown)


fmeum:
The buildkite build fails for the examples since they explicitly reference `rules_nodejs` 4.4.6, which of course does not yet include the required change to the `esbuild_repositories` macro introduced by this PR. Is this check expected to fail?

alexeagle:
There's a bit of magic, the integration test runner replaces those versions with the `local_repository(path=../..)` like you'd expect. Your code changes should be showing up inside the example, so probably something is actually wrong.

fmeum:
> There's a bit of magic, the integration test runner replaces those versions with the `local_repository(path=../..)` like you'd expect. Your code changes should be showing up inside the example, so probably something is actually wrong.

The version of skylib fetched in `esbuild_repositories` was too old. I updated all reference to `bazel_skylib` to 1.1.1.

fmeum:
Is there a way to get the testlogs of the inner Bazel execution in an e2e test? The Windows test fails at `//:test_acorn`, but without the logs its hard to tell why.

fmeum:
> Is there a way to get the testlogs of the inner Bazel execution in an e2e test? The Windows test fails at `//:test_acorn`, but without the logs its hard to tell why.

I noticed that the `e2e/core` `WORKSPACE` referred to a `bazel_skylib` patch, which I apparently dropped with the update. I now changed my commit to only update the `bazel_skylib` dependency in `esbuild_repositories` and the tests pass. This could potentially clash with `rules_nodejs_dependencies` now, but maybe that macro has been deprecated? I don't see it mentioned in the docs anymore.

alexeagle (approved):
Thanks!

Issue #3124, opened 2021-12-07
select in toolchains causes eager fetch for all platforms


alexeagle:
This should be removed:

https://github.com/bazelbuild/rules_nodejs/blob/stable/toolchains/node/BUILD.bazel#L94-L125

we have a `resolved_toolchain` mechanism in 5.x instead.

mattem:
@alexeagle believes this is fixed now

mattem:
Closed via #3172

PR #3125, opened 2021-12-08
feat(node): use multiple versions of node, can run old and new toolch…

dymart:
…ains and default behaviour is not broken

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Can only use one nodejs version for a workspace

## What is the new behavior?
Able to use multiple nodejs versions in the same workspace

- [ ] Yes
- [x] No

## Other information
Hopefully backward compatible to make transitions easier

### DIFF
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -248,7 +248,9 @@ fi
     #
     # Rules such as nodejs_image should use only ctx.toolchains["@build_bazel_rules_nodejs//toolchains/node:toolchain_type"].nodeinfo
     # when building the image as that will reflect the selected --platform.
-    node_tool_files = ctx.files._node[:]
+    node_tool_files = ctx.files.node[:]
+
+    # this should be resolved the same as above
     node_tool_files.extend(ctx.toolchains["@build_bazel_rules_nodejs//toolchains/node:toolchain_type"].nodeinfo.tool_files)
 
     node_tool_files.append(ctx.file._link_modules_script)
@@ -258,7 +260,7 @@ fi
     node_tool_files.append(ctx.file._lcov_merger_script)
     node_tool_files.append(node_modules_manifest)
 
-    is_builtin = ctx.attr._node.label.workspace_name in ["nodejs_%s" % p for p in BUILT_IN_NODE_PLATFORMS]
+    is_builtin = ctx.attr.node.label.workspace_name in ["nodejs_%s" % p for p in BUILT_IN_NODE_PLATFORMS]
 
     runfiles = runfiles[:]
     runfiles.extend(node_tool_files)
@@ -317,7 +319,7 @@ fi
         "TEMPLATED_repository_args": _to_manifest_path(ctx, ctx.file._repository_args),
         "TEMPLATED_require_patch_script": _to_manifest_path(ctx, ctx.outputs.require_patch_script),
         "TEMPLATED_runfiles_helper_script": _to_manifest_path(ctx, ctx.file._runfile_helpers_main),
-        "TEMPLATED_vendored_node": "" if is_builtin else strip_external(ctx.file._node.path),
+        "TEMPLATED_vendored_node": "" if is_builtin else strip_external(ctx.file.node.path),
     }
 
     # TODO when we have "link_all_bins" we will only need to look in one place for the entry point
@@ -591,7 +593,7 @@ Predefined genrule variables are not supported in this context.
         default = Label("//internal/node:loader.js"),
         allow_single_file = True,
     ),
-    "_node": attr.label(
+    "node": attr.label(
         default = Label("@nodejs//:node_bin"),
         allow_single_file = True,
     ),
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -167,3 +167,28 @@ load("@build_bazel_integration_testing//tools:repositories.bzl", "bazel_binaries
 
 # Depend on the Bazel binaries
 bazel_binaries(versions = SUPPORTED_BAZEL_VERSIONS)
+
+# Importing rules_nodejs to use with nodejs_binary and nodejs_test as they transition to using toolchains
+# provided by these rules
+
+local_repository(
+    name = "rules_nodejs",
+    path = ".",
+)
+
+load("@rules_nodejs//nodejs:repositories.bzl", "nodejs_register_toolchains", "rules_nodejs_dependencies")
+
+# This just gives us bazel-skylib
+rules_nodejs_dependencies()
+
+# The order matters because Bazel will provide the first registered toolchain when a rule asks Bazel to select it
+# This applies to the resolved_toolchain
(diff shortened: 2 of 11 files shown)


examples/multiple_node_versions/WORKSPACE
```diff
@@ -0,0 +1,41 @@
...
+    path = "../../../rules_nodejs",
+)
+
+load("@build_bazel_rules_nodejs//nodejs:repositories.bzl", "nodejs_register_toolchains", "rules_nodejs_dependencies")
+load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories")
+
+# This just gives us bazel-skylib
+rules_nodejs_dependencies()
+
+# if changes are made to the order update the tests as the test_node_version_default uses this order
```
alexeagle, line 27: I'd say "The order matters because Bazel will provide the first registered toolchain when a rule asks Bazel to select it"

internal/node/node.bzl
```diff
@@ -591,7 +593,8 @@ Predefined genrule variables are not supported in this context.
         default = Label("//internal/node:loader.js"),
         allow_single_file = True,
     ),
-    "_node": attr.label(
+    "node": attr.label(
+        # the default needs to change and it shouldn't be private so that it can be set by the user or a macro
```
alexeagle, line 597: "shouldn't be private" is now out-of-date right?

alexeagle: I think https://github.com/bazelbuild/rules_nodejs/issues/3124 is where we'll be forced to change the default

internal/node/node.bzl
```diff
@@ -638,6 +641,7 @@ nodejs_binary_kwargs = {
     "implementation": _nodejs_binary_impl,
     "outputs": _NODEJS_EXECUTABLE_OUTPUTS,
     "toolchains": [
+        # what is the best way to resolve this
```
alexeagle, line 644: is that a question for me?

alexeagle (commented):
nice start!!

examples/multiple_node_versions/BUILD.bazel
```diff
@@ -0,0 +1,125 @@
+load("@bazel_skylib//rules:write_file.bzl", "write_file")
+load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
+load(":defs.bzl", "my_nodejs")
+load("@build_bazel_rules_nodejs//:index.bzl", "nodejs_binary", "nodejs_test")
+
+# Trivial test fixture: a nodejs program that writes to a file
+write_file(
+    name = "js",
+    out = "some.js",
+    content = ["require('fs').writeFileSync(process.argv[2], 'stuff')"],
```
gregmagolan, line 10: To keep from growing to much duplicate test code, lets move this all over to `e2e/core`. It looks like the additional of

```
    if ctx.attr.toolchain:
        toolchain = ctx.attr.toolchain[platform_common.ToolchainInfo].nodeinfo
    else:
        toolchain = ctx.toolchains["@rules_nodejs//nodejs:toolchain_type"].nodeinfo
```
& the new attr are the load bearing changes in `my_nodejs` rule.

I'd make those changes there, bring in multiple node versions in that WORKSPACE and then bring in a test case to test that you can specify a custom toolchain. You'll want to `select` the toolchain so that it still works on multiple platforms (since we won't be relying on bazel selecting the correct one).

Something like

```
my_nodejs(
    name = "run_15",
    out = "thing_15",
    entry_point = "version.js",
    toolchain = select({
        "@bazel_tools//src/conditions:linux_x86_64": "@node15_linux_amd64//:node_toolchain",
        "@bazel_tools//src/conditions:darwin": "@node15_darwin_amd64//:node_toolchain",
        "@bazel_tools//src/conditions:windows": "@node15_windows_amd64//:node_toolchain",
    }),               
)
```
(doesn't cover the M1 case but on an M1 you can run an intel binary in rosetta compat mode)

internal/node/node.bzl
```diff
@@ -248,7 +248,9 @@ fi
     #
     # Rules such as nodejs_image should use only ctx.toolchains["@build_bazel_rules_nodejs//toolchains/node:toolchain_type"].nodeinfo
     # when building the image as that will reflect the selected --platform.
-    node_tool_files = ctx.files._node[:]
+    node_tool_files = ctx.files.node[:]
```
gregmagolan, line 251: The changes to node.bzl are unrelated to the basic e2e test case above right?

Lets do these in a separate PR where we add the `toolchain` attribute.

gregmagolan (approved):

examples/multiple_node_versions/WORKSPACE
```diff
@@ -0,0 +1,41 @@
...
+
+# load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+# http_archive(
+#     name = "rules_nodejs",
+#     sha256 = "a54a01b883beff66da057a1a9afc3be83de566d8ba78f5139f9b370026af5868",
+#     urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/4.4.0/rules_nodejs-core-4.4.0.tar.gz"],
+# )
+
+local_repository(
```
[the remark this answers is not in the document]
dymart: Have some questions about what the workflow looks like for this. Would someone normally use local repository when they are actively developing locally and then when they have a change switch it to http archive for the pr? Or do people normally run the integration test runner locally? Or is there something else?

examples/multiple_node_versions/BUILD.bazel
alexeagle: note that the select statement will sadly download toolchains for all three platforms, I don't think it can be helped but you should mention that in a comment here

internal/node/node.bzl
dymart: yes that is out of date, this should be address in a follow up pr that helps remove the reference to @nodejs

internal/node/node.bzl
dymart: Sorry that comment slipped through. It's more relevant to a follow up pr

internal/node/node.bzl
dymart: this change is still needed to include nodejs_binary and nodejs_test in this pr where node can be specified or not. A follow up pr will address the toolchain changes. This allows a user to set the node to use but doesn't fully remove the toolchain yet. Might be better to address this in the follow up pr

examples/multiple_node_versions/BUILD.bazel
dymart: Sounds good, I'll move those things over and add the select statements

dymart:
The nodejs_binary has been removed from core but it uses the new core toolchain so the rules_nodejs core package was added to the base workspace. I want to make sure this is alright to do during the transition to core

PR #3129, opened 2021-12-09
fix(builtin): pkg_npm shouldn't assume the name of the nodejs toolchain

alexeagle:
Instead it should just ask the resolved toolchain for the info it needs


gregmagolan (approved):
🌮

nodejs/toolchain.bzl
```diff
@@ -22,6 +22,7 @@ NodeInfo = provider(
         "tool_files": """Files required in runfiles to make the nodejs executable available.
 
 May be empty if the target_tool_path points to a locally installed node binary.""",
+        "run_npm": "A template for a script that wraps npm",
```
gregmagolan, line 25: nit: mention that is a batch file if the platform is Windows, otherwise a shell script?

PR #3132, opened 2021-12-10
feat(): macros nodejs_binary_toolchains nodejs_test_toolchains input …

dymart:
…multiple toolchains nodejs_binary or nodejs_test

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Could only pass one toolchain to a rule

## What is the new behavior?

can pass multiple toolchains to a rule and it will create a new target for each toolchain

- [ ] Yes
- [x] No

### DIFF
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -250,9 +250,12 @@ fi
     # when building the image as that will reflect the selected --platform.
     node_tool_files = ctx.files.node[:]
 
-    # this should be resolved the same as above
-    node_tool_files.extend(ctx.toolchains["@build_bazel_rules_nodejs//toolchains/node:toolchain_type"].nodeinfo.tool_files)
+    if ctx.attr.toolchain:
+        node_toolchain = ctx.attr.toolchain[platform_common.ToolchainInfo]
+    else:
+        node_toolchain = ctx.toolchains["@build_bazel_rules_nodejs//toolchains/node:toolchain_type"]
 
+    node_tool_files.extend(node_toolchain.nodeinfo.tool_files)
     node_tool_files.append(ctx.file._link_modules_script)
     node_tool_files.append(ctx.file._runfile_helpers_bundle)
     node_tool_files.append(ctx.file._runfile_helpers_main)
@@ -597,12 +600,13 @@ Predefined genrule variables are not supported in this context.
         default = Label("@nodejs//:node_bin"),
         allow_single_file = True,
     ),
-    "_node_patches_script": attr.label(
-        default = Label("//internal/node:node_patches.js"),
-        allow_single_file = True,
-    ),
+    "toolchain": attr.label(),
     "_repository_args": attr.label(
         default = Label("@nodejs//:bin/node_repo_args.sh"),
+        allow_single_file = True,
+    ),
+    "_node_patches_script": attr.label(
+        default = Label("//internal/node:node_patches.js"),
         allow_single_file = True,
     ),
     "_require_patch_template": attr.label(
--- a/internal/node/test/BUILD.bazel
+++ b/internal/node/test/BUILD.bazel
@@ -604,3 +604,111 @@ nodejs_test(
         "@bazel_tools//src/conditions:windows": "@node16_windows_amd64//:node_toolchain",
     }),
 )
+
+# these rules create different targets for each toolchain that is supplied in the list for toolchains
+# this makes it easy to build or test against multiple versions of a toolchain at the same time with less duplicate work
+# this will create 2 nodejs_binary targets
+# bazel build //internal/node/test:main_multiple_loop_node16
+# bazel build //internal/node/test:main_multiple_loop_node15
+# the numbers are based on the order that the toolchains are passed to the rule
+[
+    nodejs_binary(
+        name = "main_multiple_loop_" + id,
+        entry_point = "binary_version.js",
+        node = tc,
+        toolchain = tc,
+    )
+    for id, tc in zip(
+        [
+            "node16",
+            "node15",
+        ],
+        [
+            # using the select statement will download toolchains for all platforms
+            # you can also just provide an individual toolchain if you don't want to download them all
+            select({
+                "@bazel_tools//src/conditions:linux_x86_64": "@node16_linux_amd64//:node_toolchain",
+                "@bazel_tools//src/conditions:darwin": "@node16_darwin_amd64//:node_toolchain",
+                "@bazel_tools//src/conditions:windows": "@node16_windows_amd64//:node_toolchain",
+            }),
+            select({
+                "@bazel_tools//src/conditions:linux_x86_64": "@node15_linux_amd64//:node_toolchain",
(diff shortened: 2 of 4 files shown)


internal/node/node.bzl
```diff
@@ -250,8 +250,10 @@ fi
     # when building the image as that will reflect the selected --platform.
     node_tool_files = ctx.files.node[:]
 
-    # this should be resolved the same as above
-    node_tool_files.extend(ctx.toolchains["@build_bazel_rules_nodejs//toolchains/node:toolchain_type"].nodeinfo.tool_files)
+    if ctx.attr.toolchain:
```
alexeagle, line 253: can reduce duplication by assigning to a local variable that holds either ctx.attr.toolchain or ctx.toolchains[default]

internal/node/node.bzl
```diff
@@ -658,6 +661,27 @@ def nodejs_binary_macro(name, **kwargs):
         **kwargs
     )
 
+def nodejs_binary_toolchain_macro(name, toolchains = [], **kwargs):
+    """This extens nodejs binary and allows for multiple toolchains to be passed as a list and
```
alexeagle, line 665: nit: extends

maybe get a spell checker into your editor?

internal/node/node.bzl
```diff
@@ -658,6 +661,27 @@ def nodejs_binary_macro(name, **kwargs):
         **kwargs
     )
 
+def nodejs_binary_toolchain_macro(name, toolchains = [], **kwargs):
```
alexeagle, line 664: I wonder if this is holding it's weight and worth putting in rules_nodejs code, rather than just put it in an example and let users maintain a little macro for this purpose. In all the life of the project I think we only heard of two users who want it

internal/node/test/BUILD.bazel
```diff
@@ -604,3 +604,75 @@ nodejs_test(
...
 )
+
+# these rules create different targets for each toolchain that is supplied in the list for toolchains
+# this makes it easy to build or test against multiple versions of a toolchain at the same time with less duplicate work
+
+# this will create 2 nodejs_binary targets
+# bazel build //internal/node/test:main_multiple_0
+# bazel build //internal/node/test:main_multiple_1
+# the numbers are based on the order that the toolchains are passed to the rule
+nodejs_binary_toolchains(
```
alexeagle, line 615: it might even be better to show it with a comprehension rather than a macro

```
[
  nodejs_binary(name = "x" + tc, ...)
  for tc in [...]
]
```

internal/node/test/BUILD.bazel
```diff
@@ -604,3 +604,75 @@ nodejs_test(
...
+    entry_point = "toolchain_test_v15.14.0.js",
+    toolchains = [
+        # using the select statement will download toolchains for all platforms
+        # you can also just provide an individual toolchain if you don't want to download them all
+        select({
+            "@bazel_tools//src/conditions:linux_x86_64": "@node15_linux_amd64//:node_toolchain",
+            "@bazel_tools//src/conditions:darwin": "@node15_darwin_amd64//:node_toolchain",
+            "@bazel_tools//src/conditions:windows": "@node15_windows_amd64//:node_toolchain",
+        }),
+        select({
```
alexeagle, line 649: seems like the two select's have the same content?

internal/node/node.bzl
dymart: I'm not sure what the outcome here should be. Should it switch to adding nodejs_binary to rules_nodejs then have an example there? Or just remove this macro and create an example that that shows users how to do it?

My only push back on not many users requesting it is, has there been many build tools that can make it this easy to run different toolchains on the same machine in parallel. Could be useful during upgrades to different node versions and allow better gradual migrations.

internal/node/test/BUILD.bazel
dymart: So this would be in the case we decide to remove the macro mentioned in the above comment?
Is that the preferred outcome?

alexeagle: We want to keep our maintenance burden in the upstream repo minimal, since it's not really funded. So any code maintained here ought to be broadly applicable. If only a couple users need it, then my instinct is it should live in their repo (we can have an example to help them).

dymart: Sounds Good! thanks for the clarification

internal/node/test/BUILD.bazel
```diff
@@ -604,3 +604,102 @@ nodejs_test(
...
+
+# these rules create different targets for each toolchain that is supplied in the list for toolchains
+# this makes it easy to build or test against multiple versions of a toolchain at the same time with less duplicate work
+
+# this will create 2 nodejs_binary targets
+# bazel build //internal/node/test:main_multiple_loop_0
+# bazel build //internal/node/test:main_multiple_loop_1
+# the numbers are based on the order that the toolchains are passed to the rule
+[
+    nodejs_binary(
```
gregmagolan, line 616: duplicate targets to what is stamped out below?
```
//internal/node/test:main_multiple_loop_node16
//internal/node/test:main_multiple_loop_node15
```
?

internal/node/test/BUILD.bazel
```diff
@@ -604,3 +604,102 @@ nodejs_test(
...
+
+# this will create 2 nodejs_binary targets
+# bazel build //internal/node/test:main_multiple_loop_node16
+# bazel build //internal/node/test:main_multiple_loop_node15
+# the numbers are based on the order that the toolchains are passed to the rule
+[
+    nodejs_binary(
+        name = "main_multiple_loop_" + id,
+        entry_point = "binary_version.js",
+        node = tc,
```
gregmagolan, line 646: Are you aiming to collapse node & toolchain into just one attr?

internal/node/test/BUILD.bazel
dymart: Yes they would be the same. Added to show different ways of doing the same thing. If there is a preferred way I can remove the other

internal/node/test/BUILD.bazel
dymart: I was hoping to get the use of toolchains added and then go back and clean this up. Originally it was behind a macro as well so a change collapsing the two wouldn't have been visible. But if we want to combine them now we could do that as well

internal/node/test/BUILD.bazel
gregmagolan: I think the `for id, tc in zip(` example with the clearer named targets is a better one so you can remove the `enumerate` one.

internal/node/test/BUILD.bazel
gregmagolan: A follow-up of removing `node` in favor `toolchain` sounds good to me. Seems easier to do it that way and the downstream effects will be easier to see.

gregmagolan (approved):
Looks great. Can you add a nodejs_test (like the one done in e2e/core) that the node version at runtime is what you expect. Can be done as a follow up to this.

dymart:
Added test to verify node version to this pr for multiple versions

PR #3134, opened 2021-12-10
feat(typescript): allow alternative transpilers

### LINKED ISSUE #3133: ts_project should allow alternative transpilers
Every TS project needs a typechecker and a transpiler. Currently the `ts_project` rule uses one `tsc` subprocess to do both.

However a common setup is to separate these, using some other transpiler like `babel` or `swc` to produce the .js outputs, which are consumed in the developer's workflow. The typechecks produce the `.d.ts` outputs and only need to be produced when testing the repo, and not in the developer's workflow (they probably have the TS language service in their editor so they already saw typecheck diagnostics as the coded)

The `ts_project` macro could accept a `transpiler` attribute that makes it easy to do this division of responsibility, and gives an easy path to making dev workflows 10x faster.

alexeagle:
Fixes #3133


thesayyn (approved):
🤌🏻

packages/typescript/internal/ts_project.bzl
```diff
@@ -756,7 +778,28 @@ def ts_project_macro(
...
+        tsc_map_outs = []
+        transpiler(
+            name = name + "_transpile",
+            srcs = srcs,
+            js_outs = js_outs,
+            map_outs = map_outs,
+            args = transpiler_args,
+            data = [],
+            tags = kwargs.get("tags", []),
+            visibility = kwargs.get("visibility", None),
```
gregmagolan, line 792: probably a few other well known attrs such as `test_only`. We have a list somewhere right?

packages/typescript/test/ts_project/swc/mock_swc.bzl
```diff
@@ -0,0 +1,11 @@
+"Mock swc transpiler macro"
+
+load("@bazel_skylib//rules:copy_file.bzl", "copy_file")
+
+def swc(name, srcs, js_outs, map_outs, args, data, tags, visibility):
```
gregmagolan, line 5: add test coverage for map_outs?

packages/typescript/internal/ts_project.bzl
```diff
@@ -540,6 +542,26 @@ def ts_project_macro(
...
+            can only see one set of dependencies, and so it cannot be shared between different
+            `ts_project` rules. That attribute is documented as experimental, and may never graduate
+            to a better support contract.
+
+            Instead, you can pass a rule or macro that accepts these arguments:
+            `(name, srcs, js_outs, map_outs, args, data, tags, visibility)`
+
+            The rules_nodejs authors believe that [SWC](https://swc.rs) is a great choice.
+
+        transpiler_args: if the `transpiler` attribute is a rule or macro, then this list
```
gregmagolan, line 562: what about making this a dict to support arbitrary transpiler attrs such as `swcrc`?

```
transpiler_attrs = {
    "args": [...],
    "swcrc": "...",
}
```

gregmagolan: you can check the type of a param in a macro so in theory you could overload `transpiler_args` to support either a list (passed to args) or a dict (to support multiple attrs)

packages/typescript/internal/ts_project.bzl
alexeagle: Yeah it's in aspect Bazel lib

packages/typescript/internal/ts_project.bzl
alexeagle: I did think about swcrc specifically, and a common case could be to pass it as a config setting `bazel build --@aspect_rules_swc//:rc=.swcrc` so it applies as the default for your whole build.
But of course someone will have the use case of different swcrc for two different folders of code, so yes I could see this being needed.

packages/typescript/internal/ts_project.bzl
alexeagle: actually I think we're both confusing it with https://github.com/aspect-build/bazel-lib/blob/main/docs/utils.md#propagate_well_known_tags

packages/typescript/internal/ts_project.bzl
thesayyn: People who are using this attributes  are most likely going to write a wrapper with defaults. Can't they just write a wrapper around swc that is passed to ts project which invokes swc with defaults? Does this sound too complex?

alexeagle: just changed it to transpiler_kwargs for simplicity

docs/TypeScript.md
```diff
@@ -733,6 +733,35 @@ Instructs Bazel *not* to expect `.js` or `.js.map` outputs for `.ts` sources.
...
+It also means developers must wait for the type-checking in the developer loop.
+
+In theory, Persistent Workers (via the `supports_workers` attribute) remedies the
+slow compilation time, however it adds additional complexity because the worker process
+can only see one set of dependencies, and so it cannot be shared between different
+`ts_project` rules. That attribute is documented as experimental, and may never graduate
+to a better support contract.
+
+Instead, you can pass a rule or macro that accepts these arguments:
+`(name, srcs, js_outs, map_outs, args, data, tags, visibility)`
```
gregmagolan, line 751: nit: update doc string to reflect that `testonly` is also passed

packages/typescript/internal/ts_project.bzl
```diff
@@ -756,7 +783,29 @@ def ts_project_macro(
...
+        transpiler(
+            name = name + "_transpile",
+            srcs = srcs,
+            js_outs = js_outs,
+            map_outs = map_outs,
+            data = [],
+            tags = kwargs.get("tags", []),
+            visibility = kwargs.get("visibility", None),
+            testonly = kwargs.get("testonly", None),
+            **transpiler_kwargs
```
gregmagolan, line 798: This is getting into the weeds but if you wanted to pass different `tags` to the transpiler then you could in theory write
```
ts_project(
   name = "foo",
   tags = ["foo"],
   transpiler_kwargs = {
      tags: ["bar"],
   }
```
and this could be supported if `kwargs.get("tags", [])` was merged into `transpiler_kwargs` if and only if `transpiler_kwargs` does not already have a `tags`.

gregmagolan: Complete list of "common" attributes to automatically merge into `transpiler_kwargs` is https://docs.bazel.build/versions/main/be/common-definitions.html#common-attributes

alexeagle: once we are passing through so much to the transpiler, it starts feeling like this macro is too leaky, and the user should just write their pair of typecheck(emitDeclarationOnly=true) and transpile() rules in their BUILD file.

thesayyn: How about instead of hiding this complexity inside ts_project, we expect users to write a simple wrapper and pass it to the transpiler?  in the case of `.swcrc` i believe a macro like below can be written once and reused everywhere. 

```python
def swc_with_swc_rc(**kwargs):
     return swc(swcrc = "//.swcrc", **kwargs)
```

packages/typescript/internal/ts_project.bzl
```diff
@@ -540,6 +546,27 @@ def ts_project_macro(
...
+            It also means developers must wait for the type-checking in the developer loop.
+
+            In theory, Persistent Workers (via the `supports_workers` attribute) remedies the
+            slow compilation time, however it adds additional complexity because the worker process
+            can only see one set of dependencies, and so it cannot be shared between different
+            `ts_project` rules. That attribute is documented as experimental, and may never graduate
+            to a better support contract.
+
+            Instead, you can pass a rule or macro that accepts these arguments:
+            `(name, srcs, js_outs, map_outs, args, data, tags, visibility)`
```
thesayyn, line 562: Upon my experimentation using our own swc rules, I realized that this won’t work out of the box. macro does not accept many of these attributes. and some of these attributes get declared twice. 
https://github.com/aspect-build/rules_swc/blob/33d5a39030cb87b8d5bc2f03473429d49ab8657b/swc/swc.bzl#L38

so you end up having an error that you don’t know what to do with.

alexeagle: Right, the swc_rule is what exposes this API

packages/typescript/internal/ts_project.bzl
alexeagle: @tjgq I wonder if you have thoughts about this. I think we discussed the idea once?

alexeagle: oh @thesayyn you can use https://github.com/bazelbuild/bazel-skylib/blob/main/lib/partial.bzl for this, so that the swcrc argument is bound where you make the higher-order function, then ts_project calls that partial.

(Bazel 5 will be nicer here since you could just use a lambda)

PTAL

thesayyn (approved):
using a dict is a really clever trick. looks neat!

looking forward see lambdas supported.

alexeagle:
Hmm, trying this out in a downstream project, I think we didn't consider carefully enough what the shape of the dependency graph will be. If you just `dep` on a `ts_project` with this attribute set, you get .d.ts files but no way to collect up all the `.js` outputs. @dgp1130 since Angular has been thinking about type-check independent from the dev workflow maybe you have some minutes to chat about it with me?

PR #3142, opened 2021-12-13
fix(): remove _repository_args from nodejs_binary

dymart:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

nodejs_binary requires _repository_args and has a dependency on @nodejs

## What is the new behavior?

removes the need for @nodejs for _repository_args

- [ ] Yes
- [ ] No


gregmagolan:
Lets remove it out of `nodejs/repositories.bzl` and `nodejs/private/nodejs_repo_host_os_alias.bzl` in this PR

gregmagolan:
Thinking about the current default which is to add --preserve-symlinks. I think we should just go with the nodejs default of *not* setting that flag and the releases notes for 5.0 should be for users to add `--preserve-symlinks` to their args with
```
nodejs_binary(
args = [
   "--node_options=--preserve-symlinks",
]
)
```
if they want it set.

@alexeagle do you think that would be too breaking? I suppose we'll see how CI is if we flip the default in RNJ

dymart:
I've left "--preserve-symlinks" in for right now to pass the tests. I'll keep trying to get things working with the flag removed and then flip it when that works.

alexeagle:
I think this may be a use case for a custom flag.

https://docs.bazel.build/versions/main/configurable-attributes.html#custom-flags

So users would `bazel build --@rules_nodejs//nodejs:default_args="--something"` 
the default configuration would have the `--preserve-symlinks` in there.

dymart:
Great thanks for the recommendation @alexeagle I'll look into using a custom flag

dymart:
Hey @alexeagle just wanted to follow up in the example you gave it has these flags living in the nodejs folder. Would these flags still be used in the internal/node/node.bzl and passed along to launcher.sh through template variables. Or is there a different way you were thinking that these flags would be integrated in.

alexeagle:
I would expect the custom flag in the nodejs/ folder like that example, yes. Can then be used in internal/node/node.bzl or anywhere else that wants to reference them

nodejs/BUILD.bazel
```diff
@@ -36,3 +37,12 @@ toolchain_type(
     )
     for key, values in PLATFORMS.items()
 ]
+
+# User set flags that are passed to the build
+# bazel test --//nodejs:default_args="" //...
+# allowed args ["--preserve-symlinks", ""]
+user_args(
+    name = "default_args",
+    build_setting_default = "--preserve-symlinks",
```
alexeagle, line 46: if this is the default, then why did you have to add it explicitly to a bunch of examples? seems wrong, I'd expect the examples not to change, since those illustrate the changes users would have to make as well.

dymart: The flags were added to the examples to see how things would work when the default was changed. If we wanted to go with the nodejs default. As mentioned earlier in the comments on the pr. I'm happy to remove them from this pr though and reevaluate when we want to change the default.

gregmagolan: I think keeping the default to match what node_repositories() had makes sense for 5.0. We'd want to do some more due diligence before we flip the default.

nodejs/private/user_build_settings.bzl
```diff
@@ -0,0 +1,33 @@
+"""Function for user args to be set on the command line
+"""
+
+load("//nodejs/private/providers:user_build_settings.bzl", "UserBuildSettingInfo")
+
+# this can be used to limit user input for flags and also provide useful error messages for incorrect flags
+user_args_allowed = ["--preserve-symlinks", ""]
```
gregmagolan, line 7: I think we should just open this up and let users can set whatever default they want. Guessing they can set a defaults in their .bazelrc file with this improved approach?

docs/Cypress.md
```diff
@@ -156,6 +156,10 @@ Now you can add `--config=debug` to any `bazel test` command line.
 The runtime will pause before executing the program, allowing you to connect a
 remote debugger.
 
+You can also change the default args that are sent to nodejs. This can be done through a flag. The default is --preserve-symlinks while anything
+can be passed. The flag is --@build_bazel_rules_nodejs//nodejs:default_args="" ex: bazel test --@build_bazel_rules_nodejs//nodejs:default_args="test" main
```
gregmagolan, line 160: Lets add a real flag here to better illustrate what it does. Maybe pass two such as
`"--preserve-symlinks --no-warnings"` to show that you can pass multiple. A drop a link to https://nodejs.org/api/cli.html for reference.

gregmagolan (approved):
🌮

PR #3157, opened 2021-12-20
feat: support --stamp env vars in npm_package_bin

gregmagolan:
This allows npm_package_bin build actions to get volatile & stable status files using env vars `BAZEL_VERSION_FILE` and `BAZEL_INFO_FILE`. This is the same pattern that esbuild was using.


internal/providers/node_runtime_deps_info.bzl
```diff
@@ -70,6 +70,8 @@ def run_node(ctx, inputs, arguments, executable, chdir = None, **kwargs):
         arguments: list or ctx.actions.Args object containing arguments to pass to the executable
         executable: stringy representation of the executable this action will run, eg eg. "my_executable" rather than ctx.executable.my_executable
         chdir: directory we should change to be the working dir
+        stamp_env_vars: if True ctx.version_file and ctx.info_file are supplied as inputs to the action
```
alexeagle, line 73: I think the doc here needs to indicate plus and minus of doing this, like

"Allows the build result to contain version metadata like the git SHA, but makes the result non-determistic which causes cache misses for anything depending on it"

Also, I'm surprised this doesn't seem to be conditional on whether the build is performed with `--stamp` which I think it should be

gregmagolan: The condition check is done where `run_node` is called. esbuild.bzl & npm_package_bin both have
```
stamp_env_vars = ctx.attr.node_context_data[NodeContextInfo].stamp,
```

Suppose the condition could be moved to within `run_node` itself if the convention for the conditional attribute was always `node_context_data`

gregmagolan: Added that warning to the docstring

internal/providers/node_runtime_deps_info.bzl
```diff
@@ -70,6 +70,10 @@ def run_node(ctx, inputs, arguments, executable, chdir = None, **kwargs):
...
```
alexeagle, line 73: oh this is just `stamp` which is a boolean - but users of npm_package_bin (I guess pkg_npm in your case) need a trinary field where they can say stamp=always or stamp=never or stamp=use_stamp_flag, that's what gets plumbed to here

internal/providers/node_runtime_deps_info.bzl
gregmagolan: Refactored so run_node sniffs out the "stamp" attr

internal/providers/node_runtime_deps_info.bzl
gregmagolan: Refactored so run_node sniffs out the "stamp" var which is tri-state

gregmagolan:
Going to open a new one

PR #3187, opened 2022-01-05
fix(typescript): ts_project transpiler produces js_library

alexeagle:
This is needed so the .js files produced by the alternate transpiler
are passed down the dependency graph.

### DIFF
--- a/packages/typescript/internal/ts_project.bzl
+++ b/packages/typescript/internal/ts_project.bzl
@@ -4,6 +4,7 @@ load("@build_bazel_rules_nodejs//:providers.bzl", "DeclarationInfo", "ExternalNp
 load("@build_bazel_rules_nodejs//internal/linker:link_node_modules.bzl", "module_mappings_aspect")
 load("@build_bazel_rules_nodejs//internal/node:node.bzl", "nodejs_binary")
 load("@build_bazel_rules_nodejs//third_party/github.com/bazelbuild/bazel-skylib:lib/partial.bzl", "partial")
+load("@build_bazel_rules_nodejs//:index.bzl", "js_library")
 load(":ts_config.bzl", "TsConfigInfo", "write_tsconfig")
 
 _ValidOptionsInfo = provider()
@@ -45,6 +46,7 @@ _ATTRS = {
     "srcs": attr.label_list(allow_files = True, mandatory = True),
     "supports_workers": attr.bool(default = False),
     "tsc": attr.label(default = Label(_DEFAULT_TSC), executable = True, cfg = "host"),
+    "transpile": attr.bool(doc = "whether tsc should be used to produce .js outputs"),
     "tsconfig": attr.label(mandatory = True, allow_single_file = [".json"]),
 }
 
@@ -180,10 +182,6 @@ def _ts_project_impl(ctx):
     else:
         json_outs = []
 
-        # If there are no js_outs, that implies we are producing declarations only.
-        # We must avoid tsc writing any JS files in this case, as it was only run for typings.
-        arguments.add("--emitDeclarationOnly")
-
     outputs = json_outs + ctx.outputs.js_outs + ctx.outputs.map_outs + ctx.outputs.typings_outs + ctx.outputs.typing_maps_outs
     if ctx.outputs.buildinfo_out:
         arguments.add_all([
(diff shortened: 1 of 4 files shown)


internal/js_library/js_library.bzl
```diff
@@ -223,10 +223,12 @@ def _impl(ctx):
             if DefaultInfo in dep:
                 files_depsets.append(dep[DefaultInfo].files)
 
+    # We do not include typings_depsets in the runfiles because that would cause type-check actions to occur
+    # in every development workflow.
     runfiles = ctx.runfiles(
         files = all_files,
         transitive_files = depset(
-            transitive = files_depsets + typings_depsets,
+            transitive = files_depsets,
```
alexeagle, line 231: this reverts https://github.com/bazelbuild/rules_nodejs/pull/2547
which I think was a mistake, since it makes type-checking eager for a js_library, and it shouldn't be.
/cc @Aghassi

gregmagolan: Reverted in https://github.com/bazelbuild/rules_nodejs/pull/3189 for reference

packages/typescript/internal/ts_project.bzl
```diff
@@ -568,6 +576,18 @@ def ts_project_macro(
             will be called with the remaining arguments.
             See the packages/typescript/test/ts_project/swc directory for an example.
 
+            When a custom transpiler is used (i.e. `transpiler != "tsc"`) then this macro expands to these targets:
```
gregmagolan, line 579: Would a better default for `transpiler` be `None` to avoid a the magic string `tsc`? `None` implying that the default transpiler is _not_ overridden?

packages/typescript/internal/ts_project.bzl
```diff
@@ -568,6 +576,18 @@ def ts_project_macro(
...
+
+                - `[name]` - the default target is a `js_library` which can be included in the `deps` of downstream rules.
+                    Note that it will successfully build *even if there are typecheck failures* because the `tsc` binary
+                    is not needed to produce the default outputs.
+                    This is considered a feature, as it allows you to have a faster development mode where typechecking
+                    is not on the critical path.
+                - `[name].check` - this target will fail to build if the typechecking fails, useful for CI.
+                - `_[name]_tsc` - internal target which runs the binary from the `tsc` attribute
+                - ?? - whatever targets the custom transpiler rule/macro produces.
```
gregmagolan, line 588: nit: "Any additional targets the custom transpiler rule/macro produces.", the `??` seems like you forgot something

packages/typescript/internal/ts_project.bzl
alexeagle: sure, that's fine with me. Agree it's strange for it to accept various types, and we never intend any other string to be legal here

packages/typescript/internal/ts_project.bzl
```diff
@@ -568,6 +576,18 @@ def ts_project_macro(
...
             See the packages/typescript/test/ts_project/swc directory for an example.
 
+            When a custom transpiler is used (i.e. `transpiler != "tsc"`) then this macro expands to these targets:
+
+                - `[name]` - the default target is a `js_library` which can be included in the `deps` of downstream rules.
+                    Note that it will successfully build *even if there are typecheck failures* because the `tsc` binary
+                    is not needed to produce the default outputs.
+                    This is considered a feature, as it allows you to have a faster development mode where typechecking
+                    is not on the critical path.
+                - `[name].check` - this target will fail to build if the typechecking fails, useful for CI.
```
gregmagolan, line 586: How about `[name]_typings` to describe what you're building in the target. If it fails to build typings then you've got a typings error?

packages/typescript/internal/ts_project.bzl
```diff
@@ -568,6 +576,18 @@ def ts_project_macro(
...
+                - `_[name]_tsc` - internal target which runs the binary from the `tsc` attribute
+                - ?? - whatever targets the custom transpiler rule/macro produces.
+                    This could be one target per TypeScript input file.
```
gregmagolan, line 589: nit: "For example, ..."

packages/typescript/internal/ts_project.bzl
gregmagolan: Oh. I see, the `.check` target is the filegroup so the build failure will be
```
Failed to build `//path/to:_[name]_tsc` in the current naming
```

Perhaps `_[name]_tsc` should be `[name]_typings` in that case so the error message is sound?

packages/typescript/test/ts_project/swc/BUILD.bazel
```diff
@@ -31,10 +40,17 @@ ts_project(
...
-            "sourceMap": True,
-        },
-    },
+    tsconfig = _TSCONFIG,
+)
+
+ts_project(
+    name = "transpile_with_typeerror",
+    srcs = ["typeerror.ts"],
+    # The transpile_with_typeerror.check target will have a build failure
```
gregmagolan, line 49: How does the build failure get ignored on CI with this one?

packages/typescript/internal/ts_project.bzl
gregmagolan: Accepting different types can be useful in macros. I've used it here and there to make a public API flexible and cleaner than having different attributes.

alexeagle: changing to `None` as default

packages/typescript/test/ts_project/swc/BUILD.bazel
alexeagle: the manual tag gets propagated to the build targets, so we don't run them. The test explicitly asks for the js_library which passes because typecheck action doesn't run

gregmagolan: 🤦  I was starring right at it and didn't see the `"manual"` tag

gregmagolan (approved):
🌮  🚀

PR #3191, opened 2022-01-06
chore(typescript): allow newer TS versions

alexeagle:
Now that ts_library has moved to @bazel/concatjs, the @bazel/typescript package is free of code that uses TS internals.
Instead, it contains just ts_project which uses the tsc CLI and so there's little reason to be concerned that newer TypeScript causes a problem.
So just relax our peerDep constraint to have no upper bound on the TS version.


lencioni (approved):
Thanks!

PR #3195, opened 2022-01-08
fix(builtin): detect yarn 2+ berry and adjust CLI args

### LINKED ISSUE #1599: Yarn PnP Support
Yarn 2.0 has been released. https://dev.to/arcanis/introducing-yarn-2-4eh1
This release has some massive changes, the biggest one is using PnP by default. https://next.yarnpkg.com/features/pnp
However, we can disable the use of this with a pluin for a first cut, then later on aim to support PnP.

The migration guide is here https://next.yarnpkg.com/advanced/migration
From what I can see, with keeping node_modules, there aren't too many big changes, needs more investigation.

### LINKED ISSUE #3071: yarn_versions helper only includes v1
# 🚀 Yarn Berry Support

### Relevant Rules

[yarn_install](https://github.com/bazelbuild/rules_nodejs/blob/aa9882b9d5afaeb9815a3630017fa2911eeedb0a/internal/npm_install/npm_install.bzl#L808-L968)

### Description

When trying to use Yarn berry with the yarn_install rule, an error occurs because it implicitly sets the `--mutex network`, which is not supported by Yarn Berry.
The following implementation shows that the mutex option can be disabled by setting the `use_global_yarn_cache` to false, but the `--cache-folder` is implicitly set next.
https://github.com/bazelbuild/rules_nodejs/blob/aa9882b9d5afaeb9815a3630017fa2911eeedb0a/internal/npm_install/npm_install.bzl#L825-L834

In Yarn Berry, setting the `--cache-folder` in the CLI is deprecated, so an error will occur in this case as well.
https://github.com/yarnpkg/berry/blob/83311e1d4dec7fb6c0ead21f37efa9b687be58b9/packages/plugin-essentials/sources/commands/install.ts#L211-L219

The following is an example of WORKSPACE.bazel for using Yarn Berry.

```bazel
http_archive(
    name = "vendored_yarn_3_1_0",
    build_file_content = """exports_files(["vendored_yarn_3_1_0/berry--yarnpkg-cli-3.1.0/packages/yarnpkg-cli/bin/yarn.js"])""",
    urls = ["https://github.com/yarnpkg/berry/archive/refs/tags/@yarnpkg/cli/3.1.0.tar.gz"],
)

load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories")
node_repositories(
    vendored_yarn = "@vendored_yarn_3_1_0//:berry--yarnpkg-cli-3.1.0/packages/yarnpkg-cli",
)

load("@build_bazel_rules_nodejs//:index.bzl", "yarn_install")
yarn_install(
    name = "npm",
    args = ["--immutable"],
    frozen_lockfile = False,
    use_global_yarn_cache = False,
    data = [
      "@//:.yarnrc.yml",
      "@vendored_yarn_3_1_0//:berry--yarnpkg-cli-3.1.0/packages/yarnpkg-cli/bin/yarn.js",
    ],
    yarn_lock = "//:yarn.lock",
)
```

The ideas suggested in the issue comments below are helpful.
> 1. add a flag to remove the automatic addition of --mutex network
> 2. add a bool attribute yarn2_compatible that does not use --mutex network and replaces --frozen-lockfile with --immutable
> 3. detect that yarn2 is used and behave accordingly

https://github.com/bazelbuild/rules_nodejs/issues/1599#issuecomment-816480044

Personally, I think the 1st idea is better, as it seems to be easy to implement without destructive changes.
An example is shown below.

```diff
yarn_install(
    name = "npm",
    args = ["--immutable"],
    frozen_lockfile = False,
-   use_global_yarn_cache = False,
+   use_mutex = False,
    data = [
      "@//:.yarnrc.yml",
      "@vendored_yarn_3_1_0//:berry--yarnpkg-cli-3.1.0/packages/yarnpkg-cli/bin/yarn.js",
    ],
    package_json = "//:package.json",
    yarn_lock = "//:yarn.lock",
)
```

alexeagle:
Fixes #3071
Fixes #1599

### DIFF
--- a/internal/npm_install/npm_install.bzl
+++ b/internal/npm_install/npm_install.bzl
@@ -813,6 +813,17 @@ def _yarn_install_impl(repository_ctx):
     is_windows_host = is_windows_os(repository_ctx)
     node = repository_ctx.path(get_node_label(repository_ctx))
     yarn = get_yarn_label(repository_ctx)
+    result = repository_ctx.execute(
+        [repository_ctx.path(yarn), "--version"],
+        timeout = repository_ctx.attr.timeout,
+        quiet = repository_ctx.attr.quiet,
+    )
+    if result.return_code:
+        fail("yarn --version failed: %s (%s)" % (result.stdout, result.stderr))
+    if result.stdout.startswith("1."):
+        yarn_version = "classic"
+    else:
+        yarn_version = "berry"
 
     yarn_args = []
 
@@ -820,11 +831,14 @@ def _yarn_install_impl(repository_ctx):
     # file. To perform an yarn install use the vendord yarn binary with:
     # `bazel run @nodejs//:yarn install` or `bazel run @nodejs//:yarn install -- -D <dep-name>`
     if repository_ctx.attr.frozen_lockfile:
-        yarn_args.append("--frozen-lockfile")
+        if yarn_version == "classic":
+            yarn_args.append("--frozen-lockfile")
+        else:
+            fail("--frozen-lockfile should not be used with yarn 2+. Just pass arguments like --immutable.")
 
     if not repository_ctx.attr.use_global_yarn_cache:
         yarn_args.extend(["--cache-folder", str(repository_ctx.path("_yarn_cache"))])
-    else:
+    elif yarn_version == "classic":
         # Multiple yarn rules cannot run simultaneously using a shared cache.
         # See https://github.com/yarnpkg/yarn/issues/683
         # The --mutex option ensures only one yarn runs at a time, see
@@ -832,6 +846,10 @@ def _yarn_install_impl(repository_ctx):
         # The shared cache is not necessarily hermetic, but we need to cache downloaded
         # artifacts somewhere, so we rely on yarn to be correct.
         yarn_args.extend(["--mutex", "network"])
+    else:
+        # Can't tell from documentation if Yarn Berry has any replacement for the --mutex
+        # flag. We'll have to assume it's safe to run concurrently.
+        pass
     yarn_args.extend(repository_ctx.attr.args)
 
     # Run the package manager in the package.json folder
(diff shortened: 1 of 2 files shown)


gregmagolan (approved):
:taco:

internal/npm_install/npm_install.bzl
```diff
@@ -813,25 +813,43 @@ def _yarn_install_impl(repository_ctx):
...
         # Multiple yarn rules cannot run simultaneously using a shared cache.
         # See https://github.com/yarnpkg/yarn/issues/683
         # The --mutex option ensures only one yarn runs at a time, see
         # https://yarnpkg.com/en/docs/cli#toc-concurrency-and-mutex
         # The shared cache is not necessarily hermetic, but we need to cache downloaded
         # artifacts somewhere, so we rely on yarn to be correct.
         yarn_args.extend(["--mutex", "network"])
+    else:
+        # Can't tell from documentation if Yarn Berry has any replacement for the --mutex
+        # flag. We'll have to assume it's safe to run concurrently.
```
merceyz, line 851: There is no replacement for it, we fixed the cache so it could be atomic.

sluongng: > Can't tell from documentation if Yarn Berry has any replacement for the --mutex flag.

My observation on this:

There are a lot of knobs being exposed through `.yarnrc` file https://yarnpkg.com/configuration/yarnrc#enableGlobalCache so long-term it's probably best if we stop exposing each individual flags through the attributes of `yarn_install` rule.

Instead, we can create a `yarnrc_file` attribute to raise awareness to folks that they should be passing the config file explicitly to the rule. Underneath implementation, we just merge that with `data` and retain current behavior.

sluongng (commented):
I looked at this but was afk yesterday so couldnt give it a proper review then.
I know these have been merged but I will just note down my findings so far for future references.

internal/npm_install/npm_install.bzl
alexeagle: That sounds like a good feature to me. However I don't see how to pass the config file path to `yarn install` - it looks like it's just automatically inferred from its location. Maybe this could be contributed separately. (maybe it already works just by passing `.yarnrc.yml` to the `data` of yarn_install and then it's auto-detected?)

sluongng: Yup, my understanding is that folks are already passing it through data and yarn auto detects it inside the sandbox.

lokshunhung: [yarn classic](https://github.com/yarnpkg/yarn/blob/300c47c336be0e8dbc722ef451148a79d9a35486/src/cli/index.js#L75) (v1) has flag called `--use-yarnrc` to specify the yarnrc location.

[yarn berry](https://github.com/yarnpkg/berry/blob/ce1e0c31781e3fc791cfa86a01348c4f0dba9aec/packages/yarnpkg-core/sources/Configuration.ts#L1133) searches for the rc file automatically and does not have this flag, but you could change the filename of the rc file to search for by setting the [environment variable](https://github.com/yarnpkg/berry/blob/ce1e0c31781e3fc791cfa86a01348c4f0dba9aec/packages/yarnpkg-core/sources/Configuration.ts#L869) `YARN_RC_FILENAME`

PR #3198, opened 2022-01-08
chore: cleanup windows build & test tags

gregmagolan:


.bazelci/presubmit.yml
```diff
@@ -231,20 +231,20 @@ tasks:
...
-    - "--test_tag_filters=-e2e,-examples,-fix-windows,-no-bazelci-windows,-requires-runfiles,-manual,-browser:firefox-local,-cypress"
+    - "--test_tag_filters=-e2e,-examples,-fix-windows,-no-bazelci-windows,-requires-runfiles,-browser:firefox-local,-cypress"
     test_targets:
     - "//..."
   windows_runfiles_enabled:
     name: windows_runfiles_enabled
     platform: windows
     test_flags:
-    - "--test_tag_filters=requires-runfiles"
+    - "--test_tag_filters=requires-runfiles,-fix-windows,-requires-runfiles"
```
gregmagolan, line 247: whoops. that is wrong.

docs/BUILD.bazel
```diff
@@ -61,9 +61,12 @@ _READMES = {
         failure_message = "Please run:  bazel run //docs:update",
         file1 = k + ".md",
         file2 = v,
-        # No need to run on more than one platform
-        # Also there's some delta here, probably line endings
-        tags = ["no-bazelci-windows"],
+        tags = [
+            # no need to run on more than one platform
+            "no-bazelci-windows",
+            # diff_test has line endings issues on Windows
```
alexeagle, line 67: is it diff_test or stardoc? I'm still not clear where this bug actually originates

.bazelci/presubmit.yml
gregmagolan: fixed

packages/concatjs/web_test/test/stack_trace/BUILD.bazel
```diff
@@ -33,8 +33,8 @@ karma_web_test_suite(
         "@io_bazel_rules_webtesting//browsers:firefox-local",
     ],
     tags = [
-        "manual",  # not run by CI
-        "native",
+        "manual",  # not run by CI (see comment above)
+        "native",  # rules_webesting requires the "native" tag for browsers
```
alexeagle, line 37: nit: spelling on webtesting

alexeagle (approved):

docs/BUILD.bazel
gregmagolan: hmm. let me double check

gregmagolan: checked locally. the only difference is the line endings so I'd call it a diff_test issue since we'd expect windows stardoc to output CRLF right?

packages/concatjs/web_test/test/stack_trace/BUILD.bazel
gregmagolan: 👍

PR #3201, opened 2022-01-09
refactor: stop vendoring bazel-skylib now that we depend on it

alexeagle:


alexeagle:
Maybe we should get the `copy_file` implementation from https://github.com/aspect-build/bazel-lib (along with other helpers) so we don't have copies here, and don't need to continue patching bazel-skylib which probably won't ever accept it.

PR #3202, opened 2022-01-09
chore: cleanup bazelrc and add --incompatible_exclusive_test_sandboxed to common and --keep_going for CircleCI

gregmagolan:


.bazelrc
```diff
@@ -54,3 +54,15 @@ build --incompatible_no_implicit_file_export
 # build --incompatible_struct_has_no_methods
 # TODO(alexeagle): turn on this flag when dependencies allow
 # build --incompatible_use_platforms_repo_for_constraints
+
+###############################
```
alexeagle, line 58: this already happened above when we ran
`import %workspace%/common.bazelrc`

I dunno if that's a problem or not

common.bazelrc
```diff
@@ -83,11 +83,5 @@ run --incompatible_strict_action_env
...
-# Load any settings specific to the current user.
-# .bazelrc.user should appear in .gitignore so that settings are not shared with team members
-# This needs to be last statement in this
-# config, as the user configuration should be able to overwrite flags from this file.
-# See https://docs.bazel.build/versions/master/best-practices.html#bazelrc
-# (Note that we use .bazelrc.user so the file appears next to .bazelrc in directory listing,
-# rather than user.bazelrc as suggested in the Bazel docs)
-try-import %workspace%/.bazelrc.user
+# Allow exclusive tests to run in a sandbox
+test --incompatible_exclusive_test_sandboxed
```
alexeagle, line 87: oh I see. The problem with removing it from here is that it won't apply in our other places. In particular, `@bazel/create` uses this file to bootstrap a users new project, so they won't get this either.

gregmagolan: oh I see. we were using the common.bazelrc for @bazel/create. it seems strange to me to have `.bazelrc` override the `.bazelrc.user` values. Top of `.bazelrc` has
```
# Import shared settings first so we can override below
import %workspace%/common.bazelrc
```

gregmagolan: Any suggestions?

gregmagolan: Well. We can figure that out later. I'll make a smaller diff with the `--keep_going` and `--incompatible_exclusive_test_sandboxed` flags

gregmagolan:
Will figure this out another day

PR #3203, opened 2022-01-09
chore: code review feedback for #3195

alexeagle:
cc @sluongng

### DIFF
--- a/internal/npm_install/npm_install.bzl
+++ b/internal/npm_install/npm_install.bzl
@@ -805,6 +805,15 @@ check if yarn is being run by the `npm_install` repository rule.""",
     implementation = _npm_install_impl,
 )
 
+def _detect_yarn_version(rctx, yarn):
+    result = rctx.execute([rctx.path(yarn), "--version"])
+    if result.return_code:
+        fail("yarn --version failed: %s (%s)" % (result.stdout, result.stderr))
+    if result.stdout.startswith("1."):
+        return "classic"
+
+    return "berry"
+
 def _yarn_install_impl(repository_ctx):
     """Core implementation of yarn_install."""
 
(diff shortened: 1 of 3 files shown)


internal/npm_install/npm_install.bzl
```diff
@@ -836,20 +834,20 @@ def _yarn_install_impl(repository_ctx):
...
+            # The --mutex option ensures only one yarn runs at a time, see
+            # https://yarnpkg.com/en/docs/cli#toc-concurrency-and-mutex
+            # The shared cache is not necessarily hermetic, but we need to cache downloaded
+            # artifacts somewhere, so we rely on yarn to be correct.
+            yarn_args.extend(["--mutex", "network"])
     else:
-        # Can't tell from documentation if Yarn Berry has any replacement for the --mutex
-        # flag. We'll have to assume it's safe to run concurrently.
-        pass
+        yarn_args.extend(["--cache-folder", str(repository_ctx.path("_yarn_cache"))])
```
sluongng, line 849: This unfortunately will add `--cache-folder` for folks who use Yarn Berry and set user_global_yarn_cache to False.

The flag DOES exist within Yarn Berry, but the behavior is to exit with a deprecation error which is not useful:

https://sourcegraph.com/github.com/yarnpkg/berry/-/blob/packages/plugin-essentials/sources/commands/install.ts?L207-219
https://sourcegraph.com/github.com/yarnpkg/berry@ce1e0c31781e3fc791cfa86a01348c4f0dba9aec/-/blob/packages/gatsby/content/advanced/error-codes.md?L295-301

I suggest ensure that refactor it like this:

```
 if yarn_version == "classic":
    if repository_ctx.attr.use_global_yarn_cache:
        ...
    else:
        ...

if yarn_version == "berry" and not repository_ctx.attr.use_global_yarn_cache:
    # Using Yarn Berry and non-default option for global_yarn_cache
    fail("Option global_yarn_cache has no effect while using Yarn Berry. Please configure it through a .yarnrc file through 'data' attribute.")
```

sluongng (changes requested):
There is a small logic flaw I spot. The rest looks good to me.

internal/npm_install/npm_install.bzl
alexeagle: yeah my intent here was the minimal change. thanks for looking up the yarn sources, if that flag just exits that way then I agree we shouldn't call it

sluongng:
Thanks. This revision looks good to me 👍

PR #3204, opened 2022-01-09
feat(typescript): ts_project usable without any npm install

alexeagle:
This avoids the problem of having to load ts_project from an npm-installed workspace like @npm
which causes an eager fetch of a bunch of packages just for the Bazel analysis phase to work
on a BUILD file that uses ts_project.


ts/private/tsc.js
```diff
@@ -0,0 +1,3 @@
+const runfiles = require(process.env['BAZEL_NODE_RUNFILES_HELPER'])
+const tsPath = runfiles.resolve('npm_typescript-4.3.5')
```
Toxicable, line 2: I might have missed something about this feature elsewhere, but where is the 4.3.5 version defined? it looks like @bazel/typescript has a peer dep of `>=3.0.0 <4.4.0`

alexeagle: it's in the new `ts/repositories.bzl`.
I fixed this dangling reference, now all the version info is contained within that file.

alexeagle:
This is now quite feasible since I refactored the ts helpers into the core package. This PR isn't the right approach anymore though.

PR #3174, opened 2021-12-23
docs: update for 5.0

alexeagle:

PR #3207, opened 2022-01-10
chore: remove vendored bazel-skylib used by ts_project

alexeagle:
Users now have the dependency on bazel-skylib so we can just load from there


kormide:
What's the historical context on this one?

alexeagle:
Prior to 5.x, `build_bazel_rules_nodejs` had no starlark module dependencies, so we couldn't rely on users having `@bazel_skylib` installed. https://github.com/bazelbuild/rules_nodejs/wiki/Migrating-to-5.0 goes into it a bit. Also https://github.com/bazelbuild/rules_nodejs/pull/3174

kormide (approved):

PR #3208, opened 2022-01-10
refactor: split ts_project.bzl

alexeagle:
The macro is now in the public api file, and the validate_options rule is in its own file.
Introduce a 'lib' struct for code sharing. This matches our more recently preferred code layout.


kormide (approved):
LGTM. Looks like nothing has changed in the public-facing api, right?

alexeagle:
right, this one is a pure refactoring. No changes in the generated docs, nor the integration tests or examples/e2e, so that demonstrates that users don't notice this one.

PR #3212, opened 2022-01-11
chore: set symlink_node_modules to False in examples/from_source

gregmagolan:
Likely to become the default in 5.0 in https://github.com/bazelbuild/rules_nodejs/pull/3214.

This PR is a subset of https://github.com/bazelbuild/rules_nodejs/pull/3214 and can be abandoned if that one goes green


gregmagolan:
Replaced by https://github.com/bazelbuild/rules_nodejs/pull/3214

PR #3213, opened 2022-01-11
refactor: remove package_json attribute from node_repositories

gregmagolan:
Also update all reference of `@nodejs` repository to its new name `@nodejs_host`


examples/user_managed_deps/WORKSPACE
```diff
@@ -28,19 +28,12 @@ load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories")
...
-#
-# To install the node_modules of all the listed package_json files run:
-#   bazel run @nodejs//:yarn_node_repositories
-# or
-#   bazel run @nodejs//:npm_node_repositories
-node_repositories(
-    package_json = ["//:package.json"],
-)
+#   @nodejs_host//:yarn
+node_repositories()
```
alexeagle, line 36: what's the point of this example now? I think this PR is dropping the "user managed deps" feature altogether?

alexeagle (approved):
yay

examples/user_managed_deps/WORKSPACE
gregmagolan: Not much point. Can deleted in a follow-up.

PR #3214, opened 2022-01-11
refactor: set symlink_node_modules default to False in yarn_install and npm_install

gregmagolan:
Adds section to docs to explain why using `symlink_node_modules` is not recommended.

### DIFF
--- a/examples/create-react-app/WORKSPACE
+++ b/examples/create-react-app/WORKSPACE
@@ -1,7 +1,4 @@
-workspace(
-    name = "create_react_app",
-    managed_directories = {"@npm": ["node_modules"]},
-)
+workspace(name = "create_react_app")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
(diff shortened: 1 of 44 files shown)


examples/create-react-app/WORKSPACE
```diff
@@ -20,6 +17,7 @@ load("@build_bazel_rules_nodejs//:index.bzl", "yarn_install")
 yarn_install(
     # Name this npm so that Bazel Label references look like @npm//package
     name = "npm",
+    data = ["//:patches/jest-haste-map+26.6.2.patch"],
```
alexeagle, line 20: oops, we should have updated jest so this patch isn't needed, I'll TAL

alexeagle (approved):

examples/create-react-app/WORKSPACE
gregmagolan: yea. will do in a follow-up across the repo

PR #3215, opened 2022-01-11
fix: change all cfg=host and cfg=target executable attributes to cfg=exec

gregmagolan:
These should have been "exec" all along.

### DIFF
--- a/examples/angular/src/tsconfig.json
+++ b/examples/angular/src/tsconfig.json
@@ -25,13 +25,16 @@
 
             // Add all possible bazel-out directories
             // See https://github.com/microsoft/TypeScript/issues/37257
+            "../bazel-out/darwin_arm64-fastbuild/bin/src",
             "../bazel-out/darwin-dbg/bin/src",
             "../bazel-out/darwin-fastbuild/bin/src",
-            "../bazel-out/darwin_arm64-fastbuild/bin/src",
+            "../bazel-out/darwin-opt-exec-2B5CBBC6/bin/src",
             "../bazel-out/k8-dbg/bin/src",
             "../bazel-out/k8-fastbuild/bin/src",
+            "../bazel-out/k8-opt-exec-2B5CBBC6/bin/src",
             "../bazel-out/x64_windows-dbg/bin/src",
-            "../bazel-out/x64_windows-fastbuild/bin/src"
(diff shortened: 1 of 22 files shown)


examples/angular/src/tsconfig.json
```diff
@@ -25,13 +25,16 @@
 
             // Add all possible bazel-out directories
             // See https://github.com/microsoft/TypeScript/issues/37257
+            "../bazel-out/darwin_arm64-fastbuild/bin/src",
             "../bazel-out/darwin-dbg/bin/src",
             "../bazel-out/darwin-fastbuild/bin/src",
-            "../bazel-out/darwin_arm64-fastbuild/bin/src",
+            "../bazel-out/darwin-opt-exec-2B5CBBC6/bin/src",
```
alexeagle, line 31: do you know what the `2B5CBBC6` is based on? I hope it's a stable checksum that will match on another machine.

packages/terser/index.js
```diff
@@ -26,6 +26,45 @@ function isDirectory(input) {
...
 }
 
+// Returns a single quotes version of str
+function singleQuotes(str) {
+  return `'${str.replace(/'/g, '').replace(/"/g, '')}'`;
+}
+
+// Ensures that args are well formed.
+// Work-around for an issue on Windows when exec bin path is not quoted.
+// In --source-map, base=bazel-out/x64_windows-opt-exec-2B5CBBC6/bin must
```
alexeagle, line 36: interesting we didn't have this problem before, we had hyphens in the path already right?

alexeagle (approved):

packages/terser/index.js
gregmagolan: Yeah. I was puzzled. The extra `-exec-2B5CBBC6` broke it when there were no quotes 🤷  Something fishy upstream

examples/angular/src/tsconfig.json
gregmagolan: John Cater says "the hex tag is just a hash of the execution platform" (https://github.com/bazelbuild/bazel/issues/12403#issuecomment-732952252) so it should be stable. Guessing it is a hash of "host" if the exec platform is the host since it is stable between k8, darwin & windows in these paths

PR #3219, opened 2022-01-11
feat: set exports_directories_only default to True in yarn_install & npm_install repository rules

gregmagolan:
BREAKING CHANGE

Also tagged on a 2nd commit as I forgot the bash case in the original

```
Author: Greg Magolan <gmagolan@gmail.com>
Date:   Tue Jan 11 15:40:09 2022 -0800

    feat: add src patch being copied to progress message of vendored copy_file
    
    With copy_file#is_directory being a very common code path with exports_directories_only,
    more detail in the progress messages makes for a much better DX
```


internal/npm_install/generate_build_file.ts
```diff
@@ -1101,26 +1101,6 @@ alias(
...
-
-  let mainEntryPoint = resolvePkgMainFile(pkg);
-
-  // add an `npm_umd_bundle` target to generate an UMD bundle if one does
-  // not exists
-  if (mainEntryPoint && !findFile(pkg, `${pkg._name}.umd.js`)) {
-    result +=
-        `load("@build_bazel_rules_nodejs//internal/npm_install:npm_umd_bundle.bzl", "npm_umd_bundle")
-
-npm_umd_bundle(
```
gregmagolan, line 1113: npm_umd_bundle does not work with directory artifacts so no point in stamping it out

alexeagle: would a user manually write one in their own BUILD file then?

gregmagolan: ts_library & ng_module are not compatible with exports_directories_only so the whole UMD concat js path is not as load bearing with the flag set.

If the user needed to make a UMD bundle out of an npm package with exports_directories_only set they could just use the `browserify-wrapped` npm tool directly with npm_package_bin.

The `npm_umd_bundle` rule has an explicit failure if any of the sources are directories.

```
    for s in sources:
        if s.is_directory:
            fail("npm_umd_bundle does not work with directory artifacts")
```

in `internal/npm_install/npm_umd_bundle.bzl`

alexeagle (approved):

PR #3220, opened 2022-01-11
refactor: remove user_managed_deps examples

gregmagolan:
We don't recommend that user's manually configure js_library targets
that point to their local node_modules. This is an exceptional use case
and relies on undocumented internal details of yarn_install and npm_install.


alexeagle (approved):
bye code we'll miss you 🐱

PR #3225, opened 2022-01-12
chore: remove check_bazel_version

### LINKED ISSUE #2622: Remove check_rules_nodejs_version helper
This causes us to carry around version parsing code, and we haven't used it in years - it existed largely to avoid version skew with rules_typescript which we slurped into our monorepo.
Also we are very careful with breaking change rollout so we haven't ever bumped a version requirement using this utility.

alexeagle:
Also check_rules_nodejs_version, though a stub is left behind to avoid needing a patch in rules_sass which calls it. We can use the one in bazel_skylib instead.

Fixes #2622


index.bzl
```diff
@@ -102,7 +100,6 @@ def check_build_bazel_rules_nodejs_version(minimum_version_string):
             minimum_version_string,
         ))
 
-# Export check_rules_nodejs_version symbol for pre-5.0 backward compat. This
-# will eventually get removed as we move to the core @rules_nodejs package in
-# future release.
-check_rules_nodejs_version = check_build_bazel_rules_nodejs_version
+def check_rules_nodejs_version():
+    # buildifier: disable=print
+    print("check_rules_nodejs_version has been removed. This is a no-op, please remove the call.")
```
gregmagolan, line 105: nit: prefix with "WARNING: ..."?

gregmagolan (approved):
👋  cya check_bazel_version

PR #3226, opened 2022-01-12
Changed calculation of outs in ts_project to support using rule outputs as srcs

twheys:
Hi 👋 

Fixes #2504

The goal of this PR is to make ts_project support using output files from rules in srcs without breaking the pre-calculated outputs in the macro. I understand the outputs are calculated in the macro so that they can be set as outputs on the rule which exposes the output files via a Bazel label like `//path/to/package:out-file-name`.

My changes introduce a secondary calculation of out files in the implementation of the ts_project rule so that the resolved list of files from the context `ctx.files.srcs` can be used. This makes it possible to use targets in the srcs attribute that point to other rules instead of being locked into using `glob` and source files in the same package.

Example:
```
filegroup(
  name = "ts_files",
  srcs = glob(["**/*.ts"]),
)

ts_project(
  name = "js_files",
  srcs = ["ts_files"],
  ...
)
```

Please let me know if you have any feedback or would like any changes made.

PR #3227, opened 2022-01-12
Added `absolute_root_dir` attr to ts_project to support using root_dir relative to workspace root

twheys:
Hi 👋 

This is a sort of extension to #3226 

I'd like to be able to define my rules in a package that is not in one of the parent directories of the src files. This is due to some legacy code in a big project that is not feasible to refactor before introducing Bazel as the build tool.

For some context on the goal of this feature, image a project where:

- The root of the ts project is in package `a/b/c`, the tsconfig.json file is also in this dir, `a/b/c`
- The build output should go in the directory `a/d/`

So I write my ts_project target in `a/d/BUILD` and set the attr `root_dir= "a/b/c"`

--- 

I am definitely open to suggestions on alternative ways to implement this. My primary goal is to be able to set the root_dir so that it is not relative to the Bazel package. I introduced a new attribute to do this because I didn't want to impact how others might be presently using `root_dir`.


alexeagle (commented):
Thanks for explaining the motivation of this change. I imagine you could use a `copy_file` mechanism to copy build outputs from `bazel-bin/a/b/c` to `bazel-bin/a/d` but this is a small enough change that I think it's supportable this way.

Could you add some kind of test case for it? You'll see under /packages/typescript/test/ts_project we already have a bunch of "executable examples".

As for the API, maybe `root_dir` starting with a `/` can indicate that it's workspace-relative rather than package-relative? We'd have to reason about what that would do today, I guess it would just be ignored by our path joining logic so if someone has a leading slash it's a breaking change for them.

If so that's fine, we are very close to shipping 5.0 so you could make your PR on the 5.x branch instead and be breaking.

twheys:
Alright, I admit it's kind of a weird use case. I added a test case in `packages/typescript/test/ts_project/rootdir/sibling/BUILD.bazel`
The goal is to build the sources under `packages/typescript/test/ts_project/rootdir/subdir` in the output dir `packages/typescript/test/ts_project/rootdir/sibling/` 

I got rid of the bool attribute and took your suggestion of using a leading forward slash. I also made sure it worked with the bazel format of a double leading forward slash.

twheys:
The branching is screwed up at the moment. I can fix it after #3228 is merged. I tried to rebase from that branch but it pulled in a bunch of other stuff.

alexeagle:
Your other big one landed, ready to rebase this on `stable` and have another go?

jbedard:
With the `@bazel/typescript` ruleset removed in v6 this PR is no longer relevant. For typescript-related rules we suggest a ruleset such as [aspect_rules_ts](https://github.com/aspect-build/rules_ts) going forward.

PR #3228, opened 2022-01-12
Changed calculation of outs in ts_project to support using rule outputs as srcs

### LINKED ISSUE #2504: Cannot use filegroup for srcs in ts_project

twheys:
Hi 👋 

Fixes #2504

The goal of this PR is to make ts_project support using output files from rules in srcs without breaking the pre-calculated outputs in the macro. I understand the outputs are calculated in the macro so that they can be set as outputs on the rule which exposes the output files via a Bazel label like `//path/to/package:out-file-name`.

My changes introduce a secondary calculation of out files in the implementation of the ts_project rule so that the resolved list of files from the context `ctx.files.srcs` can be used. This makes it possible to use targets in the srcs attribute that point to other rules instead of being locked into using `glob` and source files in the same package.

Example:
```
filegroup(
  name = "ts_files",
  srcs = glob(["**/*.ts"]),
)

ts_project(
  name = "js_files",
  srcs = ["ts_files"],
  ...
)
```

Please let me know if you have any feedback or would like any changes made.


packages/typescript/test/ts_project/filegroup/a.ts
```diff
@@ -0,0 +1,5 @@
+export const a: string = 'hello';
+
+// Repro #2044
```
alexeagle, line 3: I think you can simplify this test case a bit, no need to duplicate coverage we already have somewhere else. just a one-line TS file would work here

packages/typescript/test/ts_project/filegroup/tsconfig-extended.json
```diff
@@ -0,0 +1,4 @@
+// This file is an extra hop of indirection, regression test for #1754
```
alexeagle, line 1: again, can simplify

packages/typescript/test/ts_project/filegroup/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+        //    cd C:/b/uuxnwop3/execroot/build_bazel_rules_nodejs
+        //    SET COMPILATION_MODE=fastbuild
+        //    bazel-out/host/bin/external/npm/typescript/bin/tsc.bat -p packages/typescript/test/ts_project/b/tsconfig.json --outDir bazel-out/x64_windows-fastbuild/bin/packages/typescript/test/ts_project/b --declarationDir bazel-out/x64_windows-fastbuild/bin/packages/typescript/test/ts_project/b --bazel_node_modules_manifest=bazel-out/x64_windows-fastbuild/bin/packages/typescript/test/ts_project/b/_tsconfig.module_mappings.json
+        //  Execution platform: @local_config_platform//:host
+        //  packages/typescript/test/ts_project/b/b.ts(1,17): error TS6305: Output file 'C:/b/uuxnwop3/execroot/build_bazel_rules_nodejs/packages/typescript/test/ts_project/a/a.d.ts' has not been built from source file 'C:/b/uuxnwop3/execroot/build_bazel_rules_nodejs/packages/typescript/test/ts_project/a/a.ts'.
+        "outDir": "../../../../../bazel-out/x64_windows-fastbuild/bin/packages/typescript/test/ts_project/a",
```
alexeagle, line 18: shouldn't be needed since we don't depend on the `ts_project/a` package?

packages/typescript/internal/ts_project.bzl
```diff
@@ -358,6 +428,7 @@ def ts_project_macro(
         out_dir = None,
         root_dir = None,
         link_workspace_root = False,
+        js_outs = [],
```
alexeagle, line 431: why is js_outs special here? can we avoid that?
if the user has more js_outs, they might also have more map_outs?

packages/typescript/internal/ts_project.bzl
```diff
@@ -333,6 +361,48 @@ def _out_paths(srcs, outdir, rootdir, allow_js, ext_map):
...
+
+    exts = {
+        "*": ".js.map",
+        ".tsx": ".jsx.map",
+    } if preserve_jsx else {"*": ".js.map"}
+    return _out_paths(srcs, out_dir, root_dir, False, exts)
+
+def _calculate_typings_outs(srcs, typings_out_dir, root_dir, declaration, composite, allow_js):
+    typings_out = []
+    #    typings_out = [
```
alexeagle, line 387: comment could be removed?

packages/typescript/internal/ts_project.bzl
```diff
@@ -98,6 +122,19 @@ def _calculate_root_dir(ctx):
     )
 
 def _ts_project_impl(ctx):
+    srcs = [src.path for src in ctx.files.srcs]
+    typings_out_dir = ctx.attr.declaration_dir or ctx.attr.out_dir
+
+    # Recalculate outputs inside the rule implementation.
+    # The outs are first calculated in the macro in order to try to predetermine outputs so they can be declared as
+    # outputs on the rule. This provides the benefit of being able to reference an output file with a label.
+    # However, it is not possible to evaluate files in outputs other rules such as filegroup, therefore the outs a
```
alexeagle, line 131: outputs other rules -> outputs of other rules
outs a -> outs are

packages/typescript/internal/ts_project.bzl
```diff
@@ -98,6 +122,19 @@ def _calculate_root_dir(ctx):
...
+    # recalculated here.
+    js_outs = _declare_outputs(ctx, _calculate_js_outs(srcs, ctx.attr.out_dir, ctx.attr.root_dir, ctx.attr.allow_js, ctx.attr.preserve_jsx, ctx.attr.emit_declaration_only))
```
alexeagle, line 133: I'm trying to convince myself that the outputs we declare here will always include all the outputs which were predeclared. obviously they share the _calculate implementations. And I think Bazel will error if a predeclared output isn't created by any actions in the implementation. This seems simple :)

packages/typescript/internal/ts_project.bzl
```diff
@@ -64,6 +75,19 @@ def _join(*elements):
         return "/".join(segments)
     return "."
 
+def _relative_to_package(path, ctx):
```
alexeagle, line 78: nit: on the 5.x branch we can just `load("@bazel_skylib//lib:paths.bzl"` to get the relativize helper, but maybe not enough code sharing that it's worth it

alexeagle (commented):
Neat! generally looks great, thank you.

this PR targets 4.x which is pretty old. Should target `stable` or `5.x` instead

packages/typescript/internal/ts_project.bzl
twheys: The "predeclared_outpus" will only find src that end with ".ts" or ".tsx", if an input is a label pointing to a target, then they won't be predeclared. 
The predeclared outputs are provided in `ctx.outputs`, but Bazel is equally happen if you use `ctx.actions.declare_file` for the same file. It figures out it's a predeclared output file and returns the same file object.

packages/typescript/internal/ts_project.bzl
twheys: Will fix the comment

packages/typescript/internal/ts_project.bzl
twheys: Yeah, I am still working on getting all of the tests to pass. This is remnants from that.

packages/typescript/internal/ts_project.bzl
twheys: Hm, I am pretty sure is was one test build that passes in js_outs as an attribute. I think that's why I added it here. I don't have a strong opinion on this, although it could come in handy that if a use wants to predeclare some outputs that might be in a filegroup, that they can specify them here. I suppose the other outs could be derived from this. Let me know if you want me to proceed a particular way with this.

packages/typescript/test/ts_project/filegroup/a.ts
twheys: Alright, I'll add a target to another package instead.

packages/typescript/internal/ts_project.bzl
alexeagle: yeah that matches my understanding also, SGTM

packages/typescript/internal/ts_project.bzl
alexeagle: let's not introduce a special case unless we know the use case for it, I'd rather keep the attribute list the same for now

alexeagle:
can you rebase on 5.x and target that branch instead? 4.x is old (really it's still pointing to 4.0.0) and I'm about to cut 5.0.0-rc.0

twheys:
Yeah - I can get this by early next week. No need to wait on this feature if you want to move forward with 5.x, though. I have a patched version that I am using which solves my needs so I don't need it urgently.

twheys:
Sorry for the delay, got pulled away from this. Will rebase for 5.x

twheys:
@alexeagle alright I rebased with 5.x which required a bit of code changes, but I can run `bazel build //packages/typescript/test/...` successfully locally now.

Could you please take a fresh look, since the commit has changed a bit since the previous iteration?

rayman1104 (commented):
I needed the same thing while migrating from `ts_library`, so thanks, I already use your solution!
The only thing I needed to change: if we pass dict to `tsconfig` field and use label in srcs, the generated tsconfig.json fails to generate files list correctly. To fix that I've just changed `packages/typescript/index.bzl:405` line to `files = srcs`. But maybe there is a prettier solution.

packages/typescript/internal/ts_project.bzl
alexeagle: ?

packages/typescript/internal/ts_project.bzl
```diff
@@ -274,21 +325,58 @@ def _replace_ext(f, ext_map):
...
-            out = _join(outdir, f[:f.rindex(".")].replace(rootdir_replace_pattern, "") + _replace_ext(f, ext_map))
+            out = _join(out_dir, f[:f.rindex(".")].replace(rootdir_replace_pattern, "") + _replace_ext(f, ext_map))
 
             # Don't declare outputs that collide with inputs
             # for example, a.js -> a.js
             if out != f:
                 outs.append(out)
     return outs
 
+def _calculate_js_outs(srcs, out_dir, root_dir, allow_js, preserve_jsx, emit_declaration_only):
```
alexeagle, line 341: nit: I think it's a bit more descriptive to name these functions "predict_*_outs" rather than calculate...

alexeagle (commented):
great, close to merging.
thanks for dealing with the branching, you had unlucky timing working on this just as we were closing in on 5.0

packages/typescript/test/ts_project/filegroup/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
...
+    extends = "//packages/typescript/test/ts_project:tsconfig-base.json",
+    tsconfig = "tsconfig.json",
+    visibility = ["//packages/typescript/test:__subpackages__"],
+    # Use @babel/parser since the package.json is required to resolve "typings" field
+    # Repro of #2044
+    deps = [
+        "@npm//@babel/parser",
+        "@npm//@babel/types",
+    ],
+)
```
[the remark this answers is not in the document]
twheys: Since c.js comes from a source in the filegroup and is therefore not predeclared as an output, so it can't be asserted by using a filegroup like in other test packages. If the ts_project rule builds, then that is a pretty good assertion, but if we want to check that `c.js` is output we would need to write a rule implementation that checks the files resolved by a sources attr. Is there an existing utility that does this?

packages/typescript/internal/ts_project.bzl
```diff
@@ -4,7 +4,7 @@ load("@rules_nodejs//nodejs:providers.bzl", "DeclarationInfo", "declaration_info
 load("@build_bazel_rules_nodejs//:providers.bzl", "ExternalNpmPackageInfo", "run_node")
 load("@build_bazel_rules_nodejs//internal/linker:link_node_modules.bzl", "module_mappings_aspect")
 load(":ts_config.bzl", "TsConfigInfo")
-load(":validate_options.bzl", "ValidOptionsInfo", _validate_lib = "lib")
+load(":validate_options.bzl", "ValidOptionsInfo", _validate_lib = "lib", _VALIDATED_ATTRS = "VALIDATED_ATTRS")
```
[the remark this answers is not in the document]
twheys: Ok, I'll move it into the lib struct

packages/typescript/internal/ts_project.bzl
```diff
@@ -91,6 +103,30 @@ def _calculate_root_dir(ctx):
     )
 
 def _ts_project_impl(ctx):
+    srcs = [_relative_to_package(src.path, ctx) for src in ctx.files.srcs]
+
+    transpile = ctx.attr.transpile
```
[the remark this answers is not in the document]
twheys: Alright, that's how i had it at first but found it a bit verbose, but i wasn't super keen on having a bunch of local variables either. Will change it back.

packages/typescript/internal/ts_project.bzl
twheys: This function is called twice, once in the macro to determine if outputs can be predeclared, and again inside the rule impl to determine the full list of output files. In the first case, I would agree that `predict` accurately describes it, but in the rule implementation it yields a full list of the outputs that will be generated by running node, so I would argue it's not making a prediction anymore at that point. I don't have a strong opinion on the naming here though, so if you prefer it to use the verb predict I'm okay with that.

packages/typescript/internal/validate_options.bzl
```diff
@@ -52,24 +52,27 @@ def _validate_options_impl(ctx):
         ValidOptionsInfo(marker = marker),
     ]
 
+VALIDATED_ATTRS = {
```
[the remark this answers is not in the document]
twheys: Yeah, I'll rename this to SHARED_ATTRS and write a comment

packages/typescript/test/ts_project/filegroup/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
...
+    srcs = glob(["*.ts"]),
+)
+
+ts_project(
+    srcs = [":srcs"],
+    composite = True,
+    extends = "//packages/typescript/test/ts_project:tsconfig-base.json",
+    tsconfig = "tsconfig.json",
+    visibility = ["//packages/typescript/test:__subpackages__"],
+    # Use @babel/parser since the package.json is required to resolve "typings" field
```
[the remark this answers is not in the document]
twheys: It gets composite = True from the base tsconfig.json so it's simpler to keep that setting here, unless you'd prefer I write a new tsconfig.json for this package? Let me know. I'll remove the babel parser from the deps.

twheys:
@rayman1104 if there are ts(x) or json files coming in from other targets needed by the tsconfig rule, then yeah we'll probably have to just pass all of srcs to that rule. I'll make that change.

twheys:
Changing `files = srcs` in the `write_tsconfig` attr lead to a test failure since files besides json and ts files were being passed to the tsconfig rule. I fixed this by moving the same filtering logic that was previously in `index.bzl` into `ts_config.bzl` inside the rule impl where the full list of files is available. Since this required some functions from `ts_library.bzl` which already imports from `ts_config.bzl`, I moved a lot of the "lib" utility functions into a separate bzl module to avoid a cyclic import.

twheys:
@alexeagle I'm not sure how to fix this failing test (https://app.circleci.com/pipelines/github/bazelbuild/rules_nodejs/5516/workflows/4ed463a5-f399-4aca-908a-3f616f3a9cfd/jobs/64160)
Can you please provide some guidance on how to run it locally? I'm not finding any of the input files it's complaining about in the repo.

packages/typescript/internal/ts_project.bzl
twheys: Fixed now

alexeagle:
The repro command is the title of that circle ci step. It's the examples/from_src failing

alexeagle:
the bug is that something in this implementation works within the same workspace where it was declared (build_bazel_rules_nodejs) but fails when referenced from a build running in a different workspace (examples/from_source is a nested workspace)

the common cause is some pathing computation which references a package name but fails to account for its pair, the workspace name.

looking through the code here I think this is the likely culprit

```
def _relative_to_package(path, ctx):
    for prefix in (ctx.bin_dir.path, ctx.label.package):
        prefix += "/"
        if path.startswith(prefix):
            path = path[len(prefix):]
    return path
```
and I don't see where the corresponding logic is in the base commit.

alexeagle:
I'd still like to add test coverage for this, and cleanup the small amount of code duplication between ts_utils.bzl and other code in this repo and in bazel-skylib. But it's an exciting feature and we're shipping a minor today so I'm comfortable landing this as-is.

Thanks @twheys for your hard work on this one, looking forward to more collaborations :)

alexeagle (approved):

PR #3230, opened 2022-01-12
refactor: move yarn fetch to a separate external repo

alexeagle:
Previously every @node_some_platform had a yarn program in it,
which doesn't make sense since yarn is platform-independent JS.
Instead make a @yarn repo and use that for getting our yarn binary.

Trivially allows a vendored yarn by passing the yarn attribute to
yarn_install overriding the default.

BREAKING CHANGE:
vendored_yarn attribute is removed


internal/node/node_repositories.bzl
```diff
@@ -41,13 +42,15 @@ def node_repositories(**kwargs):
...
 
-    # buildifier: disable=print
-    print("""WARN: node_repositories is deprecated, please instead use:
-        load("@rules_nodejs//nodejs:repositories.bzl", "nodejs_register_toolchains")
-        nodejs_register_toolchains(name = "nodejs")
-
-        See https://github.com/bazelbuild/rules_nodejs/wiki/Migrating-to-5.0
-    """)
+    # Back-compat: allow yarn_repositories args to be provided to node_repositories
+    yarn_args = {}
```
alexeagle, line 46: based on working with the repo recently, I think it's too soon to deprecate the build_bazel_rules_nodejs repositories helper. Too spammy to log here, and too much work for users, with no benefit in the vast majority case where they don't care about toolchains

nodejs/yarn_repositories.bzl
```diff
@@ -0,0 +1,217 @@
...
+            default = "nodejs",
+            doc = """The basename for a nodejs toolchain to use for running yarn.
+            Usually this is the value of the `name` attribute given to a nodejs_register_toolchains call in WORKSPACE""",
+        ),
+        "yarn_download_auth": attr.string_dict(
+            default = {},
+            doc = """auth to use for all url requests
+    Example: {\"type\": \"basic\", \"login\": \"<UserName>\", \"password\": \"<Password>\" }
+    """,
+        ),
```
mattem, line 190: Was this here before? Seems kinda insecure

mattem (approved):
Haven't gone though super thoroughly, but looks the right shape. The `npx` mentions is kinda nice, might be worth adding a line to doc of how to call it.

Is there somewhere in the docs that really explicitly calls out the labels? eg `@yarn//:yarn` etc? I always forget those and all the many aliases.

nodejs/yarn_repositories.bzl
alexeagle: yes, this block was moved without changes from the original to avoid making it more breaking than necessary

docs/repositories.md
```diff
@@ -23,7 +23,12 @@ Some ways you can use this:
 
 - Run the Bazel-managed version of node: `bazel run @nodejs_host//:node path/to/program.js`
 - Run the Bazel-managed version of npm: `bazel run @nodejs_host//:npm`
-- Run the Bazel-managed version of yarn: `bazel run @nodejs_host//:yarn`
+
+## @yarn
+
+This repository is created by calling the `yarn_repositories` function in your `WORKSPACE` file.
+
+- Run the Bazel-managed version of yarn: `bazel run @yarn//:yarn`
```
alexeagle, line 31: @mattem here's what you were looking for

mattem: Ah nice 👍

PR #3233, opened 2022-01-13
feat: default package_path to the directory of the package.json file in yarn_install and npm_install

### LINKED ISSUE #2451: Enable multi-linker by default
This is a breaking change and needs to be sequenced with the 4.0 release.

To turn it on by default, the `package_path` should default to the directory of the `package.json` file if it is unset (`""`).
It can be overridden if necessary. To override to root the value of `"/"` could be used.

gregmagolan:
Resolves #2451

BREAKING CHANGE


alexeagle (approved):
☔ 🐫

PR #3240, opened 2022-01-13
fix(cypress): use depsets for runfiles and data

mrmeku:
## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

### DIFF
--- a/packages/cypress/internal/cypress_web_test.bzl
+++ b/packages/cypress/internal/cypress_web_test.bzl
@@ -94,26 +94,25 @@ def _cypress_web_test_impl(ctx):
         cypressinfo.cypress_bin_path,
     ]
 
-    runfiles = (
+    runfiles = depset(
         [plugin_wrapper] +
         ctx.files.config_file +
         ctx.files.cypress_npm_package +
         ctx.files.plugin_file +
         ctx.files.srcs +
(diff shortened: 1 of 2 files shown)


packages/cypress/internal/cypress_web_test.bzl
```diff
@@ -94,26 +94,25 @@ def _cypress_web_test_impl(ctx):
...
         ctx.attr.plugin_file,
-        ctx.attr.srcs,
-    ]
+    ] + ctx.attr.srcs)
 
     return nodejs_test_kwargs["implementation"](
         ctx,
-        data = data,
-        runfiles = runfiles,
+        data = data.to_list(),
```
djmarcin, line 114: For me, at least, this caused an internal error because of [this line](https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/node/node.bzl#L160) which needed to change as well, to the following to ensure that duplicate files in ctx.attr.data and this data list are not duplicated when passed to `ctx.expand_location`.
```
    data = depset(ctx.attr.data + data).to_list()
```

mrmeku: Done

alexeagle (approved):

PR #3241, opened 2022-01-14
fix: create a bazel-out node_modules tree using copy_file in the external repo when exports_directories_only is True

gregmagolan:
This will allow entry points to resolve peer node_module deps within an external yarn_install & npm_install repository

### DIFF
--- a/internal/npm_install/generate_build_file.ts
+++ b/internal/npm_install/generate_build_file.ts
@@ -144,7 +144,7 @@ export async function main() {
   // find all packages (including packages in nested node_modules)
   const pkgs: Dep[] = [];
 
-  await findPackagesAndPush(pkgs, 'node_modules', deps);
+  await findPackagesAndPush(pkgs, nodeModulesFolder(), deps);
 
   // Sort the files to ensure the order
   pkgs.sort(compareDep);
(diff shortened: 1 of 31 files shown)


internal/npm_install/generate_build_file.ts
```diff
@@ -51,6 +51,8 @@ const PUBLIC_VISIBILITY = '//visibility:public';
 
 let LEGACY_NODE_MODULES_PACKAGE_NAME = '$node_modules$';
 
+const WORKSPACE_REROOTED_PATH = '_';
```
alexeagle, line 54: strangely descriptive name for this constant that's one character, and unused?

internal/npm_install/generate_build_file.ts
```diff
@@ -206,7 +214,7 @@ function flattenDependencies(pkgs: Dep[]) {
 async function generateRootBuildFile(pkgs: Dep[]) {
   let buildFile = config.exports_directories_only ?
       printRootExportsDirectories(pkgs) :
-      printRoot(pkgs);
+      printRootLegacy(pkgs);
```
alexeagle, line 217: maybe "ExportsAllFiles" is more descriptive than "Legacy"?

alexeagle (approved):
neat

internal/npm_install/generate_build_file.ts
gregmagolan: was used at one point, does that count?

internal/npm_install/generate_build_file.ts
gregmagolan: either or

PR #3244, opened 2022-01-14
chore: fix release process so bbrnj gets right version of rnj

alexeagle:


mattem (approved):

mattem:
Failing on the worker package being out of sync?

alexeagle:
yeah I'm investigating that too

PR #3246, opened 2022-01-14
chore: renovate stop sending PRs for examples

alexeagle:
We get way too many PRs for rules_sass and rules_docker version bumps that we don't really want


gregmagolan (approved):
👍

PR #3250, opened 2022-01-15
*_install: only create nodejs or yarn repositories if we need them

alexeagle:
Previously we'd create these repositories even if the user already installed node some other way.

Note, this is minor extra breaking change after rc1, but based on testing at one client I think this may be a good idea. This PR demonstrates how it caught some bugs in our e2e/examples where the wrong node (default version) was getting used.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -63,6 +63,13 @@ nodejs_register_toolchains(
     node_version = "15.14.0",
 )
 
+load("@rules_nodejs//nodejs:yarn_repositories.bzl", "yarn_repositories")
+
+yarn_repositories(
+    name = "yarn",
+    node_repository = "node16",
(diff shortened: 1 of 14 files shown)


jbedard:
What was the reason for it? Although deleting is always good :+1:

alexeagle:
It was just to make workspace files shorter in the common case.

jbedard:
What made this good a good idea at that client though? Mainly just curious...

alexeagle:
sorry, being too terse... I meant that at one client I'm seeing the wrong node downloaded so I think it may be a good idea to carry out this PR and remove this extra bit of syntax sugar to make things easier to understand.

gregmagolan (approved):

PR #3251, opened 2022-01-15
refactor: remove unneeded file from core package

alexeagle:
It's confusing to have the bbrnj providers file in this module, since you're meant to use /nodejs/providers.bzl instead


gregmagolan (approved):
🌯

PR #3252, opened 2022-01-15
chore: remove vendored copy of skylib paths.bzl

alexeagle:
This can come from upstream since we didn't have local modifications


gregmagolan (approved):
🌮

PR #3253, opened 2022-01-15
refactor: remove vendored write_file.bzl

alexeagle:
We didn't have local modifications so we can just depend directly on upstream


gregmagolan (approved):
🌮

PR #3254, opened 2022-01-15
fix: warning logic for yarn berry attrs

alexeagle:
The default for frozen_lockfile is true so the warning should be if you try to set it false.
This API is really confusing right now though. Maybe we should have separate rules like yarn_install and yarn1_install or sth


gregmagolan (approved):
🌮

PR #3255, opened 2022-01-15
fix(yarn): when running vendored yarn, prefix command with path to node

alexeagle:
Otherwise we fall through to the system node on the PATH which may not be set, or may be skewed

rebased on #3250

### DIFF
--- a/examples/BUILD.bazel
+++ b/examples/BUILD.bazel
@@ -216,10 +216,7 @@ example_integration_test(
     npm_packages = {
         "//packages/jasmine:npm_package": "@bazel/jasmine",
     },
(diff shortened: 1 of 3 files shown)


gregmagolan:
I think we also need to put the node on the PATH before calling yarn since yarn scripts & programs called by yarn scripts can invoke "node"

alexeagle:
That seems like an existing bug actually. Nothing in our yarn repository adds such a thing to our yarn wrapper scripts, and when we invoke yarn install we don't do it there either.

gregmagolan (approved):
🚀

PR #3258, opened 2022-01-17
chore: don't constrain yarn releases we mirror

### LINKED ISSUE #3071: yarn_versions helper only includes v1

alexeagle:
fixes #3071

### DIFF
--- a/examples/jest/WORKSPACE
+++ b/examples/jest/WORKSPACE
@@ -26,10 +26,18 @@ load("@build_bazel_rules_nodejs//:repositories.bzl", "build_bazel_rules_nodejs_d
 
 build_bazel_rules_nodejs_dependencies()
 
+load("@rules_nodejs//nodejs:repositories.bzl", "nodejs_register_toolchains")
+
+nodejs_register_toolchains(
+    name = "nodejs",
+    node_version = "16.2.0",
+)
+
 load("@build_bazel_rules_nodejs//:index.bzl", "yarn_install")
 
 yarn_install(
     name = "npm",
+    exports_directories_only = True,
     package_json = "//:package.json",
     quiet = False,
     yarn_lock = "//:yarn.lock",
--- a/examples/jest/package.json
+++ b/examples/jest/package.json
@@ -15,7 +15,9 @@
         "jest-junit": "13.0.0",
         "typescript": "~4.2.0"
     },
+    
     "scripts": {
+        "postinstall": "node -e 'require(\"assert\").equal(process.version, \"v16.2.0\")'",
         "test": "bazel test ..."
     }
 }
--- a/nodejs/private/yarn_versions.bzl
+++ b/nodejs/private/yarn_versions.bzl
@@ -1,9 +1,6 @@
 """
 Generated code; do not edit
 Update by running yarn update-yarn-versions
-
-Note that we don't support Yarn 2 yet, see
-https://github.com/bazelbuild/rules_nodejs/issues/1599
 """
 
 # @unsorted-dict-items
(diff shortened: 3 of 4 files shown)


alexeagle:
This doesn't actually work.
Yarn 2 berry seems to put the release artifacts somewhere else, https://registry.npmjs.org/yarn doesn't have them. Will need someone with more resources to contribute a fix here.
I'd also note that I think pnpm is the most promising package manager to use with Bazel, and so I'm not planning to keep working on this myself.

Issue #3259, opened 2022-01-17
tags are not propagated from ts_project to validation target


djmarcin:
# 🐞 bug report

### Affected Rule

ts_project macro

### Is this a regression?

No

### Description

`tags` such as `manual` are not propagated to the generated `validate_options` rule. This makes it impossible to disable this target from running under wildcard expansion, or to disable things like caching on this target.

They need to be added to [this invocation](https://github.com/bazelbuild/rules_nodejs/blob/9bc681d14ba197752e4770d95bfc524825d6082f/packages/typescript/internal/ts_project.bzl#L742-L757)

## 🔬 Minimal Reproduction

Set `tags = ['manual']` on a ts_project invocation, `bazel build //path/to/target/...`, observe that the validate target is built despite the ts_project being marked manual.

## 🔥 Exception or Error

n/a

## 🌍  Your Environment

any

**Rules_nodejs version:**

<pre>
  <code>
4.6.1
  </code>
</pre>

**Anything else relevant?**

mattem:
Closed via #3260

PR #3260, opened 2022-01-17
fix(typescript): propagate tags to validate_options

djmarcin:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #3259 

## What is the new behavior?
Tags are propagated to the validate_options rule

- [ ] Yes
- [x] No

Strictly, the answer here might be yes, particularly if someone was doing something very esoteric with custom tags and relying on the fact that propagation to this rule was not happening, but this seems extremely unlikely.

### DIFF
--- a/packages/typescript/internal/ts_project.bzl
+++ b/packages/typescript/internal/ts_project.bzl
@@ -754,6 +754,7 @@ def ts_project_macro(
                 allow_js = allow_js,
                 tsconfig = tsconfig,
                 extends = extends,
+                tags = kwargs.get("tags", []),
             )
             tsc_deps = tsc_deps + ["_validate_%s_options" % name]
 
--- a/packages/typescript/internal/ts_project.bzl
+++ b/packages/typescript/internal/ts_project.bzl
@@ -700,6 +700,12 @@ def ts_project_macro(
         fail("As of rules_nodejs 3.0, extends should have a single value, not a list.\n" +
              "Use a ts_config rule to group together a chain of extended tsconfigs.")
 
+    common_kwargs = {
+        "tags": kwargs.get("tags", []),
+        "visibility": kwargs.get("visibility", None),
+        "testonly": kwargs.get("testonly", None),
+    }
+
     if type(tsconfig) == type(dict()):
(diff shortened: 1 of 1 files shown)


alexeagle (approved):
Thanks!

Do you think we might want `testonly` and `visibility` to be propagated as well? Just thinking to be consistent with https://github.com/bazelbuild/rules_nodejs/blob/5.x/packages/typescript/index.bzl#L507-L511

djmarcin:
> Thanks!
> 
> Do you think we might want `testonly` and `visibility` to be propagated as well? Just thinking to be consistent with https://github.com/bazelbuild/rules_nodejs/blob/5.x/packages/typescript/index.bzl#L507-L511

Seems reasonable. The target is named with a leading underscore (`_validate...`) so it seems like it's not intended to be depended on by other things, making testonly and visibility likely extraneous but it probably doesn't hurt.

djmarcin:
@alexeagle I moved `common_kwargs` up and reused it for `validate_options`. It seems like this may also be good to pass to [write_tsconfig](https://github.com/bazelbuild/rules_nodejs/blob/9bc681d14ba197752e4770d95bfc524825d6082f/packages/typescript/internal/ts_project.bzl#L725-L731) and [nodejs_binary](https://github.com/bazelbuild/rules_nodejs/blob/9bc681d14ba197752e4770d95bfc524825d6082f/packages/typescript/internal/ts_project.bzl#L762-L787), but I did not do that here as it expands the scope of the PR a bit. The `write_tsconfig` macro needs to be further modified to take & pass along **kwargs to the underlying rule. Let me know if I should add that here.

alexeagle (approved):

alexeagle:
buildifier failed - I'd suggest running `pre-commit install` in this repo so a git pre-commit hook will always format your starlark.

djmarcin:
> buildifier failed - I'd suggest running `pre-commit install` in this repo so a git pre-commit hook will always format your starlark.

fixed & force pushed the last commit (not sure what the convention for this repo is in terms of the structured commit messages and trivial typo fixes like that...)

PR #3261, opened 2022-01-18
fix(builtin): npm_package_bin include runfiles in DefaultInfo

### LINKED ISSUE #3257: Generated webpack rule doesn't build when used as a dependency
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule: 
@npm//webpack

### Is this a regression?

 Not sure, this is my first attempt with Bazel + JS

### Description

 A clear and concise description of the problem...
Please take a look at the minimal repro below. I have two rules. The first uses `@npm//webpack` to bundle JS files, and the second is a `sh_binary` rule that uses the output of the first as data attribute. Running `bazel query 'deps(//:upload)'` correctly shows that `//:src/index.js` is a dependency (along with a lot of node modules files).

When I `bazel build //:upload`, the build completes without error, but webpack is not invoked. Checking bazel-out and bazel-bin confirms that there is no webpack output.

## 🔬 Minimal Reproduction
https://github.com/geoffliu/bazel-issue

## 🔥 Exception or Error
None

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
OSX 11.6
  </code>
</pre>

<pre>
  <code>
Build label: 4.2.2-homebrew
Build time: Tue Jan 1 00:00:00 1980 (315532800)
Build timestamp: 315532800
Build timestamp as int: 315532800
  </code>
</pre>

**Rules_nodejs version:**

<pre>
  <code>
4.6.0
  </code>
</pre>

**Anything else relevant?**

alexeagle:
Otherwise you can't use a data dep on one of these to gather runtime deps

fixes #3257

### DIFF
--- a/internal/node/npm_package_bin.bzl
+++ b/internal/node/npm_package_bin.bzl
@@ -87,8 +87,11 @@ def _impl(ctx):
         exit_code_out = ctx.outputs.exit_code_out,
         link_workspace_root = ctx.attr.link_workspace_root,
     )
-
-    return [DefaultInfo(files = depset(outputs + tool_outputs))]
+    files = outputs + tool_outputs
+    return [DefaultInfo(
+        files = depset(files),
+        runfiles = ctx.runfiles(files = files),
+    )]
 
 _npm_package_bin = rule(
     _impl,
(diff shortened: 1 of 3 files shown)


gregmagolan (approved):
🌮

PR #3274, opened 2022-01-21
refactor: bbrnj re-exports rules_nodejs providers

alexeagle:
This makes the 5.0 release less breaking by restoring back-compat with
rulesets that loaded these from the old location.

I manually verified that the same symbol comes through, so even though
a provider creation and usage site load from different places, the
provided symbol still matches and works.


gregmagolan (approved):
🌮

PR #3283, opened 2022-01-25
fix(jasmine): replace deprecated Jasmine APIs that have been removed  in version 4

alan-agius4:
With this change we replace APIs that have been deprecated in version 4 and removed in version 4

```
Jasmine#onComplete is deprecated. Instead of calling onComplete, set the Jasmine instance's exitOnCompletion property to false and use the promise returned from the execute method.
```

Also this addresses the breaking change in version 4 https://github.com/jasmine/jasmine/blob/main/release_notes/4.0.0.md#changes-that-affect-custom-reporters in a backward compatible manner.

### DIFF
--- a/e2e/BUILD.bazel
+++ b/e2e/BUILD.bazel
@@ -51,15 +51,29 @@ e2e_integration_test(
     tags = ["no-bazelci-windows"],
 )
 
-e2e_integration_test(
-    name = "e2e_jasmine",
+[e2e_integration_test(
+    name = "e2e_jasmine_%s" % jasmine_version.replace(".", "_"),
     npm_packages = {
         "//packages/jasmine:npm_package": "@bazel/jasmine",
+    },
+    # use these package.json packages instead
+    package_json_substitutions = {
+        "jasmine": jasmine_version,
+        "jasmine-core": jasmine_version,
     },
     # TODO: figure out why this fails on Windows since setting
     # symlink_node_modules to False in the test WORKSPACE
     tags = ["no-bazelci-windows"],
-)
+    workspace_root = "jasmine",
+) for jasmine_version in [
+    # TODO(6.0): remove old API tests
+    # Old API
+    "2.99.x",
+    "3.9.x",
+    # New API
+    "4.0.x",
+    "3.10.x",
+]]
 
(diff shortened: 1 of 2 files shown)


alan-agius4:
Looks like with this change we can only support Jasmine 3.10+ https://jasmine.github.io/api/npm/3.10/Jasmine

e2e/BUILD.bazel
```diff
@@ -51,15 +51,28 @@ e2e_integration_test(
...
+    package_json_substitutions = {
+        "jasmine": jasmine_version,
+        "jasmine-core": jasmine_version,
+    },
     # TODO: figure out why this fails on Windows since setting
     # symlink_node_modules to False in the test WORKSPACE
     tags = ["no-bazelci-windows"],
-)
+    workspace_root = "jasmine",
+) for jasmine_version in [
```
alexeagle, line 68: nice, thanks for considering the back-compat case.
Too bad we didn't land this in time for 5.0, we could have just dropped support for older jasmine in a breaking change.

Do you think we should leave a `TODO(6.0)` here to remove old support for old Jasmine?

packages/jasmine/jasmine_runner.js
```diff
@@ -154,22 +144,66 @@ async function main(args) {
     }
   }
 
-  await jrunner.execute();
+  // Jasmine versions prior to 3.10.0 should use the old API.
+  if (/^3\.[1-9]\.|^2\./.test(jrunner.coreVersion())) {
+    console.warn(`DEPRECATED: Support for Jasmine versions prior to '3.10.x' is deprecated in '@bazel/jasmine'.`);
```
alexeagle, line 149: looks like yes, we should have the `TODO(6.0)` to drop it

alexeagle (approved):
Thanks Alan! Looks great as always, just one comment nit

e2e/BUILD.bazel
alan-agius4: Sure

packages/jasmine/jasmine_runner.js
alan-agius4: Sure just give me a couple of mins

PR #3287, opened 2022-01-26
fix(builtin): pass kwargs from node_repositories helper

alexeagle:
We were missing test coverage for attributes like node_version being passed to the new toolchain registration


gregmagolan (approved):
🌮

Issue #3280, opened 2022-01-24
jasmine_node_test and exports_directories_only = True is broken


thesayyn:
# 🐞 bug report

### Affected Rule

jasmine_node_test

### Is this a regression?

Yes

### Description

jasmine_node_test is broken due to referencing a file underneath the jasmine package which is not visible because it is neither predeclared output nor exported. 

## 🔥 Exception or Error

<pre><code>
no such target '@npm//:node_modules/@bazel/jasmine/jasmine_runner.js': target 'node_modules/@bazel/jasmine/jasmine_runner.js' not declared in package '' defined by /private/var/tmp/_bazel_thesayyn/b975b285f400016eec4905e7e4968b7f/external/npm/BUILD.bazel and referenced by '//test:test'

</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
     Darwin M1
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>

  </code>
</pre>

**Rules_nodejs version:**

(Please check that you have matching versions between WORKSPACE file and `@bazel/*` npm packages.)

<pre>
  <code>
   package.json: "@bazel/jasmine": "^5.0.0", 
   WORKSPACE: rnj: 5.0.0
  </code>
</pre>

**Anything else relevant?**

alexeagle:
Thanks for taking this on!

In your fix PR you ought to be able to turn on exports_directories_only for some/all of the examples/e2e targets which depend on jasmine, e.g. https://github.com/bazelbuild/rules_nodejs/blob/stable/examples/BUILD.bazel#L130

alexeagle:
fixed in 5.0.2 by #3293

PR #3292, opened 2022-01-26
docs: demonstrate workaround for jasmine with export_directories_only

alexeagle:
Workaround for https://github.com/bazelbuild/rules_nodejs/issues/3280


alexeagle:
#3293 has the principled fix

Issue #3277, opened 2022-01-24
ESM support


Toxicable:
# 🚀 feature request
### Relevant Rules
- `ts_project`
- `nodejs_binary`
- `jasmine_node_test`
Maybe others, but these are the ones I'm familiar with
### Description
- Add support for executing JS that is running in ESM format (nodejs_binary, jasmine_node_test)
- Add support for compiling to ESM (ts_project) (if needed)

### Describe alternatives you've considered
I don't believe there's alternate ways to run ESM JS without core support in rules_nodejs

-----

The primary motivation for this feature is for support of Angular v13 which only publishes ESM JS.
The general rule with ESM is that an ESM file can import CJS or ESM, and CJS can require CJS and import ESM only via the `import()` function. Therefore you cannot `require()` ESM from a CJS file.
This means that a codebase which runs a `jasmine_node_test` as it stands is unable to import Angular v13 (or any other ESM only packages)

(please correct me if this is wrong)
My understanding of how NodeJS picks what module format to run a file in is determined by:
First it will look at file extention: .cjs -> CJS, .mjs -> ESM
If the extension is .js then it looks at the package.json field "type", where module -> ESM, commonjs -> CJS
See "ESM_FILE_FORMAT" in the NodeJS docs for a detailed algorithm

With the above in mind I think the changes required to support ESM with nodejs_binary would be to rename internal JS files (node_patches.js, jasmine_runner.js) to be *.cjs, with the exception of loader.js which will need to be modified to be able to `import()` the users entry_point

For `ts_library` I think the most unambiguous method would be to have it output `.mjs` files, however that's not currently supported and there's a lot of discussion, see some of the threads around here https://github.com/microsoft/TypeScript/issues/18442
Therefore I think the next best option is just to output .js files then the user will need to ensure they have a package.json with type=module in order to declare the output to be ESM, this will need some docs.
That would mean no changes are needed to the rule itself.

NodeJS docs: https://nodejs.org/api/esm.html

Toxicable:
I order to support `jasmine_node_test` the spec files need to be `.mjs` in order for the jasmine loader to be able to import it correctly. Ref https://github.com/jasmine/jasmine-npm/blob/main/lib/loader.js#L13
This could be done by adding a `module` attribute to the ts_project then calling a simple rule to rename outputs. The attribute would be required when typescript adds support for `.mjs` outputs anyway since we'd need to pre-declare the `.mjs` outputs.
[...]

PR #3294, opened 2022-01-27
feat(builtin): add basic ESM support

Toxicable:
## PR Checklist

## PR Type

Issue Number: #3277

## What is the new behavior?
Adds ESM support in for `nodejs_binary`, `jasmine_node_test` and `ts_project`

- [x] No - Only changes are internal API's


packages/typescript/index.bzl
```diff
@@ -63,6 +63,7 @@ def ts_project(
         out_dir = None,
         root_dir = None,
         link_workspace_root = False,
+        module_type = "commonjs",
```
alexeagle, line 66: could you separate the TS api changes from the rest of the PR? This part worries me since it adds complexity and bazel-only behaviors, in other ways this rule "just runs `tsc`". But I think the other parts could land.

Toxicable: Yeah, really not sure what the solution is here, happy to remove and at least land the other pieces first

Toxicable:
Things to note:
- requires typescript > 4.5
- requires jasmine > 3.7
- Editor support requires a manual file association for *.mts -> typescript
- Imports are required to end with .mjs and there's not currently editor support (ticket likely exists needs to be found or created)
- Typescript reference https://www.typescriptlang.org/docs/handbook/esm-node.html

Toxicable:
I've extracted ts_project support into a seperate MR https://github.com/bazelbuild/rules_nodejs/pull/3298

Toxicable:
CI looks mostly green now
Aside from this failure, which looks like it's unrelated and only on windows
>ERROR: C:/b/bk-windows-zzbr/bazel/rules-nodejs-nodejs/internal/node/test/chdir/BUILD.bazel:20:16: output 'internal/node/test/chdir/package.json' was not created

.npmrc
```diff
@@ -1,2 +1,3 @@
 tag-version-prefix=""
 message="chore(release): %s"
+registry=https://registry.npmjs.org
```
devversion, line 3: Is this intentionally part of this PR?

packages/jasmine/jasmine_runner.js
```diff
@@ -52,7 +52,7 @@ if (TOTAL_SHARDS) {
 // the maximum (See: https://nodejs.org/api/errors.html#errors_error_stacktracelimit)
 Error.stackTraceLimit = Infinity;
 
-const IS_TEST_FILE = /[^a-zA-Z0-9](spec|test)\.js$/i;
+const IS_TEST_FILE = /[^a-zA-Z0-9](spec|test)\.(mjs|js)$/i;
```
devversion, line 55:
```suggestion
const IS_TEST_FILE = /[^a-zA-Z0-9](spec|test)\.(mjs|cjs|js)$/i;
```

devversion (approved):
This looks great to me.

.npmrc
Toxicable: Ah, this change isn't related to this PR, but I needed it so that yarn wouldn't try pull packages from another registry that I have configured.
I don't think anyone else would have this issue, but it wouldn't cause an issue leaving it in as well.
Happy to remove if you think so

devversion: No problem with me, just asked since it seemed unrelated 😄

alexeagle (approved):

PR #3298, opened 2022-01-27
ts_project ESM support

Toxicable:
## PR Checklist

## PR Type

Issue Number: #3277

## What is the new behavior?
Adds ESM support in for  `ts_project`

- [x] No - Only changes are internal API's


WORKSPACE
```diff
@@ -55,7 +55,7 @@ load("@rules_nodejs//nodejs:repositories.bzl", "nodejs_register_toolchains")
 # This applies to the resolved_toolchain
 nodejs_register_toolchains(
     name = "node16",
-    node_version = "16.5.0",
```
Toxicable, line 58: Note that this change was required to for @angular/core - they require nodejs >=16.10.0

diceride:
@Toxicable Thank you working on this, I'm also strugling with the same problem.

Your implementation introduces on a new .mts file extension. The fact that Node.js relies on a .mjs file extension to detect whether to use JavaScript module syntax is a unique feature of Node.js only. In my opinion it's a bad design choice by the Node.js team. By introducing a new .mts file extension, Node.js behaviour would spill over into the TypeScript world. I'm not sure whether this is the right approach to resolve the issue.

diceride:
@Toxicable Thank you working on this, I'm also strugling with the same problem.

Your implementation introduces on a new .mts file extension. The fact that Node.js relies on a .mjs file extension to detect whether to use JavaScript module syntax is a unique feature of Node.js only. In my opinion it's a bad design choice by the Node.js team. By introducing a new .mts file extension, Node.js behaviour would spill over into the TypeScript world. I'm not sure whether this is the right approach to resolve the issue. File extensions are ignored by web browsers and by Deno too.

Toxicable:
>Your implementation introduces on a new .mts file extension

@diceride Unfortunately the ship has sailed and this is the deisgn that has been chosen by both NodeJS and the TypeScript Teams. I'm not keen to introduce any new designs here and would prefer to stick to what has already decided on.
The goal of this PR is to support the same set of file extensions that are already in TypeScript as outlined here https://github.com/bazelbuild/rules_nodejs/pull/3298 
In saying that, there's still the option to use the `package.json#type=module` method if that helps

alexeagle:
CI failure
```
error TS18003: No inputs were found in config file 'C:/b/makknmlb/execroot/build_bazel_rules_nodejs/packages/typescript/test/ts_project/esm/tsconfig.json'. Specified 'include' paths were '["**/*"]' and 'exclude' paths were '[]'.
```
is probably because of Windows lack of sandboxing by default - do you need help to make it green?

Toxicable:
@alexeagle yeah that would be awesome if you got the time

alexeagle:
Note: https://github.com/bazelbuild/rules_nodejs/pull/3405 seems like the prerequisite but needs someone with some time to green it up...

PR #3301, opened 2022-01-31
fix(builtin): make linker aspect run in constant time per target

fmeum:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #3302 

## What is the new behavior?

With this commit, the `LinkerPackageMappingInfo` provider is restructured to hold depsets of mapping entries and node_modules root paths. In this way the aspect invocations for the individual targets never traverse the
(partially constructed) mappings.
Instead, `write_node_modules_manifest` flattens the depset only after the aspect has run. If this should ever become a performance bottleneck again, the flattening could be moved into the execution phase by letting a dedicated binary consume the depset wrapped in an Args object.

In an experiment with a single medium-sized package.json and a Node binary depending on all targets in the corresponding npm repository, this commit reduces the analysis wall time spent in the aspect implementation from ~30s to ~70ms.

- [x] Yes
- [ ] No

Dependencies providing their own `LinkerPackageMappingInfo` (an advanced use case introduced by https://github.com/bazelbuild/rules_nodejs/issues/2941) have to adapt to the structure of the provider. However, if that kind of breaking change is considered intolerable, it would be possible to add a translation layer between the old and new format.

### DIFF
--- a/internal/linker/link_node_modules.bzl
+++ b/internal/linker/link_node_modules.bzl
@@ -21,7 +21,17 @@ def _debug(vars, *args):
 LinkerPackageMappingInfo = provider(
     doc = """Provider capturing package mappings for the linker to consume.""",
     fields = {
-        "mappings": "Dictionary of mappings. Maps package names to an exec path.",
+        "mappings": """Depset of structs with mapping info.
+
+Each struct has the following fields:
+  package_name: The name of the package.
+  package_path: The root path of the node_modules under which this package should be linked.
+  link_path: The exec path under which the package is available.
+
+Note: The depset may contain multiple entries per (package_name, package_path) pair.
+Consumers should handle these duplicated appropriately. The depset uses topological order to ensure
+that a target's mappings come before possibly conflicting mappings of its dependencies.""",
+        "node_modules_roots": "Depset of node_module roots.",
     },
 )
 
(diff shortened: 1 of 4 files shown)


alexeagle:
@devversion do you have any input regarding the subtle breaking change here, and whether you think it merits the extra effort and temporary complexity to make it non-breaking?

devversion (commented):
@alexeagle From my perspective, given the scenario of manual linker mappings being advanced anyway (as stated in the PR as well), there is nothing to worry about in terms of making it non-breaking (for Angular the API change should be trivial at first glance). 

This PR looks amazing, can't wait to see it published 🎉 Thank you!

internal/linker/link_node_modules.bzl
```diff
@@ -21,7 +21,17 @@ def _debug(vars, *args):
...
     fields = {
-        "mappings": "Dictionary of mappings. Maps package names to an exec path.",
+        "mappings": """Depset of structs with mapping info.
+
+Each struct has the following fields:
+  package_name: The name of the package.
+  package_path: The root path of the node_modules under which this package should be linked.
+  link_path: The exec path under which the package is available.
+
+Note: The depset may contain multiple entries per (package_name, package_path) pair.
```
alexeagle, line 31: thanks for the docs :)

alexeagle (commented):
I think this is fine to land and be in a minor release, we should just mention in the release notes in case someone other than Angular was using the non-public API.

I'll let @gregmagolan do the approve since he's most familiar with this code.

gregmagolan (approved):
Amazing! Thank you for the major optimization. 🌮

PR #3322, opened 2022-02-10
fix(builtin): avoid unnecessary chdir to prevent worker threads from failing

kormide:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

When the `chdir` attribute is used in `nodejs_test`, the script that changes the directory will run in each worker thread. Node fails when `process.chdir` is called in a worker thread. This is because node is invoked using `--require` on the chdir script: https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/node/node.bzl#L286

## What is the new behavior?

Only change directories once in the main node process. Spawned workers won't try to change the directory again.

- [ ] Yes
- [x] No

## Other information

This is needed for an architect_test rule in angular/aio to work. That project is currently on rules_nodejs 4.6.0, so I suppose we'll want to backport this and cherry-pick it into stable?

### DIFF
--- a/internal/node/node.bzl
+++ b/internal/node/node.bzl
@@ -279,7 +279,15 @@ fi
             fail("""nodejs_binary/nodejs_test only support chdir inside the current package
                     but %s is not a subfolder of %s""" % (ctx.attr.chdir, ctx.label.package))
         chdir_script = ctx.actions.declare_file(_join(relative_dir, "__chdir.js__"))
-        ctx.actions.write(chdir_script, "process.chdir(__dirname)")
+        ctx.actions.write(chdir_script, """
+/* This script is preloaded with --require, meaning it will run for the main node process
+   as well as each worker thread that gets spawned. Calling process.chdir() in a worker
+   is an error in node, so ensure it's only called once for the main process.
+*/
+if (process.cwd() !== __dirname) {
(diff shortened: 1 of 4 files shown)


mattem:
Can this target stable and we cherry-pick back to 4.x?

alexeagle:
Sure I can cut a 4.x bugfix release with the fix if you need one.

internal/node/node.bzl
```diff
@@ -283,7 +283,11 @@ fi
             fail("""nodejs_binary/nodejs_test only support chdir inside the current package
                     but %s is not a subfolder of %s""" % (ctx.attr.chdir, ctx.label.package))
         chdir_script = ctx.actions.declare_file(_join(relative_dir, "__chdir.js__"))
-        ctx.actions.write(chdir_script, "process.chdir(__dirname)")
+        ctx.actions.write(chdir_script, """
```
alexeagle, line 286: Add a comment so we understand why you bother to check, or add some regression test.

Otherwise someone will read this code later and think the conditional is unneeded because it should be a no-op to chdir to the directory which is already the current working dir.

kormide:
Updated with comment, test, and rebased onto `stable`.

PR #3326, opened 2022-02-13
{npm|yarn}_install: Add attribute for verbose output

davido:
This simplifies tracking down installation problems.

This option currently has only effect in *nices and activates -x bash
option. But conceptually, it can also be implemented on Windows using
SET ON directive.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No


mattem:
Do we need an extra attribute? Repo rules generally hook into `quiet`, which is defined here: https://github.com/bazelbuild/rules_nodejs/blob/40b58f1c0584bae2b80fb8e05f99449bbc29ead5/internal/npm_install/npm_install.bzl#L262-L265

https://docs.bazel.build/versions/main/skylark/lib/repository_ctx.html#execute

What do you think about hooking the behaviour off that?

alexeagle (commented):
we already have logic that keys on whether `VERBOSE_LOGS` is in the environment - could we just use that? It's our rules_nodejs equivalent of a missing "verbose actions" feature in Bazel

PR #3331, opened 2022-02-16
fix(runfiles): use normalized paths when guarding runfiles root and node_modules on Windows

mistic:
## PR Checklist

Please check if your PR fulfils the following requirements:

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

## What is the current behaviour?

After https://github.com/bazelbuild/rules_nodejs/pull/3060 being merged in order to fix a linker problem the paths we collecte to guard the `runfiles` and the `runfiles node_modules` at https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/node/launcher.sh#L237 are crashing the node resolution when running on Windows as it is using non normalized paths.

## What is the new behaviour?
That PR makes sure we normalize the collected paths on Windows so that we can later build a correct `BAZEL_PATCH_ROOTS`.

- [ ] Yes
- [x] No

## Other information

Closes https://github.com/bazelbuild/rules_nodejs/issues/3054 and fixes the build node_modules resolutions when on Windows.


alexeagle:
Windows CI Failure is already present on stable, not introduced here...

internal/node/launcher.sh
```diff
@@ -49,7 +77,9 @@ source "${RUNFILES_DIR:-/dev/null}/$f" 2>/dev/null || \
 # Case 6a is handled like case 3.
 if [[ -n "${RUNFILES_MANIFEST_ONLY:-}" ]]; then
   # Windows only has a manifest file instead of symlinks.
-  RUNFILES=${RUNFILES_MANIFEST_FILE%/MANIFEST}
+  # Normalizing the path and case insensitive removing the `/MANIFEST` part of the path
+  NORMALIZED_RUNFILES_MANIFEST_FILE_PATH=$(normalizeWindowsPath $RUNFILES_MANIFEST_FILE)
+  RUNFILES=$(sed 's|\/MANIFEST$||i' <<< $NORMALIZED_RUNFILES_MANIFEST_FILE_PATH)
```
alexeagle, line 82: can we avoid depending on `sed`? do we know where that comes from on a windows machine? (does every bazel user have it?)

mistic: Good point. I know `sed` is not POSIX compliant but in that use case I really think its okay to use it. In order to get all of this to work looks like we rely on `Git bash` to be installed so we can run the bash scripts inside of it when running the generated `bat` executable. As this is the case the GNU `sed` will be installed and we can be sure it will work as expected.

mistic: @gregmagolan are you okay with this change?

gregmagolan: hmmm. can we limit the code path changes to Windows here to limit the potential blast radius?
```
RUNFILES=${RUNFILES_MANIFEST_FILE%/MANIFEST}
```
changes to 
```
  NORMALIZED_RUNFILES_MANIFEST_FILE_PATH=$(normalizeWindowsPath $RUNFILES_MANIFEST_FILE)
  RUNFILES=$(sed 's|\/MANIFEST$||i' <<< $NORMALIZED_RUNFILES_MANIFEST_FILE_PATH)
```
in linux and osx... and `normalizeWindowsPath` being called on all platforms is a bit confusing as you have to read that code to know it is a no-op on non-windows platforms.

Something like
```
if windows
    RUNFILES=$(normalizeWindowsRunfilesPath)
else
    RUNFILES=${RUNFILES_MANIFEST_FILE%/MANIFEST}
fi
```
would be clearer to read and would change the code path on Windows only.

As for using `sed` on Windows, your reasoning sounds ok given that we're already relying on git bash to run this bash script.

gregmagolan: It might also good to make this an opt-in via a config flag if you think there is risk here; tho from the looks of it it seems pretty safe

mistic: Thanks for the review @gregmagolan ! I've applied the refactors you've suggested

gregmagolan (approved):

internal/node/launcher.sh
gregmagolan: Looks good. Thanks for updating

gregmagolan:
> Windows CI Failure is already present on stable, not introduced here...

These failures @alexeagle 
```


(22:29:51) ERROR: C:/b/bk-windows-p9mw/bazel/rules-nodejs-nodejs/internal/node/test/chdir/BUILD.bazel:24:16: output 'internal/node/test/chdir/package.json' was not created
--
  | (22:29:51) ERROR: C:/b/bk-windows-p9mw/bazel/rules-nodejs-nodejs/internal/node/test/chdir/BUILD.bazel:24:16: Action internal/node/test/chdir/package.json failed: not all outputs were created or valid


```

mistic:
@alexeagle could we get this one in? 😃

alexeagle:
oops sorry it missed today's release!! we finally greened up HEAD again...

mistic:
@alexeagle do you think we can also backport this into `4.x`?

alexeagle:
sure can you send a PR against the 4.x branch? I'll merge and release a patch

mistic:
I've opened the PR 👍

Issue #2992, opened 2021-09-30
Cannot set `node_version = "14.18.0"` for manual NodeJS install with `4.3.0` release.


matthewt-codaio:
This is with the 4.3.0 release on macOS.

We have in our `WORKSPACE` this rule,

```

node_repositories(
    node_repositories = {
        "14.18.0-darwin_amd64": ("node-v14.18.0-darwin-x64.tar.gz", "node-v14.18.0-darwin-x64", "6b9b4d60bcb4eba95488380be8c4da4af98fce3f4a01c9a76db881cbb736656d"),
        "14.18.0-linux_amd64": ("node-v14.18.0-linux-x64.tar.gz", "node-v14.18.0-linux-x64", "f411b8aee36d6dc6a5435906f42bd4ea59d6f678894cf562beaf115b58a318ee"),
        "14.18.0-windows_amd64": ("node-v14.18.0-win-x64.zip", "node-v14.18.0-win-x64", "2883e83ac3b1e1cb9a9bf65554043640849b39e86761e7c7ac50b664f42f20ff"),
    },
    node_urls = ["https://nodejs.org/dist/v{version}/{filename}"],
    node_version = "14.18.0",
    package_json = ["//:package.json"],
    yarn_repositories = {
        "1.22.10": ("yarn-v1.22.10.tar.gz", "yarn-v1.22.10", "7e433d4a77e2c79e6a7ae4866782608a8e8bcad3ec6783580577c59538381a6e"),
    },
    yarn_urls = ["https://github.com/yarnpkg/yarn/releases/download/v{version}/{filename}"],
    yarn_version = "1.22.10",
)
```

When the bazel build is run it falls back to `node_version` `14.17.5` which is the default. In the 4.3.0 release `14.18.0` is not in the built-ins, though I thought these attributes are sufficient for a manual install and the node version should not matter.

Same seems to happen with `yarn_version`, i.e. we can only install 'manually' yarn builds whose versions are in the built-in lists.

Is this user error?

alexeagle:
Yeah, looks like I broke this in https://github.com/bazelbuild/rules_nodejs/commit/7f85e9f21388f5449daa543b67638c76c56d69d1#diff-f381b656a6df80113b279e147e93a3fc60037f2d5aef1affbbc376476ea5bef0
it assumes our local NODE_VERSIONS mapping is sufficient to lookup what versions exist. We skip downloads of anything we don't know about.

[...]

PR #3339, opened 2022-02-25
fix(builtin): take custom node_repositories value into account when checking if Node version exists for the platform

mattem:
When setting `node_repositories` attribute on the `node_repositories` rule (well that's kinda confusing), ensure this is taken into account when checking if the requested node version exists for the platform.

This can be useful when trying to trick rnj into downloading a different version of Node for a given OS (eg, a newer version of Node on darwin arm64), or when setting this attr to a version of Node that isn't included in the default list of `NODE_VERSIONS` yet.

### DIFF
--- a/nodejs/private/os_name.bzl
+++ b/nodejs/private/os_name.bzl
@@ -72,12 +72,17 @@ def is_linux_os(rctx):
     name = os_name(rctx)
     return name == OS_NAMES[3] or name == OS_NAMES[4] or name == OS_NAMES[5] or name == OS_NAMES[6]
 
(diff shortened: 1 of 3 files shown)


alexeagle (approved):
cool thanks

matthewt-codaio:
Awesome. Thank you. Will this make it to a 5.x.x update? I see it is in 4.6.3 now.

mattem:
Oh wow I didn't realize we already had an open issue for this. Yes this resolves #2992. Yeah this will make a 5.x release (we always just cut from stable)

PR #3346, opened 2022-02-28
docs: add data attribute to ts_project macro

alexeagle:
This doesn't change behavior; previously it was included in the **kwargs passed to the underlying ts_project rule.
However it makes the documentation less confusing


mattem:
CI failure looks legit for docs update

mattem (approved):

PR #3356, opened 2022-03-07
test(builtin): test showing duplicate module load when using workspace root

joeljeske:
This PR is a test case showing  a bug in the linker when using link_workspace_root.  

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

When using `link_workspace_root = True`, any file that is loaded using the root import link will see a difference node_modules. This means that the same node_modules are available but at a different path, thus asking node to load the library twice. This is a bug not just for performance, but that many libraries require the same module being used and not multiple copies due to state management, etc... 

## What is the new behavior?

This bug should be fixed. I am not sure of the best strategy as I am not very familiar with the linker internals 

- [ ] Yes
- [X] No

### DIFF
--- a/internal/linker/test/double_workspace_link/test.js
+++ b/internal/linker/test/double_workspace_link/test.js
@@ -1,0 +1,13 @@
+const npm = require("npm");
+
+describe("linker", () => {
+  it("should load the same copy of a npm dep when link_workspace_root is True and loading from workspace root path", () => {
+    const foo = require("build_bazel_rules_nodejs/internal/linker/test/double_workspace_link/foo");
+    expect(foo.foo_npm === npm).toBe(true, "Expected foo_npm to be npm");
+  });
+
+  it("should load the same copy of a npm dep when link_workspace_root is True and loading from relative path", () => {
+    const bar = require("./bar");
+    expect(bar.bar_npm === npm).toBe(true, "Expected bar_npm to be npm");
+  });
+});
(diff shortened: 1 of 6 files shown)


internal/linker/test/double_workspace_link/test.js
```diff
@@ -0,0 +1,13 @@
...
+
+describe("linker", () => {
+  it("should load the same copy of a npm dep when link_workspace_root is True and loading from workspace root path", () => {
+    const foo = require("build_bazel_rules_nodejs/internal/linker/test/double_workspace_link/foo");
+    expect(foo.foo_npm === npm).toBe(true, "Expected foo_npm to be npm");
+  });
+
+  it("should load the same copy of a npm dep when link_workspace_root is True and loading from relative path", () => {
+    const bar = require("./bar");
+    expect(bar.bar_npm === npm).toBe(true, "Expected bar_npm to be npm");
```
joeljeske, line 11: This test case passes, as `foo` should pass after the linker is fixed

internal/linker/test/double_workspace_link/test.js
```diff
@@ -0,0 +1,13 @@
+const npm = require("npm");
```
joeljeske, line 1: The module itself does not really matter, its any module that is in `node_modules`. I picked `npm` at random from the set of deps in package.json in this repo

alexeagle:
Looks same or adjacent to bugs @gregmagolan is planning to track down soon

PR #3360, opened 2022-03-10
fix: sharded test never fail when using Jasmine < 4

alan-agius4:
See: https://github.com/angular/dev-infra/pull/463

### DIFF
--- a/packages/jasmine/jasmine_runner.js
+++ b/packages/jasmine/jasmine_runner.js
@@ -146,9 +146,9 @@ async function main(args) {
 
   // TODO(6.0): remove support for deprecated versions of Jasmine that use the old API &
   // remember to update the `peerDependencies` as well.
-  // Jasmine versions prior to 3.10.0 should use the old API.
-  if (/^3\.[1-9]\.|^2\./.test(jrunner.coreVersion())) {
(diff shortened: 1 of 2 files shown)


packages/jasmine/jasmine_runner.js
```diff
@@ -146,9 +146,9 @@ async function main(args) {
 
   // TODO(6.0): remove support for deprecated versions of Jasmine that use the old API &
   // remember to update the `peerDependencies` as well.
-  // Jasmine versions prior to 3.10.0 should use the old API.
-  if (/^3\.[1-9]\.|^2\./.test(jrunner.coreVersion())) {
-    console.warn(`DEPRECATED: Support for Jasmine versions prior to '3.10.x' is deprecated in '@bazel/jasmine'.`);
+  // Jasmine versions prior to 4.0.0 should use the old API.
+  if (jrunner.coreVersion().charAt(0) !== '4') {
```
alexeagle, line 150: Not: what happens when they release 5.0?

alexeagle: Can we use a peerDep as an additional warning at install time?

alexeagle (approved):

PR #3382, opened 2022-04-01
fix: use python3 instead of python in unittest.bash

gregmagolan:
python 2 is dead, https://www.idownloadblog.com/2022/01/28/python-macos-monterey-no-more/


alexeagle (approved):
curious, was it fixed in the upstream copy?
now that build_bazel_rules_nodejs depends on bazel_skylib it would be nice to remove this vendored copy...

gregmagolan:
> curious, was it fixed in the upstream copy? now that build_bazel_rules_nodejs depends on bazel_skylib it would be nice to remove this vendored copy...

It wasn't yet, but maybe this is a PR they would accept

alexeagle:
it is fixed upstream already
```
local PYTHON=python
      command -v python3 &> /dev/null && PYTHON=python3
      LC_ALL=C "${PYTHON}" -c 'import time; print(int(round(time.time() * 1000)))'
```

Issue #3193, opened 2022-01-06
Usage of `ts_*` causes a load-time resolution of `@npm` for non-FE target analysis


vpanta:
# 🚀 feature request

### Relevant Rules

`ts_config`
And possibly anything else loaded out of `@bazel/typescript`

### Description

 Using a rule such as `ts_config` shouldn't require a load of the `@npm` external repository, which is a heavy load for all packages and cannot be done piecemeal (AFAIK).

Because of this, any `BUILD.bazel` file loading `ts_*` requires a `@npm` load-time resolution even if the target in question does not require `@npm`. For example, we have a base `tsconfig.json` and have generated `ts_config` rules for all of them, however that root `BUILD.bazel` also exports a config file which is used by some Go binaries, and as such we have Go builds which require loading `@npm` as a whole.

### Describe the solution you'd like

 Specifically for `ts_config` (or any other rule which doesn't require `@npm`), it seems inappropriate to load it via an `npm` package. Meanwhile, though I think it would be doable for other rules to be loaded straight from `rules_nodejs` as well, `ts_library`/`ts_project` might not make a big impact as I believe their hardcoded dep on `tsc` would keep the load-time resolution of `@npm`?

### Describe alternatives you've considered

 For now we're adjusting our auto-generation as we shouldn't actually need the base `tsconfig.json` to be available via Bazel (but rather, it exists more for IDE integration). However, it still felt like this might deserve some discussion and thus thought I should file this issue.

alexeagle:
Yup, `load` statements in any BUILD file read during the analysis phase do cause eager fetches, and so that's a reason to organize code carefully avoiding a mix of `load` statements in a single package.

For typescript in particular, this is a conscious design choice. The `ts_*` rules depend on the typescript npm package, so we distribute them as an npm package. That allows .js files in the `@bazel/typescript` package to `require('typescript')` and resolve to the correct version of typescript that you have installed, and ensures that Bazel uses the same TypeScript version as your other tools like the editor.

You don't have to fetch the `@bazel/typescript` package in the same `npm_install`/`yarn_install` as all the other packages though. I think it's a good practice to have a separate `package.json` for the tools that are run during the build, as opposed to the runtime dependencies. Then you use a separate `npm_install(name = "npm_tools")` or something, and then your load statements cause an eager fetch of a smaller set of packages.

[...]

PR #3386, opened 2022-04-04
refactor: extract typescript support files to core

alexeagle:
Allows for them to be reused in 3rdparty rules that want to compile TS
without our @bazel/typescript package.

Part of #3193


kormide (approved):

docs/TypeScript.md
```diff
@@ -555,7 +555,7 @@ An example can be found in the rules_nodejs repo under `packages/typescript/test
 &gt; To apply the fix, you can use the patch-package package to apply this patch:
 &gt; https://gist.github.com/alexeagle/ba44b2601bd7c953d29c6e8ec44d1ef9
 
-Defaults to `None`
+Defaults to `Label("@npm//typescript/bin:tsc")`
```
jbedard, line 558: Was this true before but the docs were wrong?

`_DEFAULT_TSC` had the `@npm` stripped though?

alexeagle: previously the default was in the rule implementation, but we run stardoc against the wrapper macro, so it wasn't seen. Now the default is in the macro so it appears here.

The published package uses the `#BEGIN/END-INTERNAL` fencing to strip the `@npm` prefix, but stardoc runs against the sources so that stripping isn't visible to it.

Either way I think the documentation isn't very useful about what the default is.

nodejs/private/ts_lib.bzl
```diff
@@ -0,0 +1,231 @@
...
+    ),
+    "typing_maps_outs": attr.output_list(
+        doc = "Locations in bazel-out where tsc will write `.d.ts.map` files",
+    ),
+    "typings_outs": attr.output_list(
+        doc = "Locations in bazel-out where tsc will write `.d.ts` files",
+    ),
+}
+
+def _join(*elements):
```
gregmagolan, line 106: surprising this isn't in skylib yet

nodejs/private/ts_project.bzl
```diff
@@ -57,40 +23,10 @@ def _declare_outputs(ctx, paths):
...
-    return _lib.join(
-        root_path,
-        ctx.label.workspace_root,
-        ctx.label.package,
-        ctx.attr.root_dir,
-    )
-
-def _ts_project_impl(ctx):
+# This function has two extra arguments that are particular to how it's called locally.
+# - run_action is used with the build_bazel_rules_nodejs linker, but other rule authors might just use ctx.actions.run
```
gregmagolan, line 27: Hmm. `ctx.actions.run` wouldn't accept a `link_workspace_root` parameter

nodejs/private/ts_project_options_validator.js
```diff
@@ -0,0 +1,144 @@
+"use strict";
```
gregmagolan, line 1: How did this transpiled file get here? Source of truth is still `packages/typescript/internal/ts_project_options_validator.ts`?

alexeagle: correct, and there's a `diff_test` in the BUILD file next to that `.ts` file.

The reasoning is: we no longer have the `#BEGIN/END-INTERNAL` fencing trick because `rules_nodejs` promises to be equally usable from sources vs. some distro artifact. We don't want to have `load` statements here in `rules_nodejs` that try to bring in typescript compilation, because `rules_nodejs` by itself can't do it.

nodejs/private/ts_validate_options.bzl
```diff
@@ -56,32 +62,16 @@ def _validate_options_impl(ctx):
...
-    attrs = dict(SHARED_ATTRS, **{
-        "has_local_deps": attr.bool(doc = "Whether any of the deps are in the local workspace (not starting with '@')"),
-        "target": attr.string(),
-        "ts_build_info_file": attr.string(),
-        "tsconfig": attr.label(mandatory = True, allow_single_file = [".json"]),
-        "validator": attr.label(default = Label("//packages/typescript/bin:ts_project_options_validator"), executable = True, cfg = "exec"),
-    }),
-)
+_ATTRS = dict(COMPILER_OPTION_ATTRS, **{
+    "has_local_deps": attr.bool(doc = "Whether any of the deps are in the local workspace (not starting with '@')"),
```
gregmagolan, line 66: technically `@//path/to/local:file` is in the local workspace and starts with `@`

nodejs/private/ts_project.bzl
alexeagle: yes that's true, it's compensated over here
https://github.com/aspect-build/rules_js/pull/27/files#diff-8bddb9b75b98520465cccb6b6633cb8e51d69f2b208047b6a4082b91e1440494R30

maybe there's a better way 🤔

nodejs/private/ts_project_options_validator.js
gregmagolan: I see ```diff_test(
    name = "check_validator.js",
    # We are copying across packages here, so don't use generated_file_test, just print a command
    failure_message = "\nRun:\n    cp dist/bin/packages/typescript/internal/*.js nodejs/private",
    file1 = "compile",
    file2 = "//nodejs/private:ts_project_options_validator.js",
)```

packages/typescript/internal/BUILD.bazel
```diff
@@ -33,18 +33,13 @@ tsc(
...
-    ),
-    visibility = ["//:__subpackages__"],
-    deps = [
-        "@build_bazel_rules_nodejs//:bzl",
-        "@build_bazel_rules_nodejs//internal/common:bzl",
-        "@rules_nodejs//nodejs:bzl",
-    ],
+diff_test(
+    name = "check_validator.js",
+    # We are copying across packages here, so don't use generated_file_test, just print a command
```
gregmagolan, line 38: why not just move the .ts file to nodejs/private as well?

alexeagle: because then we'd need rules in the BUILD file there to build it, and rules_nodejs has no way to compile typescript on its own

nodejs/private/ts_validate_options.bzl
alexeagle: that's true, `_is_external_label` is what we use to set this attribute, which is just
`len(_to_label(param).workspace_root) > 0`

nodejs/private/ts_project.bzl
alexeagle: It would have to accept a `partial` so that the link_workspace_root argument to `run_node` could be applied before passing the helper into the library code. I think it's too complex to add that here...

alexeagle: Ah, I found a better way - make the run_action helper default to None, so a TS rule implementation can just do

```
load("@rules_nodejs//nodejs/private:ts_project.bzl", _ts_project_lib = "ts_project")

ts_project = rule(
    implementation = _ts_project_lib.implementation,
    attrs = _ts_project_lib.attrs,
)
```

and when the `run_action` is not None, then we know we can pass the link_workspace_root as well.

Also moved link_workspace_root attr to be on the build_bazel_rules_nodejs implementation.

Issue #3067, opened 2021-11-06
Loading from multiple node_modules with version > 4.4.1


duarten:
# 🐞 bug report

### Affected Rule

`nodejs_binary` / `nodejs_test`

### Is this a regression?

Yes, 4.4.1 works fine, but 4.4.2 and 4.4.3 are broken. This is likely related to the linker refactor.

### Description

When a package is symlinked under node_modules due to, for example, the `packge_name` attribute of `js_library`, then node modules will be loaded from a different directory.

I noticed this with react, which complains when multiple instances are loaded.

We can see the problem from the following (edited) stack trace:

```
      <snip>
      at useMemo (node_modules/repro/node_modules/react/cjs/react.development.js:1531:20)
      at Component (../component/component.tsx:4:12)
      <snip>
      at act (node_modules/react-dom/cjs/react-dom-test-utils.development.js:1042:14)
      <snip>
      at runCLI (node_modules/@jest/core/build/cli/index.js:173:3)
```

Notice how we first load react from `node_modules/react-dom`, and later from `node_modules/repro/node_modules`.

The component package contains the following rule:

```starlark
js_library(
    name = "component",
    package_name = "repro/component",
    visibility = ["//visibility:public"],
    deps = ["component-ts"],
)
```

Omitting the workspace name (repro) from the package_name seems to solve the problem (i.e., changing it to "component" or "blah/component").

## 🔬 Minimal Reproduction

I committed a reproduction [here](https://github.com/duarten/rules_nodejs/tree/feature/repro-multiple-node-modules/examples/multiple-node-modules).

## 🔥 Exception or Error

In this case, react complains and the test fails.

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
  macOS
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
  4.2.1
  </code>
</pre>

**Rules_nodejs version:**

<pre>
  <code>
  4.4.3
  </code>
</pre>

duarten:
This is still happening with version 4.6.0.

@alexeagle, @gregmagolan any tips on how to fix this would be greatly appreciated :)

alexeagle:
hi sorry, we've been 100% heads-down on the 5.x breaking changes and getting that release out. I doubt this is fixed there but we should at least verify it's still reproducible on 5.0.0-rc.0 which I'm going to cut todayish.

duarten:
[...]

PR #3390, opened 2022-04-05
fix: symlink to runfiles only when running under ibazel

thesayyn:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Linker links into runfiles which causes nodejs to load from multiple node_modules.

Issue Number: #3067 

## What is the new behavior?
Linker does not link into runfiles tree unless it runs under `ibazel run` which was the original intention of us to link into runfiles. (i need to confirm this with @gregmagolan)

- [ ] Yes
- [x] No

### DIFF
--- a/internal/linker/index.js
+++ b/internal/linker/index.js
@@ -291,6 +291,8 @@ function main(args, runfiles) {
         log_verbose('isExecroot:', isExecroot.toString());
         const isBazelRun = !!process.env['BUILD_WORKSPACE_DIRECTORY'];
         log_verbose('isBazelRun:', isBazelRun.toString());
+        const isiBazel = !!process.env['IBAZEL'];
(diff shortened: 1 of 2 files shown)


alexeagle:
I find it hard to believe there's behavior specific to `ibazel run` that wouldn't be the same as `bazel run` - it's just repeating the run more than once

thesayyn:
> I find it hard to believe there's behavior specific to `ibazel run` that wouldn't be the same as `bazel run` - it's just repeating the run more than once

That's true if the target has no `ibazel_notify_changes` tag.  Here's the docs; https://github.com/bazelbuild/bazel-watcher#running-a-target and the logic seem to be put in there for an ibazel specific issue is what I understood from the comments above the logic.

PR #3407, opened 2022-04-13
Allow including rules_nodejs without npm

sashamor:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Build related changes
- [ ] CI related changes

Currently, there is no way to include `rules_nodejs` as a regular GitHub dependency; it has to be downloaded with `npm` and imported with something like:

```starlark
load("@npm//@bazel/typescript:index.bzl", "ts_project")
```

This means that any `BUILD.bazel` files that use any of these rules have to download and install all of `@npm`, even if the typescript target isn't being built.

This makes queries and builds of non-typescript code significantly slower, since a project may have a large number of `@npm` dependencies.

With this fix, you have the option of importing the rules without `npm`, e.g.

```starlark
load("@build_bazel_rules_nodejs//packages/typescript:index.bzl", "ts_project")
```

Using the GitHub Archive at the tag rather. than the release artifact:

```starlark
http_archive(
    name = "build_bazel_rules_nodejs",
    patch_args = ["-p1"],
    sha256 = "580779e8f9ea752210b10bf1fd717103d6600640fcdf7fe977dfec0549181bd8",
    strip_prefix = "rules_nodejs-5.3.0",
    urls = ["https://github.com/bazelbuild/rules_nodejs/archive/refs/tags/5.3.0.tar.gz"],
)
```

The only issue is, if you do this, you get the error:

> ERROR: /foo/BUILD.bazel:29:11: in validate_options rule //foo:foo:_validate_google_ts_options: target '@build_bazel_rules_nodejs//packages/typescript/bin:ts_project_options_validator' is not visible from target '//common/proto/google:_validate_google_ts_options'. Check the visibility declaration of the former target if you think the dependency is legitimate

With this patch, the error is gone. After that, everything works as expected.

Issue Number: N/A

- [ ] Yes
- [X] No

### DIFF
--- a/packages/typescript/bin/BUILD.bazel
+++ b/packages/typescript/bin/BUILD.bazel
@@ -10,5 +10,5 @@ nodejs_binary(
     name = "ts_project_options_validator",
     data = ["@npm//typescript"],
     entry_point = "//packages/typescript/internal:ts_project_options_validator.js",
-    visibility = ["//:__subpackages__"],
+    visibility = ["//visibility:public"],
 )


alexeagle:
I just recently wrote about that problem https://blog.aspect.dev/avoid-eager-fetches
https://github.com/bazelbuild/rules_nodejs/issues/3193 is the issue you're hitting.

This fix is okay with me - I think for 6.0 we will just want to drop the `@bazel/typescript` npm package so there's only one path documented.

One thing missing is to assert that the version of typescript the rule sees matches the one the editor uses.

alexeagle (approved):
Note that over here
https://github.com/aspect-build/rules_js/commit/5a19105bb03ab0ed73797e92a7fd16ce2d8530e7#diff-f4c74353cd37ed836dc2bcd3281a9e43cede40cf841af2f29756d60edb8f552cR44
we have a similar "run it without npm install" and that didn't hit the same problem because it didn't use this nodejs_binary, it declared its own.

sashamor:
Thanks @alexeagle 🚀

alexeagle:
Sure, if you'd like to help push this along that would be great!

PR #3410, opened 2022-04-14
Don't remote-cache npm directory copies

alexeagle:
These are always the same as the files in the external repository so it's wasteful to upload them to a cache


mistic:
I can confirm this reduces by close to **10** times the amount of traffic exchanged between a given build on a local machine without any local cache and a remote cache server 🎉

PR #3414, opened 2022-04-16
feat(concatjs): allows the usage of @bazel/concatjs rules without Karma

diceride:
Seperates the @bazel/concatjs rules from the Karma web testing rules, by moving the `karma_web_test` rule and `karma_web_test_suite` rule into a dedicated subdirectory in packages/concatjs/web_test.

As a user, I want to use the @bazel/concatjs devserver rules without the requirement to install the [io_bazel_rules_webtesting](https://github.com/bazelbuild/rules_webtesting) dependency. This dependency is only required by Karma web testing rules.

Before

```python
load("@npm//@bazel/concatjs:index.bzl", "karma_web_test_suite")
```

After

```python
load("@npm//@bazel/concatjs/web_test:index.bzl", "karma_web_test_suite")
```

The documentation was updated accordingly. The Karma web testing documentation in docs/Karma.md was updated. Previously, the documentation mentioned the @bazel/karma package.


diceride:
Reverted the changes in the examples

alexeagle:
It sounds good to reduce the accidental coupling here.
However it's a breaking change to the public API so it has to wait for the next major release, following semver.

PR #3420, opened 2022-04-22
fix(concatjs): resolve error with TypeScript 4.7

crisbeto:
In TypeScript 4.7 the first argument of `resolveTypeReferenceDirectives` can be a `FileReference[]`, in addition to a `string[]`. These changes resolve a runtime error that was happening as result.


crisbeto:
I'm not sure why CircleCI timed out here, but I don't have permissions to restart it.

alexeagle (approved):
Thanks!
Feels like we should somehow give the concatjs package back to Angular/Google teams since it's nearly deprecated externally...

devversion:
@alexeagle Yes. This year, I have plans to move the ts library portion at least into the Angular dev-infra repo, w/ some
changes there. Not super concrete yet, but I agree it would make sense to not bother RNJ with that

PR #3433, opened 2022-04-28
fix(concatjs): adding devmode to BazelOpts

soldair:
adds devmode to bazelOpts type definition imported by angular ngc-wrapped for <internal> options migration.

In the future, a better plan should be made to move away from lightly supported legacy corners of this repo.
related to https://github.com/angular/angular/pull/45804


alexeagle (approved):
Hi @soldair !

PR #3434, opened 2022-04-29
feat: expose @nodejs_host//:bin/node without using alias

alexeagle:
This makes it possible for repository rules to reliably find a path to a node interpreter using rctx.path,
without having to hard-code any repository name.
Needed to workaround rctx.path not understanding alias, and repository rules not having any toolchain resolution.


gregmagolan:
Should we a 2nd alias repository in this PR with our preferred name
```
    nodejs_repo_host_os_alias(
        name = name,
        user_node_repository_name = name,
    )
```
and then remove
```
    nodejs_repo_host_os_alias(
        name = name + "_host",
        user_node_repository_name = name,
    )
```
for 6.0?

kormide (approved):

PR #3435, opened 2022-04-29
feat(rollup): support esm configurations to be provided

devversion:
Adds support for ESM (.mjs) configurations when using `@bazel/rollup`.
This is useful for using e.g. ESM rollup plugins, or the Angular linker
Babel plugin (as of v13+ Angular packages ship as strict ESM).

cc. @JamesHollyer


jbedard:
Looks like a lot of unnecessary whitspace changes?

packages/rollup/rollup_bundle.bzl
```diff
@@ -312,7 +312,8 @@ def _rollup_bundle(ctx):
 
     stamp = ctx.attr.stamp[StampSettingInfo].value
 
-    config = ctx.actions.declare_file("_%s.rollup_config.js" % ctx.label.name)
+    config_extension = ctx.file.config_file.extension
+    config = ctx.actions.declare_file("_%s.rollup_config.%s" % (ctx.label.name, config_extension))
```
jbedard, line 316: Can we just change it to mjs 100% of the time?

devversion: I don't think so, since configs may also be CommonJS. If you want to enforce ESM rollup configs, that works, but it's no cost shifting that responsibility of supporting ESM or CJS to rollup itself.

packages/rollup/test/esm_config/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
...
+
+rollup_bundle(
+    name = "esm_config",
+    config_file = "rollup.config.mjs",
+    entry_point = "input.js",
+    sourcemap = "false",
+    supports_workers = True,
+    deps = [
+        # Fake test ESM package from `npm_deps.bzl`.
+        "@npm//esm_pkg",
```
jbedard, line 12: Do we need to use an npm package to reproduce the issue? Can't just be plain local imports or a `js_library`?

devversion: I wasn't able to reproduce actual ESM issues with a local script because rollup attempts to bundle the rollup config as it seems (ironically). I think having such a package is a good thing for other future ESM additions as well.

devversion: Happy to explore this more, but I believe this should be fine, or do you have any concerns?

jbedard: Would just be nice to avoid polluting the root package.json, although there's already a `testy` package in there so I guess this just follows that same pattern. Maybe just name it `testy_esm` so it's obvious they serve the same purpose?

docs/Rollup.md
```diff
@@ -347,7 +347,7 @@ Defaults to `@npm//@bazel/rollup/bin:rollup-worker`
 
 (*Boolean*): Whether to execute the rollup binary with the --silent flag, defaults to False.
 
-Using --silent can cause rollup to [ignore errors/warnings](https://github.com/rollup/rollup/blob/master/docs/999-big-list-of-options.md#onwarn)
```
jbedard, line 350: Lets revert this...

devversion: Sorry, can you elaborate why? The trailing whitespace here is a noop and inconsistent with the rest of the file.

packages/rollup/test/esm_config/BUILD.bazel
devversion: good idea, agreed. btw: there are many other test packages, some named just `hello` 😄

package.json
```diff
@@ -46,6 +46,7 @@
         "core-util-is": "^1.0.2",
         "date-fns": "1.30.1",
         "esbuild-plugin-svg": "0.1.0",
+        "test_esm_pkg": "file:./tools/npm_packages/test_esm_pkg",
```
jbedard, line 49: lets put this beside the other `test` target. Also seems like the list is alphabetical.

docs/Rollup.md
devversion: Personally I would disagree with this. This is a good drive-by fix, similar to fixing typos while spotting them. It's unlikely these things would ever be addressed if people should send a dedicated PR. A separate commit is another story, but quite a nitpick.

jbedard: If you were changing the file already then sure, but you're not evening touching the file 🤷

devversion: This file is auto-generated from the `rollup_bundle.bzl` file and had to be updated..

jbedard: Well nm then 👍 

Interesting it got checked in like this at all then 🤔

jbedard (approved):

alexeagle (approved):
thanks!

alexeagle:
Checked that rules_rollup doesn't seem to have the same problem https://github.com/aspect-build/rules_rollup/blob/main/rollup/private/rollup_bundle.bzl#L241

alexeagle:
Checked that rules_rollup doesn't seem to have the same problem https://github.com/aspect-build/rules_rollup/blob/main/rollup/private/rollup_bundle.bzl#L241

Issue #3399, opened 2022-04-08
Wrong nodejs path on windows


flemminglundahl:
# 🐞 bug report

### Affected Rule

rules_nodejs

### Is this a regression?

Don't know. This is the first time we use nodejs like this.

### Description

build_bazel_rules_nodejs throws an error on windows when running _yarn_install_impl.

## 🔬 Minimal Reproduction

Initialitze nodejs as done below.

<pre><code>
http_archive(
    name = "vendored_node_windows_amd64",
    build_file_content = """exports_files(["node.exe"])""",
    sha256 = "2883e83ac3b1e1cb9a9bf65554043640849b39e86761e7c7ac50b664f42f20ff",
    strip_prefix = "node-v14.18.0-win-x64",
    urls = ["https://nodejs.org/dist/v14.18.0/node-v14.18.0-win-x64.zip"],
)

yarn_install(
    name = "npm",
    exports_directories_only = False,
    node_repository = "vendored_node",
    package_json = "//:package.json",
    yarn = "@vendored_yarn_1_22_0//:bin/yarn.js",
    yarn_lock = "//:yarn.lock",
)
</code></pre>

## 🔥 Exception or Error

![image](https://user-images.githubusercontent.com/17703575/162442492-60025abd-f2b4-4a58-8d8e-8e8985521307.png)

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Windows 10
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
5.0.0
  </code>
</pre>

**Rules_nodejs version:**

<pre>
  <code>
5.3.1
  </code>
</pre>

**Anything else relevant?**
Patch has been created as a workaround: [0001-fix-wrong-nodejs-path-on-windows.zip](https://github.com/bazelbuild/rules_nodejs/files/8452097/0001-fix-wrong-nodejs-path-on-windows.zip)

I intend to create a PR with my patch changes.

davschne:
I'm running into the same problem. I'm curious to know why the Node executable is expected to be `bin/node.cmd` when every Node package for Windows I've seen has the executable as `node.exe` (in the root directory). I assume it's an error, but I wonder if it's working as-is for anyone and if making the correction will break their code.

@flemminglundahl , thanks for the patch! Any plans to submit the PR soon? If not, do you mind if I submit it? I'd like to start using `rules_nodejs` as soon as possible, but this issue is blocking me.

davschne:
[...]

PR #3459, opened 2022-05-17
Correct node and npm paths on Windows

davschne:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: [3399](https://github.com/bazelbuild/rules_nodejs/issues/3399)

## What is the new behavior?

Uses the correct paths to the `node` and `npm` executables on Windows

- [ ] Yes
- [x] No

## Other information

Change authored by [flemminglundahl](https://github.com/flemminglundahl)

### DIFF
--- a/examples/vendored_node_and_yarn/WORKSPACE
+++ b/examples/vendored_node_and_yarn/WORKSPACE
@@ -61,12 +61,19 @@ http_archive(
 
 register_toolchains("//toolchains:all")
 
-load("@build_bazel_rules_nodejs//:index.bzl", "yarn_install")
+load("@build_bazel_rules_nodejs//:index.bzl", "node_repositories", "yarn_install")
+
+node_repositories(
+    vendored_node_labels = {
+        "linux_amd64": "@vendored_node_linux_amd64//:bin/node",
+        "darwin_amd64": "@vendored_node_darwin_amd64//:bin/node",
+        "windows_amd64": "@vendored_node_windows_amd64//:node.exe",
+    },
+)
 
 yarn_install(
     name = "npm",
     exports_directories_only = False,
-    node_repository = "vendored_node",
     package_json = "//:package.json",
     yarn = "@vendored_yarn_1_10_0//:bin/yarn.js",
     yarn_lock = "//:yarn.lock",
--- a/nodejs/repositories.bzl
+++ b/nodejs/repositories.bzl
@@ -100,6 +100,9 @@ and `{filename}` with the matching entry from the `node_repositories` attribute.
         default = DEFAULT_NODE_VERSION,
         doc = "the specific version of NodeJS to install",
     ),
+    "vendored_node_label": attr.label(
+        doc = "Label of the Node executable in a vendored Node distribution",
+    ),
     "use_nvmrc": attr.label(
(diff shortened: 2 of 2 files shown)


davschne:
@mattem, @gregmagolan, @alexeagle, I'm trying to figure out why the Windows NodeJS packages being used in BuildKite CI have the executable as `bin/node.cmd` rather than `node.exe`. Every Windows Node package I've seen has `node.exe` in the root directory and no `bin/node.cmd` at all. Can you shed some light on this? Where do you get your Node packages?

alexeagle:
We generate the node.cmd as a wrapper around node.exe somewhere. Does that help?

alexeagle:
The Buildkite instance is run by the bazel team (using https://github.com/bazelbuild/continuous-integration )
so the config is https://github.com/bazelbuild/rules_nodejs/blob/stable/.bazelci/presubmit.yml

The `SCRIPT_DIR` and `PATH` variables in node.cmd are load-bearing, because otherwise subprocesses will run the node binary they find on the machine instead of the bazel-managed one
https://github.com/bazelbuild/rules_nodejs/blob/stable/nodejs/repositories.bzl#L246-L249
so I don't think it's as simple as ditching that wrapper and running node.exe directly instead.

There's probably some way to keep that four-line wrapper, but still call the node.exe that's vendored into the repo...

davschne:
Cool, thanks for pointing me to the CI config, @alexeagle, and for providing clarity regarding the wrapper. I'll look into the problem further and see if I can find a solution that makes the wrapper work for vendored Node.

davschne:
@alexeagle, it looks like the `node` and `npm` wrappers get created in the repository rule `node_repositories`. At present, it looks like `node_repositories` isn't designed to be used when using local vendored Node. I'm thinking the most straightforward solution is to always use `node_repositories` (or `yarn_repositories`), even for local vendored Node, if we want to ensure the wrappers get created. I'll look into what modifications to these repository rules would be required, but I did want to check with you to see what you think about this approach.

While investigating, I also noticed that `examples/vendored_node_and_yarn` fails to build for me on Windows. It gives essentially the same error message as documented in #3399. Seems that it would be good to have a test for this in CI.

alexeagle:
Yup I think that solution can work. At some point we got too overwhelmed and dropped that windows CI coverage. Maybe even due to the same bug... sorry about that

davschne:
I'm working on a revision in which `node_repositories` can take a new attribute, `node_label`, which references the Node executable in each platform-specific vendored Node repository. That way the wrapper script can delegate to the correct executable after setting up the `PATH` variable. It could be used like this:

```starlark
node_repositories(
    node_label = {
        "linux_amd64": "@vendored_node_linux_amd64//:bin/node",
        "darwin_amd64": "@vendored_node_darwin_amd64//:bin/node",
        "windows_amd64": "@vendored_node_windows_amd64//:node.exe",
    },
)
```

`node_repositories` would find the `npm-cli.js` and `npx-cli.js` scripts in the vendored repos relative to the executable referenced by the label, since Node packages use a predictable layout.

@alexeagle, any thoughts on this interface?

PR #3485, opened 2022-06-15
fix(concatjs): mark peerDependencies as optional

alan-agius4:
These peerDependencies are not mandatory for the package to run and therefore should be marked as such.


alexeagle:
We've discussed handing the concatjs package back to Google since the community doesn't really use it and it's going unmaintained, any thoughts/progress? I can help carve it into a standalone repo

alan-agius4:
@alexeagle it seems the failures are unrelated to my change. 

With regards to handling the package back to Google, I don't have any information about this. Probably @josephperrott would be the better person to give you an update.

PR #3492, opened 2022-06-23
fix(builtin): fix an bug where a nodejs_binary would fail to resolve an npm package when the linker is disabled

kormide:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

@gregmagolan 

Under a specific set of conditions where the linker is disabled, export_directories_only is false, and you have multiple yarn_installs, npm packages from the subdir node_modules will fail to be resolved.

## What is the new behavior?

The bug is fixed.

- [ ] Yes
- [X] No


e2e/BUILD.bazel
```diff
@@ -178,3 +178,8 @@ e2e_integration_test(
     # TODO: figure out why this fails on Windows
     tags = ["no-bazelci-windows"],
 )
+
+e2e_integration_test(
+    name = "linker_disabled_regression",
+    bazel_commands = ["run //subdir:binary"],
```
alexeagle, line 184: nit: should this be two-element argv?

alexeagle (approved):

e2e/BUILD.bazel
kormide: Nope, all usages I see have a full command per string.

PR #3493, opened 2022-06-23
fix(builtin): remove unnecessary loader script

kormide:
FYI: @devversion 

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: Related to https://github.com/bazelbuild/rules_nodejs/issues/3277

ESM scripts couldn't be used with `nodejs_binary` when `--nobazel_run_linker` was set. This is because the loader script is always a `.cjs` file.

## What is the new behavior?

Remove the loader script as it no longer seems to be necessary. It would be helpful if someone with more historical context could verify this.

- [ ] Yes
- [x] No


kormide:
I thought this was working but it was a false positive. The problem is that our require patch only works for node's cjs loader, not their esm loader. If we want to support esm scripts with the linker disabled, we'll need a similar monkey patch for esm.

PR #3495, opened 2022-06-27
fix(builtin): remove unnecessary loader script

kormide:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

This is a reincarnation of https://github.com/bazelbuild/rules_nodejs/pull/3493, and related to https://github.com/bazelbuild/rules_nodejs/issues/3277.

ESM scripts couldn't be used with `nodejs_binary` when `--nobazel_run_linker` was set. This is because the loader script is always a `.cjs` file and fails when it tries to load an mjs entrypoint.

## What is the new behavior?

Remove the loader script as it no longer seems to be necessary.

Note that loading third-party dependencies in an esm script when the linker is disabled does not work as our require patch only applies to the cjs loader and not the mjs loader. However this change still provides incremental benefit as you can run an mjs script with no dependencies.

A solution that people can use to load third-party deps is to provide a custom loader using node's `--loader` option.

- [ ] Yes
- [x] No


gregmagolan (approved):
:taco:

kormide:
The test failure is caused by this bug in `rules_sass`: https://github.com/bazelbuild/rules_sass/pull/142

kormide:
Added a patch for `rules_sass`. The tests on buildkite are passing now.

PR #3498, opened 2022-07-03
fix(create): make `--packageManager` flag work

mgred:
## PR Checklist

## PR Type

- [x] Bugfix

When running `yarn create @bazel my_test_workspace --typescript --packageManager=npm` the script will create a new workspace with `yarn` as a package manager instead of `npm`.

## What is the new behavior?

The flag `--packageManager` now is respected and takes effect. This also improves the script by checking, if 'yarn' or 'npm' is given with that flag.

- [x] No


alexeagle (approved):
Thanks!

PR #3499, opened 2022-07-04
feat(create): introduce `--workspaceDir` flag

mgred:
users now can set a different name for the directory to create a workspace in

the behaivior also changed, so it is possible to create a workspace e.g. in the current dir

## PR Checklist

## PR Type

Users cannot set a different directory name when creating a new workspace.

## What is the new behavior?

Users can set a different directory name or create a workspace in an existing directory.

- [x] No

### DIFF
--- a/packages/create/index.js
+++ b/packages/create/index.js
@@ -53,6 +53,8 @@ function usage(error) {
     --packageManager=[yarn|npm]   Select npm or yarn to install packages
                                   (default: npm if you ran npm/npx, yarn if you ran yarn create)
     --typescript                  Set up the workspace for TypeScript development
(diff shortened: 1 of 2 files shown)


mgred:
One thing that looks a bit weird: when running e.g. `yarn create @bazel my_test_workspace --typescript --workspaceDir=.` The _Next steps_ section would suggest to `1. cd .` which is obviously not necessary. We could avoid this by automatically running the install process after worksapce creation. Any ideas on that?

alexeagle:
Hey, sorry no one looked at your PR!

> suggest to 1. cd .
sure, it could do something clever for that case

alexeagle (approved):

PR #3510, opened 2022-07-26
fix(concatjs): sync with internal change to ensure it works with `tsickle` host

devversion:
Tsickle now requires a `generateExtraSuppressions` public property on the TS host. This change has been implemented internally but hasn't been synced upstream, making it difficult to update to the latest `tsickle` externally.


devversion:
Re-created from #3508 which seems to be closed by accident

gregmagolan (approved):
🌮

PR #3535, opened 2022-08-16
fix: `ts_project` fails if `root_dir` used deep in source tree

vpanta:
## PR Checklist

    - Unsure if necessary 

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

If `ts_project` is used with the `root_dir` parameter, and has a set of source files where the name of the `root_dir` is repeated within the source tree, then it incorrectly calculates output files (and subsequently breaks, as the filename it calculated was not produced).

For example: using a `root_dir` of `src` with the following tree:
```
- src
  - foo
    - src
      - foo.ts
 ```
would cause a bazel error of the form:
```
ERROR: BUILD.bazel:1:1: output 'foo/foo.js' was not created
```

The build should create (and expect the output) to be at `foo/src/foo.js` instead.

- [ ] Yes
- [x] No

## Other information
The fix is small but also somewhat magic-number, so I'm happy to re-evaluate the most clear way to fix this for future developers.

### DIFF
--- a/nodejs/private/ts_lib.bzl
+++ b/nodejs/private/ts_lib.bzl
@@ -169,7 +169,7 @@ def _out_paths(srcs, out_dir, root_dir, allow_js, ext_map):
     outs = []
     for f in srcs:
         if _is_ts_src(f, allow_js):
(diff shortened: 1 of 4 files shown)


alexeagle (approved):
thanks! makes sense that replace operation should have limited number of matches

PR #3538, opened 2022-08-19
fix: make catch-all node_modules target name configurable in yarn_install and npm_install

gregmagolan:
The default target name obfuscates the "node_modules" source directory for use in `manual_build_file_contents` for custom use cases.

### DIFF
--- a/internal/npm_install/npm_install.bzl
+++ b/internal/npm_install/npm_install.bzl
@@ -359,6 +359,15 @@ Using managed_directories will mean that
         default = 3600,
         doc = """Maximum duration of the package manager execution in seconds.""",
     ),
+    "all_node_modules_target_name": attr.string(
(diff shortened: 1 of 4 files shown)


internal/npm_install/npm_install.bzl
```diff
@@ -359,6 +359,10 @@ Using managed_directories will mean that
         default = 3600,
         doc = """Maximum duration of the package manager execution in seconds.""",
     ),
+    "all_node_modules_target_name": attr.string(
+        default = "node_modules",
+        doc = """The target name of the node_modules directory in one catch-all js_library target.""",
```
alexeagle, line 364: maybe include here why you'd ever want a value other than the default?

alexeagle (approved):

PR #3550, opened 2022-09-14
fix(builtin): ensure bazel config directories are expanded to prevent escapes

joeljeske:
… escapes

## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Build related changes
- [ ] CI related changes

The runtime symlink patches on FS are not currently catching node_module roots in the bazel gendirs  in bazel-out. There appears to have been an attempt to catch these roots but a faulty shell expansion was preventing these roots from being guarded. 

Issue Number: N/A

## What is the new behavior?

Symlinks are guarded properly from escaping from the desired module roots

- [ ] Yes
- [X] No

### DIFF
--- a/internal/node/launcher.sh
+++ b/internal/node/launcher.sh
@@ -294,14 +294,16 @@ if [[ -n "${BAZEL_NODE_MODULES_ROOTS:-}" ]]; then
   OLDIFS="${IFS}"
   IFS=","
   roots=(${BAZEL_NODE_MODULES_ROOTS})
-  IFS="${OLDIFS}"
   for root_path in "${roots[@]}"; do
     if [[ "${root_path}" ]]; then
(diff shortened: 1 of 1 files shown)


gregmagolan:
Hmm. The `*` was not intended to be matched in the launcher but rather in the fs patches themselves https://github.com/bazelbuild/rules_nodejs/blob/e743399631bca9ce2a3aaa08b481508f616778db/packages/node-patches/src/fs.ts#L480. Perhaps there is an issue with the patches that is causing the issue this fixes. However, using bash expansion in the launcher seems like a good alternative. Suppose all of the platform folders that need to be protected from escapes are guaranteed to be there when the launcher runs.

gregmagolan:
@joeljeske What the the symptom you were seeing that this resolved?

PR #3555, opened 2022-09-23
fix: canonicalize @platforms//cpu:aarch64

alexeagle:
This is defined as an alias to :arm64:
https://github.com/bazelbuild/platforms/blob/212a486d66569b29c95b00364e2584e80fd08614/cpu/BUILD#L16-L20 but toolchain resolution doesn't understand the two to be equivalent.

Fixes https://github.com/aspect-build/rules_js/issues/469


gregmagolan (approved):

alexeagle:
Verified downstream in rules_js: https://github.com/aspect-build/rules_js/pull/479

PR #3562, opened 2022-09-27
fix(builtin): fix a bug where the launcher produces incorrect runfiles

kormide:
path on windows

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

During a test the path isn't normalized at this point, causing the following condition to set the wrong path.

## What is the new behavior?

Correct path on windows.

- [ ] Yes
- [x] No


kormide:
This is causing more issues. I think I'll need a pair.

gregmagolan (approved):
🌮

gregmagolan:
lgtm; looks like you just have to update an env test

kormide:
It's not setting `BAZEL_PATCH_ROOTS` correctly, which I don't really understand,

gregmagolan:
> It's not setting `BAZEL_PATCH_ROOTS` correctly, which I don't really understand,

Looks like the issue was already there on HEAD with the first change but somehow CI was green on that one. Not sure what happened; but I'll fix the test to remove the windows exception. TL;DR is that `runfilesRoot` is now equal to `RUNFILES` on Windows so the exception is no longer needed.

PR #3569, opened 2022-10-03
feat: add exports_source_directories to npm_install

gregmagolan:
This allows users to opt-out or the additional copy_directory action for exported directories if they are not using RBE. With a large set of dependencies, the additional copy_directory action overhead is noticeable.


npm_deps.bzl
```diff
@@ -326,6 +340,21 @@ js_library(
         package_lock_json = "//:tools/fine_grained_deps_npm_directory_artifacts/package-lock.json",
     )
 
+    npm_install(
```
jbedard, line 343: Should these be `testonly`?

internal/js_library/js_library.bzl
```diff
@@ -73,6 +73,11 @@ If package_path is set, the linker will link this package under <package_path>/n
 If package_path is not set the this will be the root node_modules of the workspace.""",
     ),
     "srcs": attr.label_list(allow_files = True),
+    "source_directory": attr.bool(
```
jbedard, line 76: wdyt about making this private/internal and only use it from `npm_install`? One less thing to support etc 🤷

gregmagolan:
@mistic Are you still interested in this change?

PR #3570, opened 2022-10-06
feat: add npm binary & files to toolchain provider

gregmagolan:
Downstream rules that use the toolchain can choose to include the npm binary & files specified by the toolchain as inputs for actions.

### DIFF
--- a/nodejs/toolchain.bzl
+++ b/nodejs/toolchain.bzl
@@ -20,10 +20,14 @@ See /examples/toolchain for usage example.
 NodeInfo = provider(
     doc = "Information about how to invoke the node executable.",
     fields = {
-        "target_tool_path": "Path to the nodejs executable for the target platform.",
+        "target_tool_path": "Path to the nodejs executable for this target's platform.",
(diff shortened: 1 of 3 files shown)


nodejs/toolchain.bzl
```diff
@@ -89,6 +105,19 @@ node_toolchain = rule(
             doc = "Path to an existing nodejs executable for the target platform.",
             mandatory = False,
         ),
+        "npm": attr.label(
+            doc = "A hermetically downloaded npm executable target for the target platform.",
```
alexeagle, line 109: nit: should these really say for the "target platform"?

gregmagolan: right, the node fields are `target_tool` and `target_tool_path`.

`target_npm`, `target_npm_path` and `target_npm_files`?

alexeagle: no I think the fields are named fine, it's just the docstring that's a tad misleading (for node as well - it was an existing problem here)

alexeagle (approved):

PR #3602, opened 2022-11-23
fix: check RUNFILES env variable in @bazel/runfiles helper

gregmagolan:
Resolves https://github.com/aspect-build/rules_js/issues/553 and https://github.com/aspect-build/rules_js/issues/612.


alexeagle (approved):
feels like we keep touching this, I hope it's right this time :)

PR #3610, opened 2022-12-12
feat: provide @nodejs repository

### LINKED ISSUE #3375: Rename generated repo nodejs_host to nodejs
Same reasoning as https://github.com/bazelbuild/rules_python/pull/656 where we changed rules_python.

/cc @UebelAndre

alexeagle:
Generally it's a bad practice that the user calls a repository rule (nodejs_register_toolchains) with a name, and this doesn't result in a repository with that name.

Today you call `nodejs_register_toolchains(name="my_node")` and get `external/my_node_host`

This adds a new call in the macro to the underlying repository rule without the suffix. Leaves the old `_host` variant around to make this a non-breaking change.

We did the same fix in python: https://github.com/bazelbuild/rules_python/pull/656

Fixes #3375


gregmagolan-codaio (approved):
👍

gregmagolan (approved):

PR #3625, opened 2023-02-16
fix(concatjs): resolve error with TypeScript 5.0

crisbeto:
In the concatjs `CompilerHost.resolveTypeReferenceDirectives` method the `undefined` values are excluded, despite the return value being `(ts.ResolvedTypeReferenceDirective|undefined)[]`. Until now TS was handling this gracefully, but it appears to trigger an error after 5.0.

These changes add some code that should be both backwards and forwards compatible.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -120,84 +120,85 @@ tasks:
     - "--test_output=streamed"
     test_targets:
     - "//..."
-  macos:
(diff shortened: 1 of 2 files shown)


crisbeto:
The MacOS CI failures seem like flakes, but I don't have the permissions to restart it.

alexeagle (approved):

PR #3626, opened 2023-02-22
fix: allow root repositories to override node toolchain version under bzlmod

kormide:
Addresses https://github.com/aspect-build/rules_js/issues/864.

### DIFF
--- a/nodejs/extensions.bzl
+++ b/nodejs/extensions.bzl
@@ -1,12 +1,19 @@
 "extensions for bzlmod"
 
-load(":repositories.bzl", "DEFAULT_NODE_VERSION", "nodejs_register_toolchains")
+load(":repositories.bzl", "DEFAULT_NODE_REPOSITORY", "DEFAULT_NODE_VERSION", "nodejs_register_toolchains")
 
 def _toolchain_extension(module_ctx):
     registrations = {}
     for mod in module_ctx.modules:
         for toolchain in mod.tags.toolchain:
+            if toolchain.name != DEFAULT_NODE_REPOSITORY and not mod.is_root:
+                fail("Only the root module may provide a name for the node toolchain.")
+
             if toolchain.name in registrations.keys():
+                if toolchain.name == DEFAULT_NODE_REPOSITORY:
+                    # Prioritize the root-most registration of the default node toolchain version and
+                    # ignore any further registrations (modules are processed breadth-first)
+                    continue
                 if toolchain.node_version == registrations[toolchain.name]:
                     # No problem to register a matching toolchain twice
                     continue
@@ -17,6 +24,7 @@ def _toolchain_extension(module_ctx):
                 ))
             else:
                 registrations[toolchain.name] = toolchain.node_version
+
     for name, node_version in registrations.items():
         nodejs_register_toolchains(
             name = name,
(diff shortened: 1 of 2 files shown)


nodejs/extensions.bzl
```diff
@@ -2,21 +2,33 @@
...
 def _toolchain_extension(module_ctx):
     registrations = {}
     for mod in module_ctx.modules:
         for toolchain in mod.tags.toolchain:
-            if toolchain.name in registrations.keys():
-                if toolchain.node_version == registrations[toolchain.name]:
+            if toolchain.name != DEFAULT_NODE_REPOSITORY_NAME and not mod.is_root:
+                fail("Only the root module may provide a name for the node toolchain.")
+
+            name = toolchain.name if toolchain.name else DEFAULT_NODE_REPOSITORY_NAME
```
alexeagle, line 14: how about just setting the `default` of the `name` attribute below?

kormide: Good point. Fixed.

nodejs/extensions.bzl
```diff
@@ -2,11 +2,20 @@
 
 load(":repositories.bzl", "DEFAULT_NODE_VERSION", "nodejs_register_toolchains")
 
+DEFAULT_NODE_REPOSITORY = "nodejs"
```
alexeagle, line 5: sorry, another nit, maybe this should live next to DEFAULT_NODE_VERSION?

kormide: Updated.

nodejs/repositories.bzl
```diff
@@ -374,7 +377,7 @@ node_repositories = repository_rule(
 )
 
 # Wrapper macro around everything above, this is the primary API
-def nodejs_register_toolchains(name, register = True, **kwargs):
+def nodejs_register_toolchains(name = DEFAULT_NODE_REPOSITORY, register = True, **kwargs):
```
kormide, line 380: Wdyt of defaulting it for the workspace rule as well?

nodejs/extensions.bzl
```diff
@@ -1,12 +1,19 @@
...
     registrations = {}
     for mod in module_ctx.modules:
         for toolchain in mod.tags.toolchain:
+            if toolchain.name != DEFAULT_NODE_REPOSITORY and not mod.is_root:
+                fail("Only the root module may provide a name for the node toolchain.")
+
             if toolchain.name in registrations.keys():
+                if toolchain.name == DEFAULT_NODE_REPOSITORY:
+                    # Prioritize the root-most registration of the default node toolchain version and
+                    # ignore any further registrations (modules are processed depth-first)
```
aherrmann, line 15:
```suggestion
                    # Prioritize the root-most registration of the default node toolchain version and
                    # ignore any further registrations (modules are processed starting from the root module)
```
The "depth-first" part doesn't seem right as the [docs specify breadth-first iteration](https://docs.bazel.build/versions/5.0.0/skylark/lib/module_ctx.html#modules):
> The iteration order of this dictionary is guaranteed to be the same as breadth-first search starting from the root module.

kormide: Oops, typo. Thanks for catching.

nodejs/repositories.bzl
alexeagle: hm, I don't think we need to change that old one, and explicit name is probably better there since there's no transitive semantics

kormide: Removed

alexeagle (approved):

PR #3635, opened 2023-03-30
Update NodeJS Versions

paullewis:
[Node 19 is now available](https://nodejs.org/en), so I thought it might be helpful to make it available in rules_nodejs for testing. I don't _think_ it affects the default version that's brought in, which AFAICT is v16.19.0.

@alexeagle not sure if I've done this the right way, would you be happy to check?


paullewis:
Ah, looks like I'm not going to be able to sign a CLA quickly or easily; I'll close this PR for now.

PR #3646, opened 2023-04-02
chore: update release automation

alexeagle:
It should match bazel-contrib/rules-template


.github/ISSUE_TEMPLATE/BUG-REPORT.yaml
```diff
@@ -3,18 +3,6 @@ description: File a bug report to help us improve
 title: "[Bug]: "
 labels: ["bug"]
 body:
-  - type: markdown
```
jbedard, line 6: Do we want to keep any of this?

jbedard (approved):

.github/ISSUE_TEMPLATE/BUG-REPORT.yaml
alexeagle: Everything remaining in 6.x is maintained by us.

PR #3647, opened 2023-04-02
chore: bzlmod

alexeagle:
Previous rules_nodejs entries on bazel-central-registry were created by hand. This moves our 'toolchain' example which is the natural e2e for bare rules_nodejs. We test it on BCR in the normal way.
I'll need to make sure the Publish To BCR app is enabled for the bazelbuild/rules_nodejs repo.


e2e/smoke/deps.bzl
```diff
@@ -0,0 +1,29 @@
...
+
+http = module_extension(
+    implementation = _http_impl,
+    tag_classes = {"archive": tag_class(attrs = {
+        "name": attr.string(),
+        "build_file_content": attr.string(),
+        "sha256": attr.string(),
+        "urls": attr.string_list(),
+    })},
+)
```
kormide, line 29: Do we need to wrap this in a module extension? Can the `http_archive`s just go in `WORKSPACE.bzlmod`?

alexeagle: Yeah good idea, we don't actually mean to test this module extension

alexeagle:
I also now depend on aspect_bazel_lib copy_directory instead of the patched copy_file that we vendored from bazel-skylib, hopefully means we can drop that for 6.0

kormide (approved):

PR #3649, opened 2023-04-03
chore(6.0): remove internal typescript tools

jbedard:


.github/CODEOWNERS
```diff
@@ -3,7 +3,3 @@
 #    yarn update-codeowners
 
 * @mattem @gregmagolan @alexeagle
```
alexeagle, line 5: may as well add yourself here while you're editing it

docs/Built-ins.md
alexeagle, line 1: I'm surprised this file is deleted, we still have some built-in rules?

alexeagle (approved):

PR #3650, opened 2023-04-04
chore(6.0): remove @bazel/create

jbedard:


alexeagle (commented):
@gregmagolan worth pausing for a second before we delete here, it is almost like `aspect init` if the latter knew anything about JS. Would a `npx create @bazel` workflow still be attractive to someone?

jbedard:
Note that we have to drop this or build the npm package different in order to delete things like `npm_package`.

alexeagle (approved):

PR #3652, opened 2023-04-04
chore(6.0): move providers to build_bazel_rules_nodejs

jbedard:
I think they belong in `build_bazel_rules_nodejs` since `rules_nodejs` doesn't use them?


alexeagle (commented):
Yeah, the original intent of rules_nodejs bazel module was to include both the toolchain and the interop providers, but later in rules_js development @gregmagolan decided to have all the providers in rules_js. I think it's worth one more discussion of whether any of those providers could be in rules_nodejs.
We will delete build_bazel_rules_nodejs before 6.0 so assuming we don't want to allow interop between JS rules, I think we just delete these.

alexeagle (approved):

PR #3653, opened 2023-04-05
test: fix and enable e2e smoke test

jbedard:


alexeagle (commented):
it's still red though?

I wonder if we can get rid of the integration_test runner setup here. Either something akin to what we do in other rulesets on GitHub actions, or something built-in to bazelci?

Or maybe we should use bazel-contrib/bazel-integration-test

jbedard:
Sorry I should have left this as a draft for now. Still some errors that I need CI to test.

jbedard:
I'll look into bazel-contrib/bazel-integration-test though, the more we can delete the better 👍

alexeagle:
Oh gross, Windows failing :(

jbedard:
Going to just jump to the deletion instead... https://github.com/bazelbuild/rules_nodejs/pull/3655

PR #3657, opened 2023-04-11
refactor: remove rules_nodejs dependency on bazel_skylib

jbedard:


nodejs/repositories.bzl
```diff
@@ -236,8 +233,9 @@ def _prepare_node(repository_ctx):
     node_entry = "bin/node%s" % entry_ext
     npm_entry = "bin/npm%s" % entry_ext
 
-    node_bin_relative = paths.relativize(node_bin, "bin")
-    npm_script_relative = paths.relativize(npm_script, "bin")
+    node_bin_relative = node_bin[len("bin/"):]
```
gregmagolan, line 236: perhaps check if the paths start with `bin/` and fail if they don't as a defence in depth?

jbedard:
The failures are all from docs being out of date which I'd rather not fix in this PR...

gregmagolan:
> The failures are all from docs being out of date which I'd rather not fix in this PR...

`bazel run //docs:update` should do the trick... looks like the node version got updated in another PR but the docs didn't 🤔

jbedard:
Do we want to in this PR? I'd rather just merge so I can cherry-pick into 6.x...

gregmagolan:
> Do we want to in this PR? I'd rather just merge so I can cherry-pick into 6.x...

Pre-factor that in another PR and rebase this one once 5.x is green? We may as well keep 5.x green at head.

gregmagolan (approved):

PR #3666, opened 2023-04-19
Update NodeJS versions

paullewis:
There are some features of node19/20 that I'd like to try out via Bazel & `rules_nodejs`/`rules_js`. This PR updates the available versions so that's possible.


gregmagolan (approved):
Thanks for the update @paullewis

PR #3691, opened 2023-08-30
fix(jasmine): update jasmine core version check to support versions older then 4

alan-agius4:
Currently, with jasmine core version 5 the old API is being used which causes errors.

### DIFF
--- a/packages/jasmine/jasmine_runner.js
+++ b/packages/jasmine/jasmine_runner.js
@@ -147,7 +147,7 @@ async function main(args) {
   // TODO(6.0): remove support for deprecated versions of Jasmine that use the old API &
   // remember to update the `peerDependencies` as well.
   // Jasmine versions prior to 4.0.0 should use the old API.
-  if (jrunner.coreVersion().charAt(0) !== '4') {
+  if (+jrunner.coreVersion().charAt(0) < 4) {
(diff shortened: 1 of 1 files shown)


packages/jasmine/jasmine_runner.js
```diff
@@ -147,7 +147,7 @@ async function main(args) {
   // TODO(6.0): remove support for deprecated versions of Jasmine that use the old API &
   // remember to update the `peerDependencies` as well.
   // Jasmine versions prior to 4.0.0 should use the old API.
-  if (jrunner.coreVersion().charAt(0) !== '4') {
```
alexeagle, line 150: I guess this conditional just never worked at all?

Also since the conditional only guards a `warn` statement, I don't see how this change makes any difference in behavior of your build.

alexeagle (approved):

packages/jasmine/jasmine_runner.js
alan-agius4: It does work if Jasmine 4 is used as this code portion will not be trigged. But when using Jasmine 5 it is trigged incorrectly as the check only supported `4`. 

There is other code that is executed other than the `warn`. https://github.com/bazelbuild/rules_nodejs/blob/a8cd1a1b944f9d8a9507e36bc93fd8752447413f/packages/jasmine/jasmine_runner.js#L153-L173

alexeagle (approved):

PR #3724, opened 2024-03-29
fix: check JS_BINARY__RUNFILES env variable in @bazel/runfiles helper

gregmagolan:
rules_js exports JS_BINARY__RUNFILES as an alias to RUNFILES. This change gives us the option to drop exporting RUNFILES in the future which overly generic and easily confused with an env var that Bazel export when it doesn't.

## PR Type

- [x] Bugfix

- [ ] Yes
- [x] No

### DIFF
--- a/packages/runfiles/runfiles.ts
+++ b/packages/runfiles/runfiles.ts
@@ -29,11 +29,13 @@ export class Runfiles {
       this.manifest = this.loadRunfilesManifest(_env['RUNFILES_MANIFEST_FILE']!);
     } else if (!!_env['RUNFILES_DIR']) {
       this.runfilesDir = path.resolve(_env['RUNFILES_DIR']!);
+    } else if (!!_env['JS_BINARY__RUNFILES']) {
+      this.runfilesDir = path.resolve(_env['JS_BINARY__RUNFILES']!);
     } else if (!!_env['RUNFILES']) {
       this.runfilesDir = path.resolve(_env['RUNFILES']!);
     } else {
       throw new Error(
-          'Every node program run under Bazel must have a $RUNFILES_DIR, $RUNFILES or $RUNFILES_MANIFEST_FILE environment variable');
+          'Every node program run under Bazel must have a $RUNFILES_MANIFEST_FILE, $RUNFILES_DIR, $JS_BINARY__RUNFILES or $RUNFILES environment variable');
     }
     // Under --noenable_runfiles (in particular on Windows)
     // Bazel sets RUNFILES_MANIFEST_ONLY=1.


packages/runfiles/runfiles.ts
```diff
@@ -31,9 +31,11 @@ export class Runfiles {
       this.runfilesDir = path.resolve(_env['RUNFILES_DIR']!);
     } else if (!!_env['RUNFILES']) {
       this.runfilesDir = path.resolve(_env['RUNFILES']!);
+    } else if (!!_env['JS_BINARY__RUNFILES']) {
```
jbedard, line 34: Should we put `JS_BINARY__RUNFILES` before `RUNFILES` so they are in priority order? The lower the priority the more likely we are to drop it if things like https://groups.google.com/g/bazel-discuss/c/DsVivJhU7Bw/m/trgF86CqDwAJ move forward?

gregmagolan: Sounds good. I'll update.

alexeagle (commented):
According to this
https://github.com/bazelbuild/rules_python/blob/main/python/runfiles/runfiles.py#L81-L84
we should be using `RUNFILES_DIR` exclusively

Note that this matters when you have one tool call another. A `js_binary` could be in the data of a `some_other_lang_binary` and ideally if these languages agreed on what the environment variable should be, then they both can resolve runfiles?

gregmagolan:
> According to this https://github.com/bazelbuild/rules_python/blob/main/python/runfiles/runfiles.py#L81-L84 we should be using `RUNFILES_DIR` exclusively
> 
> Note that this matters when you have one tool call another. A `js_binary` could be in the data of a `some_other_lang_binary` and ideally if these languages agreed on what the environment variable should be, then they both can resolve runfiles?

I don't remember the reason we added RUNFILES originally but I suspect there were context where RUNFILES_DIR is _not_ set. It will be an archeology project to track that one down all the way back to rules_nodejs ages ago.

jbedard:
There's also the bash one: https://github.com/bazelbuild/bazel/blob/7.1.1/tools/bash/runfiles/runfiles.bash#L87-L95

jbedard:
So we don't know the case where `RUNFILES_DIR` is not set? It is set in every test we do have and know of?

packages/runfiles/runfiles.ts
```diff
@@ -29,11 +29,13 @@ export class Runfiles {
       this.manifest = this.loadRunfilesManifest(_env['RUNFILES_MANIFEST_FILE']!);
     } else if (!!_env['RUNFILES_DIR']) {
       this.runfilesDir = path.resolve(_env['RUNFILES_DIR']!);
+    } else if (!!_env['JS_BINARY__RUNFILES']) {
+      this.runfilesDir = path.resolve(_env['JS_BINARY__RUNFILES']!);
     } else if (!!_env['RUNFILES']) {
       this.runfilesDir = path.resolve(_env['RUNFILES']!);
```
jbedard, line 35: Should we declare `RUNFILES` as deprecated and output a warning?

PR #3727, opened 2024-04-01
refactor: update node toolchain to provider File objects instead of paths

gregmagolan:
**BREAKING CHANGES** intended for rules_nodejs 7.0 and rules_js 2.0.

To-date the node toolchain has provided paths (strings) to node and npm. This complicates downstream usage as they need to convert these paths depending on they are used (between execpath and rootpath for example). This change configures the toolchain to instead provide File objects for node and the npm entry point script so downsteam uses can choose to get either the `.path` or `.short_path` as needed.

- [x] Yes
- [ ] No


gregmagolan:
Tested this on rules_js 2.0 branch: https://github.com/aspect-build/rules_js/pull/1557.

alexeagle:
I think a 7.0 of rules_nodejs is very undesirable. Breaking changes low in the depgraph have always been a major hassle for us and don't work well for users. It would need a lot of devrel care and attention that we don't have much time for.
Can we avoid it?

gregmagolan:
Replaced by https://github.com/bazelbuild/rules_nodejs/pull/3736 which maintains backward compat

PR #3728, opened 2024-04-02
chore: upgrade to Bazel 7.1.1

gregmagolan:
Fixes up the e2e/smoke bzlmod test as well. The default Node.js version of 18 breaks on Bazel CI due to GLIBC incompatibility with the runner so default Node.js must be < 18.


e2e/smoke/MODULE.bazel
```diff
@@ -18,12 +18,7 @@ local_path_override(
 
 node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node")
 
-# Note, this gets the default version of Node.js from
-# https://github.com/bazelbuild/rules_nodejs/blob/5.8.0/nodejs/repositories.bzl#L11
-node.toolchain(
-    name = "node16",
```
jbedard, line 24: Why drop this when upgrading bazel?

gregmagolan: I mentioned it in the PR summary. Getting to bazel means the bzlmod by default and that means the default toolchain is used which is Node 18. Node 18 doesn't run on the Bazel CI runners because it requires a newer version of GLIBC.

jbedard: The PR summary says nothing about the toolchain name. Why change the name making the diff 10x larger?

PR #3730, opened 2024-04-02
fix: support node_version_from_nvmrc with bzlmod

gregmagolan:
Add new `node_version_from_nvmrc` attribute to bzlmod and deprecate old naming of `use_nvmrc` used in WORKSPACE. `node_version_from_nvmrc` naming follows the convention from rules_ts where a similar attribute is named `ts_version_from` and is more accurate and self-documenting.


gregmagolan:
> Can we reduce the diff to just the core change? I'm not sure what half the diff is from...

I explained in the PR summary. The corresponding change to adding `node_version_from_nvmrc` to bzlmod is deprecating the `use_nvmrc` for WORKSPACE and auto-converting that usage to `node_version_from_nvmrc`. That keeps this from being a breaking change. If I just `use_nvmrc` to bzlmod then it would be more work to then make the rename non-breaking for bzlmod.

alexeagle (approved):

jbedard:
@gregmagolan I understand what the change is, that's simple, I don't understand why the diff on the PR needs to be so big though?

PR #3731, opened 2024-04-02
fix: use default node url if the user happens to explicitly set []

gregmagolan:
Corner case I hit in rules_js. The current error message if you accidentally set `node_urls = []` is not great. This bit of code restores the default value if the user happens to clear it with an empty list.

### DIFF
--- a/nodejs/repositories.bzl
+++ b/nodejs/repositories.bzl
@@ -9,6 +9,9 @@ load("//nodejs/private:toolchains_repo.bzl", "PLATFORMS", "toolchains_repo")
 # created by the module extension
 DEFAULT_NODE_REPOSITORY = "nodejs"
 
+# Default Node.js URL used as the default for node_urls
+DEFAULT_NODE_URL = "https://nodejs.org/dist/v{version}/{filename}"
+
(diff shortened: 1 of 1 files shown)


jbedard:
What is the corner case you hit?

Generally "correcting" bad API usage just makes things more complicated imo :/

gregmagolan:
> What is the corner case you hit?
> 
> Generally "correcting" bad API usage just makes things more complicated imo :/

The corner case was a wrapper macro. I think the default should be `[]` personally because that implies that you want to canonical upstream source but having the default set to the canonical URL is useful for the doc generator.

nodejs/repositories.bzl
```diff
@@ -166,6 +167,9 @@ def _download_node(repository_ctx):
         return
 
     node_urls = repository_ctx.attr.node_urls
+    if not node_urls:
```
alexeagle, line 170: don't you also want to check `len(node_urls)` per the commit message?

alexeagle (approved):

nodejs/repositories.bzl
gregmagolan: `if not []:` is falsy

Issue #3720, opened 2024-03-04
[Bug]: `nodejs_repo_host_os_alias` produces dangling symlink when Bzlmod is enabled


Wyverald:
### What happened?

https://github.com/bazelbuild/rules_nodejs/blob/943800d8d601872dae53f5c1e17414d316834cac/nodejs/private/nodejs_repo_host_os_alias.bzl#L34-L38

This logic fabricates a path into another repo, but the (canonical) name of that repo is different when Bzlmod is enabled.

Ideally this should be fixed by using a label somehow, but it's hard to get right and has the side effect of causing the symlink target repo to be fetched. So maybe just appending `_{os_name}` to `repository_ctx.name` is the safest bet.

### Version

present on main; only an issue with Bzlmod enabled

Wyverald:
@alexeagle could you confirm if this repo rule is still used (especially by rules_js etc)? If not, we could probably close this.

PR #3752, opened 2024-06-06
chore: enable and green up e2e/smoke and e2e/nodejs_host tests on GitHub Actions

gregmagolan:
The coverage on Bazel CI doesn't cover all cases. This is a pre-factor for fixing the https://github.com/bazelbuild/rules_nodejs/issues/3720 bug.

### DIFF
--- a/e2e/nodejs_host/version_test.sh
+++ b/e2e/nodejs_host/version_test.sh
@@ -40,9 +40,4 @@ if ! [[ $OUT =~ .*node@$EXPECTED_NODE_VERSION.* ]]; then
   exit 1
 fi
 
-if ! [[ ${OUT//\\//} =~ .*"${1}"/nodejs/(bin/)?node.* ]]; then
-  echo "Error: No match on expected node path $1/nodejs/node"
-  exit 1
-fi
-
 exit 0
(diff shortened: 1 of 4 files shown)


e2e/nodejs_host/version_test.sh
```diff
@@ -40,9 +40,4 @@ if ! [[ $OUT =~ .*node@$EXPECTED_NODE_VERSION.* ]]; then
   exit 1
 fi
 
-if ! [[ ${OUT//\\//} =~ .*"${1}"/nodejs/(bin/)?node.* ]]; then
-  echo "Error: No match on expected node path $1/nodejs/node"
```
gregmagolan, line 44: This relied on a the output of npm & npm including a path such as 
```
npm verb cli /private/var/tmp/_bazel_greg/9702de74751d6a716c3e528c1fdaa7b3/external/node16_darwin_arm64/bin/nodejs/bin/node
```

but this doesn't hold true in all cases as sometimes the linux sandbox in Bazel 7 has paths such as,

```
npm verb cli /tmp/bazel-source-roots/2/bin/nodejs/bin/node
```

instead. I don't know why that is but I'm removing this assertion as its redundant and brittle

jbedard (approved):

PR #3753, opened 2024-06-06
fix: remove unused bad symlink in nodejs_repo_host_os_alias

gregmagolan:
Fixes https://github.com/bazelbuild/rules_nodejs/issues/3720.

This repository symlink is dangling under bzlmod since the repository name the symlink is pointing to is not the same. It is safe to remove it as it is not in use anywhere.

The dangling symlink was identified due to a crash in bazel 7.1.0rc1 where the non-existant file was watched:. See https://github.com/bazelbuild/bazel/issues/21483#issuecomment-1972820810.


nodejs/private/nodejs_repo_host_os_alias.bzl
```diff
@@ -18,20 +18,13 @@ alias(name = "node_files",              actual = "@{node_repository}_{os_name}//
...
     "index.bzl",
-    "bin/node{ext}",
 ])
 """.format(
         node_repository = repository_ctx.attr.user_node_repository_name,
         os_name = os_name(repository_ctx),
         ext = ext,
     ))
 
-    repository_ctx.symlink("../{node_repository}_{os_name}/bin/node{ext}".format(
```
jbedard, line 29: Is anything testing this? No tests require changing or update? :/

jbedard (approved):

nodejs/private/nodejs_repo_host_os_alias.bzl
gregmagolan: Nothing needs updating. It was legacy and unused.

PR #3755, opened 2024-06-07
chore: misc cleanup

gregmagolan:
- remote old unused RBE configuration from when Angular RBE was used
- bump to buildifier 6.4.0
- add `//:buildifier` target that can be `bazel run`
- add Buildifier GHA check
- bazelrc cleanup
- bazelignore cleanup
- gitignore cleanup


.bazelrc
```diff
@@ -1,38 +1,17 @@
...
-build:docker-sandbox --spawn_strategy=docker --strategy=Javac=docker --genrule_strategy=docker
-build:docker-sandbox --define=EXECUTOR=remote
-build:docker-sandbox --experimental_docker_verbose
-build:docker-sandbox --experimental_enable_docker_sandbox
-# This is the same image used on BazelCI rbe_ubuntu1604 job
-build:docker-sandbox --experimental_docker_image=gcr.io/cloud-marketplace/google/rbe-ubuntu16-04
-
 # Incompatible flags to run with
-build --incompatible_no_implicit_file_export
+# TODO: flip this once @buildifier_prebuilt is compatible with it
```
jbedard, line 12: Is there a bug to link to?

jbedard (approved):

.bazelrc
gregmagolan: Not that I know of. First time I've came across this but `--incompatible_no_implicit_file_export` is also not a flag that I've seen flipped anywhere else so makes not surprising that `@buildifier_prebuilt` is not compatible. Not worth making an issue over it.

Issue #3722, opened 2024-03-15
[FR]: Separate toolchain for for C++ headers


Silic0nS0ldier:
### What is the current behavior?

As of #3694 the `NodeInfo` provider includes `CcInfo` for the NodeJS N-API headers.

While useful for native addon development, this made the `node_toolchain` depend on the C++ toolchain as that `CcInfo` instance is pulled from the `cc_library` target `:headers`.

For workflows that don't need those headers, they incur the penalty of fetching the C++ toolchain. This can play out as;
- A sub-second delay while the repo fetches.
- A much more substantial delay (e.g. if using `nixpkgs_cc_configure` from https://github.com/tweag/rules_nixpkgs)
- An error due to the C++ toolchain not being defined.

### Describe the feature

Introduce a separate toolchain targeted at native addon development so that plain JS workflows do not implicitly depend on C++ toolchains.

jesses-canva:
A specific use case that is broken by this is embedding a `js_binary` in a Linux container image with `rules_docker` or `rules_oci`, but building on macOS or Windows, or Linux of a different arch, as in these cases a cc toolchain targeting the container image platform isn't available and a `Error in fail: Unable to find a CC toolchain using toolchain resolution.` results.

dzbarsky:
Hmm can this be optional/lazy somehow? 

Having it be a separate toolchain might end up with it not matching the selected target toolchain. I'm surprised we haven't heard similar issues with py_binary, I think it's the same setup?

jesses-canva:
> I'm surprised we haven't heard similar issues with py_binary, I think it's the same setup?

[py_runtime](https://bazel.build/reference/be/python#py_runtime) doesn't contain a label to a `cc_library` like `node_toolchain` does, so doesn't have this specific issue.

Although there is a similar issue in https://github.com/bazelbuild/bazel/issues/8751 that causes a `py_binary` to require a C++ toolchain, we have patched that out for the same reason.

jesses-canva:
One option would be
1. Use a `filegroup` instead of a `cc_library`
2. Include just a `DeafaultInfo` of the header files in the `NodeInfo`, instead of a `DefaultInfo` and a `CcInfo`, this way analysing the toolchain doesn't require analysing a `cc_library`.
3. In `current_node_cc_headers` return the `DefaultInfo` out of the current `NodeInfo.headers` instead of a `CcInfo`.
[...]

PR #3760, opened 2024-06-08
fix: make CcInfo/cc dep in nodejs toolchain opt-in via include_headers attribute

gregmagolan:
Fixes https://github.com/bazelbuild/rules_nodejs/issues/3722.

Having a `cc_library` target always included in the `headers` field of the `NodeInfo` provided by the Node.js toolchain means that there must be always c++ toolchain available when using rules_nodejs which is not desirable. This change makes the `cc_library` target in headers an opt-in via the `include_headers` attribute.

Users can set this attribute as follows,

WORKSPACE:
```
load("@rules_nodejs//nodejs:repositories.bzl", "nodejs_register_toolchains")
nodejs_register_toolchains(include_headers = True, ...)
```

MODULE.bazel
```
node = use_extension("@rules_nodejs//nodejs:extensions.bzl", "node", dev_dependency = True)
node.toolchain(include_headers = True, ...)
```


.bazelci/presubmit.yml
```diff
@@ -36,9 +46,28 @@ tasks:
       - "--test_tag_filters=-skip-on-bazelci-ubuntu"
     test_targets:
       - "//..."
-  macos-smoke:
+  macos:
```
jbedard, line 49: We're adding macos testing on the root `//...`? Is that related to the headers?

jbedard (approved):

.bazelci/presubmit.yml
gregmagolan: Not strictly related but good to add the coverage along with this change either way

PR #3763, opened 2024-06-24
feat: add `node_urls` parameter to bzlmod `toolchain` in the `node` extension

redsun82:
## PR Checklist

* I don't think I have the capacity to also add tests for this.
* I don't think there's currently any bzlmod documentation, so not sure where I would be adding this.

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

`node_urls` can be provided to add nodejs mirrors to the `WORKFLOW` based setup, but the same can't be done when using bzlmod.

Issue Number: N/A

## What is the new behavior?

When using bzlmod, `node_urls` can be provided to the `toolchain` tag class in the `node` extension, behaving just like the `WORKFLOW` counterpart.

- [ ] Yes
- [X] No


nodejs/extensions.bzl
```diff
@@ -62,6 +72,10 @@ If set then the version found in the .nvmrc file is used instead of the one spec
 This setting creates a dependency on a c++ toolchain.
 """,
             ),
+            "node_urls": attr.string_list(
+                doc = "Custom list of URL formats to use to download NodeJS, containing {version} and {filename}",
```
jbedard, line 76: Can you copy the same docs that are in the repo rule version?
```
List of URLs to use to download Node.js.

Each entry is a template for downloading a node distribution.

The `{version}` parameter is substituted with the `node_version` attribute,
and `{filename}` with the matching entry from the `node_repositories` attribute.
```

nodejs/extensions.bzl
```diff
@@ -1,40 +1,50 @@
...
             else:
-                registrations[toolchain.name] = struct(
-                    node_version = toolchain.node_version,
-                    node_version_from_nvmrc = toolchain.node_version_from_nvmrc,
-                    include_headers = toolchain.include_headers,
-                )
+                registrations[toolchain.name] = toolchain
 
-    for k, v in registrations.items():
+    for k, t in registrations.items():
```
jbedard, line 41: Any reason for this? Revert if not...

nodejs/extensions.bzl
```diff
@@ -1,40 +1,50 @@
...
+    "nodejs_register_toolchains",
+)
 
 def _toolchain_extension(module_ctx):
     registrations = {}
     for mod in module_ctx.modules:
         for toolchain in mod.tags.toolchain:
             if toolchain.name != DEFAULT_NODE_REPOSITORY and not mod.is_root:
                 fail("Only the root module may provide a name for the node toolchain.")
-
```
jbedard, line 11: Revert? 🤷

nodejs/extensions.bzl
redsun82: Will do so tomorrow!

nodejs/extensions.bzl
redsun82: it's now a `toolchain` rather than a `version`, but I can revert it.

nodejs/extensions.bzl
redsun82: Done!

nodejs/extensions.bzl
redsun82: Done!

nodejs/extensions.bzl
redsun82: Done!

redsun82:
@jbedard thanks for the review.

While going over the code again I figured there was also a problem in the existing conflict check, in that it didn't check a difference on the `include_headers` setting. I've refactored the check so that it takes that into account and any new attributes we might add in the future. It makes the diff bigger, but I think it's worth it.

jbedard:
Are there any tests in the WORKSPACE version that can maybe be copied for the extension?

redsun82:
@jbedard I'll try to take a quick look.

By the way: CI failed on a temporary nodejs.org failure, which exactly the kind of problem I want to solve with this in my code base with mirrors :slightly_smiling_face: I didn't see any way I could rerun the check, but maybe I didn't look enough. On the other hand if I commit something more I'll get the reruns any way.

nodejs/extensions.bzl
```diff
@@ -14,54 +26,65 @@ def _toolchain_extension(module_ctx):
                     # Prioritize the root-most registration of the default node toolchain version and
                     # ignore any further registrations (modules are processed breadth-first)
                     continue
-                if toolchain.node_version == registrations[toolchain.name].node_version and toolchain.node_version_from_nvmrc == registrations[toolchain.name].node_version_from_nvmrc:
+                toolchain_info = _toolchain_info(toolchain)
+                registered_toolchain_info = _toolchain_info(registrations[toolchain.name])
+                if toolchain_info != registered_toolchain_info:
```
jbedard, line 31: Lets change this to `if not toolchains_equal(toolchain, registrations[toolchain.name])` just to hide most of this stuff (the `_toolchain_info` calls) in one method that describes what's going on.

jbedard:
Sorry I forgot about this.

If @redsun82 doesn't respond maybe you can update this? I assume my latest suggestion is easy enough, and would have saved me more time trying to remember what this PR is doing that I think it's worth it 😅

jbedard (approved):

PR #3767, opened 2024-07-09
feat: expose latest known node version

### LINKED ISSUE #3765: [FR]: Expose known Node versions (as starlark variable)
### What is the current behavior?

The list of versions that is supported is private.

https://github.com/bazelbuild/rules_nodejs/blob/6f9c4ad26b4558712a8d921c76a3038311b1594f/nodejs/private/node_versions.bzl#L6

Also, it is somewhat unstructured, as can be seen by how `DEFAULT_NODE_VERSION` is calculated:

https://github.com/bazelbuild/rules_nodejs/blob/6f9c4ad26b4558712a8d921c76a3038311b1594f/nodejs/repositories.bzl#L19-L24

### Describe the feature

We are struggling to integrate Renovate and `rules_nodejs` properly: Since `rules_nodejs` always lags a bit behind in node versions, we get a lot of very noisy updates (that fail, because `rules_nodejs` doesn't recognize the version yet).

Our plan is to automatically update the Renovate configuration (via bazel `write_source_files` rules) to include appropriate configuration (we do a similar thing with the `pnpm` version for `rules_js`).

However, since the supported Node versions are not exported, we cannot do this.

I'd like to propose the following (happy to do the implementation work, just wanted to check in first):

- Refactor `NODE_VERSIONS` from `Map[String, String]` to `Map[String, Map[String, String]]` (version -> platform -> repository info).
- Expose the keys of that dictionary as `KNOWN_NODE_VERSIONS` (in `//nodejs:repositories.bzl`).

gzm0:
Note: The term "known" (as opposed to supported) is deliberate to avoid giving the impression that the version is "supported" by rules_nodejs any more than having the SHA on file.

Closes #3765

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #3765

- [ ] Yes
- [x] No

### DIFF
--- a/nodejs/repositories.bzl
+++ b/nodejs/repositories.bzl
@@ -20,6 +20,12 @@ DEFAULT_NODE_VERSION = [
     if v.startswith("18.")
 ][-1]  # Versions are sorted increasing, so last one is the latest version
 
+NEWEST_KNOWN_NODE_VERSION = [
+    # 16.18.1-windows_amd64 -> 16.18.1
+    v.split("-")[0]
+    for v in NODE_VERSIONS.keys()
+][-1]  # Versions are sorted increasing, so last one is the latest version
+
 BUILT_IN_NODE_PLATFORMS = PLATFORMS.keys()
 
 _DOC = """To be run in user's WORKSPACE to install rules_nodejs dependencies.
--- a/nodejs/repositories.bzl
+++ b/nodejs/repositories.bzl
@@ -20,7 +20,7 @@ DEFAULT_NODE_VERSION = [
     if v.startswith("18.")
 ][-1]  # Versions are sorted increasing, so last one is the latest version
 
-NEWEST_KNOWN_NODE_VERSION = [
+LATEST_KNOWN_NODE_VERSION = [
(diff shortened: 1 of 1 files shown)


gzm0:
- I have deliberately not added a test, feels like it would be more brittle than useful.
- I have deviated from the original plan laid out in #3765 because the shape of `NODE_VERSIONS` is exposed through the `node_repositories` attribute. We could re-shape it, but it doesn't seem worth the effort. Instead I have settled for the minimum API necessary to support our use case. (happy to widen this, e.g. if you feel we should expose all node versions).

jbedard:
IMO this should be "latest", not "newest". This is probably debatable but I think "newest" could be a new patch on an old version, where "latest" is most up to date?

gzm0:
Huh, good point.

My intuition is actually flipped w.r.t. latest / newest. Maybe "highest" is what we are looking for?

gzm0:
FWIW: @jbedard I've switched this to `LATEST`. I do not feel very strongly about the name.

jbedard (approved):

PR #3783, opened 2024-09-02
chore: remove mention of yarn

alexeagle:
Switch root package manager to pnpm.
Remove clang tools, we'll just use prettier.
Remove stray reference to rules_codeowners which is already removed.
Installed pre-commit locally, so some files got formatted too.


dist/bin
```diff
@@ -0,0 +1 @@
+/private/var/tmp/_bazel_alexeagle/f366b2ea298a1eb16435a50617f648a1/execroot/rules_nodejs/bazel-out/darwin_arm64-fastbuild/bin
```
jbedard, line 1: These need to be dropped

scripts/update-nodejs-versions.js
```diff
@@ -77,26 +80,26 @@ async function main() {
   const versions = await getNodeJsVersions();
   const nodeRepositories = await Promise.all(versions.map(getNodeJsVersion));
   console.log('""" Generated code; do not edit');
-  console.log('Update by running yarn update-nodejs-versions\n"""\n');
+  console.log('Update by running npm run update-nodejs-versions\n"""\n');
```
jbedard, line 83: Shouldn't this be `pnpm run`?

jbedard (approved):

PR #3794, opened 2024-10-11
fix(runfiles): @bazel/runfiles usage with non-bzlmod repos

jbedard:
When in non-bzlmod and no `_repo_mapping` file exists the logic repo mapping logic should have no side-effects and should take the same code path as before https://github.com/bazel-contrib/rules_nodejs/commit/21b56da9f20c9c5e8b8c1ac19052f1f7cc7b5920

### DIFF
--- a/packages/runfiles/runfiles.ts
+++ b/packages/runfiles/runfiles.ts
@@ -168,7 +168,7 @@ export class Runfiles {
       // If the repository mappings were loaded ensure the source repository is valid.
       if (!(sourceRepo in this.repoMappings)) {
         throw new Error(
-          `source repository ${sourceRepo} not found in repo mappings: ${JSON.stringify(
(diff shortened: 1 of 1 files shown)


gregmagolan (approved):

jbedard:
There still seem to be issues and differences with non-bzlmod in bazel 6 vs 7.

In bazel 7 the `_repo_mapping` seems to exist with non-bzlmod, but the mappings are essentially no-ops. While in bazel 6 the `_repo_mapping` does not exist at all. I can't get bazel 6 or 7 working properly with non-bzlmod in the [rules_js e2e/runfiles tests](https://github.com/aspect-build/rules_js/tree/main/e2e/runfiles).

But I believe @jfirebaugh reports this fix is working so I think it is good to land for now...

Issue #3795, opened 2024-10-15
[FR][Bug]: Split `@rules_nodejs//nodejs:toolchain_type` to appropriately cover all use cases


Silic0nS0ldier:
### What is the current behavior?

Currently `@rules_nodejs//nodejs:toolchain_type` is covering 2 separate use cases.
1. Direct usage within a rule.
2. Runtime for executable outputs.

When execution and target platforms are the same (as is usually the case) scenarios 1 and 2 can be satisfied by the same toolchain implementation.

However when the execution and target platforms are different, problems arise. For example;

```starlark
# From /___/external/rules_nodejs~~node~nodejs_toolchains/BUILD.bazel
toolchain(
    name = "linux_amd64_toolchain",
    exec_compatible_with = ["@platforms//os:linux", "@platforms//cpu:x86_64"],
    toolchain = "@nodejs_linux_amd64//:toolchain",
    toolchain_type = "@rules_nodejs//nodejs:toolchain_type",
)
toolchain(
    name = "linux_amd64_toolchain_target",
    target_compatible_with = ["@platforms//os:linux", "@platforms//cpu:x86_64"],
    toolchain = "@nodejs_linux_amd64//:toolchain",
    toolchain_type = "@rules_nodejs//nodejs:toolchain_type",
)
toolchain(
    name = "darwin_arm64_toolchain",
    exec_compatible_with = ["@platforms//os:macos", "@platforms//cpu:arm64"],
    toolchain = "@nodejs_darwin_arm64//:toolchain",
    toolchain_type = "@rules_nodejs//nodejs:toolchain_type",
)
toolchain(
    name = "darwin_arm64_toolchain_target",
    target_compatible_with = ["@platforms//os:macos", "@platforms//cpu:arm64"],
    toolchain = "@nodejs_darwin_arm64//:toolchain",
    toolchain_type = "@rules_nodejs//nodejs:toolchain_type",
)
```

Suppose I am on macOS and building for Linux.
* `:linux_amd64_toolchain` will be skipped;
  * 🔴 `exec_compatible_with`
  * 🟢 `target_compatible_with` (no constaints)
* `:linux_amd64_toolchain_target` _can_ be selected;
  * 🟢 `exec_compatible_with` (no constaints)
  * 🟢 `target_compatible_with`
* `:darwin_arm64_toolchain` _can_ be selected;
  * 🟢 `exec_compatible_with`
  * 🟢 `target_compatible_with` (no constaints)
[...]

PR #3800, opened 2024-11-03
Split `:toolchain_type` to correctly cover all use cases

Silic0nS0ldier:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #3795

See issue for details.

## What is the new behavior?

* Split `@rules_nodejs//nodejs:toolchain_type` into `@rules_nodejs//nodejs:runtime_toolchain_type` and `@rules_nodejs//nodejs:exec_runtime_toolchain_type`.
* `@rules_nodejs//nodejs:runtime_toolchain_type` is suited towards usage in the target environment (e.g. runtime for `*_binary` rule outputs).
* `@rules_nodejs//nodejs:exec_runtime_toolchain_type` is suited towards usage within rule actions (e.g. running NodeJS directly).
* Split behaviour is disabled by default, controlled by a new attribute `incompatible_split_toolchains` on;
  * `node.toolchain` module extension tag.
  * `nodejs_register_toolchains` macro.
  * `nodejs_toolchains_repo` repository rule (private API).
* `:resolved_toolchain` target moved into `@rules_nodejs//nodejs`. This makes it possible to access a NodeJS toolchain implementation without needing to know the name of a user controlled generated repository (e.g. for use inside a `genrule` defined in another ruleset).
* Add new toolchain type `@rules_nodejs//nodejs:compilation_toolchain_type` suited to usage of NodeJS as a compiler (e.g. for snapshot generation) when the exec and target platforms need to be aligned. The exact platforms can still be different, but they should _at least_ be compatible across OS and CPU.

- [ ] Yes
- [x] No (but does add deprecations)

## Extra TODO

- [ ] Update issue references to point to this PR.


alexeagle (commented):
I imagine this is approximately what needs to happen, thanks for working on it!

Is there an issue (or maybe several) which could be linked from here?
I'm not sure who's best suited to do the review, among folks in the bazel-contrib org there are certainly a few experts on toolchain types and resolution.

fmeum:
Instead of an incompatible attribute on repo rules and module extensions, it may be possible to temporarily make the new toolchain types `label_flag`s that by default point to the old toolchain type for backwards compatibility. Enabling the new behavior would then require setting the `label_flag` to a new distinct type. That's a much smaller API surface and dropping the flag would only require `.bazelrc` changes, no source changes.

mortenmj:
@Silic0nS0ldier any chance this might get the attention it needs to land? For context, we've been running with this for a few months now, and it solved a lot of the issues we had cross-compiling from Mac.

Silic0nS0ldier:
With the exception of `@rules_nodejs//nodejs:compilation_toolchain_type` (for the uncommon scenario where OS+CPU need to match across exec and target, like for snapshot generation) yes, although I think you were after #3795.

Thanks @guw for getting the important bits over the line!

PR #3830, opened 2025-05-06
feat: change nodejs default to 20

alexeagle:
Goal is to match github actions `ubuntu-latest` https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md


gregmagolan (approved):
🌮

PR #3839, opened 2025-06-24
feat: lazy initialize runfiles

nickschaap:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: [3837](https://github.com/bazel-contrib/rules_nodejs/issues/3837)

- [ ] Yes
- [x] No

### DIFF
--- a/packages/runfiles/runfiles.ts
+++ b/packages/runfiles/runfiles.ts
@@ -29,7 +29,6 @@ export class Runfiles {
 
   private _initialize() {
     if (this._initialized) return;
-    this._initialized = true;
     // If Bazel sets a variable pointing to a runfiles manifest,
     // we'll always use it.
     // Note that this has a slight performance implication on Mac/Linux
@@ -69,6 +68,7 @@ export class Runfiles {
       // //path/to:target -> path/to
       this.package = target.split(':')[0].replace(/^\/\//, '');
     }
+    this._initialized = true;
   }
 
   /** Resolves the given path from the runfile manifest. */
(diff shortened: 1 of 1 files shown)


packages/runfiles/runfiles.ts
```diff
@@ -23,20 +23,26 @@ export class Runfiles {
...
   repoMappings: RepoMappings|undefined;
 
-  constructor(private _env = process.env) {
+  private _initialized = false;
+
+  constructor(private _env = process.env) {}
+
+  private _initialize() {
+    if (this._initialized) return;
+    this._initialized = true;
```
jbedard, line 32: This means the public methods such as `loadRunfilesManifest()` will only throw the first time, then `this._initialized = true` will prevent it from throwing again and the behaviour of `loadRunfilesManifest()` is then inconsistent.

Maybe we should keep the initialization in the `constructor()`, but move the `throw` into the public methods?

nickschaap: I could also move the flag change to be at the end of the _initialize method so that it doesn't ever mark it as initialized if it throws in the middle of the initialization

nickschaap: Ended up going with your initial suggestion and making sure that proper environment variables are set when the user calls `resolve()` either directly or as a result of calling another public method that calls it internally.

packages/runfiles/runfiles.ts
```diff
@@ -152,6 +159,8 @@ export class Runfiles {
 
   /** Resolves the given module path. */
   resolve(modulePath: string, sourceRepo?: string): string {
+    this._assertRunfilesResolved();
```
jbedard, line 162: Are you certain this is the only public method where this should be invoked?

jbedard: I think you're correct 👍

jbedard (approved):

Issue #3854, opened 2025-08-19
[Bug]: Execution toolchain defined without `target_compatible_with` makes it a candidate to selection


guw:
### What happened?

`rules_nodejs` registers two toolchains:
https://github.com/bazel-contrib/rules_nodejs/blob/356c5ff7e8804164bf9d13c3e9536579704b3b0f/nodejs/private/nodejs_toolchains_repo.bzl#L110-L121

One is for execution on the host (execution toolchain) and another one for cross-compilation. However, the execution toolchain is defined without `target_compatible_with`, which makes it applicable to all possible targets (no constraints). In our situation this leads to an issue where NodeJS Linux AMD64 is picked wrongly instead of Linux ARM64 when building multi-platform containers on AMD64 machine.

This issue appeared after migrating to bzlmod. In our pre-bzlmod `WORKSPACE` setup we manually registered NodeJS toolchains ourselves and the problem does not occur.

### Version

Development (host) and target OS/architectures:
* Host: Linux AMD64
* Target: Linux ARM64
* bzlmod

Output of `bazel --version`:
* 7.6.1

Version of rules_nodejs, or other relevant rules from your
`WORKSPACE` or `MODULE.bazel` file:
* 6.4.0

### How to reproduce

* needs a multi-platform oci image build with transitions to produce arm64 image on amd64 host

### Any other information?

Note, the second target specific toolchains was introduced in #3373. It seems to rely purely on lexicographic ordering. I do wonder if this is the correct approach and if both toolchain definitions are needed.

fmeum:
Forget what I said above, the two toolchains use the same toolchain type. That's just wrong - target and exec toolchain need different toolchain types. https://github.com/bazel-contrib/rules_nodejs/pull/3800 is the right approach.

lucasteixeira-cb:
I've found the workaround here: https://github.com/bazel-contrib/rules_nodejs/pull/3855/files and it does work. Thanks @guw!

PR #3855, opened 2025-08-19
Set target_compatible_with for host execution toolchain

guw:
Fix for #3854

## PR Checklist

## PR Type

- [x] Bugfix

Issue Number: #3854

## What is the new behavior?

The execution toolchain will not be selected for incompatible target platform.

- [x] No

### DIFF
--- a/nodejs/private/nodejs_toolchains_repo.bzl
+++ b/nodejs/private/nodejs_toolchains_repo.bzl
@@ -107,9 +107,27 @@ resolved_toolchain(name = "resolved_toolchain", visibility = ["//visibility:publ
 
     for [platform, meta] in PLATFORMS.items():
         build_content += """
+# Toolchain resolution is quite complex when it has to choose between both target and exec platform.
+# When building any target it has both exec and target platform set.
+# On the other hand the toolchain resolution tries to find the first one that matches. The order is lexicographical
+# based on native.toolchain's name.
+# This means that when you build a target for linux-arm64 on linux-amd64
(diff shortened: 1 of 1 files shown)


guw:
It sounds like https://github.com/bazel-contrib/rules_nodejs/pull/3800 will take sometime to bake. Can we merge this meanwhile to mitigate #3854?

nodejs/private/nodejs_toolchains_repo.bzl
```diff
@@ -107,9 +107,27 @@ resolved_toolchain(name = "resolved_toolchain", visibility = ["//visibility:publ
 
     for [platform, meta] in PLATFORMS.items():
         build_content += """
+# Toolchain resolution is quite complex when it has to choose between both target and exec platform.
```
jbedard, line 110: Should this comment be in the generated code like this or outside? Should it be repeated for each platform like it is atm?

guw: Where would you like it to put? I can move it outside.

alexeagle (approved):
@fmeum is my expert here. I'm pretty sure we had this code years ago and removed it.

fmeum:
This fix would make the situation more incorrect until we land a fix similar to #3800: we don't need two different kinds of constraints on any of the two toolchains. 

@guw Could you share more details on how Node ends up in the image? That ultimate consumer of the Node toolchain should be updated to consume the existing target toolchain type rather than the exec toolchain type.

PR #3880, opened 2025-11-15
chore: test on bazel 9

alexeagle:
Requires adapting to the Bazel 9 breaking changes (finally)

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -16,6 +16,17 @@ workspace(name = "rules_nodejs")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
+http_archive(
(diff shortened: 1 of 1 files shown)


.bcr/presubmit.yml
```diff
@@ -1,8 +1,8 @@
 bcr_test_module:
   module_path: "e2e/smoke"
   matrix:
-    bazel: ["7.x", "8.x", "rolling"]
-    platform: ["debian11", "macos", "ubuntu2204", "windows"]
+    bazel: ["7.x", "8.x", "9.*"]
+    platform: ["debian10", "macos", "ubuntu2004", "windows"]
```
jbedard: Why downgrade OSs?

PR #3911, opened 2026-04-09
feat: add node_data attribute to nodejs_toolchain

djedward:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Only the node binary itself can be provided for a toolchain. No supporting files may be included.

Issue Number: N/A

## What is the new behavior?

This introduces a new, optional, node_data attribute to nodejs_toolchain to allow for additional runtime files to be provided with the node binary. Though this is likely not a common need, we ran into issues for some of our node toolchains which have wrapper scripts to handle special setups. I'm working on a corresponding update to [rules_js](https://github.com/djedward/rules_js) to use this new attribute as well.

- [ ] Yes
- [x] No

## Other information
All of the links in the existing documentation to nodejs_toolchain appear to go to non-existent places, so outside of the documetation in the toolchain.bzl, I didn't add any.

### DIFF
--- a/e2e/smoke/BUILD.bazel
+++ b/e2e/smoke/BUILD.bazel
@@ -1,7 +1,8 @@
 load("@bazel_skylib//rules:build_test.bzl", "build_test")
 load("@bazel_skylib//rules:diff_test.bzl", "diff_test")
 load("@bazel_skylib//rules:write_file.bzl", "write_file")
-load(":defs.bzl", "my_nodejs")
+load("@rules_nodejs//nodejs:toolchain.bzl", "nodejs_toolchain")
+load(":defs.bzl", "check_node_data", "my_nodejs")
 
 not_windows = select({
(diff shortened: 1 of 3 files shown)


jbedard:
Can you provide an example of when this would be used instead of simply using `js_binary(data)` on the binary target itself?

djedward:
Our use-case is that our node team produces our node toolchain with a wrapper script:
```
bin/
├── node (shell script with internal-specific setup)
├── node_real (the real node binary, can't run without the shell setup)
```

So when we setup the toolchain for other teams to consume, we need a way for both of these files to be present in the sandbox. One hack we played with was including it in the `npm_srcs` and then all js_binary uses could specify `include_npm`. It works, but quite hacky/confusing to consumers.  Similarly, providing a way for it to be specified in js_binary `data` could potentially work, but since it is part of the toolchain, it felt wrong to push onto consumers.

PR #3916, opened 2026-06-12
feat: use facts to auto-retrieve node versions if possible

gzm0:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Node versions / digests are hardcoded and shipped with rules_nodejs.

## What is the new behavior?

If the bazel facts feature is available (bazel >= 8.5.0) rules_nodejs retrieves required Node digests as needed and stores them in facts.

- [ ] Yes
- [x] No

## Other information

Sorry for the out of the blue PR, I felt it was faster to just build this and discuss whether this is worth including on the PR directly.

### DIFF
--- a/e2e/nodejs_host/BUILD.bazel
+++ b/e2e/nodejs_host/BUILD.bazel
@@ -1,3 +1,4 @@
+load("@bazel_features//:features.bzl", "bazel_features")
 load("@rules_shell//shell:sh_test.bzl", "sh_test")
 
 # Dummy test target that imports various targets from the
@@ -23,8 +24,9 @@ load("@rules_shell//shell:sh_test.bzl", "sh_test")
         "node16",
         "node16_nvmrc",
         "node17_custom",
-        "node26_facts",
-    ]
+    ] + (
+        ["node26_facts"] if bazel_features.external_deps.has_facts else []
+    )
 ]
 
 [
@@ -67,7 +69,10 @@ load("@rules_shell//shell:sh_test.bzl", "sh_test")
         ("node16_nvmrc", "npx"),
         ("node17_custom", "npm"),
         ("node17_custom", "npx"),
-        ("node26_facts", "npm"),
-        ("node26_facts", "npx"),
-    ]
+    ] + (
+        [
+            ("node26_facts", "npm"),
+            ("node26_facts", "npx"),
+        ] if bazel_features.external_deps.has_facts else []
+    )
 ]
--- a/e2e/nodejs_host/BUILD.bazel
+++ b/e2e/nodejs_host/BUILD.bazel
@@ -23,6 +23,7 @@ load("@rules_shell//shell:sh_test.bzl", "sh_test")
         "node16",
         "node16_nvmrc",
         "node17_custom",
+        "node26_facts",
     ]
 ]
 
@@ -66,5 +67,7 @@ load("@rules_shell//shell:sh_test.bzl", "sh_test")
         ("node16_nvmrc", "npx"),
         ("node17_custom", "npm"),
         ("node17_custom", "npx"),
+        ("node26_facts", "npm"),
+        ("node26_facts", "npx"),
     ]
 ]
--- a/nodejs/extensions.bzl
+++ b/nodejs/extensions.bzl
@@ -9,6 +9,9 @@ use_repo(node, "nodejs_toolchains")
 ```
 """
 
+load("//nodejs/private:fetch_node_repositories.bzl", "fetch_node_repositories")
+load("//nodejs/private:nodejs_toolchains_repo.bzl", "PLATFORMS")
+load("//nodejs/private:version_from_attr.bzl", "version_from_attr")
 load(
     ":repositories.bzl",
     "DEFAULT_NODE_REPOSITORY",
@@ -28,6 +31,57 @@ def _toolchains_equal(lhs, rhs):
         if getattr(lhs, attr) != getattr(rhs, attr):
             return False
     return True
+
+def _use_repository_facts(registration):
+    """Whether we should use facts (if available) for the given registration."""
+
+    if registration.node_repositories.items():
+        return False  # custom repositories, do not mess with them.
+    if registration.node_urls != [DEFAULT_NODE_URL]:
+        return False  # we do not know how to fetch shas with custom URLs.
+
+    return True
+
+def _update_repository_facts(module_ctx, registrations):
+    """Fetch / update the necessary repository facts for the given registrations.
+
+    * Takes into account existing facts.
+    * Returns empty dict if facts are not supported.
+    """
+
+    if not hasattr(module_ctx, "facts"):
+        # facts not supported.
+        # repository rules will fallback to builtin NODE_VERSIONS.
+        return {}
+
+    new_facts = {}
+
+    for registration in registrations.values():
+        if not _use_repository_facts(registration):
+            continue
+
+        version = version_from_attr(module_ctx, registration)
+
+        should_fetch = False
+
+        # Get repository values for the default PLATFORMS.
+        # These are the only platforms supported by the extension anyways.
+        for platform in PLATFORMS:
+            key = version + "-" + platform
+
+            existing = new_facts.get(key) or module_ctx.facts.get(key)
+            if existing:
+                new_facts[key] = existing
+            else:
+                should_fetch = True
+
+        if should_fetch:
+            # Note: Even after fetching, it is possible that keys are not in facts:
+            # Old node versions might not have all platforms.
+            # In that case, we want to fail lazily when the repository is actually used.
+            new_facts.update(fetch_node_repositories(module_ctx, version))
+
+    return new_facts
 
 def _toolchain_extension(module_ctx):
     registrations = {}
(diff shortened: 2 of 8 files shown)


nodejs/repositories.bzl
gzm0, line 1: Happy to split this change out into another PR if desired.

nodejs/extensions.bzl
```diff
@@ -29,6 +32,57 @@ def _toolchains_equal(lhs, rhs):
...
+        version = version_from_attr(module_ctx, registration)
+
+        should_fetch = False
+
+        # Get repository values for the default PLATFORMS.
+        # These are the only platforms supported by the extension anyways.
+        for platform in PLATFORMS:
+            key = version + "-" + platform
+
+            existing = new_facts.get(key) or module_ctx.facts.get(key)
```
gzm0: Alternative: Get from NODE_VERSIONS if available.

nodejs/extensions.bzl
```diff
@@ -29,6 +32,57 @@ def _toolchains_equal(lhs, rhs):
...
+            if existing:
+                new_facts[key] = existing
+            else:
+                should_fetch = True
+
+        if should_fetch:
+            # Note: Even after fetching, it is possible that keys are not in facts:
+            # Old node versions might not have all platforms.
+            # In that case, we want to fail lazily when the repository is actually used.
+            new_facts.update(fetch_node_repositories(module_ctx, version))
```
gzm0: Note: This will trigger fetch on each evaluation for old node versions where not all platforms are available.

However, the only alternative requires a different fact layout (where all the available platforms in a version are represented in the fact). I opted to keep the fact layout the same as the node repository layout, because that's already a public API.

nodejs/extensions.bzl
```diff
@@ -29,6 +32,57 @@ def _toolchains_equal(lhs, rhs):
...
+        # Get repository values for the default PLATFORMS.
+        # These are the only platforms supported by the extension anyways.
+        for platform in PLATFORMS:
+            key = version + "-" + platform
+
+            existing = new_facts.get(key) or module_ctx.facts.get(key)
+            if existing:
+                new_facts[key] = existing
+            else:
+                should_fetch = True
```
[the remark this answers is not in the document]
gzm0: ... this comment is actually kind of impressive :)

Indeed a trade-off to be made, see https://github.com/bazel-contrib/rules_nodejs/pull/3916#discussion_r3402824337

nodejs/extensions.bzl
gzm0: Resolving. Let's continue this discussion on this thread: https://github.com/bazel-contrib/rules_nodejs/pull/3916#discussion_r3402838476

gzm0:
The new tests fail on bazel 7.x (expected). Any help on how to exclude is appreciated :) I'll start digging.

nodejs/extensions.bzl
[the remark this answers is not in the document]
jbedard: I think we need to reverse this and only use facts if the version is not in `NODE_VERSIONS`. Basically as a fallback if (a) rules_nodejs hasn't included yet or (b) you don't want to upgrade or (c) you prefer hashes generated into your `MODULE.bazel.lock` instead of manually written in your `MODULE.bazel`.

Basically what was done for [fetching pnpm in rules_js](https://github.com/aspect-build/rules_js/blob/v3.2.2/npm/private/pnpm_extension.bzl#L145-L169)

gzm0:
@jbedard, I have updated this to only use facts for unknown node versions. For this to be easy, I've opened #3917.

To review this PR in isolation, only look at the last three commits:
https://github.com/bazel-contrib/rules_nodejs/pull/3916/changes/463749c99bb52184b632dd1ed4dd38f7c094a9d3..0db48b3bac8d50b5b264ab78fe169aeb3a0e2f4f

Note that I'm still working on fixing / excluding the new test for bazel 7.7.

nodejs/extensions.bzl
gzm0: Resolving. The node version based retrieval fixes this.

gzm0:
I'm trying to get `has_facts` added to bazel features so we can adjust the tests (https://github.com/bazel-contrib/bazel_features/pull/147).

My plan is to exclude usage of the new node26 repositories from the e2e test, but *not* the definition. Keeping the definition ensures that versions that would fall back to facts don't try to call `module_ctx.facts`.

gzm0:
- Added a fix for the tests that pulls in https://github.com/bazel-contrib/bazel_features/pull/147 via git_override.
- Marking as work in progress to prevent accidental merging.

It seems that the git_override does not work on CI. I've tested locally that `e2e/nodejs_host$ bazel test ...` passes.

gzm0:
This is ready for final review.

e2e/nodejs_host/BUILD.bazel
```diff
@@ -23,7 +24,9 @@ load("@rules_shell//shell:sh_test.bzl", "sh_test")
         "node16",
         "node16_nvmrc",
         "node17_custom",
-    ]
+    ] + (
+        ["node26_facts"] if bazel_features.external_deps.has_facts else []
```
jbedard, line 28: Can't we still run the tests, it just won't be reproducible?

nodejs/extensions.bzl
```diff
@@ -52,17 +55,39 @@ def _toolchain_extension(module_ctx):
...
+            node_version = node_version,
             node_urls = v.node_urls,
-            node_repositories = v.node_repositories,
+            node_repositories = node_repositories,
             include_headers = v.include_headers,
             register = False,
         )
 
+    if supports_facts:
+        return module_ctx.extension_metadata(
```
jbedard: I normally prefer writing these such that when we drop support for bazel versions that don't `support_facts` you only have to delete code, without moving other code around. So I think:

```
if not supports_facts:
    return  # buildifier: disable=return-value (allow no value)

return module_ctx.extension_metadata(...)
```

Then the `git blame` won't change, you will forever be blamed for this change 👍

nodejs/extensions.bzl
```diff
@@ -52,17 +55,39 @@ def _toolchain_extension(module_ctx):
...
+            reproducible = True,
```
jbedard: It would be nice to do this even if `not supports_facts`, although that's not directly related to your feature so only if its super simple maybe...?

gzm0: Good point. I've added a commit that returns as much as we can.

nodejs/extensions.bzl
gzm0: Heh, that's a neat trick indeed. Nice! (I updated).

e2e/nodejs_host/BUILD.bazel
gzm0: No. With the current PR, we'll hit this line eventually:

https://github.com/bazel-contrib/rules_nodejs/blob/e6706c5410be89e95a4114e384732bb63a58ae08/nodejs/repositories.bzl#L90-L92

(because we pass empty node_repositories).

What we can do, is to call `fetch_node_repositories` in the extension even if facts are not supported.
Then, if we fetched something, we can return `reproducible = False`. This will snapshot the repository rule invocation parameters in the lockfile (at least with sufficiently new bazel versions when I checked this last). I will take a stab at this (I'll open a separate PR).

What I would advise against is to make the repository rule itself not reproducible (especially w/o explicit opt in). IMHO this would be a step back.

gzm0: Alternative: https://github.com/bazel-contrib/rules_nodejs/pull/3919

jbedard: > What we can do, is to call fetch_node_repositories in the extension even if facts are not supported.
Then, if we fetched something, we can return reproducible = False. This will snapshot the repository rule invocation parameters in the lockfile (at least with sufficiently new bazel versions when I checked this last). I will take a stab at this (I'll open a separate PR).

This sounds ideal to me to me I think, IIUC.

> What I would advise against is to make the repository rule itself not reproducible (especially w/o explicit opt in). IMHO this would be a step back.

That would only be a step back when (1) there checksums are not in rules_nodejs so your new feature is used, but (2) there is no lockfile (it's not checked in) or the bazel version is too old to persist facts. So it would only be a step back when your new feature is used with old bazel versions and no lockfile? That doesn't sound horrible to me, and encourages the use of lockfiles...

We could potentially output a warning or even `fail()` if we had to fetch node to determine checksums but there is no facts API? That would only be for old bazel versions basically though...

gzm0: > This sounds ideal to me to me I think, IIUC.

That's what #3919 does. So let's focus on that one.

Regarding the point about the repository rule not being reproducible:

The execution of repository rules does not get snapshotted to the lockfile (this is in contrast to module extensions). So even when running on an up-to-date bazel version with a checked-in lockfile, a non-reproducible repository rule is kind of bad.

Just to be clear: Another way of saying what I'm trying to say is we should not call `fetch_node_repositories` in a repository rule. We should always call it in the module extension and then call `nodejs_register_toolchains` with the resolved digests (both this PR and #3919 do that).

I have tested with older bazel versions down to 7.1.0 and they all snapshot the repository rule invocations and their parameters to the lockfile ([like this](https://gist.github.com/gzm0/777a23ce5201d2c70a3b18bde4b83d4a#file-module-bazel-lock-L749-L797)). So we'll only end up with something non-reproducible if the lockfile is not checked in (which always was the case).

gzm0:
Closing. Obsoleted by #3919

PR #3919, opened 2026-06-18
feat: auto-fetch node versions in module extension

gzm0:
More aggressive alternative to #3916: We even do the fetching if facts are not available.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Node versions / digests are hardcoded and shipped with rules_nodejs.

## What is the new behavior?

rules_nodejs retrieves required Node digests as needed and stores them in facts or in generatedRepoSpecs in the bzlmod lockfile.

- [ ] Yes
- [x] No

## Other information

- #3916 (fact only alternative)
- [Example lockfile with bazel 8.1.0](https://gist.github.com/gzm0/777a23ce5201d2c70a3b18bde4b83d4a)
- [Example lockfile with bazel 9.1.1](https://gist.github.com/gzm0/eeb1e54658e78259a33f763376a79004)

### DIFF
--- a/nodejs/extensions.bzl
+++ b/nodejs/extensions.bzl
@@ -9,6 +9,10 @@ use_repo(node, "nodejs_toolchains")
 ```
 """
 
+load("@bazel_features//:features.bzl", "bazel_features")
+load("//nodejs/private:fetch_node_repositories.bzl", "fetch_node_repositories")
+load("//nodejs/private:node_versions.bzl", "NODE_VERSIONS")
+load("//nodejs/private:version_from_attr.bzl", "version_from_attr")
 load(
     ":repositories.bzl",
     "DEFAULT_NODE_REPOSITORY",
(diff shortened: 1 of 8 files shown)


nodejs/extensions.bzl
```diff
@@ -52,17 +56,50 @@ def _toolchain_extension(module_ctx):
...
+    new_repository_facts = {}
+
     for k, v in registrations.items():
+        node_version = version_from_attr(module_ctx, v)
+        node_repositories = v.node_repositories or NODE_VERSIONS.get(node_version, {})
+
+        if not node_repositories:
+            node_repositories = (
+                new_repository_facts.get(node_version) or
+                (module_ctx.facts.get(node_version) if supports_facts else None) or
```
jbedard, line 69: can we just do `old-facts = module_ctx.facts if supports_facts else {}` up above alongside the `new_repository_facts = {}`?

gzm0: Nope:

> Error: can only assign an element in a dictionary or a list, not in a 'Facts'

`module_ctx.facts` is not a dict, its a [Facts](https://bazel.build/rules/lib/builtins/Facts). (tl;dr: immutable, cannot be enumerated).

Note that we wouldn't want this anyways: We do not want to keep old, now unnecessary facts. For example, if we change the node version from 24.0.0 to 24.0.1 in `MODULE.bazel`, upon first invocation of the extension, we want to throw away the facts for 24.0.0; otherwise the facts end up growing unbounded.

This property is somewhat neatly enforced by Facts not being enumerable.

jbedard (approved):

jbedard:
I think this LGTM, are you in favour of merging this one over #3916?

gzm0:
Yes, I'm in favor of this one.

gzm0:
@jbedard any chance we could get this released? :innocent:

jbedard:
I've tagged a release which has [kicked things off](https://github.com/bazel-contrib/rules_nodejs/actions/runs/28082502534/job/83140532790)

alexeagle:
Does it mean we could turn off the job that mirrors node versions and hashes into this repo?

jbedard:
I would prefer if we didn't force users to depend on lockfiles or facts being persisted to make things `reproducible`, or forcing them to hardcode hashes.

I would like to consider:
* dropping old major versions (do we really need the node 8 hashes?)
* only persisting _some_ versions, maybe the latest patch of each minor release?

Then users can use avoid facts/.lock/hashes if they are willing to use the latest patch version.

WDYT? I'm debating the same for everywhere we hardcode hashes like this...

alexeagle:
Facts are only in some bazel versions, so it sounds right to me to avoid relying on them until this repo stops supporting Bazel version without it.

Once we're sure users have the feature available, I feel we could consolidate on the new way of pinning that Bazel intends via the lockfile - having hashes provided by rulesets feels like it was a workaround prior to that being possible.

jbedard:
Once we drop bazel 7 **and** we are confident that people are checking lockfiles in...

gzm0:
My 2 cts, since I saw this conversation :shrug:

- Using repository rules directly still requires `NODE_VERSIONS` and IIUC there is no feature on the horizon that would allow otherwise (and direct repository usages are not always easily replaceable by extensions).
- I'm not sure not checking lockfiles in is something we should be concerned about: if users do not check in the lockfile, rules_nodejs itself is also not pinned.

jbedard:
>  if users do not check in the lockfile, rules_nodejs itself is also not pinned

But rules_nodejs and other modules are at least cached properly, where IIUC fetching the node hashes wouldn't be?

PR #3929, opened 2026-09-16
fix(nodejs): expose headers as a standalone target

alexeagle:
## Problem

`include_headers = True` currently creates the generated `:headers` `cc_library` and also wires it into `NodeInfo`. That makes every consumer of the Node toolchain depend on C++ toolchain resolution.

Some native consumers only need to reference the headers explicitly and should not add that dependency to unrelated Node targets.

## Change

Always generate the standalone `:headers` target, while preserving the existing conditional `NodeInfo.headers` behavior. Users can depend directly on `@nodejs_<platform>//:headers` without enabling `include_headers`.

The existing `include_headers = True` behavior remains unchanged.

## Test

- `bazel test //:using_headers_test --test_output=errors` from `e2e/headers`


nodejs/repositories.bzl
```diff
@@ -252,8 +252,7 @@ filegroup(
         npx_entry = npx_entry,
     )
 
-    if repository_ctx.attr.include_headers:
-        build_content += """
+    build_content += """
 load("@rules_cc//cc:defs.bzl", "cc_library")
 cc_library(
```
jbedard: Do you just want access to the files or are you explicitly looking for the compiled `cc_library`?

We could _always_ put the files into a `filegroup` but keep the `cc_library` behind the flag 🤷

jbedard: Should this be tagged as manual? Even without your change, but especially after it...

